PR #1, opened 2016-11-14
bolt07: Adding P2P routing gossiping

cdecker:
This is my first draft of the gossiping protocol. It implements a
staggered broadcast, and forces nodes to have a channel open in order to
get announcements forwarded.

All announcements are signed by the announcing node, the messages are
smaller than FLARE, has metadata for nodes and the signed state of the
announcing node can be reconstructed from the node table and channel
table that we need anyway. The downside is that we disseminate the
network topology to every participant in the network, but we should be
able to get up to a few tens of thousand nodes before this becomes an
issue.

We need to merge this with FLARE light: https://docs.google.com/document/d/1sgghubGy2X23CivR8ndUjDnGvCYczcPwbq8amrtyhDo/edit

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -1,0 +1,54 @@
+<h1 id="bolt-7-p2p-node-and-channel-discovery">BOLT #7: P2P Node and Channel Discovery</h1>
+<p>This specification describes a simple initial node and channel discovery mechanism that does not rely on a third-party to disseminate the information.</p>
+<p>Node and channel discovery serve two different purposes:</p>
+<ul>
+<li>Channel discovery allows the creation and maintenance of a local view of the network's topology such that the node can discover routes to desired destination.</li>
+<li>Node discovery allows nodes to broadcast their ID, host and port, such that other nodes can open connections and establish payment channels.</li>
+</ul>
+<p>Peers in the network exchange <code>route_announcement</code> messages that contain information about a node and about its outgoing channels. The node creating an announcement is referred to as the <em>announcing node</em>. Announcements MUST have at least one associated channel whose existence can be proven by inspection of the blockchain, i.e., the anchor transaction creating the channel MUST have been confirmed.</p>
+<p>Nodes receiving an announcement verify that this is the latest update from the announcing node by comparing the included timestamp and update their local view of the network's topology accordingly. The receiving node removes all channels that are signaled to be removed or are no longer included in the announcement, adds any new channels and adjusts the parameters of the existing channels. Notice that this specification does not include any mechanism for differential announcements, i.e., every announcement ships the entire final state for that node.</p>
+<p>Once the announcement has been processed it is added to a list of outgoing announcements to the processing node's peers, which will be flushed at regular intervals. This store and delayed forward broadcast is called a <em>staggered broadcast</em></p>
+<p>Notice that each announcement will only announce a single direction of the channel, i.e., the outgoing direction from the point of view of the announcing node. The other direction will be announced by the other endpoint of the channel. This reflects the willingness of the announcing node to forward payments over the announced channel.</p>
+<h2 id="message-format">Message Format</h2>
+<p>This specification defines one message type with identifier 256 (<code>0x00000100</code>), called a <code>MSG_ROUTING_UPDATE</code>. It comprises both the sending node's identity and contact information as well as all of its outgoing channels.</p>
+<pre><code>[4:timestamp]
+[33:node_id]
+[16:ipv6]
+[2:port]
+[3:rgb_color]
+[var:alias]
+[4+X*46:channels]
+[64:signature]</code></pre>
+<p>The timestamp is required to be monotonically increasing from one update to another, it MAY correspond to unix timestamp when the sending node created the announcement. The node ID is a 33 byte compressed public key that uniquely identifies the node in the network and matches the signature. The IPv6 field is used for both IPv4 and IPv6 addresses, using the IPv4-Mapped IPv6 Address format defined in <a href="https://tools.ietf.org/html/rfc4291#section-2.5.5.2">RFC 4291 section 2.5.5.2</a>:</p>
+<pre><code>|                80 bits               | 16 |      32 bits        |
++--------------------------------------+--------------------------+
+|0000..............................0000|FFFF|    IPv4 address     |
++--------------------------------------+----+---------------------+</code></pre>
+<p>The alias is a length prefixed UTF-8 encoded string. The UTF-8 encoded string is prefixed by a 4 byte unsigned int in network byte order, defining the length of the string. The RGB color and alias MAY be used by the node operator to customize their node's appearance in maps and graphs, and MAY be used to reference nodes in interfaces. Notice however that there is no collision protection for aliases, hence the node ID MUST be verified before initiating any transfer.</p>
+<p>The <code>channels</code> field comprises a 4 byte unsigned int count of channels that follow. Each channel has the following format:</p>
+<pre><code>[33:destination]
+[4:blockheight]
+[4:blockindex]
+[1:expiry]
+[2:fee_base]
+[2:fee_proportional]</code></pre>
+<p>The <code>destination</code> field is the node ID of the remote node from the point of view of the announcing node. Thus the <code>node_id</code> from the enclosing announcement and the <code>destination</code> identify the two endpoints of the channel. The <code>blockheight</code> and <code>blockindex</code> MAY be combined to form a globally unique <code>channel_id</code> that MAY be used to identify the channel in future. The remaining fields determine the minimum timelock difference that HTLCs have to respect (<code>expiry</code>) and the fees that the announcing node expected when routing a payment through the announced channel from <code>node_id</code> to <code>destination</code>.</p>
+<h2 id="creation-and-processing-of-announcements">Creation and Processing of Announcements</h2>
+<p>The announcing node creates the message with the node's information and all its channels. Normal removal of a channel is done by omitting the channel in the <code>channels</code> field. Notice that this does not allow removing a channel if no active channels are left open, since an announcement requires at least one channel in the <code>channels</code> field to be valid. An explicit removal of all channel MAY be signaled by creating an announcement that includes only one last channel to be closed, and setting the <code>expiry</code> field of that channel to <code>0xFF</code>.</p>
+<p>The announcing node serializes the message omitting the <code>signature</code> field, and computes the signature on the partial serialized message using the private key matching the <code>node_id</code> in the message. The signature is then serialized and appended to the message, completing the message.</p>
+<p>The announcing peer then sends the announcement to its peers, initiating the broadcast.</p>
+<p>Upon receiving an announcement the nodes verifies the validity of the announcement. In order to be valid the following conditions MUST be satisfied:</p>
+<ul>
+<li>The timestamp MUST be larger than the last valid announcement from the announcing node.</li>
+<li>The signature in the announcement MUST be a valid signature from the public key in the <code>node_id</code> field for the message up to the signature itself.</li>
+<li>The announcement MUST have at least on valid channel. The validity of the channel can be verified by inspecting the anchor transaction specified in the announcement.</li>
+<li>If the <code>expiry</code> field of any channel is set to <code>0xFF</code> then it MUST be the only channel in the announcement.</li>
+</ul>
+<p>If an announcement is not valid, it MUST be discarded, otherwise the node applies it to its local view of the topology: the receiving node removes all channels from its local view that match the <code>node_id</code> as the origin of the channel, i.e., all channels that have been previously announced by that node, and adds all channels in the announcement unless they have an <code>expiry</code> field of <code>0xFF</code>.</p>
+<p>If, after applying the changes from the announcement, there are no channels associated with the announcing node, then the receiving node MAY purge the announcing node from the set of known nodes. Otherwise the receiving node updates the metadata and stores the signature associated with the announcement. This will later allow the receiving node to rebuild the announcement for its peers.</p>
+<p>After processing the announcement the receiving node adds the announcement to a list of outgoing announcements. The list of outgoing announcement MUST NOT contain multiple announcements with the same <code>node_id</code>, duplicates MUST be removed and announcements with lower <code>timestamp</code> fields MUST be replaced. This list of outgoing announcements is flushed once every 60 seconds, independently of the arrival times of announcements, resulting in a staggered announcement and deduplication of announcements.</p>
+<p>Nodes MAY re-announce their channels regularly, however this is discouraged in order to keep the resource requirements low. In order to bootstrap nodes that were not online at the time of the broadcast nodes will announce all known nodes and their associated channels at the time of connection establishment. The individual announcements can be reconstructed from the set of known nodes, containing the metadata and signatures for the announcements, and the routing table, containing the channel information. The broadcast is stopped after the first hop since the peers of the newly joined node already have the announcement and the timestamp check will fail.</p>
+<h2 id="references">References</h2>
+<ul>
+<li><a href="https://tools.ietf.org/html/rfc4291">RFC 4291</a></li>
+</ul>
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -87,7 +87,7 @@ Upon receiving an announcement the nodes verifies the validity of the announceme
 In order to be valid the following conditions MUST be satisfied:
 
  - The timestamp MUST be larger than the last valid announcement from the announcing node.
- - The anchor transactions for each channel, identified by the (`blockheight`, `blockindex`)-tuple must have reached at least the configurable minimum number of confirmations.
+ - The anchor transactions for each channel, identified by the (`blockheight`, `blockindex`)-tuple must have reached at least a minimum number of 6 confirmations. The node MAY require more confirmations before processing and/or forwarding the announcement.
  - The signature in the announcement MUST be a valid signature from the public key in the `node_id` field for the message up to the signature itself.
  - The announcement MUST have at least on valid channel. The validity of the channel can be verified by inspecting the anchor transaction specified in the announcement.
  - If the `expiry` field of any channel is set to `0xFF` then it MUST be the only channel in the announcement.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -1,54 +1,113 @@
-<h1 id="bolt-7-p2p-node-and-channel-discovery">BOLT #7: P2P Node and Channel Discovery</h1>
-<p>This specification describes a simple initial node and channel discovery mechanism that does not rely on a third-party to disseminate the information.</p>
-<p>Node and channel discovery serve two different purposes:</p>
-<ul>
-<li>Channel discovery allows the creation and maintenance of a local view of the network's topology such that the node can discover routes to desired destination.</li>
-<li>Node discovery allows nodes to broadcast their ID, host and port, such that other nodes can open connections and establish payment channels.</li>
-</ul>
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -0,0 +1,54 @@
...
+<h2 id="message-format">Message Format</h2>
+<p>This specification defines one message type with identifier 256 (<code>0x00000100</code>), called a <code>MSG_ROUTING_UPDATE</code>. It comprises both the sending node's identity and contact information as well as all of its outgoing channels.</p>
+<pre><code>[4:timestamp]
+[33:node_id]
+[16:ipv6]
+[2:port]
+[3:rgb_color]
+[var:alias]
+[4+X*46:channels]
+[64:signature]</code></pre>
```
Roasbeef: If we only use a single signature generated using a node's identity public key, then it is possible for nodes to advertise non-canonical versions of the network topology by generating several ID's and re-advertising the channel with each. 

In order to eliminate this possibility (if we deem it undesirable), then we can add a single signature to each of the advertised channels. This signature would be generated by adding the multi-sig channel key, and a node's identity channel key to create a channel authentication key. A signature over the channel description would then be generated using this key. 

Nodes can verify this signature by performing public key recovery on the signature in order to yield the channel authentication key, then verify that the result of adding the multi-sig key for the channel and a node's identity key yields the same point. After this recovery process, signature validation continues as normal, with nodes rejecting the channel if the output referenced on the blockchain is spent, or the sig is invalid.

07-routing-gossip.md
```diff
@@ -0,0 +1,54 @@
...
+<p>After processing the announcement the receiving node adds the announcement to a list of outgoing announcements. The list of outgoing announcement MUST NOT contain multiple announcements with the same <code>node_id</code>, duplicates MUST be removed and announcements with lower <code>timestamp</code> fields MUST be replaced. This list of outgoing announcements is flushed once every 60 seconds, independently of the arrival times of announcements, resulting in a staggered announcement and deduplication of announcements.</p>
+<p>Nodes MAY re-announce their channels regularly, however this is discouraged in order to keep the resource requirements low. In order to bootstrap nodes that were not online at the time of the broadcast nodes will announce all known nodes and their associated channels at the time of connection establishment. The individual announcements can be reconstructed from the set of known nodes, containing the metadata and signatures for the announcements, and the routing table, containing the channel information. The broadcast is stopped after the first hop since the peers of the newly joined node already have the announcement and the timestamp check will fail.</p>
```
Roasbeef: Announcing the _full_ routing table upon each reconnection may become a bit large (hundreds of megs) as the network grows. 

Instead, what if nodes collected the advertise channel state into a merkle tree. With this, upon reconnection nodes can exchange hashes then decide if they need to re-send the state or not. If the hashes differ, then the nodes can bisect the routing state tree in order to discover where their state diverged, sending the sub-tree(s) necessary for full synchronization. 

One issue with the proposal above is that it requires nodes to maintain the exact same order w.r.t the routing state so the can reconstruct an identical graph topology root. To remedy this, we can use a merkle patricia tree instead. Such a data structure has a desirable trait in that the structure of the final tree doesn't depend on insertion order. Therefore, nodes can process updates in an arbitrary order, yet still conduct the same tree root in the end.

07-routing-gossip.md
```diff
@@ -0,0 +1,54 @@
...
+<p>Nodes receiving an announcement verify that this is the latest update from the announcing node by comparing the included timestamp and update their local view of the network's topology accordingly. The receiving node removes all channels that are signaled to be removed or are no longer included in the announcement, adds any new channels and adjusts the parameters of the existing channels. Notice that this specification does not include any mechanism for differential announcements, i.e., every announcement ships the entire final state for that node.</p>
+<p>Once the announcement has been processed it is added to a list of outgoing announcements to the processing node's peers, which will be flushed at regular intervals. This store and delayed forward broadcast is called a <em>staggered broadcast</em></p>
+<p>Notice that each announcement will only announce a single direction of the channel, i.e., the outgoing direction from the point of view of the announcing node. The other direction will be announced by the other endpoint of the channel. This reflects the willingness of the announcing node to forward payments over the announced channel.</p>
```
Roasbeef: One down-side with this approach compared to the "both must advertise", is that we now have a good bit of redundant data flying around the network as each channel is essentially advertised twice: once by each node.

cdecker: Yeah that is a bit unfortunate, however it also brings some advantages:

 - we require no coordination by the end-points to advertise a channel in one direction
 - we sidestep the whole business of having to a canonical direction ordering and having to associate parameters with one direction or the other, e.g., the fee from `A` to `B` are `X`, while from `B` to `A` the fees are `Y`, it is always clear which direction we are talking about
 - channels can be removed uncooperatively, e.g., one node dies and I still get forward requests for that channel hours later, and it allows for uncooperative parameter changes, e.g., the other endpoint does not sign an announcement in which we increase our fees

I think it's a tradeoff between bytes in flight (storage at node is identical without OWAS), and the simplicity and the advantages I mentioned.

07-routing-gossip.md
cdecker: Very nice, I like the patricia tree idea. That was exactly my worry with merkle tree, missing a node somewhere in the middle would invalidate all following entries as well.

I'd still keep a node and it's channels together as the leafs of the tree, simply because having a single signature cover the latest state forces some consistency and we can reuse the hash from the signature as the leaf entry.

cdecker:
Oops, it seems I uploaded the `pandoc`d version of the markdown file. I'll commit the markdown source and squash upon merge. Rebase would kill inline comments...

07-routing-gossip.md
```diff
@@ -0,0 +1,113 @@
...
+The `destination` field is the node ID of the remote node from the point of view of the announcing node.
+Thus the `node_id` from the enclosing announcement and the `destination` identify the two endpoints of the channel.
+The `blockheight` and `blockindex` MAY be combined to form a globally unique `channel_id` that MAY be used to identify the channel in future.
+The remaining fields determine the minimum timelock difference that HTLCs have to respect (`expiry`) and the fees that the announcing node expected when routing a payment through the announced channel from `node_id` to `destination`.
+
+## Creation and Processing of Announcements
+
+The announcing node creates the message with the node's information and all its channels.
+Normal removal of a channel is done by omitting the channel in the `channels` field.
+Notice that this does not allow removing a channel if no active channels are left open, since an announcement requires at least one channel in the `channels` field to be valid.
```
pm47: What not just allow messages with no channels?

07-routing-gossip.md
```diff
@@ -0,0 +1,113 @@
...
+This specification describes a simple initial node and channel discovery mechanism that does not rely on a third-party to disseminate the information.
+
+Node and channel discovery serve two different purposes:
+
+ - Channel discovery allows the creation and maintenance of a local view of the network's topology such that the node can discover routes to desired destination.
+ - Node discovery allows nodes to broadcast their ID, host and port, such that other nodes can open connections and establish payment channels.
+ 
+Peers in the network exchange `route_announcement` messages that contain information about a node and about its outgoing channels.
+The node creating an announcement is referred to as the _announcing node_.
+Announcements MUST have at least one associated channel whose existence can be proven by inspection of the blockchain, i.e., the anchor transaction creating the channel MUST have been confirmed.
```
pm47: You mean that the transaction should have reached min_conf right?

07-routing-gossip.md
```diff
@@ -0,0 +1,113 @@
...
+
+If, after applying the changes from the announcement, there are no channels associated with the announcing node, then the receiving node MAY purge the announcing node from the set of known nodes.
+Otherwise the receiving node updates the metadata and stores the signature associated with the announcement.
+This will later allow the receiving node to rebuild the announcement for its peers.
+
+After processing the announcement the receiving node adds the announcement to a list of outgoing announcements.
+The list of outgoing announcement MUST NOT contain multiple announcements with the same `node_id`, duplicates MUST be removed and announcements with lower `timestamp` fields MUST be replaced.
+This list of outgoing announcements is flushed once every 60 seconds, independently of the arrival times of announcements, resulting in a staggered announcement and deduplication of announcements.
+
+Nodes MAY re-announce their channels regularly, however this is discouraged in order to keep the resource requirements low.
```
pm47: As a basic dos protection, shouldn't we just not propagate re-announcements?

07-routing-gossip.md
cdecker: The problem is not so much accepting messages without channels the problem is that we'd have to forward them as well, flooding the network with messages that are basically free to create.

An attacker can just generate any number of node IDs and announce its existence. And we can't just drop empty announces for which we don't know the corresponding node, because other nodes may know it and require the removal message.

If we require a channel, even a closed one, he must at least make a bitcoin transaction, which makes the attack non-free. I agree that signaling via the `expiry` is a bit wonky, but keep in mind that this is voluntary signaling, we can still remove channels if the anchor TX gets spent (modulo effort of tracking those).

07-routing-gossip.md
cdecker: Good point, I had a heartbeat in mind, allowing nodes to purge channels and nodes that it hasn't seen for a long time (24h?), but it might cause a lot of traffic.

We require timestamps to increase, so that re-announcements actually have to start from the announcing node. But dropping identical re-announcements is not hard, on the other hand it is trivial for the announcing node to slightly modify its announcement making it count as new.

07-routing-gossip.md
cdecker: Well we might just set this at 6 confirmations, which should be safe. Keep in mind though that the endpoints of the channel may have completely different policies, some accepting 0-conf anchors for channels and some being pedantic and waiting for 120 confirmations for example. It should be safe to just a lower bound enforced by nodes.

07-routing-gossip.md
Roasbeef: I see, in `lnd` we currently generate a fresh key for the multi-sig output in each channel. Within the codebase the identity keys and multi-sig keys are completely separated with the rationale that in the future the multi-sig keys may be completely offline. 

Yeh, that attaching a signature for each channel would indeed seriously blow up the size of the channel advertisements. However, since the advertising node controls the private key(s) used within all their funding outputs, and also their identity key, we can easily aggregate the signatures since nodes advertise info about their channel direction unilaterally. 

In order to compress the advertisements, a can node add together the private keys of all multi-sig keys and their identity (reduced modulo `N`, where `N` is the order of the curve), then use that to generate a single signature over all the advertisements. To validate, node's would then similarly perform pub-key recovery, add together all the pubkeys contained int he advertisement, ensure that the sum of those keys is what has been recovered, then verify the signature, and finally ensure the outputs is still unspent. This allows us to maintain the size of the advertisements as you've currently drafted them. 

I think this can also be adapted to be used for incremental updates, meaning an update the includes new channels that have been added.

07-routing-gossip.md
pm47: 6 confirmations seems fine. My point was that if we just wait for one confirmation (which is what the spec seems to be saying) we might have issues with some nodes not having received the new block yet.

07-routing-gossip.md
cdecker: Well in that case we need to also include the per-channel pubkey for both endpoints, in order to link the channel announcement to the anchor TX, which means an additional 66 bytes per channel announcement, since they are completely detached from the ID key.

But the aggregation scheme is nice, let's keep that in mind (and let's mix in the ID key as well), though I'm not sure whether this'd allow a forwarding node to mix in old announcements with new ones if we go with the incremental version.

07-routing-gossip.md
pm47: Right, we will probably need some kind of rate limiting logic and blacklisting then, à la bitcoind.

07-routing-gossip.md
cdecker: :+1: I'll add the 6 confirmations to the spec.

07-routing-gossip.md
rustyrussell: There's natural asymmetry here as you point out.  Ideally both sides should have to advertise creation, and then either can shutdown.  It's a little more complex, though:

1.  Channel announce.  This contains both node ids, and the channel.  Both node ids sign; you don't need to know anything about the node to validate and forward these.  Conflicting (ie. different other than sigs) channel announces also get forwarded, and the channel and nodes blacklisted.  
2.  Node announce. Only valid after channel announce has identified the node.
3.  Channel updates.  These are one-sided.  Only valid after channel announce.

On startup, you dump all the channel announce, all the node announces, and the latest channel update for each.

07-routing-gossip.md
rustyrussell: Bisecting is horrible for latency.  Fortunately, the rsync algorithm solves this pretty well, and is easy to implement.  You just need a canonical ordering of the data, which is trivial to come up with, and then you can do this in 1RTT.

Roasbeef:
As the prior comment change was getting a bit buried as the commits were updated, I'd like to discuss an approach for efficiently synchronizing the routing table state between two nodes. 

Earlier up in the chain Christian and I were discussing possibly introducing a merkle particia tree over the entire/known routing state. Upon connect, nodes would then use this tree to compare root hashes. If the root hashes differ, then they would engage in a bisection protocol in order to find the point where their state diverged, sending the appropriate sub-tree(s) in conclusion. This approach is nice as with a single RTT both sides can determine if their state is synchronized , and since the structure of merkle patricia trees aren't dependent on the order of insertion, nodes don't need to agree on a canonical leaf sorting order as they would for merkle trees. 

However, a major downside that Rusty pointed out is that the above protocol can lead to pretty poor latency to reach a synced state as it requires many round trips back and forth to reconcile the difference between the routing table state. It still may be worth looking into an authenticated data structure of the entire/known channel state as that might allow us to do some interesting things w.r.t to fragmentation/sharding of the state and associated updates (and I think it can be combined with what I'm about to suggest). As counter to eliminate the excessive latency, Rusty proposed that we use the `rsync` algorithm to find the difference in a single round-trip. I think we can take this a step further without getting _too_ fancy. 

Y'all remember IBLT's? I think they're perfectly applicable to the problem we're attempting to solve, namely: efficient set reconciliation. In our context, the set itself is the routing table. The reconciliation can be done in a single round-trip, if Alice (the initiator) doesn't have any entries that Bob isn't aware of and in 1.5 RTT's if Alice has entires that isn't present in Bob's routing table. The approach outlined in [this paper](https://www.ics.uci.edu/~eppstein/pubs/EppGooUye-SIGCOMM-11.pdf) (IBLT's plus a heuristic to estimate the size of the set difference) seems ripe for integrating into our system. 

In the past Rusty did a bit of research into searching for optimal parameters applicable to the previous use-case: synchronizing the mempool between Bitcoin full-nodes. I think much of that research can be re-used as we figure out the correct parameterization for our use-case. 

This can possibly be combined with the root state hash idea, to first compare the hashes of the root routing state, falling back to set reconciliation if the hashes don't match. The tree would be keyed by node pubkey, while the items inserted into the IBLT would be some digest of the channel authentication information. So...[why not both](http://imgur.com/gallery/c7NJRa2)?

07-routing-gossip.md
pm47: Btw I just realized that 0-conf anchors are not possible with ids based on block height 😁

BitfuryLightning:
@cdecker 
We have some comments on routing PR:

1. If node has a lot of channels than probably it closes/opens them
frequently. That process will need significant bandwith usage, since
message contains full RT. Sending diff is prefered in this case,
because it will decrease bandwith significantly.
2. Blockheight + transaction height is not unique in the case of
forks, and using these as id is not that stable as using transaction's
hash

So our propositions are:
1. Send messages about only newly created channels. There is proposal
in Bitfury's routing schema with ideas how to acheive that, you can
use them. If you come up with any specs how to implement incremental
changes - it would be very nice.
2. Not send messages about closed channels because this can be
determined from blockchain.
3. Node retransmit message/part if it contained information not known
to node before
4. Use transaction hash instead  pair blockheight + transaction height to make it more stable to blockchain forks.

What do you think?

rustyrussell:
@Roasbeef Good thinking, please open a wishlist PR to track it.  For the moment, let's just blast on opening.

@BitfuryLightning Yes, see my proposal, which was buried above, and contains three separate messages:

1. Channel announce. This contains both node ids, and the channel. Both node ids sign; you don't need to know anything about the node to validate and forward these. Conflicting (ie. different other than sigs) channel announces also get forwarded, and the channel and nodes blacklisted.
2. Node announce. Only valid after channel announce has identified the node.
3. Channel updates. These are one-sided. Only valid after channel announce.

Happy with txid in channel announce.  Happy to ignore closing for now, and assume everyone has UTXO set.  Agreed with rexmit of previously unknown messages as simplest solution. 

For the moment, just send entire state on first connection, otherwise changes since last connection.  That  will scale fine for the first few thousand nodes, while @Roasbeef specs up IBLT...

cdecker:
Good points by @BitfuryLightning, let me try to address them as I originally thought about them.

In the serialized format the channels are very light, i.e., 46 bytes each, so a single TCP/UDP packet can transport about 30 channels with the minimal MTU of 1500. Upon updates we do not forward the entire routing table, but we do forward all channels of a single node, so the update size is proportional to the updating node, not the size of the network.

Regarding the frequency, I agree that large hubs with many connections will have many updates, that can't be avoided, and this full dump of the updating node's state is wasteful, however the staggering serves to aggregate these updates along the propagation path, by replacing superseded updates in the send buffer of forwarding nodes. The staggering delays the announcement by 1 minute * network diameter, however we also get rid of DoS attacks and the flapping node states that'd result from too frequent updates.

Finally, it is relatively easy for nodes to be out of sync with most differential synchronization mechanisms, which would then either revert to a full dump or a more involved resynchronization mechanism, taking a number of roundtrips to resolve. I simply feel that sending the node's infos as a whole is the most robust way of dealing with this currently, and if we eventually upgrade to a sophisticated differential synchronization mechanism we shouldn't bother optimizing this first iteration at all, after all the nodes will be small for now, and we want to discourage large hubs anyway :-)

Re passive removal: I know I came up with it, but on further thought it's very ugly, sorry about that. It actually requires nodes to actively monitor the blockchain for spends of any anchor transaction they've seen so far, and it makes voluntary removals that do not result in the channel being torn down impossible. Take for example the case of a user that goes offline at night, he'll probably want to keep the channel established, and its peer is happy with that, but without voluntary removals they'd either still get forwarding requests or they'd have to close the channel.

Re Short-ID: The channel needs to be established and considered open by both endpoints for it to be announced, this already requires a number of confirmations, and by @pm47's suggestion I added a `min_conf` of 6 for channels to be forwarded at all. I feel that the short-id is relatively stable, and a reorg would lead to receiving nodes to have to look up the TX in the blockchain anyway, so a changing short-ID is not that bad. It's a nice little trick to reduce the size of the updates, once we go for the differential schemes, and cuts down on redundant information. But I'm also happy to drop it if a majority feel it introduces too much complexity :-)

The much more pressing matter I think is the signed unit of information that we try to synchronize on the nodes and not so much how we do the synchronization. We want to have the information authenticated by the origin, i.e., the announcing node, so we probably want to have that state signed by it. This means we have to come up with a format of the information that we agree on for the foreseeable future, since any synchronization mechanism will have to be able to rebuild that format from its local information in order to verify the signature. We probably want the signature to commit to the state of the node as well as its channels. So we can either sign the entire state with a single signature, or we can sign each incremental change, or we can sign node info and individual channels independently. My proposal is into the direction of the first option, it has very few signatures and everyone can verify that its local state is either in sync with what the node announced or not. This needs to be addressed now, and it will have an influence on how we build the remaining protocol, e.g., @Roasbeef's comment about the anchor pubkeys not being related to the node ID will require us to add more info in that serialization format. What do you guys think?

rustyrussell:
OK, I typed up my counter-proposal, stealing lots of content from here, see #11

cdecker:
Closing in favor of #11

PR #3, opened 2016-11-15
key generation: use standard blinding instead.

rustyrussell:
We use the counter as a blinding factor for payments to either side,
and the per-commitment-secret for the revocation keys.

We were also missing a place where we sent the first revocation key,
so fix that (has to be after initial open handshake, once we know their
base point).


rustyrussell:
@adiabat @Roasbeef

rustyrussell:
Turns out this is *too* simple, as it requires 33 bytes additional storage per tx for a watcher.  Closing.

PR #4, opened 2016-11-15
minor: only 2 bytes for open_channel.max-num-htlcs

pm47:
do we really want to have more than 65 536 oustanding htlcs?


cdecker:
I think the argument was that we want to have all the `num-XXX` fields to be 4 bytes, we even have some limit on outstanding HTLCs (300 IIRC), but this is for uniformity's sake :-)

rustyrussell:
> I think the argument was that we want to have all the `num-XXX` fields to be 4 bytes, we even have some limit on outstanding HTLCs (300 IIRC), but this is for uniformity's sake :-)

Yes.bytes.  Let me start a new spec on how to extend the protocol which
should explain this and much else that's assumed...

Thanks!
Rusty.

pm47:
Great! closing this one

PR #7, opened 2016-11-16
minor: fixed hierarchy and added a table of content

pm47:
I spotted some inconsistencies in naming between channel establishment and channel closing.

I'd like to do:
- `shutdown` -> `close_channel`
- `close_signature` -> `closing_signed`

what do you think?

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -11,8 +11,12 @@ The messages described in this document are grouped logically into 4 groups by t
  - Routing (types `256`-`511`): node and channel announcements, as well as any active route exploration.
 
 # Table of Contents
+=================
 
-FIXME
+  * [Channel](#channel)
+    * [Channel Establishment](#channel-establishment)
+    
+  
 
 # Channel
 
@@ -262,7 +266,7 @@ given in [BOLT #3].
 1. type: 37 (`MSG_UPDATE_FEE`)
 2. data:
    * [8:channel-id]
-   * [4:fee-per-kilobyte]
+   * [4:feerate-per-kilobyte]
 
 ### Requirements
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -11,9 +11,25 @@ The messages described in this document are grouped logically into 4 groups by t
  - Routing (types `256`-`511`): node and channel announcements, as well as any active route exploration.
 
 # Table of Contents
-
-FIXME
-
+  * [Channel](#channel)
```
Roasbeef: Nice addition with the table of contents! 👍

rustyrussell: Hmm, there's shutdown, which is where we don't add anything new, and closing, which is where we're negotiation the final close tx.  I used to call shutdown "clearing" but that has a specific finance meaning which was deeply sucky.

Ah, I see you reverted that AFAICT?  I'll squash and merge.

rustyrussell:
Squashed into three patches, and merged.  Thanks!

02-peer-protocol.md
pm47: > Ah, I see you reverted that AFAICT? I'll squash and merge.

Yeah, I didn't want to mix two separate topics in the same PR. I am just OCDing on `funding_signed` vs `close_signature`  😵

rustyrussell:
>
>
>
>> @@ -11,9 +11,25 @@ The messages described in this document are grouped logically into 4 groups by t
>   - Routing (types `256`-`511`): node and channel announcements, as well as any active route exploration.
>  
>  # Table of Contents
> -
> -FIXME
> -
> +  * [Channel](#channel)
>
>> Ah, I see you reverted that AFAICT? I'll squash and merge.
>
> Yeah, I didn't want to mix two separate topics in the same PR. I am just OCDing on `funding_signed` vs `close_signature`  😵 

Agreed.  Pick one to change and submit another PR :)

Thanks!
Rusty.

Issue #2, opened 2016-11-15
Pick a licence for the specs


josephpoon:
CC-BY?
GNU FDL?
WTFPL?

Roasbeef:
My vote is for the [WTFPL](http://www.wtfpl.net/) 🙌🏾. 

![WTFPL](http://www.wtfpl.net/wp-content/uploads/2012/12/wtfpl-strip.jpg)

cdecker:
I quite like CC-BY but I have no idea of the implications :-)

rustyrussell:
Joseph Poon <notifications@github.com> writes:
> CC-BY?
> GNU FDL?
> WTFPL?

CC-BY would be my default guess, but I'll ask People Who Know This Stuff
and see if there's some reason to prefer something else here.

Cheers,
Rusty.

harrycheng8:
Duplicate of #

PR #9, opened 2016-11-19
Reject HTLC if amount is below dust-limit-satoshis.

AndrewSamokhvalov:
Hi all! At the Lightning Summit in Milan you agreed on a mechanism to allow sub-dust `HTLC's` on the network. But one of the drawbacks of such decision might be the case when attacker (`Alice #1`) sends a lots of dust `HTLCs` over some intermediary node (`Carol`) to himself (`Alice #2`). In this case if `dust-limit-satoshis` in channel `Alice #1 => Carol` lower than `dust-limit-satoshis` in channel `Carol => Alice #2`, attacker might turn all `bitcoins` in channel `Carol => Alice #2` to miners fee without any consequences for himself. 

In this case we may add additional field `max_num_dust_htlcs` and after exceeding this number we still should reject sub-dust `HTLCs`. So, for me it looks like it would be better to just reject `HTLCs` below dust limit. 

Is accepting dust `HTLCs` is something that end user really needs in first version? I mean as far as I know dust limit in `Bitcoin` is `5460 sat` which is `5460 sat / 10^8 sat * 746$ = 0,041$` which is pretty low amount. Sorry if that was already discussed :)

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -413,8 +413,7 @@ estimate of cost of inclusion in a block.
 
 The sender MUST set `signature` to the bitcoin signature of the close
 transaction with the node responsible for paying the bitcoin fee
-paying `fee-satoshis`, without populating any output which is below
-the receiver's `dust-limit-satoshis`.
+paying `fee-satoshis`.
 
 The receiver MUST check `signature` is valid for the close transaction
 with the given `fee-satoshis` as detailed above, and MUST fail the
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -32,7 +32,7 @@ commitment transaction.
 
 ### Commitment Transaction Outputs
 
-The amounts for each output are rounded down to whole satoshis.  If this amount is less than the `dust-limit-satoshis` set by the owner of the commitment transaction, the output is not produced (thus the funds add to fees).
+The amounts for each output are rounded down to whole satoshis.  If this amount is less than the `dust-limit-satoshis` set by the owner of the commitment transaction, the `HTLC` SHOULD be rejected.
 
 To allow an opportunity for penalty transactions in case of a revoked commitment transaction, all outputs which return funds to the owner of the commitment transaction (aka "local node") must be delayed for `to-self-delay` blocks.  This delay is done in a second stage HTLC transaction.
 


cdecker:
IIRC this we punted this for a later version. We have a local policy on the node, not to accept any new HTLCs if there are many active HTLCs, this limits the scope of the attack significantly, since we now can only have a `dustlevel * num_max_htlcs` satoshis locked in (still bad enough).

Eventually we probably want to go with a probabilistic scheme, that is with a `amount/dustlevel` chance we actually require the sender to commit to `dustlevel` satoshis instead of the actual amount, thus creating the output. If we repeat this many times then the effect is the same as actually committing to the sub-dust HTLCs, thus the attack is no longer free. We simply haven't come up with a scheme to do that in a fair way :-)

rustyrussell:
You can just set htlc-minimum-msat to match dust-limit-satoshis is you don't want any dust HTLCs?

If may make sense to add a total-dust-limit parameter, but I think we can probably punt that for now; we've got a lot of parameters already....

AndrewSamokhvalov:
@cdecker , @rustyrussell ok, thanks!

Roasbeef:
The current dust-limit widely used on the network is now `546 satoshis`, at the current exchange rate (`$738`), that comes out to `$0.004` or `4/10 th` of 1 cent. With segwit, the dust-limit changes slightly since data within witnesses gets a slight discount, applying this discount, we arrive at a dust-limit of `294 satoshis` or `$0.002` putting the HTLC floor at `2/10 th` of a cent.

rustyrussell:
> The current dust-limit widely used on the network is now `546 satoshis`, at the current exchange rate (`$738`), that comes out to `$0.004` or `4/10 th` of 1 cent. With segwit, the dust-limit changes slightly since data within witnesses gets a slight discount, applying this discount, we arrive at a dust-limit of `294 satoshis` or `$0.002` putting the HTLC floor at `2/10 th` of a cent. 

In practical terms, each HTLC output adds 172 to the weight (not even
considering the cost to actually spend it).  At a fee rate of 0.00056159
that costs 2414 satoshi in fees.

We should be trimming according to the fee rate, but that can wait until
v1.1 I think.  Needs careful consideration about limiting the number of
such dust HTLCs.

Cheers,
Rusty.

PR #11, opened 2016-11-21
BOLT #7: More complex proposal, using three separate message types.

rustyrussell:
Contents stolen from @cdecker 's draft, but tightened requirements and split into three messages.

I added bitcoin-to-node cross-signing, require both sigs for announcing a channel, expanded expiry from 1 to two bytes to match BOLT #2.  I think the result is pretty simple, though I can see us later switching to schnorr and having a bulk-update message.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -23,17 +23,17 @@ its fee levels and expiry using `channel_update`.
 
 1. type: 256 (`MSG_CHANNEL_ANNOUNCEMENT`)
 2. data:
-	* [4:blockheight]
-	* [3:blockindex]
-	* [1:outputindex]
-	* [33:node-id-1]
-	* [33:node-id-2]
-	* [33:bitcoin-key-1]
-	* [33:bitcoin-key-2]
-	* [64:bitcoin-signature-1]
-	* [64:bitcoin-signature-2]
-	* [64:node-signature-1]
-	* [64:node-signature-2]
+    * [4:blockheight]
+    * [3:blockindex]
+    * [1:outputindex]
+    * [33:node-id-1]
+    * [33:node-id-2]
+    * [33:bitcoin-key-1]
+    * [33:bitcoin-key-2]
+    * [64:bitcoin-signature-1]
+    * [64:bitcoin-signature-2]
+    * [64:node-signature-1]
+    * [64:node-signature-2]
 
 ### Requirements
 
@@ -51,14 +51,14 @@ ascending numerical order, and MUST set `bitcoin-key-1` and
 respectively.
 
 The creating node MUST set `bitcoin-signature-1` to the signature of
-the double-SHA of `node-id-1` using `bitcoin-key-1`, and set
-`bitcoin-signature-2` to the signature of the double-SHA of
+the double-SHA256 of `node-id-1` using `bitcoin-key-1`, and set
+`bitcoin-signature-2` to the signature of the double-SHA256 of
 `node-id-2` using `bitcoin-key-2`
 
 The creating node MUST set `node-signature-1` to the signature of the
-double-SHA of every data field up to and including
+double-SHA256 of every data field up to and including
 `bitcoin-signature-2` using `node-id-1`, and `node-signature-2` to the
-signature of the double-SHA of every data field up to and including
+signature of the double-SHA256 of every data field up to and including
 `bitcoin-signature-2` using `node-id-2`.
 
 The receiving node MUST ignore the message if the output `outputindex`
@@ -78,9 +78,9 @@ Otherwise, if the transaction referred to was not previously announced
 as a channel, the receiving node SHOULD queue the message for
 rebroadcasting.  If it has previously received a valid
 `channel_announcement` for the same transaction in the same block, but
-different `node-id-1` or `node-id-2`, it SHOULD blacklist the the
+different `node-id-1` or `node-id-2`, it SHOULD blacklist the
 previous message's `node-id-1` and `node-id-2` as well as this
-`node-id-1` and `node-id-2` and forget channels conntected to them,
+`node-id-1` and `node-id-2` and forget channels connected to them,
 otherwise it SHOULD store this `channel_announcement`.
 
 The receiving node SHOULD forget a channel once its funding output has
@@ -115,8 +115,7 @@ nodes for which a channel is not already known are ignored.
    * [33:node-id]
    * [3:rgb-color]
    * [2:pad]
-   * [4:len]
-   * [len:alias]
+   * [32:alias]
 
 The `timestamp` allows ordering in the case of multiple announcements;
 the `ipv6` and `port` allow the node to announce its willingness to
@@ -130,21 +129,19 @@ The creating node MUST set `timestamp` to be greater than any previous
 `node_announcement` it has created.  It MAY base it on a UNIX
 timestamp.  It MUST set the `ipv6` and `port` fields to all zeroes, or
 a non-zero `port` and `ipv6` set to a valid IPv6 address or an IPv4-Mapped IPv6 Address format as defined in [RFC 4291 section 2.5.5.2](https://tools.ietf.org/html/rfc4291#section-2.5.5.2).  It MUST set `signature` to the signature of
-the double-SHA of the entire remaining packet after `signature` using the
-key given by `node-id`.  It MUST set `pad` to zero.  It MAY set `alias` and `rgb-color` to customize their node's appearance in maps and graphs, where the first byte of `rgb` is the red value, the second byte is the green value and the last byte is the blue value.  It MUST set `alias` to a valid UTF-8 string.
-
-The receiving node SHOULD fail the connection if the message is of
-insufficent length to contain `alias`.
+the double-SHA256 of the entire remaining packet after `signature` using the
+key given by `node-id`.  It MUST set `pad` to zero.  It MAY set `alias` and `rgb-color` to customize their node's appearance in maps and graphs, where the first byte of `rgb` is the red value, the second byte is the green value and the last byte is the blue value.  It MUST set `alias` to a valid UTF-8 string of up to 21 bytes in length, with all `alias` bytes following equal to zero.
 
 The receiving node SHOULD fail the connection if `signature` is
 invalid or incorrect for the entire message including unknown fields
 following `alias`, and MUST NOT further process the message.  The
-receiving node SHOULD ignore `ipv6` if `port` is zero.  It MUST ignore
+receiving node SHOULD ignore `ipv6` if `port` is zero.  It SHOULD fail
+the connection if the final byte of `alias` is not zero.  It MUST ignore
 the contents of `pad`.
 
 The receiving node SHOULD ignore the message if `node-id` is not
 previously known from a `channel_announcement` message, or if
-`timestamp` is not greater than than the last-received
+`timestamp` is not greater than the last-received
 `node_announcement` from this `node-id`.  Otherwise, if the
 `timestamp` is greater than the last-received `node_announcement` from
 this `node-id` the receiving node SHOULD queue the message for
@@ -173,22 +170,22 @@ it wants to change fees.
 
 1. type: 258 (`MSG_CHANNEL_UPDATE`)
 2. data:
-	* [64:signature]
-	* [4:blockheight]
-	* [3:blockindex]
-	* [1:outputindex]
-	* [4:timestamp]
-	* [1:side]
-	* [1:pad]
+    * [64:signature]
+    * [4:blockheight]
+    * [3:blockindex]
+    * [1:outputindex]
+    * [4:timestamp]
+    * [1:side]
+    * [1:pad]
     * [2:expiry]
-	* [4:htlc-minimum-msat]
-	* [4:fee-base-msat]
-	* [4:fee-proportional-millionths]
+    * [4:htlc-minimum-msat]
+    * [4:fee-base-msat]
+    * [4:fee-proportional-millionths]
 
 ### Requirements
 
 The creating node MUST set `signature` to the signature of the
-double-SHA of the entire remaining packet after `signature` using its own `node-id`.
+double-SHA256 of the entire remaining packet after `signature` using its own `node-id`.
 
 The creating node MUST set `blockheight`, `blockindex` and `outputindex` to
 match those in the already-sent `channel_announcement` message, and MUST set `side` to 0 if the creating node is `node-id-1` in that message, otherwise 1.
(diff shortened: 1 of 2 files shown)


07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+`node-signature-2` are invalid or not correct.
+
+Otherwise, if `node-id-1` or `node-id-2` are blacklisted, it SHOULD
+ignore the message.
+
+Otherwise, if the transaction referred to was not previously announced
+as a channel, the receiving node SHOULD queue the message for
+rebroadcasting.  If it has previously received a valid
+`channel_announcement` for the same transaction in the same block, but
+different `node-id-1` or `node-id-2`, it SHOULD blacklist the the
```
pm47: > the the

typo

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+previous message's `node-id-1` and `node-id-2` as well as this
+`node-id-1` and `node-id-2` and forget channels conntected to them,
```
pm47: > conntected 

typo

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+1. type: 257 (`MSG_NODE_ANNOUNCEMENT`)
+2. data:
+   * [64:signature]
+   * [4:timestamp]
+   * [16:ipv6]
+   * [2:port]
+   * [33:node-id]
+   * [3:rgb-color]
+   * [2:pad]
+   * [4:len]
```
pm47: We should probably enforce a max value on `len`. FWIW I'd rather put a fixed size (e.g. 32B) null-terminated string.

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+
+The creating node MUST set `timestamp` to be greater than any previous
+`node_announcement` it has created.  It MAY base it on a UNIX
+timestamp.  It MUST set the `ipv6` and `port` fields to all zeroes, or
+a non-zero `port` and `ipv6` set to a valid IPv6 address or an IPv4-Mapped IPv6 Address format as defined in [RFC 4291 section 2.5.5.2](https://tools.ietf.org/html/rfc4291#section-2.5.5.2).  It MUST set `signature` to the signature of
+the double-SHA of the entire remaining packet after `signature` using the
+key given by `node-id`.  It MUST set `pad` to zero.  It MAY set `alias` and `rgb-color` to customize their node's appearance in maps and graphs, where the first byte of `rgb` is the red value, the second byte is the green value and the last byte is the blue value.  It MUST set `alias` to a valid UTF-8 string.
+
+The receiving node SHOULD fail the connection if the message is of
+insufficent length to contain `alias`.
```
pm47: > insufficent

typo

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+1. type: 258 (`MSG_CHANNEL_UPDATE`)
+2. data:
+	* [64:signature]
+	* [4:blockheight]
+	* [3:blockindex]
+	* [1:outputindex]
+	* [4:timestamp]
+	* [1:side]
+	* [1:pad]
+    * [2:expiry]
```
pm47: >     * [2:expiry]

leading spaces

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+
+The receiving node SHOULD fail the connection if `signature` is
+invalid or incorrect for the entire message including unknown fields
+following `alias`, and MUST NOT further process the message.  The
+receiving node SHOULD ignore `ipv6` if `port` is zero.  It MUST ignore
+the contents of `pad`.
+
+The receiving node SHOULD ignore the message if `node-id` is not
+previously known from a `channel_announcement` message, or if
+`timestamp` is not greater than than the last-received
```
pm47: > than than

typo

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+is not greater than than the last-received `channel_announcement` for
+this channel and `side`.  Otherwise, if the `timestamp` is equal to
+the last-received `channel_announcement` and the fields other than
+`signature` differ, the node MAY blacklist this node-id and forget all
+channels associated with it.  Otherwise the receiving node SHOULD
+queue the message for rebroadcasting.
+
+## Rebroadcasting
+
+Nodes receiving an announcement verify that this is the latest update from the announcing node by comparing the included timestamp and update their local view of the network's topology accordingly.
```
pm47: I think you mean that the receiving node should keep the latest update per message type and per channel right ?

Which means for announcing node A:
- latest `node_announcement` message with `node-id`=A if any
- all `channel_announcement` messages with `node-id-1`=A OR `node-id-2`=A not previously broadcast
- latest `channel_update` for each unique  `blockheight` \ `blockindex` \ `outputindex`

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+   * [4:len]
+   * [len:alias]
+
+The `timestamp` allows ordering in the case of multiple announcements;
+the `ipv6` and `port` allow the node to announce its willingness to
+accept incoming network connections, the `rgb-color` and `alias` allow
+intelligence services to give their nodes cool monikers like IRATEMONK
+and WISTFULTOLL and use the color black.
+
+### Requirements
```
pm47: Should a `node_announcement` message be sent after every new `channel_announcement`, or once after the first `channel_announcement`?

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+Otherwise the receiving node updates the metadata and stores the signature associated with the announcement.
+This will later allow the receiving node to rebuild the announcement for its peers.
+
+After processing the announcement the receiving node adds the announcement to a list of outgoing announcements.
+
+### Requirements
+
+Each node SHOULD flush outgoing announcements once every 60 seconds, independently of the arrival times of announcements, resulting in a staggered announcement and deduplication of announcements.
+
+Nodes MAY re-announce their channels regularly, however this is discouraged in order to keep the resource requirements low.
```
pm47: Should we add some kind of blacklisting based on a max number of `node_announcement` or `channel_update` that a peer is allowed to send per second, considering that those messages are costless to produce?

cdecker:
Very nice proposal, it's rather similar to what I had in mind initially, having the individual classes of objects created and updated in separate messages. The reason I moved away from that idea was that it introduces  interdependence between messages, i.e., a node announcement is only valid if it is preceded by a channel announcement, and the same goes for the channel update.

Having the node ID sign the pubkey used in the anchor transaction is very nice because it addresses @Roasbeef 's usecase in which every anchor transaction may use completely separate keys. On the other hand it blows up the message size, with the delegation signatures, the pubkeys and the channel signatures. For comparison in order to fully announce a channel and the corresponding node, i.e., `MSG_CHANNEL_ANNOUNCEMENT` + `MSG_NODE_ANNOUNCEMENT` + `MSG_CHANNEL_UPDATE`, with no alias, this results in 616 bytes, while the scheme in #1 requires just 176 bytes, but admittedly lacks the delegation, so this comparison may not be fair :-)

It's not just that we need to transfer all of the information around, but that is also the minimal set that we need to keep in memory at each node, since we might need to bootstrap another node and provide it with the signed state we received. This is also why I dislike having padding in the messages, since we now need to either define that to be null or keep it in memory since it is signed.

As mentioned in #1 I dislike the passive closure, which forces all nodes to track all anchor transactions, in order to be able to cleanup closed channels, and it does not allow for voluntary disabling of channels. But we can simply redefine the `pad` byte as an `option` bitfield and use one bit to signal removal :-)

Another difference is that this proposal signs channels independently, which results in more signatures to send and store, but allows a per-channel sync, instead of the per-node sync in #1. This is important when looking at the extensibility of the routing system, since it is a different signed unit of information that we need to recreate when syncing and verifying. I'm not sure which one is nicer.

07-routing-gossip.md
```diff
@@ -0,0 +1,252 @@
...
+The creating node MUST set `blockheight`, `blockindex` and `outputindex` to
+match those in the already-sent `channel_announcement` message, and MUST set `side` to 0 if the creating node is `node-id-1` in that message, otherwise 1.
+
+The creating node MUST set `timestamp` to greater than zero, and MUST set it to greater than any previously-sent `channel_update` for this channel, and MUST set `pad` to zero.
+
+It MUST set `expiry` to the number of blocks it will subtract from an incoming HTLC's `expiry`.  It MUST set `htlc-minimum-msat` to the minimum HTLC value it will accept, in millisatoshi.  It MUST set `fee-base-msat` to the base fee it will charge for any HTLC, in millisatoshi, and `fee-proportional-millionths` to the amount it will charge per millionth of a satoshi.
+
+The node SHOULD accept HTLCs which pay a fee equal or greater than:
+
+	fee-base-msat + htlc-amount-msat * fee-proportional-millionths / 1000000
```
cdecker: Right, but that is the node's local policy isn't it? Either he has a grace period, or he is rejecting things that'd make him some fees. So I'm not sure we need to specify this here.

07-routing-gossip.md
cdecker: We have the dependence between the `channel_announcement` and the `node_announcement`, i.e., a `node_announcement` that is not preceded by a `channel_announcement` is invalid and discarded. So we don't forward these free messages, which is the real danger here. If our peer wants to spam us with a `node_announcements` that we'll just drop, that's fine, and can be handled more generically, e.g., close a channel if our peer sends us more than 5kb/s, independently of the message type.

07-routing-gossip.md
pm47: > you can get away with just the channel_announcement + channel_update if you don't care about nodes

Right! I had totally missed that

rustyrussell:
>
> I like this proposal better, the separation between `channel_announcement` and `channel_update` is very similar to a prototype we built into eclair a few weeks ago.
>
> I also like the fact that we rely on the blockchain to find out which channels are closed instead of announcing both open and close, which was a little cumbersome in PR #1 (and it can still be added).
>
> On the downside, I think this design is more vulnerable to DoS attacks because messages `node_announcement` and `channel_update` are costless to produce once a single valid channel has been created. Maybe we could add some rate limiting logic to mitigate this?

I think stepped propagation automatically controls this for the moment,
doesn't it?  You can only propagate one every 60 seconds.

> +The node SHOULD accept HTLCs which pay a fee equal or greater than:
> +
> +	fee-base-msat + htlc-amount-msat * fee-proportional-millionths / 1000000
>
> I think there need to be a tolerance on the fee, because we need to take into account the time it takes for the information to propagate. Otherwise a node will systematically reject payments received right after it increased its relaying fee.

Yes, and that's why it's a should.

I needed to stick that formula somewhere in the doc, but wasn't sure
where.

I've moved it out to a separate section:

## HTLC Fees

The node creating `channel_update` SHOULD accept HTLCs which pay a fee equal or greater than:

	fee-base-msat + htlc-amount-msat * fee-proportional-millionths / 1000000

The node creating `channel_update` SHOULD accept HTLCs which pay an
older fee for some time after sending `channel_update` to allow for
propagation delay.

> +The creating node MUST set `bitcoin-signature-1` to the signature of
> +the double-SHA of `node-id-1` using `bitcoin-key-1`, and set
>
> Should we specify what kind of SHA?

Thanks,  I s/SHA/SHA256/ everywhere.

>> the the
>
> typo
..
>> conntected 
>
> typo

Both fixed, thanks.

>> +## The `node_announcement` message
> +
> +This allows a node to indicate extra data associated with it, in
> +addition to its public key.  To avoid trivial denial of service attacks,
> +nodes for which a channel is not already known are ignored.
> +
> +1. type: 257 (`MSG_NODE_ANNOUNCEMENT`)
> +2. data:
> +   * [64:signature]
> +   * [4:timestamp]
> +   * [16:ipv6]
> +   * [2:port]
> +   * [33:node-id]
> +   * [3:rgb-color]
> +   * [2:pad]
> +   * [4:len]
>
> We should probably enforce a max value on `len`. FWIW I'd rather put a fixed size (e.g. 32B) null-terminated string.

Good idea.  The longest NSA codename I could find is "EVOLVED MUTANT BROTH"
which fits comfortably in 32 bytes.

>> insufficent
>
> typo

Removed, due to fixed len.

>> +After a channel has been initially announced, each side independently
> +announces its fees and minimum expiry for HTLCs.  It uses the 8-byte
> +channel shortid which matches the `channel_announcement` and one byte
> +to indicate which end this is.  It can do this multiple times, if
> +it wants to change fees.
> +
> +1. type: 258 (`MSG_CHANNEL_UPDATE`)
> +2. data:
> +	* [64:signature]
> +	* [4:blockheight]
> +	* [3:blockindex]
> +	* [1:outputindex]
> +	* [4:timestamp]
> +	* [1:side]
> +	* [1:pad]
> +    * [2:expiry]
>
>>     * [2:expiry]
>
> leading spaces

Tabs v. spaces.  Replaced with spaces everywhere.

>> than than
>
> typo

Fixed.

> +## Rebroadcasting
> +
> +Nodes receiving an announcement verify that this is the latest update from the announcing node by comparing the included timestamp and update their local view of the network's topology accordingly.
>
> I think you mean that the receiving node should keep the latest update per message type and per channel right ?

Yes, I cut and paste from @cdecker overzealously here.

> Which means for announcing node A:
> - latest `node_announcement` message with `node-id`=A if any
> - all `channel_announcement` messages with `node-id-1`=A OR `node-id-2`=A not previously broadcast
> - latest `channel_update` for each unique  `blockheight` \ `blockindex` \ `outputindex`

How about:

        Nodes receiving a new `channel_announcement` or a
        `channel_update` or `node_update` with an updated timestamp
        update their local view of the network's topology accordingly.

        Once the announcement has been processed it is added to a list
        of outgoing announcements (perhaps replacing older updates) to
        the processing node's peers, which will be flushed at regular
        intervals.  This store and delayed forward broadcast is called a
        _staggered broadcast_

> +The `timestamp` allows ordering in the case of multiple announcements;
> +the `ipv6` and `port` allow the node to announce its willingness to
> +accept incoming network connections, the `rgb-color` and `alias` allow
> +intelligence services to give their nodes cool monikers like IRATEMONK
> +and WISTFULTOLL and use the color black.
> +
> +### Requirements
>
> Should a `node_announcement` message be sent after every new `channel_announcement`, or once after the first `channel_announcement`?

It only needs to be sent once, unless details change.  I thought about
banning changing of monikers, but that would be less fun.

>> +Nodes receiving an announcement verify that this is the latest update from the announcing node by comparing the included timestamp and update their local view of the network's topology accordingly.
> +
> +Once the announcement has been processed it is added to a list of outgoing announcements to the processing node's peers, which will be flushed at regular intervals.
> +This store and delayed forward broadcast is called a _staggered broadcast_
> +
> +If, after applying the changes from the announcement, there are no channels associated with the announcing node, then the receiving node MAY purge the announcing node from the set of known nodes.
> +Otherwise the receiving node updates the metadata and stores the signature associated with the announcement.
> +This will later allow the receiving node to rebuild the announcement for its peers.
> +
> +After processing the announcement the receiving node adds the announcement to a list of outgoing announcements.
> +
> +### Requirements
> +
> +Each node SHOULD flush outgoing announcements once every 60 seconds, independently of the arrival times of announcements, resulting in a staggered announcement and deduplication of announcements.
> +
> +Nodes MAY re-announce their channels regularly, however this is discouraged in order to keep the resource requirements low.
>
> Should we add some kind of blacklisting based on a max number of `node_announcement` or `channel_update` that a peer is allowed to send per second, considering that those messages are costless to produce?

They only waste local bandwidth, since we'll only remember the last one.

Eventually we may want a more realistic limit, perhaps based on
N-updates-per-block which is simple to agree on.  Meanwhile, staggered
broadcast limits us to one per 60 seconds.

Cheers,
Rusty.

rustyrussell:
> +The node SHOULD accept HTLCs which pay a fee equal or greater than:
> +
> +	fee-base-msat + htlc-amount-msat * fee-proportional-millionths / 1000000
>
> Right, but that is the node's local policy isn't it? Either he has a grace period, or he is rejecting things that'd make him some fees. So I'm not sure we need to specify this here.

Well, the formula needs to go somewhere.  I moved it to its own section.

Cheers,
Rusty.

BitfuryLightning:
Very nice proposal. I like this proposal better than the previous one because it doesn't send duplicating info each time.

PR #13, opened 2016-11-22
BOLT #1: make ERROR data field fixed length.

rustyrussell:
So far, it's the only variable-length field we have in the protocol, so that weighs me in the direction of simply nailing it.

Although a protocol violation, you could send a longer message: all nodes will ignore the extra bytes unless we ever extend the MSG_ERROR definition to add a field, which I can't see happening.


rustyrussell:
No, actually the shutdown outputscript is also variable length, so not worth it.

PR #14, opened 2016-11-22
tools/extract-formats.py: produce structure definitions and check alignment

rustyrussell:
You made me write python!  Shame on you!

This doesn't enforce that total message length meet alignment constraints!  C implementers beware that they will have junk at the end of their structures, since we have some odd lengths.

### DIFF
--- a/01-messaging-crypto-and-init.md
+++ b/01-messaging-crypto-and-init.md
@@ -682,7 +682,7 @@ features supported or required by this node.  Odd features are
 optional, even features are compulsory ("it's OK to be odd!").  The
 meaning of these bits will be defined in future.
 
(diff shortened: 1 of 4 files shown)


rustyrussell:
Rebased on top of #19  and made it output CSV, so it's completely language neutral.

cdecker:
LGTM :+1:

Roasbeef:
It looks like this might need to be rebased? Or maybe some other PR's were combined into this one? 

An LGTM from me once that's cleared up. I don't think I personally will make use of the attached script, but I can certainly see how it can be super useful for other implementers to possibly write some auto-code generation for their structs/objects and encode/decode methods.

PR #15, opened 2016-11-22
'to-local output': add revocation key

sstone:
I may have missed something, but I don't see how the 'to-local' output can be spent by a penalty tx when a revoked commit tx is published: it seems that its script should be the same as the htlc timeout/htc success tx output script ?


rustyrussell:
Good catch.  I now need to make some minor fixes to size calculations I think...

PR #21, opened 2016-11-23
BOLT 3: fix witness scripts

sstone:
Add a short paragraph on P2WSH (the witness script is omitted for brevity in all descriptions) and fix witness scripts for spending the different tx outputs.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -44,7 +44,7 @@ The amounts for each output are rounded down to whole satoshis.  If this amount,
 
 #### To-Local Output
 
-This output sends funds back to the owner of this commitment transaction, thus must be timelocked using OP_CSV. If can be claimed, without delay, by the other party if they know the revocation key. The output is a version 0 P2WSH, with a witness script:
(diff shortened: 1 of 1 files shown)


Roasbeef:
What's the rationale behind moving from CMSV to CMS? Unintended clean-stack interaction perhaps?

rustyrussell:
Thanks for the review!

Tadge caught my VERIFY error before and I didn't fix these.  And the witnesses were from an older variant.  As you can tell, they're still untested...

rustyrussell:
@sstone Hmm, want to review my math on fee calculation as well?  See #22 

Similarly, the Penalty Transaction Weight Calculation in BOLT 5.

rustyrussell:
Squashed and merged, with slightly more detailed commit msg.  Thanks!

sstone:
@Roasbeef yes, you need a final stack with a single "1" and since CMSV cleans the stack this "1" had to be added to the witness script of the spending tx, it is cleaner with CMS

@rustyrussell I'm reviewing the fee calculation and will create a PR if needed

PR #23, opened 2016-11-24
BOLT 2: clarify what feerate-per-kw stands for and how it's calculated.

rustyrussell:
Bitcoin still uses feerate-per-kb, which needs to be multiplied by 4
in segwit.  It's more than a little confusing.

Split the paragraphs, too, as it's a little overwhelming.

@Roasbeef


rustyrussell:
Rebased.

PR #24, opened 2016-11-24
[RFC] BOLT 1, BOLT 2, BOLT 5: 2-byte lengths everywhere.

rustyrussell:
Since our cryptopacket limits us to 2 bytes, and since people will send 1-message-per-crypto-packet and nobody will test the multiple-messages-in-one-cryptopacket code, let's just restrict to 64k messages.

1. Make cryptopacket length not include the HMAC, so we can actually send
   64k messages.
2. Remove len prefix from packet, add padding to restore alignment.
3. Change message internal lengths/counts from 4 to 2 bytes, since more
   is nonsensical anyway, and this removes a need to check before allocating:
    - init feature bitfield length
    - error message length
    - shutdown scriptpubkey length
    - commit_sig number of HTLC signatures
    - revoke_and_ack number of HTLC-timeout signatures
4. Also change max-num-htlcs to a single byte, which ensures that commit_sig
   will always be under 64k (max around 32k).

### DIFF
--- a/01-messaging-crypto-and-init.md
+++ b/01-messaging-crypto-and-init.md
@@ -18,26 +18,22 @@ the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
 
 ## Message Format and Handling
 
-All messages are of form:
+All protocol messages are of the form:
 
-1. `4-byte` big-endian data length.
-2. `4-byte` big-endian type.
-3. Data bytes as specified by the length.
+1. `2-byte` big-endian type.
+3. Data bytes as specified by the total packet length.
 
 All data fields are big-endian unless otherwise specified.
 
 ### Requirements
 
-A node MUST NOT send a message with data length greater than `8388608`
+A node MUST NOT send a message with more than `65529` data
 bytes.  A node MUST NOT send an evenly-typed message not listed here
-without prior negotiation.
-
-A node MUST disconnect if it receives a message with data length
-greater than `8388608` bytes; it MUST NOT fail the channels in that case.
+without prior negotiation.  A node MUST set the padding to zeroes.
 
 A node MUST ignore a received message of unknown type, if that type is
 odd.  A node MUST fail the channels if it receives a message of unknown
-type, if that type is even.
+type, if that type is even.  A node MUST ignore the padding.
 
 A node MUST ignore any additional data within a message, beyond the
 length it expects for that type.
@@ -48,13 +44,19 @@ The standard endian of `SHA2` and the encoding of bitcoin public keys
 are big endian, thus it would be unusual to use a different endian for
 other fields.
 
-Length is limited to avoid memory exhaustion attacks, yet still allow
-(for example) an entire bitcoin block to be comfortable forwarded as a
-reasonable upper limit.
+Length is limited by the cryptographic wrapping, and messages are never
+more than 65535 bytes anyway.  That message has a 2 byte length, so
+the data bytes here are aligned if decrypted in-place.
 
 The "it's OK to be odd" rule allows for future optional extensions
 without negotiation or special coding in clients.  The "ignore
 additional data" rule similarly allows for future expansion.
+
+Implementations may prefer have message data aligned on an 8-byte
+boundary (the largest natural alignment requirement of any type here),
+but adding a 6-byte padding after the type field was considered
+wasteful: alignment may be achieved by decrypting the message into
+a buffer with 6 bytes of pre-padding.
 
 ## Cryptographic Messaging Overview
 
@@ -138,13 +140,11 @@ prefix thereby causing a node to erroneously read an incorrect number of bytes.
 ## Protocol Message Encapsulation
 
 Once both sides have entered the transport message exchange phase (after a
-successful completion of the handshake), Lightning Network protocol messages
-will be encapsulated within the exchanged `AEAD` ciphertexts. The maximum size
-of transport messages is `65535-bytes`. Node MUST NOT send a transport message
-which exceeds this size. Note that this is only a cryptographic messaging limit
-within the protocol, and not a limit on the message size of Lightning Network
-protocol messages. A Lightning Network message which exceeds this size can be
-chunked into several messages before being sent.
+successful completion of the handshake), each Lightning Network protocol message
+will be encapsulated within a single `AEAD` ciphertext. The maximum size
+of these transport messages is `65535-bytes`, so the largest message data
+possible is 65529 bytes.  If larger messages are needed in future, a
+fragmentation method will be defined.
 
 ### Noise Protocol Instantiation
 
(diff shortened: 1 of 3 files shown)


rustyrussell:
So I realized riding home yesterday that "max-num-htlcs" is really "max-num-accepted-htlcs" (you can't control the total, since that's racy).  Which is why commitsig.num-htlcs can be up to 510, hence 2 bytes.  revoke_and_ack.num-htlcs can be 1 byte, but for as the convention for "[num] [num*items]" has [num] as 2 bytes anyway, I think we should just make that 2 bytes.

01-messaging-crypto-and-init.md
```diff
@@ -20,24 +20,21 @@ the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
 
 All messages are of form:
 
-1. `4-byte` big-endian data length.
-2. `4-byte` big-endian type.
-3. Data bytes as specified by the length.
+1. `2-byte` big-endian type.
```
cdecker: So we are moving the length of the packet into the message itself? I prefer having a framing mechanism that takes care of both the `type` and the `length`, which also means we can actually reuse the two bytes wasted on padding for length.

01-messaging-crypto-and-init.md
```diff
@@ -20,24 +20,21 @@ the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
...
+2. 4 bytes of padding.
```
cdecker: This does no longer align to a multiple of 4.

01-messaging-crypto-and-init.md
```diff
@@ -590,21 +578,25 @@ creating a decryption oracle.
...
+|                                  |
+|                                  |
++----------------------------------
+|         16-byte MAC of the       |
+|            ciphertext            |
++----------------------------------
 ```
 
-The prefixed packet lengths are encoded as a `2-byte` big-endian integer.
+The prefixed packet lengths are encoded as a `2-byte` big-endian integer,
```
cdecker: As mentioned above I'd prefer having the length in the message framing, this also means moving the MAC of the length into the framing.

01-messaging-crypto-and-init.md
cdecker: The reason I'm bringing this up is that I always stumble over the mention of message length, which has not been defined yet.

Roasbeef:
Nice, excluding the MAC from the packet size limit makes sense. Also as you pointed out on IRC, it lets us avoid some possibly implementation quirks/bugs related to underflowing the integer due to an arithmetic error. 

However, I'm against restricting `max-num-htlcs/max-accepted-htlcs` (they're currently inconsistent within the doc) to only a single byte, further restricting he _total_ number of pending HTLC's within a commitment transaction to a hard cap of `512`. Instead, I think the integer for encoding the max `max-num-htlcs` paramter, and the integer within the HTLC add message should be of length `2-bytes`. 

In my opinion we shouldn't arbitrarily restrict the number of HTLC's in flight due to an affinity for small numbers. In contrast, the _hard_ limits (protocol enforced limits) should instead be based on the raw constraints of the layer below us, meaning Bitcoin's (widely used policy) weight limits on transactions.

Factoring the weight of the base commitment transaction, and the weight of each HTLC, this number is ~2k HTLCs. Some work has been done by @AndrewSamokhvalov to create accurate estimates which we can use to pin-point an exact number. Below this hard limit, clients can the use the _soft_ limits defined in the `max-num-htlcs/max-accepted-htlcs` limit which are advertised at the creation of the channel. If a side goes over this limit (the _soft_ limit), the the channel should fail. Additionally, if a client attempts to go over the _hard_ limit the channel should also fail. 

Aside from the rational for sticking to hard-limits above, in my opinion having such a low _soft_ limit may: constrain quick bursts of bandwidths caused by asymmetries in channel capacity, or possibly eliminate/limit higher level constructions which rely on higher limits.

01-messaging-crypto-and-init.md
Roasbeef: The rationale for having the packet length size be _outside_ of the protocol message framing is that with this, the encrypted transport is completely decoupled from the protocol message framing. 

Currently, the protocol message themselves are _encapsulated_ within the encrypted transport. Furthermore, we can change the messaging framing arbitrarily without having to modify the encrypted transport messaging. 

Also other chains/coins can freely re-use the encrypted message transport without having to make any additional modifications outside of possibly modifying the inter-chain protocol messages.

01-messaging-crypto-and-init.md
```diff
@@ -20,24 +20,21 @@ the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
...
+1. `2-byte` big-endian type.
+2. 4 bytes of padding.
+3. Data bytes as specified by the total packet length.
 
 All data fields are big-endian unless otherwise specified.
 
 ### Requirements
 
-A node MUST NOT send a message with data length greater than `8388608`
+A node MUST NOT send a message with more than `65529` data
```
Roasbeef: I believe  the `65529` here should instead be `65569` (as if defined below).

01-messaging-crypto-and-init.md
```diff
@@ -20,24 +20,21 @@ the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
 
 All messages are of form:
```
Roasbeef: I think this should be made a bit more explicit, instead reading: 
> All protocol message are of the form:

01-messaging-crypto-and-init.md
Roasbeef: Hmm, yeah I also find this a bit ambiguous. It seems like the message header can be restricted solely to having `2/4 bytes` for the message type, with implementations relying on a lookup take which tells the how many additional bytes to read for the message.

rustyrussell: @Roasbeef we still need explicit lengths:
1) we are allowed to append data to messages in future revisions, and
2) you're supposed to totally ignore odd-typed messages you don't know.

01-messaging-crypto-and-init.md
rustyrussell: Agreed.  Fixed.

01-messaging-crypto-and-init.md
rustyrussell: @cdecker Yes, I like this idea.  Add 2 bytes to the framing header, meaning a naive implementation will have this naturally-aligned (a decrypt-in-place will be 4-byte aligned, but will anyone really do that?).

What does @Roasbeef think?

01-messaging-crypto-and-init.md
rustyrussell: It does if we decrypt-in-place.  But I agree that's unlikely; see above.

My unmarshalling implementation doesn't actually care about alignment, though it could be optimized a little if it did.

01-messaging-crypto-and-init.md
rustyrussell: That's lightly true: an implementation could simply prepend the twp byte type to the encapsulation of data though.

Certainly pushing it down a layer like this simplifies the alignment problem, which @adiabat was concerned with in Milan IIRC.

01-messaging-crypto-and-init.md
rustyrussell: Yep, that looks like a definition slip.

We should probably spell out terms and use them consistently, eg:
1. Header: Encrypted 2 byte length (+2 byte type if we decide to go that way)
2. Header MAC: 16 byte mac of header.
3. Ciphertext: encrypted message as defined in other BOLTs.
4. Cipertext MAC: 16 byte mac of ciphertext.

(Better terms welcome!)

cdecker: Sounds good to me, and naming things should help clarify some things.

rustyrussell:
Rebased on top of master, squashed into a single commit.

PR #25, opened 2016-11-24
Adding a first-per-commitment-point field to open_channel

pm47:
```
   A                                B
   |                                |
   |       (1) open_channel         |
   |------------------------------->|
   |                                |
   |       (2) accept_channel       |
   |<-------------------------------|
   |                                |
   |      (3) funding_created       |
   |------------------------------->|
   |                                |
   |       (4) funding_signed       |
   |<-------------------------------|
```
B needs A's `first-per-commitment-point` to compute the signature that B is supposed to return in step (4).

On a related note, I am not sure why A includes his signature in `funding_created`. At first I thought the reason for this is that the first commitment transaction can pay both parties, because of the possibility to have a `pushMsat` > 0. But on second thought I don't think B needs to have it: as soon as the funding tx is published, the only tx that can spend it is A's first commitment tx, which does pay `pushMsat` to B. Sure if A disappears, then B will never get his money, but should he get paid if nothing ever happened in the channel? I guess that depends on the rationale behind `pushMsat`?

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -84,6 +84,7 @@ desire to set up a new channel.
    * [33:revocation-basepoint]
    * [33:payment-basepoint]
    * [33:delayed-payment-basepoint]
+   * [33:first-per-commitment-point]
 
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
Yes, I left the initial signature in there for symmetry, and so it can count on push-msat (once the funding tx is locked).  @adiabat might have an opinion here?

PR #26, opened 2016-11-26
BOLT 3: add more detailed commitment transaction estimation

AndrewSamokhvalov:
Hi, I come up with slightly different estimation for commitment transaction weight. Would you consider to correct my numbers and add more detailed description? I think it might be helpful especially if you read the spec at first time.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -191,10 +191,10 @@ The *expected weight* of a commitment transaction is calculated as follows:
 		- previous_out_point: 36 bytes
 			- hash: 32 bytes
 			- index: 4 bytes
-		- OP_DATA: 1 byte (ScriptSigLength)
+		- var_int: 1 byte (script_sig length)
 		- script_sig: 0 bytes
-		- witness <----	we use "Witness" instead of "ScriptSig" for
-	 			transaction validation, but "Witness" is stored
+		- witness <----	we use "witness" instead of "script_sig" for
+	 			transaction validation, but "witness" is stored
 	 			separately and cost for it size is smaller. So
 	 			we separate the calculation of ordinary data
 	 			from witness data.
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -158,22 +158,89 @@ Thus we use a simplified formula for *expected weight*, which assumes:
 
 The *expected weight* of a commitment transaction is calculated as follows:
 
-	transaction core: 4 + 1 + 1 + 4
-	transaction input: 32 + 4 + 1 + 4
-	transaction input witness: 1 + 74 + 74 + 1 + 1 + 34 + 34 + 1 + 1
-	transaction output: 8 + 1
-	transaction htlc output script: 34
-	transaction to-local output script: 34
-	transaction to-remote output script: 25
+	p2wsh: 34 bytes
+		- OP_0: 1 byte
+		- OP_DATA: 1 byte (witness_script_SHA256 length)
+		- witness_script_SHA256: 32 bytes
 
+	p2wpkh: 22 bytes
+		- OP_0: 1 byte
+		- OP_DATA: 1 byte (public_key_HASH160 length)
+		- public_key_HASH160: 20 bytes
+
+	multi_sig: 71 bytes
+		- OP_2: 1 byte
+		- OP_DATA: 1 byte (pub_key_alice length)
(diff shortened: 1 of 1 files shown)


03-transactions.md
```diff
@@ -158,22 +158,89 @@ Thus we use a simplified formula for *expected weight*, which assumes:
...
+		- sig_bob_length: 1 byte
+		- sig_bob: 73 bytes
+		- witness_script_length: 1 byte
+		- witness_script (multi_sig)
+		
+	funding_input: 41 bytes
+		- previous_out_point: 36 bytes
+			- hash: 32 bytes
+			- index: 4 bytes
+		- OP_DATA: 1 byte (ScriptSigLength)
```
Roasbeef: This isn't an `OP_DATA`, it's actually a var-int in the serialized form.

03-transactions.md
```diff
@@ -158,22 +158,89 @@ Thus we use a simplified formula for *expected weight*, which assumes:
...
+		- tx_out: 74 + 43 * num-htlc-outputs bytes
+			output_paying_to_them,
+			output_paying_to_us,
+			....htlc_output's...
+		- lock_time: 4 bytes
+	
 Multiplying non-witness data by 4, this gives a weight of:
+	
+	// 500 + 172 * num-htlc-outputs weight
+	commitment_transaction_weight = 4 * commitment_transaction
```
Roasbeef: The weight shouldn't simply be multiplied by `4` here. In order to properly calculate the weight of the transaction, the `"stripped-size"` (the size without witness data) needs to be multiplied by `3` then added to the _total_ size (size with the witness data) of the transaction.

03-transactions.md
```diff
@@ -158,22 +158,89 @@ Thus we use a simplified formula for *expected weight*, which assumes:
...
+	multi_sig: 71 bytes
+		- OP_2: 1 byte
+		- OP_DATA: 1 byte (pub_key_alice length)
+		- pub_key_alice: 33 bytes
+		- OP_DATA: 1 byte (pub_key_bob length)
+		- pub_key_bob: 33 bytes
+		- OP_2: 1 byte
+		- OP_CHECKMULTISIG: 1 byte
+
+	witness: 222 bytes
```
Roasbeef: Seems like this is actually calculating the `weight` instead of the raw size in bytes.

Roasbeef: Actually, nvm didn't see that the size of the witness script was factored in also!

03-transactions.md
```diff
@@ -158,22 +158,89 @@ Thus we use a simplified formula for *expected weight*, which assumes:
...
+	 htlc_output: 43 bytes
+		- value: 8 bytes
+		- var_int: 1 byte (pk_script length)
+		- pk_script (p2wsh): 34 bytes
+
+	 witness_header: 2 bytes
+		- flag: 1 byte
+		- marker: 1 byte
+
+	 commitment_transaction: 125 + 43 * num-htlc-outputs bytes
```
Roasbeef: I think this should actually be `125`: as `4+2+1+41+1+74+4 = 125` (expanding using the variables defined above).

03-transactions.md
rustyrussell: @roasbeef Well, weight = witness-bytes + non-witness-bytes*4.  Which is equivalent to your "weight = non-witness-bytes * 3 + (witness-bytes + non-witness-bytes)".

If you're retrofitting on top of existing code, your formula makes sense.  But I hope(!) it will look quaint in a few years when All The World's Segwit...

03-transactions.md
AndrewSamokhvalov: As far as I understand `witness_header` is a part of witness data and its cost equals to `1`. To make calculation cleaner I decided to not include it in `commitment_transaction` size as far as latter I still would have to split it on `125 * 4 + 2`.

rustyrussell:
Thank you *so* much for doing this!!

And thanks for catching my errors!  Esp. P2WPKH size and witness size.

You assumed a `to-remote` output in your calculations.  That means our fee rate is about 18% too high (worst-cast) if we don't have one, but I like the simplification.

I've created a follow-up pull request which documents this and similarly expands (and corrects!) the weights for the htlc transactions: #30

PR #27, opened 2016-11-27
BOLT 7: sign only public key rather than whole message

AndrewSamokhvalov:


07-routing-gossip.md
```diff
@@ -51,9 +51,54 @@ the double-SHA256 of `node-id-1` using `bitcoin-key-1`, and set
...
+4. Create cross reference 'node-id-1'<->'bitcoin-key-1'
+5. Prove ownership of 'node-id-2'
+6. Create cross reference 'node-id-2'<->'bitcoin-key-2'
+7. Check cross reference 'bitcoin-key-1'<->'bitcoin-key-2'
+
+    (3)       (4)         (1)         (7)         (2)         (6)       (5)
+[node-id-1] <- ? -> [bitcoin-key-1] <- ? -> [bitcoin-key-2] <- ? -> [node-id-2]
+
+
+D(e,k) - decryption function, which takes:
```
Roasbeef: What decryption algorithm is used? Why is encryption needed at all?

rustyrussell:
Indeed, the rationale behind signing the entire message (including any appended data in future) is to ensure you've relayed it reliably.  In particular, we *will* need to add a feature bitmask.

I agree with @Roasbeef that the spec was under-defined though, particularly the rationale.

cdecker:
Agreed, we'd probably want to have the signature prefix the rest of the message and have them sign the entire message. It is good though to make the relationship between keys explicit in the documentation.

PR #28, opened 2016-11-28
Splitting BOLT01 into messaging and transport specs

cdecker:
The separation between between framing, type and payload HMAC, while
logically nice, make it quiet hard to get a complete picture. Merging
these parts results in a clearer picture of how packets look like. I
also wrote a small intro to get a quick overview. The reader can then
drill down into the specifics of the individual components.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -1,50 +1,31 @@
-# BOLT #1: Message Format, Encryption, Authentication and Initialization
+# BOLT #1: Base Protocol
 
-## Communication Protocols
+## Overview
+This protocol assumes an underlying authenticated and ordered transport mechanism that takes care of framing individual messages.
+[BOLT 08](08-transport.md) specifies the canonical transport layer used in Lightning, though it can be replaced by any transport that fulfills the above guarantees.
 
-This protocol is written with TCP in mind, but could use any ordered,
-reliable transport.
-
-The default TCP port is `9735`.  This corresponds to hexadecimal `2607`,
-the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
+The default TCP port is 9735. This corresponds to hexadecimal `0x2607`, the unicode code point for LIGHTNING.<sup>[2](#reference-2)</sup>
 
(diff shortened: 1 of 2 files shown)


rustyrussell:
Crap, I should have checked PRs before doing mine (#29) .  I like your organization even better, can we combine?

cdecker:
No problem, let's merge yours first, since you are moving blocks of text around and I can rebase on top of that. Should not be too difficult. I was actually also contemplating about splitting bolt01 into two specs, one for the lower level crypto transport and one for the higher protocol layer. Since we have the logical separation we could just split there and make this a bit easier to follow along :-)

Roasbeef:
Nice, I like the reorganization as it gives the reader some initial context w.r.t to the lightning message encapsulation within the crypto transport. 

However, if the packet type is _only_ within the header of the crypto packets, then I don't see how it would be possible (as specified now) to seamlessly use another crypto transport with message bodies being the same lightning message framing (tor hidden services for example). As it is now, the lightning messages are fully encapsulated within the body of the crypto packets, this independence allows the possibility of using an entirely different crypto transport without adding special case context-aware framing modifications in the lightning message layer. 

I think splitting up bolt01 one document describing the set up and management of the default crypto transport and another giving an overview of the lightning message framing is a great idea. It would make the independence of the two layers more apparent.

cdecker:
Yeah, forgot about that, I'll amend that separation later, I'll have to wait for the split until @rustyrussell merged his moves. If only we had SVN locks :-)

cdecker:
Ok, starting to split into two separate specs, I didn't want to shift all numbers so I just made BOLT08 the transport layer. We can certainly switch the two numbers around but I wouldn't shift. If people agree I'd like to merge this as soon as possible, otherwise this is going to be conflict hell...

I'll do a few more cleanup passes later :-)

rustyrussell:
Yes, please merge.

cdecker:
Thanks guys, will do a few more passes to dedup, fix markup and make some things explicit :-)

PR #29, opened 2016-11-29
BOLT 01: reorganize, tighten nomenclature.

rustyrussell:
Major changes:
1. Move crypto first, then cover contents
2. Use "lightning messages" explicitly to refer to upper layer, use
   "packet" in the one case where we refer to the whole thing.
3. Collapse "Transport Message Exchange" and "Protocol Message Encapsulation"
   into "Lightning Message Specification" which already contains much overlap.

Minor changes:
1. Put the big-endian requirement in the general overview, since the crypto
   layer uses that too.
2. Say "is" instead of "should be" encrypted: it's not an option.
3. Note why we separate local and global features.


cdecker:
LGTM :+1:

PR #30, opened 2016-11-29
Better calcs

rustyrussell:
Based upon @AndrewSamokhvalov PR #26.


cdecker:
Very nice, this is much more readable. Compared to #26 this adds the details for the HTLC transaction, right? Otherwise the numbers seem to match.

PR #33, opened 2016-12-05
BOLT 8: minor clarifications and fixes.

rustyrussell:
1. More than three variables.
2. Make nonce encoding explicit (Big-endian, like everything else).
3. `e` is a keypair in this context; use `re` for their pubkey.
4. Explicitly note `re` and `le` notation.
5. Fix RFC reference.
6. Close `` around HKDF.
7. Use `ss` for the shared secret; `s` is already taken.
8. Use `re` for their pubkey.
9. Note IETF variant of chacha-poly
10. Note seralization of pubkey over the wire.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -36,13 +36,27 @@ The messages described in this document are grouped logically into 4 groups by t
 
 
 Channel establishment begins immediately after authentication, and
-consists of each node sending an `open_channel` message, followed by
-one node sending `open_funding`, the other providing its
-`open_commit_sig` then both sides waiting for the funding transaction
-to enter the blockchain and reach their specified depth, at which
-point they send `open_complete`.  After both sides have sent
-`open_complete` the channel is established and can begin normal
-operation.
+consists of the funding node sending an `open_channel` message,
+followed by the responding node sending `accept_channel`. With the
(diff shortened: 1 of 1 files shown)


08-transport.md
```diff
@@ -248,7 +248,7 @@ and `16 bytes` for the `poly1305` tag.
        handshake digest.
 
 
-  * Send `m = 0 || e || c` to the responder over the network buffer.
+  * Send `m = 0 || e.pub.serializeCompressed().pub || c` to the responder over the network buffer.
```
Roasbeef: Hmm, thought I described the behavior/return-value of `.serializeCompressed()`, but looking at it again it only declares the function rather than specifying it's behavior.

rustyrussell: Oops, yeah, typo.

08-transport.md
```diff
@@ -340,7 +340,7 @@ for the `poly1305` tag.
      * Finally, the generated ciphertext is accumulated into the authenticating
        handshake digest.
 
-  * Send `m = 0 || e || c` to the initiator over the network buffer.
+  * Send `m = 0 || e.pub.serializeCompressed().pub || c` to the initiator over the network buffer.
```
rustyrussell: Yep, will fix and push.

PR #34, opened 2016-12-05
BOLT 3: Add test vectors for per-commitment-secret generation

rustyrussell:
These were generated (and recovery tested) by the command-line tool at:

	https://github.com/rustyrussell/ccan/tree/master/ccan/crypto/shachain/tools


sstone:
Thanks! We've added them to our own tests and everything passes.

PR #35, opened 2016-12-05
Moved msg type description from BOLT#2 to BOLT#1

pm47:
Makes more sense I think, since some messages are described in BOLT #1.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -2,7 +2,7 @@
 
 ## Overview
 This protocol assumes an underlying authenticated and ordered transport mechanism that takes care of framing individual messages.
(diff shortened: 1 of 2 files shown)


01-messaging.md
```diff
@@ -23,6 +23,13 @@ A node MUST NOT send an evenly-typed message not listed here without prior negot
 A node MUST ignore a received message of unknown type, if that type is odd.
 A node MUST fail the channels if it receives a message of unknown type, if that type is even.
 
+The messages are grouped logically into 4 groups by their most significant set bit:
+
+ - Setup & signalling (types `0`-`31`): comprises setup of the cryptographic transport, communication of supported features and error reporting. These are described below.
```
rustyrussell: You can remove "setup of the cryptographic transport," as that doesn't use these codes.

PR #36, opened 2016-12-06
BOLT 8: clarify nonces.

rustyrussell:
Use a temporary nonce with temp_k, and spell out the send/receive nonces for the normal message transport.

This seems to change the protocol (which as spec'ed is broken, since one side doesn't reset the nonce in one place), but I think is the intent?

### DIFF
--- a/08-transport.md
+++ b/08-transport.md
@@ -118,7 +118,7 @@ Throughout the handshake process, each side maintains these variables:
    zero-length AEAD payloads at the end of each handshake message.
 
  * `n`: A **counter-based nonce** which is to be used with `temp_k` to encrypt
-   each message with a new nonce.  It is encoded as a big-endian number.
+   each message with a new nonce.  It is encoded as a 96-bit big-endian number.
 
  * `e`: A party's **ephemeral keypair**. For each session a node MUST generate a
    new ephemeral key with strong cryptographic randomness.
@@ -140,11 +140,11 @@ The following functions will also be referenced:
        `HKDF`.
 
   * `encryptWithAD(k, n, ad, plaintext)`: outputs `encrypt(k, n++, ad, plaintext)`
-     * where `encrypt` is an evaluation of `ChaCha20-Poly1305` with the passed
+     * where `encrypt` is an evaluation of `ChaCha20-Poly1305` (IETF variant) with the passed
        arguments.
 
   * `decryptWithAD(k, n, ad, ciphertext)`: outputs `decrypt(k, n++, ad, ciphertext)`
-     * where `decrypt` is an evaluation of `ChaCha20-Poly1305` with the passed
+     * where `decrypt` is an evaluation of `ChaCha20-Poly1305` (IETF variant) with the passed
        arguments.
 
   * `generateKey()`
@@ -248,7 +248,7 @@ and `16 bytes` for the `poly1305` tag.
        handshake digest.
 
 
-  * Send `m = 0 || e || c` to the responder over the network buffer.
+  * Send `m = 0 || e.pub.serializeCompressed() || c` to the responder over the network buffer.
 
 
 **Receiver Actions:**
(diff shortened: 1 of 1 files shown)


08-transport.md
```diff
@@ -329,10 +323,9 @@ for the `poly1305` tag.
   * `ck, temp_k = HKDF(ck, ss)`
      * This phase generates a new temporary encryption key (`temp_k`) which is
        used to generate the authenticating MAC.
-     * The nonce `n` should be reset to zero: `n = 0`.
 
 
-  * `c = encryptWithAD(temp_k, n, h, zero)`
```
Roasbeef: The nonce used for each call to `encryptWithAD` _should_ be zero. Each `temp_k` is only used for a single encryption/decryption within the handshake.

08-transport.md
```diff
@@ -408,11 +399,11 @@ construction, and `16 bytes` for a final authenticating tag.
 **Sender Actions:**
 
 
-  * `c = encryptWithAD(temp_k, n, h, s.pub.serializeCompressed())`
+  * `c = encryptWithAD(temp_k, 2, h, s.pub.serializeCompressed())`
     * where `s` is the static public key of the initiator.
 
 
-  * `h = SHA-256(h || c)`
```
Roasbeef: The prior version (as written currently) is correct. The ciphertext gets mixed in, instead of the raw public key. The handshake digest is meant to authenticate everything that's sent over the wire. In this case the encrypted pubkey, not the pubkey itself is being sent.

08-transport.md
```diff
@@ -442,6 +432,8 @@ construction, and `16 bytes` for a final authenticating tag.
      * This step generates the final encryption keys to be used for sending and
        receiving messages for the duration of the session.
 
+  * `rn = 0, sn = 0`
```
Roasbeef: Good addition by making the sender/receiver nonces explicit!

08-transport.md
```diff
@@ -371,13 +364,11 @@ for the `poly1305` tag.
...
      * This phase generates a new temporary encryption key (`temp_k`) which is
        used to generate the authenticating MAC.
-     * The nonce `n` should be reset to zero: `n = 0`.
 
 
-  * `p = decryptWithAD(temp_k, n, h, c)`
+  * `p = decryptWithAD(temp_k, 1, h, c)`
     * If the MAC check in this operation fails, then the initiator MUST
       terminate the connection without any further messages.
-    * The nonce `n` should be reset to zero: `n = 0`.
```
Roasbeef: Yeah, so this line (the nonce reset after decrypt) shouldn't be there, nice catch!

08-transport.md
Roasbeef: So within this diff  `encryptWithAD/decryptWithAD` should always be called with a nonce of zero during the handshake.

08-transport.md
```diff
@@ -459,7 +451,7 @@ construction, and `16 bytes` for a final authenticating tag.
     abort the connection attempt.
 
 
-  * `rs = decryptWithAD(temp_k, n, h, c)`
+  * `rs = decryptWithAD(temp_k, 2, h, c)`
```
Roasbeef: (commenting here as I can't on the line below) 

The modification on line 406 is incorrect. The actual error is on line 459, the receiver should mix the  _ciphertext_ into the handshake rather than the decrypted material (the pubkey).

rustyrussell:
OK, new version.  The first three patches are non-controversial.

Number 4 spells out the exact nonce values; there's a reuse of temp_k between Act 2 and 3, so there I assumed you wanted a nonce increment, otherwise they're 0.

Number 5 avoids reusing temp_k, and spells out clearly that there are three separate ones throughout the process.  I'm not sure about this: we're happy to overwrite ck every time without comment.  But OTOH doing that made the nonce-can't-always-be-0 problem clear.

08-transport.md
```diff
@@ -483,15 +470,18 @@ construction, and `16 bytes` for a final authenticating tag.
...
 
        `rk` is the key to be used by the responder to decrypt the messages sent
-       by the responder,
+       by the initiator,
 
 
-       and `sk` is the key to be used by the initiator to encrypt messages to
-       the responder,
+       and `sk` is the key to be used by the responder to encrypt messages to
+       the initiator,
```
cdecker: Good catch, these were indeed twisted.

cdecker:
Looking good, I'll let @Roasbeef pull the trigger :wink:

rustyrussell:
Squashing fixes now, no changes....

PR #38, opened 2016-12-06
BOLT 2: Move update_fee message from 'Channel Establishment' to 'Normal Operation'

pm47:
From an implementor's point of view, I think the `update_fee` message is much closer to other `update_*` messages than to channel setup or teardown, and belongs to the Normal operation section. `update_fee` messages will be queued along with htlc-related messages, and applied in order when signing a new commitment.

NB: a consequence of this is that the `type` of the message changes from `37` to `134`.

edit:
Actually I wonder if we should reserve types:
- `32`-`127` to channel creation, establishment, and in the future: cash in, cash out, reanchoring, ...
- `128`-`255` to commitment update (htlc, fees, signatures)

But I am not sure what the rationale was for the current logical grouping?

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -11,7 +11,6 @@ operation, and closing.
       * [The `funding_created` message](#the-funding_created-message)
       * [The `funding_signed` message](#the-funding_signed-message)
       * [The `funding_locked` message](#the-funding_locked-message)
-      * [Updating Fees: `update_fee`](#updating-fees-update_fee)
     * [Channel Close](#channel-close)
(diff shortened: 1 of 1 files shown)


rustyrussell:
Agreed that operationally the new grouping is logical, but it was @cdecker who grouped them that way so I've assigned this to him.

cdecker:
Good catch, I think my rationale was that the 32-128 `type`s concerned the channel's lifecycle, while the 128-256 were HTLC lifecycle related, and since `update_fee` concerns the channel, and not an individual HTLC I grouped them into the channel lifecycle.

But I agree that `update_fee` cuts across these concerns, so I'm fine with both options.

PR #43, opened 2016-12-08
BOLT 8: use an alternate ECDH function

sstone:
If we change the definition of our ECDH function to return the actual point in compressed format, I believe that we should be able to reuse existing Noise implementations without any modifications. 

The actual code that I've used is there: https://github.com/sstone/noise/tree/wip-secp256k1.

### DIFF
--- a/08-transport.md
+++ b/08-transport.md
@@ -127,8 +127,7 @@ The following functions will also be referenced:
   * `ECDH(rk, k)`: Performs an Elliptic-Curve Diffie-Hellman operation using
     `rk` which is a `secp256k1` public key and `k` which is a valid private key
     within the finite field as defined by the curve paramters.
-      * The returned value is the raw big-endian byte serialization of
-        `x-coordinate` (using affine coordinates) of the generated point.
+      * The returned value is the serialization of the generated point in compressed format (33 bytes)
 
   * `HKDF`: a function is defined in [5](#reference-5), evaluated with a
     zero-length `info` field.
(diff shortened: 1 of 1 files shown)


cdecker:
Is there actually any advantage in using the point/coordinate in an unhashed fashion? `libsecp256k1` returns the `sha256` hash of the compressed representation of the point instead, so maybe we could use that?

sstone:
I think so: if we return the actual point then we can just use the output of the Write() construct as defined in the Noise specs and feed it to the Read() construct.

rustyrussell:
Yech, I implemented this wrong too, and used libsecp's ECDH function, which as @cdecker points out, doesn't do either! 

Since everyone has should be probably eventually be using libsecp256k1, even if via language wrappers, my preference would be to use its ECDH function, which is:

     SHA256(y-first-byte || x)

Where y-first-byte is the first byte of y's compressed representation (ie. 2 or 3) and x is x's compressed representation.

Let's see if I can pull @apoelstra into this conversation...

rustyrussell:
gmaxwell says on sepc256k1 (https://botbot.me/freenode/secp256k1/2016-12-09/?msg=77695797&page=1 )

> gmaxwell: Returning the point is philosophically at odds with the design of our interface. Doing so would be unsafe by default and we've seen broken cryptography created by people directly using the point arising from ECDH for things. (e.g. the electrum original 'message encryption' feature).
> gmaxwell: I believe we discussed taking a hasher function, similar to how we do nonce validation, to enable alternative constructions while remaining secure by default.

Implementations of Noise need to define ECDH over sepc256k1 anyway, so from that POV why does it matter what the definition is?  What am I missing?

sstone:
It does not matter very much. I was just being lazy and it seems that the Noise specs assume that dhLen == pubkeyLen (i.e. when processing "e" Write produces e.pub || xxx and Read says re = first dhLen bytes etc...), and I've seen ECDH libraries return either compressed point, point.x or sha256(point.x).
I'm happy with secp256k1's ECDH and changing the description of ECDH in bolt #8 to match it.

Roasbeef:
Yeah, so in either case, implementers which are instantiating new noise schemes will need to define the semantics of their `ECDH` function. You should be able to use existing noise implementations adding in one minor modification: the existence of our version byte which is the _prefix_ of each handshake sent.

I agree with Rusty, I think we should redefine our `ECDH` function within `BOLT0008` to better align ourselves with `libsecp256k1` as that's a popular, well reviewed crypto library in the space. Modifying my implementation to use `SHA(compressedPoint)` is a pretty trivial change, which should resolve this issue.

sstone:
Fine with me, closing this PR. Thanks!

PR #52, opened 2016-12-12
Bolt 7 cleanups

rustyrussell:
- Remove useless padding.
- Spell out exact signature checks for receivers.
- Add explicit feature bitfields for future use
- Allow non-broadcast of spam packets.
- Spell out proof chain logic as inspired by Andrew's PR #27


rustyrussell:
@AndrewSamokhvalov

07-routing-gossip.md
```diff
@@ -15,12 +15,34 @@ information), and at least two `channel_update` messages are expected.
...
+[BOLT #3](03-transactions.md#funding-transaction-output).
+
+The second two are done by explicit signatures (`bitcoin-signature-1`
+and `bitcoin-signature-2`, generated by each `bitcoin-key` and signing
+the corresponding `node-id`.
+
+We also need to prove that `node-1` and `node-2` both agree on this
+announcement message; that is done by having a signature from each
+`node-id` signing the entire message (`node-signature-1` and
+`node-signature-2`).
```
cdecker: This is a bit underdefined, we can either set the signatures to a fixed value (all zeros) and then sign the entire message, or we can sign the message starting from byte 128, i.e., skipping the signature fields themselves. I have a slight preference for the latter since then we can just store the signatures in-place and skip serializing twice. In the former case we'd have to store the signatures somewhere else while doing the two signatures and then copy them over.

07-routing-gossip.md
```diff
@@ -94,6 +125,15 @@ While channels shouldn't be advertised before they are sufficiently
...
 transaction hasn't moved to a different block.
 
+To avoid having to store excessive-sized messages, yet allow
+reasonable expansion in future, nodes are allowed to restrict
+rebroadcasting (perhaps statistically).
+
+New channel features are possible in future; backwards compatible (or
+optional) ones will have odd feature bits, incompatible ones will have
+even feature bits.  These will be propagated by nodes even if they
+can't use the announcements themselves.
```
cdecker: Maybe we can add a reference to the _it's ok to be odd_ rule in the glossary?

07-routing-gossip.md
cdecker: I see that further down it specifies that signatures skip the initial 128 bytes, but maybe we can make it explicit here as well.

rustyrussell: Yeah, this is the discussion of why we're doing it, the requirements are below.   Hmm, I removed the word "entire", to make it deliberately more vague.

PR #56, opened 2016-12-13
BOLT04: specify per-hop-payload format

### LINKED ISSUE #54: BOLT 4: hop payload contents should be specified
[1:realm][8:amt_to_forward][2:cltv_value][9:unused_with_v0_version_on_header]

There was discussion on why we want to duplicate CLTV here (after all, it's in the HTLC itself); ISTR it was to avoid probe attacks using differing HTLC values?

Roasbeef:
So here's a first stab at specifying the format and forwarding related error conditions around the per-hop-payload that's now a part of the Sphinx packet. 

The current proposed format is: `[1:realm][8:amt_to_forward][2:incoming_cltv_value][9:unused_for_v0]`. Details surrounding the exactly meaning and cross-validation of the payload contents can be found within the sole commit within this PR. 

I'm a bit conflicted about exactly _where_ in the specs the format+handling specification should go. Currently, within the PR I've included it within `BOLT04`. However, the semantics kinda of leak into `BOLT02` as improper values (payload doesn't match the HTLC) within the `per-hop-payload` may result in HTLC failure messages. Also in order to validate the information within the `per-hop-payload` the packet processor would also need access to the `update_add_htlc` which included the HTLC packet. 

I guess the question here is do we want to keep `BOLT04` lightning agonistic? Or are we OK within mixing in context/details from other parts of the spec also. Personally, I lean towards the self-containment of all details w.r.t onion routing residing within `BOLT04`, we've already began to tweak the packet construction/processing enough where it's no longer a generic Sphinx protocol. 

Fixes #54.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -128,6 +128,53 @@ With the `hopfiller` being constructed in the same way as the routing
 info `filler` and each payload being incrementally obfuscated at each
 hop.
 
+### Per Hop Payload Format
+
+Using the per-hop-payload, the sender is able precisely specify the path and
+structure of the HTLC's forwarded at each hop. As the per-hop-payload is
+protected under the packet-wide HMAC, the information within the
+per-hop-payload is fully authenticated with each pair-wise relationship between
+the HTLC sender, and each intermediate node in the path.  Using this end-to-end
+authentication, forwarding nodes are able to ensure that the incoming node
+didn't forward an ill-crafted HTLC by cross-checking the HTLC parameters with
+the values as specified within the per-hop-payload.
+
+The format of the per-hop-payload for a version 0 packet is as follows: 
+```
+┌──────────────┬─────────────────────────┬───────────────────────────────┬───────────────────────────────────────────┐
+│realm (1 byte)│amt_to_forward (8 bytes) │ incoming_cltv_value (2 bytes) │unused_with_v0_version_on_header (9 bytes) │
+└──────────────┴─────────────────────────┴───────────────────────────────┴───────────────────────────────────────────┘
+```
+
+Field Description: 
+
+   * `realm` - The realm (chain) link that the outgoing HTLC should be forwarded
+     on. Within a version 0 packet, this field currently carries no meaning as
+     cross-chain forwarding has not yet been fully specified.
+
+   * `amt_to_forward` - The amount in milli-satoshi to forward to the next
+     (outgoing) hop specified within the routing information. 
+     
+     This value factors in the computed fee for this particular hop. When
+     processing an incoming Sphinx packet, if the `amt_to_forward` field
+     unconverted does not match the value of the HTLC extended, then the HTLC
+     should be failed and rejected as this indicates the node has deivated from
+     the specified paramters.
+
+   * `incoming_cltv_value` - The CLTV value that the _incoming_ HTLC carrying
+     the packet should have. 
+     
+     Inclusion of this field allows a node to authtenicate the information
+     specified by the original sender and the paramters of the HTLC forwarded.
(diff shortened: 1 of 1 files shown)


04-onion-routing.md
```diff
@@ -128,6 +128,53 @@ With the `hopfiller` being constructed in the same way as the routing
...
+the HTLC sender, and each intermediate node in the path.  Using this end-to-end
+authentication, forwarding nodes are able to ensure that the incoming node
+didn't forward an ill-crafted HTLC by cross-checking the HTLC parameters with
+the values as specified within the per-hop-payload.
+
+The format of the per-hop-payload for a version 0 packet is as follows: 
+```
+┌──────────────┬─────────────────────────┬───────────────────────────────┬───────────────────────────────────────────┐
+│realm (1 byte)│amt_to_forward (8 bytes) │ incoming_cltv_value (2 bytes) │unused_with_v0_version_on_header (9 bytes) │
+└──────────────┴─────────────────────────┴───────────────────────────────┴───────────────────────────────────────────┘
```
cdecker: We probably should stick with one ASCII art style for tables, just for consistency's sake.

04-onion-routing.md
```diff
@@ -128,6 +128,53 @@ With the `hopfiller` being constructed in the same way as the routing
...
+     cross-chain forwarding has not yet been fully specified.
+
+   * `amt_to_forward` - The amount in milli-satoshi to forward to the next
+     (outgoing) hop specified within the routing information. 
+     
+     This value factors in the computed fee for this particular hop. When
+     processing an incoming Sphinx packet, if the `amt_to_forward` field
+     unconverted does not match the value of the HTLC extended, then the HTLC
+     should be failed and rejected as this indicates the node has deivated from
+     the specified paramters.
```
cdecker: We should make this explicit: `incoming_htlc_amt - fee >= amt_to_forward`

04-onion-routing.md
```diff
@@ -128,6 +128,53 @@ With the `hopfiller` being constructed in the same way as the routing
 info `filler` and each payload being incrementally obfuscated at each
 hop.
 
+### Per Hop Payload Format
+
+Using the per-hop-payload, the sender is able precisely specify the path and
```
cdecker: s/able precisely/able to precisely/

cdecker:
I share your concern about mixing context, however I'm starting to suspect that keeping them separate may lead to a harder to understand spec, since we push details into BOLT02, which forces future implementors to switch back and forth between the specs. So I'm happy to have this in BOLT04, maybe we eventually come up with a cleaner separation, e.g., have a wrapper spec that specifies all the Lightning-specific things and where to put them, but that's for a distant future (v2.0? :wink: )

04-onion-routing.md
Roasbeef: Good point, I've modified the ASCII diagram to be consistent with the style used within the rest of the document.

04-onion-routing.md
Roasbeef: Indeed, I update the test to be more explicit w.r.t to the exact check a node should assert before forwarding.

04-onion-routing.md
Roasbeef: Fixed.

cdecker:
Very good, thanks :smile:

rustyrussell:
OK, I'm going to squash and merge this.  But we'll need to revisit to tighten the requirements later.  In general, *data* doesn't have requirements, those *creating* and *receiving* it have requirements (and that maps better to implementation).  Saying field X must be Y doesn't say what to do if it isn't...

PR #58, opened 2016-12-13
[RFC] BOLT 7: allow multiple address descriptors.

rustyrussell:
This lets us have both IPv4 and IPv6 (and Tor etc in future) address.


rustyrussell:
Closing; a better format is 2:len,len:addrbytes. Then addrbytes is 1:type then the length, with types always in ascending order.

That ensures we can add other fields later.

Will resubmit tomorrow.

PR #60, opened 2016-12-14
BOLT 7: allow multiple address descriptors. (Take II)

rustyrussell:
This lets us have both IPv4 and IPv6 (and Tor etc in future) address,
while still preserving the ability to add other fields to the message
in future.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -160,7 +160,7 @@ and WISTFULTOLL and use the color black.
 `addresses` allows the node to announce its willingness to accept
 incoming network connections: it contains series of `address
 descriptor`s for connecting to the node.  The first byte describes the
-address type, followed by the number of bytes for that type.
+address type, followed by the appropriate number of bytes for that type.
 
 The following `address descriptor` types are defined:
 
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -144,30 +144,50 @@ nodes for which a channel is not already known are ignored.
...
 and WISTFULTOLL and use the color black.
 
+`addresses` allows the node to announce its willingness to accept
+incoming network connections: it contains series of `address
+descriptor`s for connecting to the node.  The first byte describes the
+address type, followed by the number of bytes for that type.
+
+The following `address descriptor` types are defined:
+
+1. `0`: padding.  data = none (length 0).
```
cdecker: This allows a node to bloat the announcement with padding, and every node will have to store it in case they'd like to forward it. That's also the reason why we dropped all paddings from BOLT07 and/or repurposed them.

07-routing-gossip.md
```diff
@@ -144,30 +144,50 @@ nodes for which a channel is not already known are ignored.
...
 `node_announcement` it has created.  It MAY base it on a UNIX
-timestamp.  It MUST set the `ipv6` and `port` fields to all zeroes, or
-a non-zero `port` and `ipv6` set to a valid IPv6 address or an IPv4-Mapped IPv6 Address format as defined in [RFC 4291 section 2.5.5.2](https://tools.ietf.org/html/rfc4291#section-2.5.5.2).  It MUST set `signature` to the signature of
+timestamp.  It MUST set `signature` to the signature of
 the double-SHA256 of the entire remaining packet after `signature` using the
 key given by `node-id`.  It MAY set `alias` and `rgb-color` to customize their node's appearance in maps and graphs, where the first byte of `rgb` is the red value, the second byte is the green value and the last byte is the blue value.  It MUST set `alias` to a valid UTF-8 string, with any `alias` bytes following equal to zero.
 
-The creating node SHOULD set `len` to the minimum length required to
+The creating node SHOULD fill `addresses` with an address descriptor
+for each distinct network address which expects incoming connections.
```
cdecker: I think it'd reasonable to limit this to 1 instance for each address class. First of all it's going to be by far the most common case that a node just announces one address, and secondly a node can still announce different addresses by announcing a different `node_id` associated with that address.

07-routing-gossip.md
```diff
@@ -178,7 +198,10 @@ valid signature using `node-id` of the double-SHA256 of the entire
 message following the `signature` field (including unknown fields
 following `alias`), and MUST NOT further process the message.
 
-The receiving node SHOULD ignore `ipv6` if `port` is zero.
+The receiving node SHOULD ignore the first `address descriptor` which
```
cdecker: So `addrlen` needs to be the size in bytes of the `address` array, and not the number of addresses, otherwise attaching new fields to the end would be broken. I first assumed it would be the latter case, maybe we can make this explicit above?

rustyrussell: Yes, it's implied in the "[addrlen:addresses]" description, but I've made it explicit.

cdecker:
Great, looking good. I'll leave this open till Monday so others can review as well :smile:

PR #61, opened 2016-12-16
bolt04: Added additional information field to return message

### LINKED ISSUE #53: BOLT 4: failure messages should contain channel_update (maybe + a message string?)
Every failure should come down to one of three things:
(1) it violates requirements set out in channel_update (eg. amount too low, fee too low), OR
(2) it's using an old channel_update, OR
(3) this channel is transiently limited (in which case we need a signed channel update which says that: TODO)

Eventually, we'll want a robust chain-of-custody for these messages, to ensure they don't get corrupted (and if they do, we can prove it so that node is blacklisted).

But ignoring that, we don't want nodes to be able to publish a channel_update and not honor it.

cdecker:
Added an `additional` field to the return message, so that we can
include any protocol level message to inform the sender about the
cause of the failure. This could for example be a `channel_update` if
the channel has become unusable. The message is no longer fixed size,
as hopefully the failure is a rare event, in which case timing
analysis becomes easy anyway.

Closes #53

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -390,8 +390,12 @@ The node returning the message builds a return packet consisting of the followin
  - `hmac` (20 bytes): an HMAC authenticating the remainder of the packet, with a key using the above key generation with key type "_um_".
  - `failure code` (4 bytes)
  - `message length` (2 bytes): the length of the message field in bytes
- - `message` (variable): a string representation. The message is
-   padded to 128 bytes with `0x00` bytes.
+ - `message` (variable): a string representation of the message.
+ - `additional length` (2 bytes): the length of any additional protocol level message attached as additional information.
+ - `additional` (variable): a protocol level message, preceeded by the
+   its `type`, as defined by other specifications. For example a route
+   failure MAY include a `channel_update` message to inform the sender
+   that a channel is no longer usable.
 
 The node then generates a new key, using the key type `ammag`.
 This key is then used to generate a 154 byte pseudo-random stream, which is then applied to the packet using `XOR`.


rustyrussell:
I think we should *always* send the latest channel_announce, making us fixed-lengt-ish again, and SHA256() of the onion packet as we received it.

pm47:
Just to be sure, can you elaborate a little bit on when this facility is to be used as opposed to the `data` field in the `Error` message (defined in BOLT 1)?

FWIW, one thing that was critical when we worked on eclair/lightning-c compatibility is the ability to know the raw commitment tx in case of a signature error. This helps a lot debugging a whole variety of interoperability issues from tx format, protocol (what htlc is to be included), to fee calculation bugs, especially if you are not familiar with the language the other implementation is written in. Should we agree on providing the raw commit tx in the first versions of the implementations, and if so should that be done using this additional information field?

rustyrussell:
> Just to be sure, can you elaborate a little bit on when this facility is to be used as opposed to the `data` field in the `Error` message (defined in BOLT 1)?

ERROR is used when something is terminally wrong on a channel.

> FWIW, one thing that was critical when we worked on eclair/lightning-c compatibility is the ability to know the raw commitment tx in case of a signature error. This helps a lot debugging a whole variety of interoperability issues from tx format, protocol (what htlc is to be included), to fee calculation bugs, especially if you are not familiar with the language the other implementation is written in. Should we agree on providing the raw commit tx in the first versions of the implementations, and if so should that be done using this additional information field?

Agreed.  We should suggest that it include the hex-encoded tx in the
case where the error is a signature validation failure, something like:

diff --git a/01-messaging.md b/01-messaging.md
index 95760ed..f16f108 100644
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -124,7 +124,9 @@ errors which make channels unusable or further communication unusable.
 fail the channel referred to by the `channel-id`, or if `channel-id`
 is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
-A node MUST NOT set `len` to greater than the data length.
+A node MUST NOT set `len` to greater than the data length.  A node
+SHOULD include the raw, hex-encoded transaction in the case where failure
+was caused by an invalid signature check.
 
 A node receiving `error` MUST fail the channel referred to by
 `channel-id`, or if `channel-id` is `0xFFFFFFFFFFFFFFFF` it MUST fail

If you agree, please PR :)

Meanwhile, this is for a more distant failure, and it's severely
underspecified at the moment.  I've been looking at our code, and will
come back with something which tries to specify all the different
failure cases...

Thanks!
Rusty.

cdecker:
Closing in favor of #67

PR #62, opened 2016-12-17
Fix outputs limit and funding output script

fivepiece:
- change outputs limit to 252 as more would mean `output-index` > 1 byte
- remove `0` from the funding output script as it is not part of a normal 2-of-2 multisig script

I believe these two corrections are appropriate.

BOLT02 says: "The `output-index` can only be 1 byte.." , this means compact size int up to 0xFC, or 252.

BOLT03 describes the "Funding Transaction Output" as a P2WSH of a 2-of-2 multisig redeemscript of the two commit keys' pubkeys.
This is just a normal 2-of-2 multisig, so `0` isn't really used.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -215,7 +215,7 @@ The recipient MUST fail the channel if `signature` is incorrect, and MUST ignore
 
 #### Rationale
 
-The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 255 outputs should not be overly burdensome.
+The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 252 outputs should not be overly burdensome.
 
 ### The `funding_signed` message
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -215,7 +215,7 @@ The recipient MUST fail the channel if `signature` is incorrect, and MUST ignore
 
 #### Rationale
 
-The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 252 outputs should not be overly burdensome.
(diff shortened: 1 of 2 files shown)


02-peer-protocol.md
```diff
@@ -215,7 +215,7 @@ The recipient MUST fail the channel if `signature` is incorrect, and MUST ignore
 
 #### Rationale
 
-The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 255 outputs should not be overly burdensome.
+The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 252 outputs should not be overly burdensome.
```
Roasbeef: The protocol doesn't use any compact integers, just regular full big-endian integers everywhere on the wire. With that said, `255` is the correct limit, I don't see the rationale behind reducing it to `252`.

03-transactions.md
```diff
@@ -13,7 +13,7 @@ Most transaction outputs used here are P2WSH outputs, the segwit version of P2SH
 ## Funding Transaction Output
 
 * The funding output script is a pay-to-witness-script-hash<sup>[BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#witness-program)</sup> to:
-   * `0 2 <key1> <key2> 2 OP_CHECKMULTISIG`
+   * `2 <key1> <key2> 2 OP_CHECKMULTISIG`
```
Roasbeef: Nice catch, it seems the zero was there in order to indicate an extra item on the stack to consume the extra pop as a result of the CMS bug within Script. The extra (dummy) item appears on the _witness_, not the witness script.

02-peer-protocol.md
fivepiece: Reverted, but accidentally replied in the commit message (oops)

03-transactions.md
rustyrussell: Yep, good catch!  I expect more of these as we implement and create test vectors...

PR #64, opened 2016-12-19
BOLT 3: fix fee computations

sstone:
1. HTLCs must be able to pay for their HTLC success/timeout transaction fees
2. The amount in the transaction output for a given HTLC should match the HTLC amount
3. The amount of the HTLC success/timeout transaction should be the HTLC amount minus the transaction fee
4. add examples

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -369,18 +369,38 @@ The fee for a commitment transaction MUST BE calculated to match:
 
 1. Start with `weight` = 724, and `fee` = 0.
 
-3. For every offered HTLC, if the HTLC amount plus the HTLC-timeout
-   transaction fee is greater or equal to the local node's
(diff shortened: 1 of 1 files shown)


sstone:
In the currents specs it seems that each HTLC results in an output of "HTLC amount" + fee in the commitment transaction, and an output of "HTLC amount" in the HTLC timeout/success transaction. This feels odd since it would mean, in the case of an offered HTLC, that the other party gets "HTLC amount" + fee when they spend the HTLC output directly.
So I propose to have HTLC outputs that match HTLC amounts exactly in the commitment transaction, and subtract the relevant fees from the HTLC success/timeout transactions.

rustyrussell:
Yes, your reading is correct, the calculation line was wrong.   You really are supposed to output HTLC amount, and the recipient loses the fees.

PR #65, opened 2016-12-20
BOLT 1: recommend full tx in error pkt if signature fails.

rustyrussell:
This will usually be the commitment tx, but could also be the HTLC
tx.

Reported-by: pm47 <pm.padiou@gmail.com>

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -124,9 +124,7 @@ errors which make channels unusable or further communication unusable.
 A node MAY send an empty [data] field.  A node sending `error` MUST
 fail the channel referred to by the `channel-id`, or if `channel-id`
 is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
-A node MUST NOT set `len` to greater than the data length.  A node
-SHOULD include the raw, hex-encoded transaction in the case where failure
-was caused by an invalid signature check.
+A node MUST NOT set `len` to greater than the data length.  A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitsig` message when failure was caused by an invalid signature check.
 
 A node receiving `error` MUST fail the channel referred to by
 `channel-id`, or if `channel-id` is `0xFFFFFFFFFFFFFFFF` it MUST fail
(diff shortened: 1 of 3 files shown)


01-messaging.md
```diff
@@ -124,7 +124,9 @@ errors which make channels unusable or further communication unusable.
 A node MAY send an empty [data] field.  A node sending `error` MUST
 fail the channel referred to by the `channel-id`, or if `channel-id`
 is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
-A node MUST NOT set `len` to greater than the data length.
+A node MUST NOT set `len` to greater than the data length.  A node
```
pm47: I am curious why you don't say "A node MUST set len to exactly the data length"?

01-messaging.md
```diff
@@ -124,7 +124,9 @@ errors which make channels unusable or further communication unusable.
...
+SHOULD include the raw, hex-encoded transaction in the case where failure
+was caused by an invalid signature check.
```
pm47: Maybe:
"A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitsig` message when failure was caused by an invalid signature check"?

btw, could we rename `commitsig` -> `commit_signed` or `commitment_signed`? I know, I am obsessed with consistency 😸

rustyrussell:
>
> Thanks!
>
>> @@ -124,7 +124,9 @@ errors which make channels unusable or further communication unusable.
>  A node MAY send an empty [data] field.  A node sending `error` MUST
>  fail the channel referred to by the `channel-id`, or if `channel-id`
>  is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
> -A node MUST NOT set `len` to greater than the data length.
> +A node MUST NOT set `len` to greater than the data length.  A node
>
> I am curious why you don't say "A node MUST set len to exactly the data length"?

That's a really good question.  Hmm, let's fix that too.

>> @@ -124,7 +124,9 @@ errors which make channels unusable or further communication unusable.
>  A node MAY send an empty [data] field.  A node sending `error` MUST
>  fail the channel referred to by the `channel-id`, or if `channel-id`
>  is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
> -A node MUST NOT set `len` to greater than the data length.
> +A node MUST NOT set `len` to greater than the data length.  A node
> +SHOULD include the raw, hex-encoded transaction in the case where failure
> +was caused by an invalid signature check.
>
> Maybe:
> "A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitsig` message when failure was caused by an invalid signature check"?

Yep, I like that better.  Pushed.

>
> btw, could we rename `commitsig` -> `commit_signed` or `commitment_signed`? I know, I am obsessed with consistency 😸

commitment_signed I think.  Why not...


Cheers,
Rusty.

PR #66, opened 2016-12-20
BOLT 3: Add test vectors for transactions

pm47:
I have moved the details of weight computation to the Appendix section and added another Appendix containing a few commitment and htlc tx examples.

This should allow the testing of:
- the tx scripts
- the ordering and selection of outputs (`dust-limit` parameter)
- the fee computation ~~(note that fees are implemented as per currently specified, as opposed to what @sstone proposed in PR #64)~~
- fee splitting among nodes

There are about a million ways to create bugs with this, so I am pretty sure you'll end up with (hopefully slightly) different results... Looking forward to your feedback!


pm47:
I've updated the PR to reflect PR #64, and added a section about how to split the fee between parties.

I am not to sure about the case where fee can't be paid because "not enough money": should that be related to `channel-reserve-satoshis` parameter or is it correct as-is?

rustyrussell:
Wait, we're not splitting fees.  It was agreed in Milan that originator pays (for v1).

Now, it *can* happen that originator can't pay fee (other side adds too many HTLCs): see BOLT 2:

    There is a possibility of a race: the recipient can add new HTLCs
    before it receives the `update_fee`, and the sender may not be able to
    afford the fee on its own commitment transaction once the `update_fee`
    is acknowledged by the recipient.  In this case, the fee will be less
    than the fee rate, as described in [BOLT #3](03-transactions.md#fee-calculation).

Though BOLT 3 doesn't actually say that we extract the fees from the funder!

Now, we can decide to terminate the channel if this happens, but that's a unilateral decision; I think we should stick with "fees will be extracted from originator's amount, or if that is insufficient, will use the entire amount of the originator's output.  A node MAY fail the channel if the resulting fee rate is too low."

pm47:
> Wait, we're not splitting fees. It was agreed in Milan that originator pays (for v1).

Gosh!! Apparently I [didn't get the memo](https://www.youtube.com/watch?v=jsLUidiYm0w)!

Time to take some vacations I think...

rustyrussell:
Oh, and I wanted to write a tool to extract the test vectors, so could we use a consistent format?

I used: [4 spaces][fieldname]:[value]

Every unique test starting with "name: <brief title of test>", then one value per line.

The field names are obviously test-specific, and it's kind of messy with multiple htlcs, I suggest:

    htlc 1 direction: local->remote
    htlc 1 amount_msat: 10 000 000
    htlc 1 expiry: 443 210

Unf I'm a fair way from implementing this, so I can't test your test vectors.  I'd say we commit them and fix if later it turns out there are issues...

pm47:
> could we use a consistent format?

Sure! I should also probably get rid of the spacing in numeric values for easier parsing.

> I'm a fair way from implementing this, so I can't test your test vectors. I'd say we commit them and fix if later it turns out there are issues...

Ok, but first I need to revert the splitting of fees. I am actually going on vacation tomorrow but I'll get back to it next year! 🎉

rustyrussell:
OK, this looks good, but really needs to be collapsed into fewer commits: say, 1 which updates index and fixes section headers, 1 which adds section about fee (maybe just do that in same commit if easier), then one which adds the test vectors.

pm47:
Closing this PR (replaced by #74)

PR #67, opened 2016-12-21
Detailed error format

### LINKED ISSUE #53: BOLT 4: failure messages should contain channel_update (maybe + a message string?)

rustyrussell:
OK, this took way longer than I thought, but I think this works. Adds onto #61 and closes #53

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -475,7 +475,7 @@ to be used for an incoming HTLC, leaving the node with unexpected
 liability.
 
 
-Thus the effective timeout of the HTLC is the `expiry`, plus some
+Thus the effective timeout of the HTLC is the `cltv-expiry`, plus some
 additional delay for the transaction which redeems the HTLC output to
 be irreversibly committed to the blockchain.
 
@@ -505,7 +505,7 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
    * [8:channel-id]
    * [8:id]
    * [4:amount-msat]
-   * [4:expiry]
+   * [4:cltv-expiry]
    * [32:payment-hash]
    * [1254:onion-routing-packet]
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -618,7 +618,11 @@ A receiving node MUST check that the `payment-preimage` value in
 A receiving node MUST fail the channel if the `BADONION` bit in
 `failure-code` is not set for `update-fail-malformed-htlc`.
 
-A receiving node which has an outgoing HTLC canceled by
+A receiving node MAY check the `sha256-of-onion` in
+`update-fail-malformed-htlc` and MAY retry or choose an alternate
+error response if it does not match the onion it sent.
+
+Otherwise, a receiving node which has an outgoing HTLC canceled by
 `update-fail-malformed-htlc` MUST return an error in the
 `update-fail-htlc` sent to the link which originally sent the HTLC
 using the `failure-code` given and setting the `additional` data to
@@ -632,6 +636,12 @@ A node which doesn't time out HTLCs risks channel failure (see
 If the onion is malformed, the upstream node won't be able to extract
 a key to generate a response, hence the special failure message which
 makes this node do it.
+
+The node can check that the SHA256 the upstream is complaining about
+does match the onion it sent, which may allow it to detect random bit
+errors.  Without re-checking the actual encrypted packet sent, however,
+it won't know whether the error was its own or on the remote side, so
+such detection is left as an option.
 
 ### Committing Updates So Far: `commitment_signed`
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -566,14 +566,17 @@ still be under the maximum message size.  It also ensures that
 a single penalty transaction can spend the entire commitment transaction,
 as calculated in [BOLT #5](05-onchain.md#penalty-transaction-weight-calculation).
 
-### Removing an HTLC: `update_fulfill_htlc` and `update_fail_htlc`
+### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc` and `update_fail_malformed_htlc`
 
 For simplicity, a node can only remove HTLCs added by the other node.
 There are three reasons for removing an HTLC: it has timed out, it has
 failed to route, or the payment preimage is supplied.
 
 The `reason` field is an opaque encrypted blob for the benefit of the
-original HTLC initiator as defined in [BOLT #4](04-onion-routing.md).
+original HTLC initiator as defined in [BOLT #4](04-onion-routing.md),
+but there's a special malformed failure variant for the case where
+our peer couldn't parse it; in this case the current node encrypts
+it into a `update_fail_htlc` for relaying.
 
 1. type: 130 (`update_fulfill_htlc`)
 2. data:
@@ -590,6 +593,15 @@ For a timed out or route-failed HTLC:
    * [2:len]
    * [len:reason]
 
+For a unparsable HTLC:
+
+1. type: 133 (`update_fail_malformed_htlc`)
+2. data:
+   * [8:channel-id]
+   * [8:id]
+   * [4:failure-code]
+   * [32:sha256-of-onion]
+
 #### Requirements
 
 A node SHOULD remove an HTLC as soon as it can; in particular, a node
@@ -603,17 +615,23 @@ A receiving node MUST check that the `payment-preimage` value in
 `update-fulfill_htlc` SHA256 hashes to the corresponding HTLC
 `payment-hash`, and MUST fail the channel if it does not.
 
-A receiving node which closes an incoming HTLC in response to an
-`update-fail-htlc` message on an offered HTLC MUST copy the `reason`
-field to the outgoing `update-fail-htlc`.
+A receiving node MUST fail the channel if the `BADONION` bit in
+`failure-code` is not set for `update-fail-malformed-htlc`.
 
+A receiving node which has an outgoing HTLC canceled by
+`update-fail-malformed-htlc` MUST return an error in the
+`update-fail-htlc` sent to the link which originally sent the HTLC
+using the `failure-code` given and setting the `additional` data to
+`sha256-of-onion`.
 
 #### Rationale
-
 
 A node which doesn't time out HTLCs risks channel failure (see
 "Risks With HTLC Timeouts").
 
+If the onion is malformed, the upstream node won't be able to extract
+a key to generate a response, hence the special failure message which
+makes this node do it.
 
 ### Committing Updates So Far: `commitment_signed`
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -587,7 +587,8 @@ For a timed out or route-failed HTLC:
 2. data:
    * [8:channel-id]
    * [8:id]
-   * [154:reason]
+   * [2:len]
+   * [len:reason]
 
 #### Requirements
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -429,13 +429,13 @@ Blinding is done by computing a blinding factor from the node's public key and t
 The blinding factor is the result of serializing the node's public key into its compressed format, appending the shared secret and computing the `SHA256` hash.
 The blinded EC point then is the result of the scalar multiplication between the EC point and the blinding factor.
 
-## Returning Messages
+## Returning Errors
 
-The protocol includes a simple mechanism to return messages to the origin node.
-The returned messages may either be failures reported by any intermediate hop, or a success message from the final node.
-The format of the forward packet is not usable for the return path, since no node other than the origin has the required information.
+The protocol includes a simple mechanism to return encrypted error messages to the origin node.
+The returned messages may either be failures reported by any intermediate hop, or the final node.
+The format of the forward packet is not usable for the return path, since no node other than the origin has the required information.  Note that these error messages are not reliable, as they are not placed on-chain in the case of node failure.
 
-Intermediate hops store the shared secret on from the forward path along and reuse it to obfuscate the return packets on each hop.
+Intermediate hops store the shared secret from the forward path and reuse it to obfuscate the error packet on each hop.
 In addition each node locally stores the previous hop it received the forward packet from, in order to determine where to send an eventual return packet.
 The node returning the message builds a return packet consisting of the following fields:
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -442,12 +442,15 @@ Intermediate hops store the shared secret from the forward path and reuse it to 
 In addition each node locally stores the previous hop it received the forward packet from, in order to determine where to send an eventual return packet.
 The node returning the message builds a return packet consisting of the following fields:
 
- - `hmac` (20 bytes): an HMAC authenticating the remainder of the packet, with a key using the above key generation with key type "_um_".
- - `failure code` (4 bytes)
- - `additional length` (2 bytes): the length of any additional protocol level message attached as additional information.
- - `additional` (variable): as specified under "Failure Codes" below.
- - `pad length` (2 bytes): the length of `pad`.
- - `pad` (variable): extra bytes to conceal length.
+1. data:
+   * [20:hmac]
+   * [4:failure-code]
+   * [2:additional-len]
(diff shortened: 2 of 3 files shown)


04-onion-routing.md
```diff
@@ -463,6 +463,163 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+The following failure codes are supported.  A node MUST select one of
+these failure codes when creating an error message, and MUST include the
+appropriate `additional` data.
+
+In the case of more than one error, a node SHOULD select the first one
+listed.
+
+The top byte of the failure code can be read as a set of flags:
+* 0x80000000 (BADONION): unparsable onion, encrypted by previous node.
+* 0x80000000 (PERM): permanent failure (otherwise transient)
```
cdecker: `PERM` and `BADONION` have the same prefix. And is there a reason we use the most significant nibble?

04-onion-routing.md
```diff
@@ -463,6 +463,163 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+
+1. type: BADONION|PERM|4 (`invalid_onion_version`)
+2. data:
+   * [32:sha256-of-onion]
+
+If the ephemeral key in the onion is invalid:
+
+1. type: BADONION|PERM|5 (`invalid_onion_key`)
+2. data:
+   * [32:sha256-of-onion]
```
cdecker: I don't think we can distinguish a tampered onion packet from a tampered ephemeral key, since they both may end up producing garbage. We capture both cases with the HMAC.

04-onion-routing.md
```diff
@@ -463,6 +463,163 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+
+If an otherwise unspecified transient error occurs for the outgoing
+channel (eg. peer unresponsive):
+
+1. type: 6 (`temporary_channel_failure`)
+
```
cdecker: Would this also include channel exhaustion?

04-onion-routing.md
```diff
@@ -463,6 +463,163 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+1. type: PERM|14 (`incorrect_payment_amount`)
+
+If the `cltv-expiry` is too low, the final node MUST fail the HTLC:
+
+1. type: 15 (`final_expiry_too_soon`)
+
+### Receiving Failure Codes
+
+If the sending node is the final node:
```
cdecker: We should specify that this is the node sending the error, not the node initiating the transfer.

04-onion-routing.md
```diff
@@ -463,6 +463,163 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+  * The origin node MAY queue the `channel_update` for broadcast.
+* Otherwise, the origin node SHOULD eliminate the channel outgoing
+  from the sending node from consideration.
+  * If the PERM bit is not set, the origin node SHOULD restore the
+    channel as it sees a new `channel_update`.
+* The origin node SHOULD then retry routing and sending the payment.
+
+The origin node MAY use the data specified in the various types of
+failure for debugging purposes.
+
```
cdecker: What happens if we receive an error with an unknown type?

02-peer-protocol.md
```diff
@@ -590,6 +593,15 @@ For a timed out or route-failed HTLC:
...
    * [len:reason]
 
+For a unparsable HTLC:
+
+1. type: 133 (`update_fail_malformed_htlc`)
+2. data:
+   * [8:channel-id]
+   * [8:id]
+   * [4:failure-code]
+   * [32:sha256-of-onion]
```
cdecker: The `onion` refers to the full serialized onionpacket in the `update_add_htlc`, not any specific part of the onion right? Alternatively we could use the HMAC internal to the onionpacket, but we'd need to hand it one layer up.

02-peer-protocol.md
```diff
@@ -607,13 +619,22 @@ A receiving node which closes an incoming HTLC in response to an
 `update-fail-htlc` message on an offered HTLC MUST copy the `reason`
 field to the outgoing `update-fail-htlc`.
 
+A receiving node MUST fail the channel if the `BADONION` bit in
+`failure-code` is not set for `update-fail-malformed-htlc`.
```
cdecker: If I remember correctly the `failure-code` is opaque to the nodes forwarding the error, due to encryption. So this would only affect the first hop, i.e., transfer origin to first node in the path, despite the failure creator setting these bits. It'd allow anybody on a path kill that first channel, potentially uncovering the true origin. The alternative is to push the `failure-code` into the cleartext, which has other issues.

02-peer-protocol.md
```diff
@@ -607,13 +619,22 @@ A receiving node which closes an incoming HTLC in response to an
...
+`additional-data` to `sha256-of-onion`.
 
+#### Rationale
 
 A node which doesn't time out HTLCs risks channel failure (see
 "Risks With HTLC Timeouts").
 
+If the onion is malformed, the upstream node won't be able to extract
+a key to generate a response, hence the special failure message which
+makes this node do it.
```
cdecker: Good catch :+1:

04-onion-routing.md
```diff
@@ -444,13 +444,13 @@ The node returning the message builds a return packet consisting of the followin
...
- - `additional` (variable): a protocol level message, preceeded by the
-   its `type`, as defined by other specifications. For example a route
-   failure MAY include a `channel_update` message to inform the sender
-   that a channel is no longer usable.
+ - `additional` (variable): as specified under "Failure Codes" below.
+ - `pad length` (2 bytes): the length of `pad`.
+ - `pad` (variable): extra bytes to conceal length.
+
+The node SHOULD set `pad` such that the `additional length` plus `pad length` is equal to 128.
+This is 30 bytes longer than then the longest currently-defined message.
```
cdecker: I assume the pad length would eventually allow us to pad to other lengths in the future?

cdecker: Looking back to 27576f79463070398841e359a67be6763639b248, it seems we are specifying the length multiple times, once as payload length in the `update_fail_htlc` and once here as `pad length`. It's not like we do anything with the trailer (except include it in the HMAC computation) so we may as well drop `pad length` here.

02-peer-protocol.md
```diff
@@ -587,7 +587,8 @@ For a timed out or route-failed HTLC:
 2. data:
    * [8:channel-id]
    * [8:id]
-   * [154:reason]
+   * [2:len]
+   * [len:reason]
```
cdecker: Nevermind, efd970fd0de0e5fc7ca9ebe56a02c7c0d29e3131 re-establishes constant size.

04-onion-routing.md
```diff
@@ -429,13 +429,13 @@ Blinding is done by computing a blinding factor from the node's public key and t
...
 
-The protocol includes a simple mechanism to return messages to the origin node.
-The returned messages may either be failures reported by any intermediate hop, or a success message from the final node.
-The format of the forward packet is not usable for the return path, since no node other than the origin has the required information.
+The protocol includes a simple mechanism to return encrypted error messages to the origin node.
+The returned messages may either be failures reported by any intermediate hop, or the final node.
+The format of the forward packet is not usable for the return path, since no node other than the origin has the required information.  Note that these error messages are not reliable, as they are not placed on-chain in the case of node failure.
 
-Intermediate hops store the shared secret on from the forward path along and reuse it to obfuscate the return packets on each hop.
+Intermediate hops store the shared secret on from the forward path along and reuse it to obfuscate the error packet on each hop.
```
rustyrussell: Wow, I didn't even see that!

04-onion-routing.md
rustyrussell: Err, yeah. Fixed.

We could really use 2-byte error codes, but since you chose 4, I just started from the top bit down.

04-onion-routing.md
rustyrussell: Yes, this just covers grossly malformed: if onion key doesn't start with 0x2 or 0x3, it's invalid, for example; we can't even pretend to use it.

We should add another BADONION code for bad HMAC.

04-onion-routing.md
rustyrussell: Yep, exactly.

04-onion-routing.md
rustyrussell: yep, reworded.

02-peer-protocol.md
rustyrussell: Only your direct peer can send you update-fail-malformed-htlc: if it receives it from an upstream node it converts it to a normal update-fail-htlc.

It should *never* be trying to induce you send an update-fail-htlc with some random failure code!

cdecker: Ah, I mixed the two update-fail variants, ok. so this seems safe.

cdecker:
TIL Github magic would have allowed you to push your changes onto the `patch-04` branch on `cdecker/lightning-rfc`, which avoids opening new PRs :wink:

04-onion-routing.md
```diff
@@ -456,6 +466,169 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+
+If a node has requirement advertised in its `node_announcement`
+`features` which were not present in the onion:
+
+1. type: PERM|NODE|3 (`required_node_feature_missing`)
+
+A forwarding node MAY return one of the following errors, the final
+node MUST NOT:
+
+If the onion version byte is unknown:
```
Roasbeef: Should the three errors below really be a permanent failure? If I'm reading this correctly, then a `PERM` failure causes the node which originally sent the HTLC to remove the link that failed from their channel graph. In this case, it seems a bit extreme as

02-peer-protocol.md
```diff
@@ -566,14 +566,17 @@ still be under the maximum message size.  It also ensures that
...
 
 For simplicity, a node can only remove HTLCs added by the other node.
 There are three reasons for removing an HTLC: it has timed out, it has
 failed to route, or the payment preimage is supplied.
 
 The `reason` field is an opaque encrypted blob for the benefit of the
-original HTLC initiator as defined in [BOLT #4](04-onion-routing.md).
+original HTLC initiator as defined in [BOLT #4](04-onion-routing.md),
+but there's a special malformed failure variant for the case where
+our peer couldn't parse it; in this case the current node encrypts
```
cdecker: Wait, didn't we discuss this in Milan? I thought we agreed just to send back the hash of the message as we got it from our predecessor, I don't see the advantage in returning the whole packet.

04-onion-routing.md
cdecker: Well I understood the `PERM` bit as saying "I will fail this in the future as well if you retry", which for `BADONION|PERM|4` is correct, because the node will always reject any future onion with this version. It doesn't really result in the edge being removed.

04-onion-routing.md
```diff
@@ -456,6 +466,169 @@ Having the shared secrets of all intermediate nodes it can unwrap the packet unt
...
+
+If the ephemeral key in the onion is unparsable:
+
+1. type: BADONION|PERM|5 (`invalid_onion_key`)
+2. data:
+   * [32:sha256-of-onion]
+
+If the onion HMAC is incorrect:
+
+1. type: BADONION|PERM|6 (`invalid_onion_hmac`)
```
cdecker: If I'm not mistaken we can't really distinguish between the onion-key being invalid and the HMAC being invalid (unless the onion-key simply is not a valid secp point), not sure whether differentiating here is sensible.

02-peer-protocol.md
rustyrussell: Let's be clear,  C will see a bad MAC, and send ipdate_fail_malformed_htlc with  BADONION|PERM|6 + sha256(onion).  B could (in theory) check this sha, see C is on crack, and re-send (or close all channels with C because it's wigging out, and send back a permanent failure).  But if it doesn't, it encapsulates this into a normal fail message and pushed it back to A.

Now, A doesn't know if it's B or C's fault that it got garbage (it sees the SHA is wrong), but it doesn't really care: it discards that B->C channel from consideration and retries.  Which is kind of the Right Thing.

02-peer-protocol.md
```diff
@@ -606,13 +619,22 @@ A receiving node which closes an incoming HTLC in response to an
 `update-fail-htlc` message on an offered HTLC MUST copy the `reason`
 field to the outgoing `update-fail-htlc`.
 
+A receiving node MUST fail the channel if the `BADONION` bit in
+`failure-code` is not set for `update-fail-malformed-htlc`.
 
-#### Rationale
+A receiving node which closes an incoming HTLC in response to an
```
rustyrussell: Yeah, wording is awkward, I prefer yours.

It currently copies the entire reason field (which is failure-code + sha256-of-onion + anything we add later).  I think that's clearer for the bitflip case, but...

Perhaps we should tighten the spec, and say all BADONION messages MUST have contain only the sha256-of-onion as `reason`, and the node receiving update-fail-malformed-htlc MAY choose a different failure (or retransmission) if it does not match the onion it sent?

That allows you do detect bitflips from C and choose what to do.  I expect most will simply punt upstream.

04-onion-routing.md
rustyrussell: onion key has a 02 or 03 prefix; basically, parse failure.

04-onion-routing.md
rustyrussell: It's a corner case, which should Never Happen.  Something has gone terribly wrong: if it's the initiator's bug, it should leave the network and go home :)  If it's the forwarding node, you don't want to use it again.

I kind of erred on the side of caution: avoid the channel altogether in future.  Perhaps once you upgrade your software you'll want to undo that, though...

rustyrussell:
> One case I think is currently missing from this PR is the case that a node receive an incoming HTLC, but the current available capacity on the outgoing link is unable to carry the value of the HTLC. Following the schema outlined in the PR, I think this would be of base type UPDATE, with the additional data being the current available capacity of the channel. Such an error type would allow node to incrementally build up a view of the available capacity of all channels as failures occur, which can be used as supplemental real-time data when performing path finding.

For the moment, this would fall under `temporary_channel_failure`.  We could add another type, but the channel broadcasts do not (currently) include any capacity information, so no point sending that.

There are two capacity limits which could be hit: max number of htlcs, and limited funds.  I can see us adding more sophisticated limits in future (eg. total dust limit, to reduce the total number of sub-dust htlcs), and even implementing Random Early Drop to try to obfuscate limits.

The danger with providing finer grained information is exactly the problem of mapping the network; I think More Research Needed before we add this?

04-onion-routing.md
cdecker: Right, that'd be the case of an invalid key :+1:

Roasbeef:
>  but the channel broadcasts do not (currently) include any capacity information

Well they implicitly carry capacity information as they reference as particular output within the chain. This implicit capacity can and should be use by nodes when selecting a potential route for a payment during path finding. As you note this doesn't include a node's link-level constraints such as max-htlcs, min-htlc, dust-limit, etc.

> think More Research Needed before we add this?

Agreed, there're quite a number of tradeoffs that need to be explored on a live network.

04-onion-routing.md
```diff
@@ -511,15 +511,15 @@ If the onion version byte is unknown:
 2. data:
    * [32:sha256-of-onion]
 
-If the ephemeral key in the onion is unparsable:
+If the onion HMAC is incorrect:
```
cdecker: I don't think HMAC is the first check, if we can't parse the ephemeral key, then we can't generate the secret needed for the HMAC verification. So the check of the ephemeral key is implicitly done first. But let's not bike-shed this, I'm fine with either order :wink:

cdecker:
ACK edd4d75add43206af15d86cba4819394bfb66c12

rustyrussell:
OK, rebased and squashed fixes, thanks everyone!

PR #70, opened 2016-12-22
witnessScript is part of witness + null dummy for OP_CHECKMULTISIG

cjamthagen:


cjamthagen:
> This last item has been omitted for brevity in the rest of this document.

Don't know how I missed this. Reverted.

PR #75, opened 2017-01-04
BOLT 8: Test vectors for transport / key rotation.

rustyrussell:
Note that we increment nonce twice every message, meaning we rotate at msg 500.


08-transport.md
```diff
@@ -845,6 +845,25 @@ The responder should produce the given output when fed this input.
...
+1001 times (we only show 6 example outputs for brevity, and to test
+two key rotations):
+
+	name: transport-message test
+    ck=0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01
+	sk=0x969ab31b4d288cedf6218839b27a3e2140827047f2c0f01bf5c04435d43511a9
+	rk=0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442
+	output 0: 0x5bed0e4d7e2bc28afff2c05dd8fd7a24da81dc17be87e87504e5266a5301529467b98884e0b269
+	output 1: 0x6f5217771111a446ba1285e0849bb19f138441bf0404bdc432d287987285016afedb559d593297
+	# 0x2934606f0a74a2e1ba3f51965038cf1451d93c336974e98e5e9f9ac375d67911, 0x0fc678348c66aef07dc0775a34df85dbf68c54382c3c61551864c9eb4e9cd666 = HKDF(0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01, 0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442)
```
Roasbeef: If this is the initiator sending the message, then why is the receiver's key shown as being used in the `HKDF` key rotation?

08-transport.md
```diff
@@ -845,6 +845,25 @@ The responder should produce the given output when fed this input.
...
+
+In this test, the initiator sends length 5 messages containing "hello"
+1001 times (we only show 6 example outputs for brevity, and to test
+two key rotations):
+
+	name: transport-message test
+    ck=0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01
+	sk=0x969ab31b4d288cedf6218839b27a3e2140827047f2c0f01bf5c04435d43511a9
+	rk=0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442
+	output 0: 0x5bed0e4d7e2bc28afff2c05dd8fd7a24da81dc17be87e87504e5266a5301529467b98884e0b269
```
Roasbeef: I'm getting completely different results. What is `output` here? The full output of the ciphertext packet (including the length)?

08-transport.md
```diff
@@ -845,6 +845,25 @@ The responder should produce the given output when fed this input.
...
+## Message Encryption Tests
+
+In this test, the initiator sends length 5 messages containing "hello"
+1001 times (we only show 6 example outputs for brevity, and to test
+two key rotations):
+
+	name: transport-message test
+    ck=0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01
+	sk=0x969ab31b4d288cedf6218839b27a3e2140827047f2c0f01bf5c04435d43511a9
+	rk=0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442
```
Roasbeef: If this is the initiator sending, then why is the receiving key (which the initiator uses to decrypt message sent by the responder) part of the test vectors?

Roasbeef: Examined the vector a little closer, and it seems that these messages are actually the result of encrypting "hello" with the _receiving_ key rather than the sending key.

08-transport.md
Roasbeef: Actually, my results match up, but only on the latter half of each of the ciphertexts. This means our packet lengths are being encrypted differently.

Can you post your first few serialized packet length's as well (and possibly also nonces)? (in this comment thread)

Roasbeef: My key rotations however, do indeed match up (if I use the receiving key rather than the sending key).

08-transport.md
```diff
@@ -845,6 +845,25 @@ The responder should produce the given output when fed this input.
     input: 0x00b9e3a702e93e3a9948c2ed6e5fd7590a6e1c3a0344cfc9d5b57357049aa22355361aa02e55a8fc28fef5bd6d71ad0c38228dc68b1c466263b47fdf31e560e139bb
     output: ERROR (ACT3_BAD_TAG)
 
+## Message Encryption Tests
+
+In this test, the initiator sends length 5 messages containing "hello"
```
Roasbeef: This is the ASCII string "hello", right?

Roasbeef:
Why was this merged so quickly? I didn't get a chance to comment after I cross referenced the test vectors with my own implementation.

08-transport.md
sstone: I guess you're right and it seems that I swapped rk and sk when I generated my own test vectors (the encryption key that I used was 0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442)

08-transport.md
sstone: For the first output, I have:
k = 0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442
encrypt(k, 0x000000000000000000000000, 0x0005 ) = 0x5bed0e4d7e2bc28afff2c05dd8fd7a24da81

encrypt(k, 0x000000000100000000000000, "hello") = 0xdc17be87e87504e5266a5301529467b98884e0b269

which gives us:
 0x5bed0e4d7e2bc28afff2c05dd8fd7a24da81dc17be87e87504e5266a5301529467b98884e0b269

08-transport.md
zauguin: According to "Encrypting Messages": 

> The nonce sn is encoded as a 96-bit big-endian number.

So I think it should be 0x0000000000000000000000001 instead of 0x000000000100000000000000.

08-transport.md
zauguin: According to "Encrypting Messages": 

> The nonce sn is encoded as a 96-bit big-endian number.

So I think it should be 0x0000000000000000000000001 instead of 0x000000000100000000000000.

sstone: Good catch! We should update the "Encrypting Messages" section and  make it consistent with our definitions of encryptWithAd/decryptWithAd (which is based on the Noise specs): the nonce is encoded as 32 zero bits followed by a little-endian 64-bit value.

Roasbeef: Ah yeah, seems we forgot to update the spec from when we agreed to use little-endian nonces as defined in the Noise spec. I'll create a PR to update BOLT08 once I reconcile these test vectors with my implementation. 

I'm one step further now: I realized I was still including the length of the MAC in the length prefix for the cipher text (as that's how it was originally). I've changed this to match the spec, and now my results match up with the test vectors up until output 500.

08-transport.md
```diff
@@ -845,6 +845,25 @@ The responder should produce the given output when fed this input.
...
+two key rotations):
+
+	name: transport-message test
+    ck=0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01
+	sk=0x969ab31b4d288cedf6218839b27a3e2140827047f2c0f01bf5c04435d43511a9
+	rk=0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442
+	output 0: 0x5bed0e4d7e2bc28afff2c05dd8fd7a24da81dc17be87e87504e5266a5301529467b98884e0b269
+	output 1: 0x6f5217771111a446ba1285e0849bb19f138441bf0404bdc432d287987285016afedb559d593297
+	# 0x2934606f0a74a2e1ba3f51965038cf1451d93c336974e98e5e9f9ac375d67911, 0x0fc678348c66aef07dc0775a34df85dbf68c54382c3c61551864c9eb4e9cd666 = HKDF(0x919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01, 0xbb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442)
+	output 500: 0xbfd031ec37bfd43f29401e2c5a465256ec7efe5258e70d7b0271200afd24239f7d3adc01e0be1f
```
Roasbeef: Hey @sstone, can you also post your parameters to the `encrypt` function at this stage? From here on all the results differ completely from what I have (after fixing my length encoding error).

sstone: This is what I get:
```
encrypt(bb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442, 00000000e603000000000000, , 0005) = (be53, 1c31b9ef1b0e57a89b86a90aebbce45a)
encrypt(bb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442, 00000000e703000000000000, , 68656c6c6f) = (79fa511e24, 4a551eb8084ad1c6bfaa7290c2a4ebe0)
output 499: 0xbe531c31b9ef1b0e57a89b86a90aebbce45a79fa511e244a551eb8084ad1c6bfaa7290c2a4ebe0

key rotation 
(2934606f0a74a2e1ba3f51965038cf1451d93c336974e98e5e9f9ac375d67911, 0fc678348c66aef07dc0775a34df85dbf68c54382c3c61551864c9eb4e9cd666) = hkdf(919219dbb2920afa8db80f9a51787a840bcf111ed8d588caf9ab4be716e42b01, bb9020b8965f4df047e07f955f3c4b88418984aadc5cdb35096b9ea8fa5c3442)

encrypt(0fc678348c66aef07dc0775a34df85dbf68c54382c3c61551864c9eb4e9cd666, 000000000000000000000000, , 0005) = (bfd0, 31ec37bfd43f29401e2c5a465256ec7e)
encrypt(0fc678348c66aef07dc0775a34df85dbf68c54382c3c61551864c9eb4e9cd666, 000000000100000000000000, , 68656c6c6f) = (fe5258e70d, 7b0271200afd24239f7d3adc01e0be1f)
output 500: 0xbfd031ec37bfd43f29401e2c5a465256ec7efe5258e70d7b0271200afd24239f7d3adc01e0be1f
```

PR #76, opened 2017-01-05
BOLT 2: Message retransmission and reconnect semantics.

rustyrussell:
1) Make it clear that `init` needs to be sent every time.
   - This means if you upgrade and no longer support an old connection, it's
     clear, plus it simplifies the question of re-transmission of `init`.
2) Spell out the retransmission requirements for reconnection.
   - We agreed in Milan to simply use retransmit and ignore-dups.
   - This needs actual testing by implementations, but this is my best guess
     on exactly how far back to retransmit.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -21,6 +21,7 @@ operation, and closing.
       * [Committing Updates So Far: `commitment_signed`](#committing-updates-so-far-commitment_signed)
       * [Completing the transition to the updated state: `revoke_and_ack`](#completing-the-transition-to-the-updated-state-revoke_and_ack)
       * [Updating Fees: `update_fee`](#updating-fees-update_fee)
+    * [Message Retransmission](#message-retransmission)
   * [Authors](#authors)
   
 # Channel
@@ -127,6 +128,11 @@ allow commitment transactions to propagate through the Bitcoin
 network.  It SHOULD set `htlc-minimum-msat` to the minimum
 amount HTLC it is willing to accept from this peer.
 
+The receiving node MUST accept a new `open-channel` message if the
+connection has been re-established after receiving a previous
+`open-channel` and before receiving a `funding-created` message.  In
+this case, the receiving node MUST discard the previous `open-channel`
+message.
 
 The receiving node MUST fail the channel if `to-self-delay` is
 unreasonably large.  The receiver MAY fail the channel if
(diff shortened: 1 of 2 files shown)


02-peer-protocol.md
```diff
@@ -778,6 +784,73 @@ it is simplest to only allow them to set fee levels, but as the same
...
+
+### Requirements
+
+A node MUST handle continuing a previous channel on a new encrypted
+transport.  On disconnection, a node MAY forget nodes which have not
+sent or received an `accept_channel` message.
+
+On disconnection, a node MUST reverse any uncommitted updates sent by
+the other side (ie. all messages beginning with `update_` for which no
+`commitment_signed` has been received).  A node SHOULD retain the `r`
```
Roasbeef: I think it'd be good to spell out the rational behind recommending the node to retain the `r` value in between re-connections. Something along the lines of

> We recommend that a node retains the `r` values from `update_fulfill_htlc` messages as it allows them to directly sweep the value in the HTLC output, sidestepping the absolute timeout in the case of a commitment broadcast.

02-peer-protocol.md
```diff
@@ -778,6 +784,73 @@ it is simplest to only allow them to set fee levels, but as the same
...
+value from the `update_fulfill_htlc`, however.
+
+On reconnection, a node MUST retransmit old messages which may not
+have been received, and MUST NOT retransmit old messages which have
+been explicitly or implicitly acknowledged.  The following table
+lists the acknowledgement conditions for each message:
```
Roasbeef: Nice breakdown! I think this table succinctly describes the retransmission/acknowledgment criteria.

02-peer-protocol.md
```diff
@@ -778,6 +784,73 @@ it is simplest to only allow them to set fee levels, but as the same
...
+
+* `open_channel`: acknowledged by `accept_channel`.
+* `accept_channel`: acknowledged by `funding_created`.
+* `funding_created`: acknowledged by `funding_signed`.
+* `funding_signed`: acknowledged by `funding_locked`.
+* `funding_locked`: acknowledged by `update_` messages, `commitment_signed`, `revoke_and_ack` or `shutdown` messages.
+* `update_` messages: acknowledged by `revoke_and_ack`.
+* `commitment_signed`: acknowledged by `revoke_and_ack`.
+* `revoke_and_ack`: acknowledged by `shutdown`, `update_` messages, or `commitment_signed`
+* `shutdown`: acknowledged by `closing_signed` or `revoke_and_ack`.
```
Roasbeef: Why would a `revoke_and_ack` message acknowledge a `shutdown` message?

02-peer-protocol.md
```diff
@@ -778,6 +784,73 @@ it is simplest to only allow them to set fee levels, but as the same
...
+the other side (ie. all messages beginning with `update_` for which no
+`commitment_signed` has been received).  A node SHOULD retain the `r`
+value from the `update_fulfill_htlc`, however.
+
+On reconnection, a node MUST retransmit old messages which may not
+have been received, and MUST NOT retransmit old messages which have
+been explicitly or implicitly acknowledged.  The following table
+lists the acknowledgement conditions for each message:
+
+* `open_channel`: acknowledged by `accept_channel`.
```
Roasbeef: Actually, do we _really_ want to force nodes to remember all the scripts, keys, and hashes required to initialize (then possibly remember) a funding workflow in-between successful reconnections? As is defined currently, a node would need to persist all the relevant funding context to disk in order to be able to resume the workflow if either side flaps before completion. 

I think the current semantics might open us up to a bit of a "disk flooding" attack, whereby an attacker spins up several identities, starts the funding process with each of them, but then never responds with a `funding_created` message. As a result, we'd possibly need to hold on to all the prior funding contexts so we can properly respond with a `funding_created` message

Instead, we could only define transmission semantics for messages sent _after_ the `funding_signed` message is received. Only at this point do both sides truly have skin in the game, therefore they should be prepared to handle retransmissions properly.

02-peer-protocol.md
rustyrussell: Because there's no explicit acknowledge: even if they reply with `shutdown` that may be them just happening to shutdown at the same time.

You know they've got it when they acknowledge a following `commitment_signed` with `revoke_and_ack` or when the actually start the closing.  I will note here that the `revoke_and_ack` must be for a `commitment_signed` sent following the `shutdown`, not before.

02-peer-protocol.md
rustyrussell: But if you don't lock like this, then it's tricky.  If you send funding_signed then forget about it (because they disconnect), what happens when they reconnect?  What if you choose different keys?

If we change `open_channel` to be acknowledged by `funding_signed`... then accept_channel still has the problem.  Do we need another step in the handshake where we actually send the tx?  Then funder really is committed...

02-peer-protocol.md
rustyrussell: Actually, that's the other node's problem.  The real reason is that knowledge is not revocable, so as soon as the expose the payment preimage, you can (and should!) use it to claim any upstream payment.  So, really, you should have already used the `r` value...

02-peer-protocol.md
rustyrussell: Note: after IRC discussion (see https://botbot.me/freenode/lightning-dev/2017-01-17/?msg=79466673&page=1 ) we agreed to leave this as-is, but suggest a timeout, as all other approached are fraught.

Issue #73, opened 2017-01-03
BOLT 2: differing channel-id fails channel


cjamthagen:
What if two peers who want to open a channel with each other sets `minimum-depth` to 1 and there are two blocks created simultaneously (same height) with the funding transaction in both of them, and each peer receives a different block. They will send the `funding_locked` message with (most likely) different `channel-id`, and they must now fail the channel according to the spec. Should the protocol not handle this case?

pm47:
This is a duplicate of #6, but AFAICT the `minconf=6` didn't actually make it to the spec. I'd be in favor of adding it to BOLT 2.

rustyrussell:
Christopher Jämthagen <notifications@github.com> writes:
> What if two peers who want to open a channel with each other sets `minimum-depth` to 1 and there are two blocks created simultaneously (same height) with the funding transaction in both of them, and each peer receives a different block. They will send the `funding_locked` message with (most likely) different `channel-id`, and they must now fail the channel according to the spec. Should the protocol not handle this case?

It does; as you said: they will fail the channel :)

blockchain.info shows 27 orphans in the last 3 months, or ~ one every 3
days.  We could change the spec to allow retransmissions, instead, such
instead of failing, you may wait for agreement.

How's this?

diff --git a/02-peer-protocol.md b/02-peer-protocol.md
index b91a3ec..56d3bdd 100644
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -263,14 +263,20 @@ The sender MUST wait until the funding transaction has reached
 `minimum-depth` before sending this message.  The sender MUST encode
 the block position of the funding transaction into `channel-id`.  If
 the sender has already received `funding_locked` from the other node,
-it MUST fail the channel if its own `channel-id` does not match the
-received.  The sender MUST set `next-per-commitment-point` to the
+it MAY fail the channel if its own `channel-id` does not match the
+received: otherwise it MUST ignore the `funding_locked` message.
+The sender MAY re-transmit `funding_locked` if the `channel-id` changes.
+
+The sender MUST set `next-per-commitment-point` to the
 per-commitment point to be used for the following commitment
 transaction, derived as specified in
 [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
 
-If the recipient has already sent `funding_locked` it MUST fail the
-channel if `channel-id` does not match the `channel-id` it sent.
+If the recipient has already sent `funding_locked` it MAY fail the
[...]

PR #77, opened 2017-01-05
BOLT 2: allow more leniancy with forks during channel establishment.

rustyrussell:
Christoper points out that two nodes with aggressive minimum-depth settings
may see different blocks and the protocol requires they close the channel
since their funding_locked messages will disagree.

This can also happen when only one side has an aggressive minimum-depth
setting: if it sends funding_locked referring to a block which is orphaned,
it can't update it.

There are three changes here, two optional.
- Allow sending of an updated funding_locked.  This fixes this case where
  one side is on an orphan and uses a v. low minimum-depth.
- Require accepting of an updated funding_locked.
- Allow waiting instead of immediate failure if funding_lock disagrees.
  eg. you might wait another block or two to see if one side reorgs.

Reported-by: Christopher Jämthagen
Closes: #73

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -263,14 +263,20 @@ The sender MUST wait until the funding transaction has reached
 `minimum-depth` before sending this message.  The sender MUST encode
 the block position of the funding transaction into `channel-id`.  If
 the sender has already received `funding_locked` from the other node,
-it MUST fail the channel if its own `channel-id` does not match the
-received.  The sender MUST set `next-per-commitment-point` to the
+it MAY fail the channel if its own `channel-id` does not match the
+received: otherwise it MUST ignore the `funding_locked` message.
+The sender MAY re-transmit `funding_locked` if the `channel-id` changes.
+
+The sender MUST set `next-per-commitment-point` to the
 per-commitment point to be used for the following commitment
 transaction, derived as specified in
 [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -263,14 +263,20 @@ The sender MUST wait until the funding transaction has reached
 `minimum-depth` before sending this message.  The sender MUST encode
 the block position of the funding transaction into `channel-id`.  If
 the sender has already received `funding_locked` from the other node,
-it MUST fail the channel if its own `channel-id` does not match the
-received.  The sender MUST set `next-per-commitment-point` to the
+it MAY fail the channel if its own `channel-id` does not match the
+received: otherwise it MUST ignore the `funding_locked` message.
+The sender MAY re-transmit `funding_locked` if the `channel-id` changes.
```
cdecker: I think we also need to specify that a node MUST retransmit the `funding_locked` message if the `channel-id` changes, otherwise we may never make progress. If both nodes silently drop the `funding_locked` message, but never issue send the changed `channel-id` then we just sit there and wait forever. This may also include sending a `funding_locked` at a bigger depth than what we were waiting for.

rustyrussell: I think we can fix this by tying the two together: If either node chooses to ignore, it must re-transmit when it changes.  Because if either node is doing the naive "fail on disagree" it will have failed.

cdecker:
ACK 83e04bf1cf718fe2732adecf6717f31594a65cab

PR #78, opened 2017-01-06
BOLT 2: fix duplicate numbers.

rustyrussell:
revoke-and-ack is already 133.


cdecker:
Simple enough, I'll just merge it

PR #81, opened 2017-01-10
bolt04: Added onion routing test vectors for forward path

cdecker:
For now I just generated the wrapping direction since the unwrapping
relies on the steps in reverse. It's a 5-hop onion packet, padded to
the standard 20 hop size. Associated data and per-hop-payloads are
filled with dummy values, not the format described in the bolt.


04-onion-routing.md
```diff
@@ -629,6 +629,114 @@ Otherwise, the node sending the error is an intermediate node:
...
+
+	hop_payload[0] = 0x4141414141414141414141414141414141414141
+	hop_payload[1] = 0x4141414141414141414141414141414141414141
+	hop_payload[2] = 0x4141414141414141414141414141414141414141
+	hop_payload[3] = 0x4141414141414141414141414141414141414141
+	hop_payload[4] = 0x4141414141414141414141414141414141414141
+
+#### Per-Hop Information
+
+	hop_shared_secret[0] = 0xa6254636b0f9fe6a871ef7057de6b1fa0f31e6f0e46990747fda49f7871a9f1f
```
sstone: it seems that you're using the "old" variant of ECDH (which returns sha256((pub.priv).x) ?

cdecker: Great catch, I must have missed that. I'll change the Go and the C implementation to the newest spec ASAP :+1:

sstone: Otherwise I get the same results. Thanks!
By the way, if think that there is a typo in the specs: line 254 reads: "... The blinding factor is the SHA256 hash of the concatenation between the node's public key nodepk_k and ...".  Shouldn't it be the ephemereal public key instead of the node's public key ?

cdecker:
Sorry about that @Roasbeef I seem to have forgotten my push to the LightningNetwork/lightning-onion repo.

PR #84, opened 2017-01-13
Fix packet format extraction tool, add failure codes as packets and add CI

cdecker:
The extraction tool was broken by the recent failure codes patch, since the failure codes description was using the same format that the tool recognized as packet types. This seems to be desirable since it allows us to reuse that tool to create parsers for the failure payloads as well. So I went ahead and fixed the tool to allow packets with no payloads and match masked types as well.

I also added travis-ci (must still be enabled by a repo owner) so that the extraction tool is run against new commits and we get feedback on breakages. Maybe we can expand this to also enforce some spec styles in the future?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,7 @@
+language: python
+python:
+  - "3.4"
+  - "3.5"
+  - "3.6"
+script:
+  - (set -e; for i in 0?-*.md; do echo "Extracting $i"; python3 tools/extract-formats.py --message-types --message-fields --check-alignment $i; done)
--- a/tools/extract-formats.py
+++ b/tools/extract-formats.py
@@ -64,15 +64,17 @@ parser.add_option("--message-fields",
 
 (options, args) = parser.parse_args()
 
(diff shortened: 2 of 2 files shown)


04-onion-routing.md
```diff
@@ -442,7 +442,8 @@ Intermediate hops store the shared secret from the forward path and reuse it to
 In addition each node locally stores the previous hop it received the forward packet from, in order to determine where to send an eventual return packet.
 The node returning the message builds a return packet consisting of the following fields:
 
-1. data:
+1. type: 255 (`error-payload`)
```
rustyrussell: Erk, let's not do this.  If we really want the tool to extract this we'll need to make the tool smarter: corrupting the doc with a fake number really doesn't sit well :(

rustyrussell:
OK, I had same forgotten script fixes sitting unpushed in my tree :(

So I merged them with your travis addition and force-pushed (I also neatened the script and added a comment  showing what example formats it accepts).

Issue #90, opened 2017-01-25
BOLT 3: Appendix B: Transactions Test Vectors invalid keys?


rustyrussell:
local_funding_key: 30ff4956bbdd3222d44cc5e8a1261dab1e07957bdac5ae88fe3261ef321f374901
    remote_funding_key: 1552dfba4f6cf29a62a0af13c8d6981d36d0ef8d61ba10fb0fe90da7634d7e1301
    local_revocation_key: 131526c63723ff1d36c28e61a8bdc86660d7893879bbda4cfeaad2022db7c10901
    local_payment_key: e937268a37a774aa948ebddff3187fedc7035e3f0a029d8d85f31bda33b02d5501
    remote_payment_key: ce65059278a571ee4f4c9b4d5d7fa07449bbe09d9c716879343d9e975df1de3301

None of these are valid keys; they should start with 03 or 02, and why do they all end in 01?

The names also don't match the names in the rest of the document, but I was hoping to figure that out from the test vectors and submit a patch...

sstone:
Ah yes, the names are confusing sorry about that: these are private keys serialized with the bitcoin convention of adding an extra 01 at the end to tag them as "compressed private keys" (i.e the public keys derived from these private keys are compressed). I will submit a patch.

rustyrussell:
Ah, OK, that makes much more sense!

rustyrussell:
Closed by #99

PR #91, opened 2017-01-25
BOLT 3: fix transaction test vectors 

sstone:
- rename keys properly (see #90)
- fix fees
- fix remote-output (it's a P2PKH output, we incorrectly used a P2WPH output)
- sign all transactions


rustyrussell:
OK, this still leaves me very confused.  This needs many more comments in the test vectors, IMHO: lines starting with # which help implementers debug where they went wrong.  As my teacher used to say "show your working!".

What commitment tx number are we producing?  0? How do we produce the obscured commitment transaction number?  That needs the payment-basepoint for both sides for the obscuring factor.  It seems you've used commitment number 42, and the local_payment_key and remote_payment_key as the basepoints?  Thats wrong...

For HTLC outputs and the to-local output, we need the delayed-localkey, and we don't have it, either.

I'm reworking this now...

sstone:
Our intent was to generate basic tx test vectors that can be used to check scripts, ordering, ... and make sure that all implementation agree on a basic level, because we believe that it makes building implementations incrementally easier. This is why we did not include key derivation (which makes generation and verification of test vectors more involved). They are meant as: "if keys look like this, then txs should look like this".
 
But you're right: the transaction number does use payment basepoints, we cannot just pick random keys, instead we need to derive them from the same basepoints. So there is really a problem here, which I missed when I added the commit tx number obfuscation and updated our test vectors :(

I've fixed the txs (the to-remote ouput has changed anyway), added comments and updated the PR

sidenote: Regarding the `delayed-localkey`, in our implementation we used it everywhere in the local tx, both for the delayed payment outputs (main+htlc) and in the multisig htlc outputs. This is not what the spec says (mea maxima culpa!) and I just realized it. OTOH, the way the spec is currently written, the same `localkey` will be used in two otherwise unrelated transactions (with the same tx index): a local tx (for the multisig htlc output) and a remote tx (for direct to-us payments). Is that really a good thing?

rustyrussell:
> This is not what the spec says (mea maxima culpa!) and I just realized it. OTOH, the way the spec is currently written, the same localkey will be used in two otherwise unrelated transactions (with the same tx index): a local tx (for the multisig htlc output) and a remote tx (for direct to-us payments). Is that really a good thing?

Yes, I considered adding another key, but it doesn't help at all.  You can tell from the forms of the script which party is being paid for both cases anyway.

rustyrussell:
After discussion, this was closed in favor of #99 which Fabrice already debugged for me :)

PR #95, opened 2017-01-31
BOLT 4: Use 4 bytes for outgoing_cltv_value

pm47:
The cltv value in the `per-hop-payload` packet is an absolute value, not a delta, therefore it needs to be encoded on 4 bytes just like `update_add_htlc`.`cltv-expiry`.

To keep the packet length at 20 bytes I removed 2 bytes from the following unused field.


rustyrussell:
Ack, good catch.

PR #97, opened 2017-01-31
BOLT 2&7: Cleaner separation of concerns wrt announcement signatures

cdecker:
So far we did not have any indication on what to do if a node does not
allow announcing the channel and we had a mix of concerns in the
`funding_locked` message, which would also transfer the signatures
needed for the announcement. This is a proposal about splitting the
signatures into their own message, so that simple omission is an
opt-out of announcements, and it does not mix announcement/gossip
stuff into the peer-protocol.

This builds on top of #96. There is some concern that pulling this
message out of the strictly controlled state machine of the channel
lifecycle may have unwanted consequences.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -23,13 +23,12 @@ It contains the necessary signatures by the sender to construct the `channel_ann
     * [64:node-signature]
     * [64:bitcoin-signature]
 
-An `announcement_signatures` message MUST NOT be sent before the channel is fully established, i.e., before both ends have received the corresponding `funding_locked` messages.
-
-Assuming both endpoints agree to announce the channel, then they SHOULD create a `channel_announcement` message, and sign it with the secrets matching their `node-id` and `bitcoin-key`, and send them using an `announcement_signatures`.
-Should one or both endpoints not want to announce the channel to the rest of the network, then they MUST NOT send an `announcement_signatures` message, and ignore any incoming `announcement_signatures` messages.
-
+The willingness of the endpoints to announce the channel is signaled during the connection setup by setting the first bit (`0x01`) in the `localfeatures` field.
+Should both endpoints have signaled that they'd like to publish the channel (`localfeatures & 0x01 == 0x01`) then the `announcement_signatures` message MUST directly sent following the `funding_locked` message that established the corresponding channel.
+The `announcement_signatures` message is created by constructing a `channel_announcement` message corresponding to the newly established channel, and sign it with the secrets matching their `node-id` and `bitcoin-key`, and send them using an `announcement_signatures`.
 The recipient MAY fail the channel if the `node-signature` or `bitcoin-signature` is incorrect.
 The recipient SHOULD queue the `channel_announcement` message for its peers if it has sent and received a valid `announcement_signatures` message.
+If either endpoints does not signal (`localfeatures & 0x01 == 0x00`) then `announcement_signatures` MUST NOT be sent.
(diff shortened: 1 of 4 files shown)


cdecker:
I definitely need new reading glasses, sorry for missing those :-)

I amended your first two points, but I am a bit wary of mixing the channel lifecycle messages with the routing messages. The exchange of the routing signatures is not time-critical, so we could just move the signature exchange at a later point in the execution, outside of the strict state machine. Retransmission is also rather simple: if we haven't seen our channel announced we can just send the signatures again.

rustyrussell:
> I definitely need new reading glasses, sorry for missing those :-)
>
> I amended your first two points, but I am a bit wary of mixing the channel lifecycle messages with the routing messages. The exchange of the routing signatures is not time-critical, so we could just move the signature exchange at a later point in the execution, outside of the strict state machine. Retransmission is also rather simple: if we haven't seen our channel announced we can just send the signatures again.

I'd agree, except you can't distinguish between "they didn't want to
announce the channel" and "they didn't get my message" or "I didn't get
their message".

The other reliable way to do it is *always* re-transmit on new
connection if we've agreed on funding locked, unless you've seen an
announcement.  But that still needs to be specified in the reconnection
paragraph, so it's no cleaner :(

Cheers,
Rusty.

cdecker:
I like the "just send signature on open/reopen" much better, it's an event we are hooking into anyway to do the initial announcement.

pm47:
>  I am a bit wary of mixing the channel lifecycle messages with the routing messages

I share Christian's concerns, but I am still slightly in favor of the existing explicit opt-out mechanism because I think it makes the channel creation flow simpler and easier to understand (even if it comes at the expense of the separation of concerns).

Also I like the fact that we know from the get-go if the channel will ever be 'public' or 'private' from a user's perspective.

07-routing-gossip.md
```diff
@@ -13,6 +13,24 @@ There can only be one valid `channel_announcement` for any channel,
...
+This is a direct message between two endpoints of a channel and serves as an opt-in mechanism to allow the announcement of the channel to the rest of the network.
+It contains the necessary signatures by the sender to construct the `channel_announcement` message.
+
+1. type: 259 (`announcement_signatures`)
+2. data:
+    * [8:channel-id]
+    * [64:node-signature]
+    * [64:bitcoin-signature]
+
+An `announcement_signatures` message MUST NOT be sent before the channel is fully established, i.e., before both ends have received the corresponding `channel_locked` messages.
```
Roasbeef: `channel_locked` should be `funding_locked`

rustyrussell:
OK, I think we want to add a feature bit.  That should be easy:
(1) document the feature bit, say bit #1 (odd is OK).  This is a localfeature, IIUC.
(2) if both sides set feature bit, `announcement_signatures` MUST follow `funding_locked`, otherwise it MUST NOT.
(3) amend retranmission section and that sentence in bolt 02 "Once both nodes have exchanged funding_locked" to add "(and optionally announcement_signatures)".

cdecker:
Ok, amended the PR to include the opt-in localfeature, added a tracking document for the assigned feature flags and added the restriction to send it inbetween the `funding_locked` and any state update.

One minor concern is that the featureflag now impacts all channels on this connection (`open_channel` and `accept_channel` currently don't have a features field), but that's likely not a problem. Still think we can do without the strict ordering requirements, but that's also a minor thing.

PR #102, opened 2017-02-07
BOLT 9: assign feature bits in pairs, give them names, clarify position.

rustyrussell:
Christian assumed first bit was 1, I assumed 0.  And we should generally
assign in pairs (so an optional understanding can later become compulsory),
though for the initial draft it's unnecessary.

By giving names we avoid smearing values over the spec, containing them in
BOLT 9.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -23,12 +23,12 @@ It contains the necessary signatures by the sender to construct the `channel_ann
     * [64:node-signature]
     * [64:bitcoin-signature]
 
-The willingness of the endpoints to announce the channel is signaled during the connection setup by setting the first bit (`0x01`) in the `localfeatures` field.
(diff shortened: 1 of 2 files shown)


cdecker:
Absolutely agree, I don't know why I just assumed that we'd be 1-basing, and we should definitely assign pairs. If we allow both bits we also need to specify all 3 behaviors: both opt in, both opt out, one opts in and one opts out. Currently the last case will kill the connection.

rustyrussell:
Hmm, not sure what you mean.  Bits are in pairs because of the "odd is OK" rule; if you don't *understand* a bit and it's even, you have to fail/not connect.  If you understand it, the meaning is defined by the spec.

So, we could say "bit 0 means you MUST support announce or I'll fail you", "bit 1 means we will announce IFF we both agree".  Or as currently specified, they're the same (except for the general "if you don't understand, you have to fail" property of feature bits).

I should spell this out properly, too...

cdecker:
Right, I'm probably overthinking this, I guess everybody will just advertise the odd flag.

PR #103, opened 2017-02-07
BOLT 7: Added flag for optional initial routing sync dump

cdecker:
Opening a lot of connections results in getting this information a
whole lot of times, so let's add an opt-in flag for the initial dump.

This will cause a conflict with #102, I'll rebase once that is merged.


09-features.md
```diff
@@ -12,5 +12,5 @@ This flags may only be used in the `init` message.
 | Bits | Description                                     | Link                                                                |
 |------|-------------------------------------------------|---------------------------------------------------------------------|
 | 1    | The sending node wishes to announce the channel | [BOLT #7](07-routing-gossip.md#the-announcement_signatures-message) |
-
+| 2/3  | `initial_routing_sync` | The sending node needs a complete routing information dump | [BOLT #7](07-routing-gossip.md#initial-sync) |)
```
Roasbeef: The format of this new entry doesn't match up with that of the one above. I thin they should be made to be consistent.

cdecker: Yup, that's what I meant by dependency on #102. Rebased and merged. Should be ready to go ^^

Issue #88, opened 2017-01-23
BOLT3: make HTLC outputs directly spendable by penalty transactions


sstone:
We would like to propose a small modification of the commiment transaction HTLC scripts, which would make it possible, when the other party publishes a revoked commit transaction, to create penalty transactions that spend the HTLC outputs. With  the current design the penalty transaction spends the second-stage HLTC-timeout and HTLC-success transactions. 

This does not change the way external channel monitors would work, but for node which do their own monitoring this has several advantages:
- for htlcs offered by the counterparty: we don't need to know the counterparty's HTLC-timeout signature, which simplifies BOLT2 and saves the burden of storing the signatures
- for htlcs received we don't need to wait for the htlc timeout refund to claim the funds 

The resulting transaction tree would be:
```
+------------+
| funding tx |
+------------+
      |
      |        +-------------+
      \--------| commit tx B |
               +-------------+
                  |  |  |  |  
                  |  |  |  | A's main output
                  |  |  |  \------------------ to A
                  |  |  |
                  |  |  |
                  |  |  |                  ,-- to B (& delay)
                  |  |  | B's main output /
                  |  |  \----------------<
                  |  |                    \ 
                  |  |                     `-- to A (& revocation key)
                  |  |
                  |  |                                                ,-- to B (& delay)
                  |  |                        +-----------------+    /
                  |  |                     ,--| HTLC-timeout tx |---<
                  |  | HTLC offered by B  /   +-----------------+    \
                  |  \-------------------<     (after timeout)        `-- to A (& revocation key)
                  |                       \
                  |                        `-- to A (& payment preimage)
                  |                        \
                  |                         `- to A (& revocation key)
                  |                   
                  |                                                   ,-- to B (& delay)
                  |                           +-----------------+    /
                  |                        ,--| HTLC-success tx |---<
                  | HTLC received by B    /   +-----------------+    \
                  \----------------------<     (w/ payment preimage)  `-- to A (& revocation key)
                                          \
                                           `-- to A (after timeout)
                                           \
                                            `- to A (& revocation key)
```

## Changes to the Offered HTLC Output script:

The solution we propose is to use a multisig 2-of-3 with `localkey`, `remotekey` and `revocationkey` for the Offered HTLC Output script:
```
<remotekey> OP_SWAP
    OP_SIZE 32 OP_EQUAL
OP_NOTIF
    # To me via HTLC-timeout transaction (timelocked) or to you with revocation key.
[...]

PR #105, opened 2017-02-10
BOLT 2,3,5: Make htlc outputs of the commitment tx spendable with revocation key

pm47:
Please see #88 for the rationale.

This PR updates:
- BOLT 3: scripts, fees, test vectors
- BOLT 2: `RevokeAndAck` message
- BOLT 5: handling of revoked tx and weight calculations

---

Tradeoffs:

:heavy_plus_sign: no more sending of htlc-timeout signatures in `RevokeAndAck`
* protocol is simplified
* no need to generate signatures

:heavy_plus_sign: no more storage of htlc-timeout signatures
* for offered htlc we don't need to store anything at all
* for received htlcs we might need to store htlc expiries, or just brute-force them (maybe with MAST we could entirely get rid of this issue?)

:heavy_plus_sign: if a recent revoked commit tx is published, if the counterparty doesn't publish her htlc-success tx, we can immediately redeem the funds without waiting for the expiry

:heavy_minus_sign: htlc scripts are larger (and htlc-received script is kind of ugly)
* HTLC-timeout weight: 635 -> 669 (+5%)
* HTLC-success weight: 673 -> 718 (+7%)

:heavy_minus_sign: maximum number of htlcs in a single penalty tx is down from 626 to 424 for the worst case scenario (although it might not be a good idea to use a single tx because it can be invalidated if the counterparty published a 2nd-stage htlc tx)

---

If anything, I think it would be worth it to just change the htlc-offered script: the modification is minimal both in script complexity and weight, and we would keep most of the benefits (but we would also loose the symmetry with htlc-received behavior).


cdecker:
I quite like this proposal, it simplifies things and comes with a moderate potential cost increase. Hopefully we don't drop to chain too often so that the actual cost increase is moderate as well. If that's the case we can always go back and revert this change.

So for me it's a :+1:

rustyrussell:
Hmm, there's a variant which removes the need for the extra signatures, but only adds ~11 bytes to the witness script for accepted HTLCs: we add a further CLV timeout (eg. locktime + to-self-delay) after which the localkey is sufficient to spend.  This avoids having to put the revocation key into the witness script.

rustyrussell:
#### Received HTLC Outputs

This output sends funds to the remote peer after the HTLC timeout, or to an HTLC-success transaction with a successful payment preimage. The output is a P2WSH, with a witness script:

    <remotekey> OP_SWAP
        OP_SIZE 32 OP_EQUAL
    OP_IF
        # To me via HTLC-success transaction.
        OP_HASH160 <ripemd-of-payment-hash> OP_EQUAL
		<localkey> OP_SWAP
		OP_IF
			2 OP_ROT 2 OP_CHECKMULTISIG
		OP_ELSE
			OP_NIP <locktime+to-self-delay> OP_CHECKLOCKTIMEVERIFY OP_DROP OP_CHECKSIG
		OP_ENDIF
    OP_ELSE
        # To you after timeout.
        OP_DROP <locktime> OP_CHECKLOCKTIMEVERIFY OP_DROP
        OP_CHECKSIG
    OP_ENDIF

rustyrussell:
OK, it's this vs. #116  ; both get rid of the sending of extra signatures (yay!), but this is 37 witness bytes larger.  On the other hand, #116 requires waiting for the timeout for the penalty transaction, which this does not.

Please vote in comments!

I slighly prefer #116 due to blockchain bloat, but I agree it's marginal.

Roasbeef:
I'm having trouble following how the additional timeout on the scripts for received HTLC's doesn't assume some ordering/delay timeout and introduce an additional race condition. It seems that if I (the HTLC receiver) broadcast a prior _revoked_ commitment that's from a sufficient number of days back (enough so the CLTV is a no-op), then I can snag the HTLC w/o any penalty. Am I missing something here?

rustyrussell:
Oops, I flipped the script :(  I'll fix that today.

For A's commitment transaction; the change only applies to HTLCs offered by B (ie. A's received HTLC), and should go to B of course, not A!

Roasbeef:
Excellent work @pm47, very clever usage of Bitcoin's scripting! I prefer this proposal over #116 for the reasons listed in my comment in the other PR.

05-onchain.md
```diff
@@ -315,32 +320,33 @@ broadcasts HTLC-timeout and HTLC-success transactions, but the
...
+    accepted_htlc_penalty_witness: 307 bytes
 
-This gives 1+3+73+1+9+66+2=155 bytes of witness data, weight 155.
+The penalty txinput itself takes 41 bytes, thus has a weight of 164, making the weight of each input:
 
-The penalty txinput itself takes 41 bytes, thus has a weight of 164,
-meaning each input adds 319 weight.
+    to_local_penalty_input_weight: 318 bytes
+    offered_htlc_penalty_input_weight: 455 bytes
+    accepted_htlc_penalty_input_weight: 471 bytes
```
Roasbeef: The variables denote weight, yet the units here are in bytes.

Roasbeef: Same with the witness weight description above.

03-transactions.md
```diff
@@ -107,13 +107,13 @@ This output sends funds to the other peer, thus is a simple P2WPKH to `remotekey
...
-This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or to the remote peer on successful payment preimage.  The output is a P2WSH, with a witness script:
+This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or to the remote peer using the payment preimage or the revocation key.  The output is a P2WSH, with a witness script:
 
     <remotekey> OP_SWAP
         OP_SIZE 32 OP_EQUAL
     OP_NOTIF
-        # To me via HTLC-timeout transaction (timelocked).
-        OP_DROP 2 OP_SWAP <localkey> 2 OP_CHECKMULTISIG
+        # To me via HTLC-timeout transaction (timelocked) or to you with revocation key.
+        OP_DROP 2 OP_SWAP <localkey> <revocationkey> 3 OP_CHECKMULTISIG
```
cdecker: Just noticed: doesn't this result in a race? Both `<local-sig> <revocation-sig>` and `<remote-sig> <revocation-sig>` can spend this output. At which point both endpoints could try to spend the output.

pm47: There is a subtlety: only remote is able to generate the `revocationkey`, local can't because it doesn't know the `revocation-basepoint-secret`  (see [BOLT3](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#revocationkey-derivation)).

cdecker: Oh, right that explains it ^^

PR #106, opened 2017-02-11
BOLT 2: specify requirement not to send fulfill until both sides locked in

rustyrussell:
Otherwise you can lose funds!

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -624,6 +624,9 @@ For a unparsable HTLC:
 A node SHOULD remove an HTLC as soon as it can; in particular, a node
 SHOULD fail an HTLC which has timed out.
 
+A node MUST NOT send `update_fulfill_htlc` until an HTLC is
+irrevocably committed in both sides' commitment transactions.
+
 A receiving node MUST check that `id` corresponds to an HTLC in its
 current commitment transaction, and MUST fail the channel if it does
 not.
@@ -649,6 +652,9 @@ using the `failure-code` given and setting the data to
 
 A node which doesn't time out HTLCs risks channel failure (see
 "Risks With HTLC Timeouts").
+
+A node which sends `update_fulfill_htlc` before the sender is also
+committed to the HTLC risks losing funds.
 
 If the onion is malformed, the upstream node won't be able to extract
 a key to generate a response, hence the special failure message which


pm47:
ACK, we actually have a bug related to this in eclair!

cdecker:
ACK 34e92aa2af9c8bcdc32d107d3f9dd332c37d5428

pm47:
Thinking more about this, should we add a warning about the risk of forwarding an htlc if the sender isn't locked in? In an A->B->C setup, an adversarial counterparty A could make the forwarding node B lock his funds in the B-C channel without having anything at stake in the A-B channel. There is no risk of loss of funds, but that could make some of B's funds temporarily unusable.

pm47:
Answering to myself: B can mitigate the risk using `max-accepted-htlcs` and `max-htlc-value-in-flight-msat`.

pm47:
I think B actually does risk losing funds by forwarding an htlc that is not signed by both sides. Consider this scenario:
```
A            B            C
|    add     |            |
|----------->|            |
|    sig     |            |
|----------->|            |
|    rev     |            |
|< ----------|            |
|            |    add     |
|            |----------->|
|            |    sig     |
|            |----------->|
|            |    rev     |
|            |<-----------|
|            |    sig     |
|            |<-----------|
|            |    rev     |
|            |----------->|
|            |            | C forwards the htlc
|            |            | and eventually gets
|            |            | the payment preimage
|            |            |
```

At this point if A publishes its commitment transaction and gets it confirmed, B will lose funds no matter what happens between B and C.

Am I overthinking this?

rustyrussell:
> Thinking more about this, should we add a warning about the risk of forwarding an htlc if the sender isn't locked in? In an A->B->C setup, an adversarial counterparty A could make the forwarding node B lock his funds in the B-C channel without having anything at stake in the A-B channel. There is no risk of loss of funds, but that could make some of B's funds temporarily unusable.

It *does* lead to loss of funds.  C gives preimage, A drops commitment
transaction without HTLC to chain.  Oops.

(Oh, you noted this later).

Yes, this is where the extra latency comes from: B needs A to be locked
in ("irrevocably committed") to the HTLC before it can commit to C.  And
the way that HTLCs work, this means B will be committed to the HTLC with
A too.

The only case where you can act before this commitment is in the case of
fulfilling an HTLC, as knowledge of the preimage can't be revoked :)

Cheers,
Rusty.

pm47:
>It *does* lead to loss of funds. C gives preimage, A drops commitment transaction without HTLC to chain. Oops.

Even if C doesn't give B the preimage, if C just publishes its commitment transaction he will pull funds from B, and even if B is able to retrieve the preimage from the blockchain, it won't be able to use it if A publishes its commitment tx.

>The only case where you can act before this commitment is in the case of fulfilling an HTLC, as knowledge of the preimage can't be revoked :) 

But it is too late already at this point I think. In my example B does not yet have the preimage and is already in danger of loosing funds. Shouldn't the spec also say something like this: "An htlc should not be "cross-committed" on the right side of a node if it isn't on the left side of a node"? And if we add this to the spec, then it *implies* the change made in this PR right? (and if you are the last node in the route the hltc needs to be cross-committed before you give the coffee to the customer)

Cheers,

Pierre

rustyrussell:
>>It *does* lead to loss of funds. C gives preimage, A drops commitment transaction without HTLC to chain. Oops.
>
> Even if C doesn't give B the preimage, if C just publishes its commitment transaction he will pull funds from B, and even if B is able to retrieve the preimage from the blockchain, it won't be able to use it if A publishes its commitment tx.

Yes.

>
>>The only case where you can act before this commitment is in the case of fulfilling an HTLC, as knowledge of the preimage can't be revoked :) 
>
> But it is too late already at this point I think. In my example B does
> not yet have the preimage and is already in danger of loosing
> funds.

Sorry, I was being unclear, and stating an exception to the general
rule.

> Shouldn't the spec also say something like this: "An htlc
> should not be "cross-committed" on the right side of a node if it
> isn't on the left side of a node"? And if we add this to the spec,
> then it *implies* the change made in this PR right? (and if you are
> the last node in the route the hltc needs to be cross-committed before
> you give the coffee to the customer)

Yes, we probably need a section on forwarding.  Is this clearer?

Insert this before "Risks With HTLC Timeouts" and rework the first
paragraph of that to match this:

### Forwarding HTLCs

In general, a node offers HTLCs for two reasons: to initiate a payment
of its own, or to forward a payment coming from another node.  In the
forwarding case, care must be taken to ensure that the *outgoing* HTLC
cannot be redeemed unless the *incoming* HTLC can be redeemed; these
requirements ensure that is always true.

The addition/removal of an HTLC is considered *irrevocably committed*
when:

1. the commitment transaction with/without it it is committed by both
   nodes, and any previous commitment transaction which without/with it
   has been revoked, OR
2. the commitment transaction with/without it has been irreversibly
   committed to the blockchain.

#### Requirements

A node MUST NOT offer an HTLC (`update_htlc_add`) in response to an
incoming HTLC until the incoming HTLC has been irrevocably committed.

A node MUST NOT fail an incoming HTLC (`update_fail_htlc`) for which it
has committed to an outgoing HTLC, until the removal of the outgoing
HTLC is irrevocably committed.

A node SHOULD fulfill an incoming HTLC for which it has committed to an
outgoing HTLC, as soon as it receives `update_fulfill_htlc` for the
outgoing HTLC.

#### Rationale

In general, we need to complete one side of the exchange before dealing
with the other.  Fulfilling an HTLC is different: knowledge of the
preimage is by definition irrevocable, so we should fulfill the incoming
HTLC as soon as we can to reduce latency.

PR #107, opened 2017-02-11
Fix test vectors minor

rustyrussell:
The test vectors themselves were fine, but they're supposed to be derived
from known basepoints (and I was actually testing this with some new code).

I accidentally used the *remote* per-commitment-point, instead of the *local*
per-commitment-point to derive the remotekey/remote_privkey; since we are
generating the local transaction, this is wrong.  We don't need to know
the remote per-commitment-point at all, in fact.

Thus, the remotekey (and signatures) in the test vectors change.

There's also a patch which adds the HTLC number to the witness scripts for extra clarity.


03-transactions.md
```diff
@@ -822,292 +820,292 @@ And here are the test vectors themselves:
...
+    remote_htlc_signature = 30450221009039b990b12951fcf180df6107b226d018caa8291f7ba753d22bf2b09b7d7f7b022060e9b05bcf6e2ab0ded7ae7aae179a4eccc7a3a39912da89a2468c292c549b59
+    # local_signature = 3045022100c9328a474ca222952a1cd25df57ea65b577cbf824752aed7fde4af1609b83b9202202888d5252a2d5949b288b0002f1019ba13c3c1a8e31ca5eb9ccab649bfebb789
+    output htlc_success_tx 0: 020000000001014a13eb73434fa97218577b183a4f2051b307dcee3712318b96eccd98be63c73a0000000000000000000121020000000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e0500473044022021413c112f8a8bb8343e5508acc11748ecc1dcf6a11b51534ce18a4af3ab252702200cda8cfaaaeb9f759ae551d3e67837fa8da6dd6396306ffeb98c3275f331546101483045022100c9328a474ca222952a1cd25df57ea65b577cbf824752aed7fde4af1609b83b9202202888d5252a2d5949b288b0002f1019ba13c3c1a8e31ca5eb9ccab649bfebb789012000000000000000000000000000000000000000000000000000000000000000006e210394854aa6eab5b2a8122cc726e9dded053a2184d88256816826d6231c068d4a5b7c8201208763a914b8bcb07f6344b42ab04250c86a6e8b75d3fdbbc688527c21030d417a46946384f88d5f3337267c5e579765875dc4daca813e21734b140639e752ae677502f401b175ac6800000000
```
sstone: Now I have doubts about this one: shouldn't the output amount be 546 satoshis instead of 546, since the fee should be rounded down to 554 satoshis ? (it's also true with the current test vectors, I'm pretty sure we were not computing the output properly)

rustyrussell: Ah, this was using the pre-corrected weights.  Let me update.

rustyrussell:
Trivial rebase, plus additional commit to update the HTLC fee calcs as per ec99f893f320e8c88f564c1c8566f3454f0f1f5f

03-transactions.md
```diff
@@ -822,292 +820,292 @@ And here are the test vectors themselves:
...
+    remote_htlc_signature = 3045022100a0d07c88f68bd9d3a721a159741b374c7a195393acded34344a9a1f4d42070630220228221c3076eb512f4115ae33fb0cd384b5fcd64dabc11d20d70dea18c6e7dd8
+    # local_signature = 30440220523a2c62ac50f4795aa01793eea6c510b20bc7051b41fd44c57219431c1a9cee022003a76a1e4a6f25a1834532f73d02cf125197f3a129bb40a273cba59aaf4e6535
+    output htlc_success_tx 0: 020000000001017644b6a9fb53bbc752147b6bb88b1a90ab5bfca44d7a87b8f993f83446b418b30000000000000000000121020000000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e0500473044022003f201eea4aa28831c8a0552e791d81a3ff454e97bdc688f97a1635692992a3602203e4c9362eaa04b97b94cf7a44161b932073dfcf683b1cf34904fa266ae47b14f014730440220523a2c62ac50f4795aa01793eea6c510b20bc7051b41fd44c57219431c1a9cee022003a76a1e4a6f25a1834532f73d02cf125197f3a129bb40a273cba59aaf4e6535012000000000000000000000000000000000000000000000000000000000000000006e210394854aa6eab5b2a8122cc726e9dded053a2184d88256816826d6231c068d4a5b7c8201208763a914b8bcb07f6344b42ab04250c86a6e8b75d3fdbbc688527c21030d417a46946384f88d5f3337267c5e579765875dc4daca813e21734b140639e752ae677502f401b175ac6800000000
```
sstone: I still think that there is a pb with the output of this tx, which should be 546 and not 545 satoshis

rustyrussell: OK, I'll annotate the calculation further, see where the mistake is.

rustyrussell: Found it, thanks.  Bug in my test code, was passing out-by-one feerate for HTLC generation.  Regening now.

sstone: Ok, there is still one last doubt: it seems that you're using the HTLC success tx weight when computing the fee paid by HTLC timeout txs (sorry for not reporting this before).

rustyrussell: Ah, good catch!  Fixed.

rustyrussell:
Latest fixup for htlc-timeout fee (it was using htlc-success rate, cut & paste bug, as pointed out by @sstone )

rustyrussell:
Squashed last three fixup commits together.

Issue #109, opened 2017-02-13
BOLT2: reconnection handling: one connection per peer vs one connection per channel


sstone:
Suppose A opens a channel with B.  funding-locked messages have been sent and received ad there are no pending acks. Now suppose A restarts, and tries to reconnect to B and restore the channel. How does B know which channel the new incoming connection is for ?
If we assume there is a single connection between peers and channel messages are multiplexed over it, then it's easy but what if we have one connection per channel ? 
A mandatory retransmit message that includes the channel id might be helpful here. 

ps. I remember that multiplexing channel messages over single connections was discussed but I don't remember that a choice had been made,  so the 'one connection per channel' model should also be valid ?
ps2: I had a discussion over this with Christian on #lightning-dev

rustyrussell:
Fabrice Drouin <notifications@github.com> writes:
> Suppose A opens a channel with B.  funding-locked messages have been sent and received ad there are no pending acks. Now suppose A restarts, and tries to reconnect to B and restore the channel. How does B know which channel the new incoming connection is for ?

Each message should have a channel-id prefix.

> If we assume there is a single connection between peers and channel messages are multiplexed over it, then it's easy but what if we have one connection per channel ? 

Well, I'm not planning on implementing multiple inter-peer channels, so
I might not be the right person to answer, but you could send any
channel message over any connection.

> A mandatory retransmit message that includes the channel id might be helpful here. 
>
> ps. I remember that multiplexing channel messages over single connections was discussed but I don't remember that a choice had been made,  so the 'one connection per channel' model should also be valid ?i

We chose multiplexing, because it would be a little painful to add
later.  The one-connection-per model would have been simpler in the
immediate term.

c-lightning currently disconnects the old connection when it gets a new
one from the same peer (because it probably means the old one is dead),
but that's a trivial detail.

Cheers,
Rusty.

Roasbeef:
>  Now suppose A restarts, and tries to reconnect to B and restore the channel. How does B know which channel the new incoming connection is for ?

Tries to restore the channel? The channel has already been opened at this point if I'm following the scenario correctly. Each channel update related messages carries the `channel_id` that it's modifying. 


> If we assume there is a single connection between peers and channel messages are multiplexed over  > it, then it's easy but what if we have one connection per channel ?
> A mandatory retransmit message that includes the channel id might be helpful here.

[...]

PR #113, opened 2017-02-16
BOLT 9: Make it clear that 'channel_public' apply to all channels

pm47:
This follows #109 and Christian's [comment ](https://github.com/lightningnetwork/lightning-rfc/pull/97#issuecomment-277683797) in #97.

I have also renamed `channel_public` to `channels_public`.

### DIFF
--- a/09-features.md
+++ b/09-features.md
@@ -14,7 +14,7 @@ Flags begin at bit 0 (ie. 0x1), and odd-numbered flags (eg. 0x2) are optional.
 
 | Bits | Name             |Description                                     | Link                                                                |
 |------|------------------|------------------------------------------------|---------------------------------------------------------------------|
-| 0/1  | `channel_public` | The sending node wishes to announce the channel | [BOLT #7](07-routing-gossip.md#the-announcement_signatures-message) |
+| 0/1  | `channels_public` | The sending node wishes to announce channels | [BOLT #7](07-routing-gossip.md#the-announcement_signatures-message) |
 | 2/3  | `initial_routing_sync` | The sending node needs a complete routing information dump | [BOLT #7](07-routing-gossip.md#initial-sync) |
 
 ## Assigned `globalfeatures` flags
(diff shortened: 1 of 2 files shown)


Roasbeef:
Hmm, why not make this more granular and instead allow nodes to signal their willingness to advertise _per-channel_? 

This route gives nodes more flexibility. With this, nodes don't need to rear down and reestablish a new connection in the case that they want an additional link to another node that has a network visibility policy opposite to that of any prior channels they have open with that node.

cdecker:
While it surely adds flexibility I don't think we want to add yet another flags field to another message. We're not even sure if people will multiplex channels at all, so adding this here is a bit of an overkill, especially since it can be amended later (featureflag on the connection, signalling that we append a features field to `channel_open` and `channel_accept`).

09-features.md
```diff
@@ -14,7 +14,7 @@ Flags begin at bit 0 (ie. 0x1), and odd-numbered flags (eg. 0x2) are optional.
 
 | Bits | Name             |Description                                     | Link                                                                |
 |------|------------------|------------------------------------------------|---------------------------------------------------------------------|
-| 0/1  | `channel_public` | The sending node wishes to announce the channel | [BOLT #7](07-routing-gossip.md#the-announcement_signatures-message) |
+| 0/1  | `channels_public` | The sending node wishes not to announce channels | [BOLT #7](07-routing-gossip.md#the-announcement_signatures-message) |
```
rustyrussell: You inverted the bit?  I'm confused....

cdecker: That's strange indeed, it also clashes with the name for the bits.

pm47: Hmmm, I don't know why I did that, how embarrassing! Fixed...

pm47:
@Roasbeef: the purpose of this PR was just to fix an inconsistency in the spec. I would also prefer to be able to set this parameter at the channel level.

FWIW I see another potential issue with setting it at the connection level: in case of a disconnection, one or both nodes may change their parameter before reconnecting, and it can happen before we finished the channel setup and did/didn't publish the announcements. Which means we probably need to store this parameter at the channel level, and also that we can have both public and private channels in the same connection independently of the current value of the parameter.

PR #114, opened 2017-02-16
BOLT 9: Added a `retransmit-channelid` feature bit

pm47:
This would allow a node to announce what it will do when there is a `channel-id` mismatch in the `funding-locked` step. I hesitated with an opposite `fail-on-channelid-mismatch` feature.

I am not too sure about this, but I think it might be better to be explicit?

Note that in the Rationale, BOLT 2 says that "we require that they send updates to avoid relying on timeouts" but there is nothing in the Requirements section about that

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -264,13 +264,14 @@ from this point onwards.
 
 The sender MUST wait until the funding transaction has reached
 `minimum-depth` before sending this message.  The sender MUST encode
-the block position of the funding transaction into `channel-id`.  If
-the sender has already received `funding_locked` from the other node,
-and its own `channel-id` does not match that received, it MUST either
-fail the channel or ignore the `funding_locked` message.  If it
-ignores the `funding_locked` message it MUST re-transmit
-`funding_locked` if the `channel-id` changes, otherwise it MAY
-re-transmit `funding_locked` if the `channel-id` changes.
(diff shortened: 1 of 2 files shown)


Roasbeef:
Just want to point out again that all these complications around conflicting `channel_id`'s, reorgs, and cross-chain incompatibility _disappears_ if we switch to using the `txid` of the funding transaction or full outpoint of the funding output as the global identifier for channels within the entire network... 

There's a minute possibility that this current scheme never converges if using 1-conf channels

rustyrussell:
No, if we use txids they fail in almost the same way: Half the forked network considers the channel invalid (double-spent tx), or too shallow to be useful (confirmed later/not-at-all in fork).  So we win nothing.

Similarly, if peers are on two sides of a fork you're in a bad way whatever you do.  If you both elected for a low-confirms, you're *going* to disagree unless both blocks contain the same tx.  And when we expand to cover SPV nodes with proofs, we fail in new and exciting ways.

Back to the core point: configuration options are always an admission of failure, and failing to fulfill our responsibility: If *we* can't decide which way is best, how will other implementers?

If you want to make wait-and-retransmit-on-disagree compulsory, that's probably better.  If you disagree, wait.  If you still disagree after (some timeout), abort.  I can write language around that, if you prefer.

PR #115, opened 2017-02-16
BOLT 2: revoke_and_ack is not acked by update messages

pm47:
I think `revoke_and_ack` is only acknowledged by `commitment_signed` and `shutdown` because in the following scenario B can send update messages while waiting for revocation):
```
A                                B
|              add               |
|<-------------------------------|
|              sig               |
|<-------------------------------|
| rev                        add |
|---------------  ---------------|
|               \/               |
|               /\               |
|<--------------  -------------->|
|              sig               |
|<-------------------------------|
|              rev               |
|------------------------------->|
```

In our current implementation, `revoke_and_ack` is not even acknowledged by `shutdown` but it might be a little bit far fetched? For example this is legal in eclair:
```
A                                B
|              add               |
|<-------------------------------|
|              sig               |
|<-------------------------------|
| rev                   shutdown |
|---------------  ---------------|
|               \/               |
|               /\               |
|<--------------  -------------->|
|              shutdown          |
|------------------------------->|
```

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -922,7 +922,7 @@ lists the acknowledgement conditions for each message:
 * `funding_locked`: acknowledged by `update_` messages, `commitment_signed`, `revoke_and_ack` or `shutdown` messages.
 * `update_` messages: acknowledged by `revoke_and_ack`.
 * `commitment_signed`: acknowledged by `revoke_and_ack`.
-* `revoke_and_ack`: acknowledged by `shutdown`, `update_` messages, or `commitment_signed`
+* `revoke_and_ack`: acknowledged by `shutdown` or `commitment_signed`
 * `shutdown`: acknowledged by `closing_signed` or `revoke_and_ack`.
 
 The last `closing_signed` (if any) must always be retransmitted, as there


rustyrussell:
Good catch, you can send more updates before receiving revoke-and-ack.

Pretty sure B cannot send shutdown in the above case, since:

> A node MUST NOT send a `shutdown` if there are updates pending on the receiving node's commitment transaction which are not yet added to the sender's commitment.

ie. it has to wait for the revoke-and-ack before sending shutdown?

pm47:
> Pretty sure B cannot send shutdown in the above case, since:
>>    A node MUST NOT send a shutdown if there are updates pending on the receiving node's commitment transaction which are not yet added to the sender's commitment.
>
>ie. it has to wait for the revoke-and-ack before sending shutdown?

I am not sure what difference it makes? A will always receive the sig before the shutdown right? (and in case of a disconnection it works too). Actually A can't even see the difference I think.

rustyrussell:
>> Pretty sure B cannot send shutdown in the above case, since:
>>>    A node MUST NOT send a shutdown if there are updates pending on the receiving node's commitment transaction which are not yet added to the sender's commitment.
>>
>>ie. it has to wait for the revoke-and-ack before sending shutdown?
>
> I am not sure what difference it makes? A will always receive the sig before the shutdown right? (and in case of a disconnection it works too). Actually A can't even see the difference I think.

B cannot send shutdown before receiving the revoke-and-ack.  Therefore,
once A receives shutdown it knows that B has seen revoke-and-ack.

Maybe the text could be clearer?

pm47:
Yes I think the text should make this explicit.

But I still can't find a good reason for not allowing a node to send a `commitment_signed` and then immediately a `shutdown`? What troubles me is that in (1) below, A needs to remember that it has to send a `shutdown` immediately after receiving a `revoke_and_ack`, while in (2) it just sends two messages in a row and is done with it.

```
          (1) (current spec)                            (2)
A                                B       A                                B
|                                |       |                                |
| add                   shutdown |       | add                   shutdown |
|---------------  ---------------|       |---------------  ---------------|
|               \/               |       |               \/               |
|               /\               |       |               /\               |
|<--------------  -------------->|       |<--------------  -------------->|
|              sig               |       |              sig               |
|------------------------------->|       |------------------------------->|
|              rev               |       |            shutdown            |
|<-------------------------------|       |------------------------------->|
|            shutdown            |       |              rev               |
|------------------------------->|       |<-------------------------------|
```

Btw I believe that `revoke_and_ack` is also acknowledged by `closing_signed`, I'll submit another PR.

rustyrussell:
> Yes I think the text should make this explicit.
>
> But I still can't find a good reason for not allowing a node to send a
> `commitment_signed` and then immediately a `shutdown`? What troubles
> me is that in (1) below, A needs to remember that it has to send a
> `shutdown` immediately after receiving a `revoke_and_ack`, while in
> (2) it just sends two messages in a row and is done with it.

OK, if we allow sig then shutdown, then `revoke_and_ack` is acknowledged
by `commitment_signed` or `closing_signed`, because now we the crossover
scenario is possible.

Let's simplify shutdown a little, and say it's only acked by
`closing_signed`.  Naively, it's acked by a revoke_and_ack *after* you
send another `commitment_signed`, but it doesn't hurt to retransmit it
on every reconnect until the other side actually starts closing.

New PR for you to review...

PR #116, opened 2017-02-21
BOLT 2,3,5: add extra term to received HTLC output script.

rustyrussell:
Since we expire HTLCs as soon as we can anyway (to avoid one-sided redemption),
we can simply allow very old unspent HTLCs to go to the counterparty.

This makes the witness script 8 bytes longer, and means we no longer need
to receive and store the signatures for each accepted HTLC output to push it
to the HTLC-timeout transaction where we can use our revocation key.  We simply
wait until we can use the extra-long HTLC timeout condition.

(The alternative is to allow spending the HTLC output directly using the
revocation key, but this increases the witness script size by 45 bytes,
though it does simplify some things.  See Pull Request #105).

This does not update the test vectors, as they are still being worked on.


rustyrussell:
OK, now I've fixed it: I hacked the wrong script :(  (Most of the descriptions were correct, though).

If A offers B an HTLC, B can collect it with preimage, A can get it back with (timelocked) HTLC-timeout transaction, *OR* B can collect it after ctlv-timeout + extra-delay.  This final term should never happen (A *must* use HTLC timeout before then or risk one-sided redemption anyway), but allows B to penalize in the case of old commitment transaction.

pm47:
Here is the corresponding tx tree:
```
+------------+
| funding tx |
+------------+
      |
      |        +-------------+
      \--------| commit tx B |
               +-------------+
                  |  |  |  |  
                  |  |  |  | A's main output
                  |  |  |  \------------------ to A
                  |  |  |
                  |  |  |
                  |  |  |                  ,-- to B (& delay)
                  |  |  | B's main output /
                  |  |  \----------------<
                  |  |                    \ 
                  |  |                     `-- to A (& revocation key)
                  |  |
                  |  |                                                ,-- to B (& delay)
                  |  |                        +-----------------+    /
                  |  |                     ,--| HTLC-timeout tx |---<
                  |  | HTLC offered by B  /   +-----------------+    \
                  |  \-------------------<     (after timeout)        `-- to A (& revocation key)
                  |                       \
                  |                        `-- to A (& payment preimage)
                  |                        \
                  |                         `- to A (after long timeout)
                  |                   
                  |                                                   ,-- to B (& delay)
                  |                           +-----------------+    /
                  |                        ,--| HTLC-success tx |---<
                  | HTLC received by B    /   +-----------------+    \
                  \----------------------<     (w/ payment preimage)  `-- to A (& revocation key)
                                          \
                                           `-- to A (after timeout)
```

rustyrussell:
> *    for htlcs B->A: A will need to wait until the long-timeout has expired (what do you think its value could be? just htlc-expiry+1?) to get her funds, unless she stores all preimages (and does some periodic housekeeping)

Draft uses the A's to-self-delay, which makes sense as it's A's "worst case, I will look at the blockchain this often".  It could me the max of A-'s to-self-delay and B's to-self-delay; longer is better than too short.  Too short means for non-revoked commitment txs, B has to perform the timeout spend in exactly that block, or risk racing with A.

And I agree with your point on the increase in conceptual complexity.  It's clever, but maybe not a good thing :)

I guess @Roasbeef and @adiabat need to decide then...

Roasbeef:
While this is a smaller net change, after reviewing the updated version of this proposal, think I prefer #105 (*insert randy jackson meme*). 

I'm uneasy with the concept of overloading time within HTLC's to server both a timeouts for unclaimed HTLC's and also as a revocation clause. It also seems to complicate the code for claiming revoked outputs, as with this HTLC's can have different timeouts, and therefore need to be handled in distinct "justice contexts", so to speak. Finally, it adds an additional timeout based parameter that'll require further tweaking and analysis. 

With #105, the semantics around claiming outputs that have been revoked are uniform, and _all_ revoked outputs can be claimed *immediately*.

03-transactions.md
```diff
@@ -107,7 +107,7 @@ This output sends funds to the other peer, thus is a simple P2WPKH to `remotekey
 
 #### Offered HTLC Outputs
 
-This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or to the remote peer on successful payment preimage.  The output is a P2WSH, with a witness script:
+This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or to the remote peer on successful payment preimage.  To allow penalties, it has an additional condition which allows the remote peer to claim the funds long after the HTLC has expired.  We use the local node's specified `to-self-delay` to determine what how long the extra HTLC delay is (rather than that specified by the remote node as used in the HTLC transactions). The output is a P2WSH, with a witness script:
```
cdecker: "what how long"

cdecker:
Ok, re-reading both proposals I think I like #116 better, despite adding some longer timeouts and some additional state tracking. The size increase of #105 is just too big. The longer timeouts are worrying, especially since many potential users I talk to bring locking funds up as a flaw, but I think if we cannot clear an agreed upon HTLC we're in serious trouble for the channel anyway.

rustyrussell:
Closed in favor of #105 which is simpler, though marginally less space-optimal.

PR #119, opened 2017-02-28
BOLT 0,1,2,7: use txout not channel-id for demuxing.

rustyrussell:
At cost of a few extra bytes, this avoids the whole "oops, we were on
a chain fork" problem.

Now we move the announcement_signature exchange to at least 6 confirms,
which makes re-xmit tricky; I resolved that by insisting on reconnect that
we send if we haven't received, and reply to the first one.

The term "channel shortid" wasn't used anywhere, so I removed it; it's
now a gossip-only thing anyway.

Closes: #114
Suggested-by: Olaoluwa Osuntokun <laolu32@gmail.com>

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -120,7 +120,7 @@ For simplicity of diagnosis, it is often useful to tell the peer that something 
 1. type: 17 (`error`)
 2. data:
    * [32:funding-txid]
-   * [1:funding-output-index]
+   * [2:funding-output-index]
    * [2:len]
    * [len:data]
 
@@ -129,8 +129,8 @@ The 2-byte `len` field indicates the number of bytes in the immediately followin
 #### Requirements
 
 The channel is referred to by `funding-txid` and `funding-output-index`,
-unless `funding-txid` is 0, in which case `funding-output-index` of 0 refers
-to the currently opening channel (ie. before `funding_signed` or `funding_created` are sent), and a `funding-output-index` of 255 refers to all channels.
+unless `funding-txid` is all-zero, in which case `funding-output-index` of 0 refers
+to the currently opening channel (ie. before `funding_signed` or `funding_created` are sent), and a `funding-output-index` of 65535 refers to all channels.
 
 A node SHOULD send `error` for protocol violations or internal
 errors which make channels unusable or further communication unusable.
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -119,8 +119,7 @@ For simplicity of diagnosis, it is often useful to tell the peer that something 
 
 1. type: 17 (`error`)
 2. data:
-   * [32:funding-txid]
-   * [2:funding-output-index]
+   * [32:channel-id]
    * [2:len]
    * [len:data]
 
@@ -128,19 +127,17 @@ The 2-byte `len` field indicates the number of bytes in the immediately followin
 
 #### Requirements
 
-The channel is referred to by `funding-txid` and `funding-output-index`,
-unless `funding-txid` is all-zero, in which case `funding-output-index` of 0 refers
-to the currently opening channel (ie. before `funding_signed` or `funding_created` are sent), and a `funding-output-index` of 65535 refers to all channels.
+The channel is referred to by `channel-id` unless `channel-id` is all-zero, in which case it refers to all channels.
 
 A node SHOULD send `error` for protocol violations or internal
 errors which make channels unusable or further communication unusable.
 A node MAY send an empty [data] field.  A node sending `error` MUST
-fail the channel referred to by the error message, or if all channels, MUST
+fail the channel referred to by the error message, or if `channel-id` is all-zero, MUST
 fail all channels and MUST close the connection.
 A node MUST set `len` equal to the length of `data`.  A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitment_signed` message when failure was caused by an invalid signature check.
 
 A node receiving `error` MUST fail the channel referred to by the message,
-or if all channels, it MUST fail all channels and MUST close the connection.  If no existing channel is referred to by the message, the receiver MAY fail all channels and close the connection. A receiving node MUST truncate
+or if `channel-id` is all-zero, it MUST fail all channels and MUST close the connection.  If no existing channel is referred to by the message, the receiver MUST ignore the message. A receiving node MUST truncate
 `len` to the remainder of the packet if it is larger.
 
 A receiving node SHOULD only print out `data` verbatim if the string is composed solely of printable ASCII characters.
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -119,7 +119,8 @@ For simplicity of diagnosis, it is often useful to tell the peer that something 
 
 1. type: 17 (`error`)
 2. data:
-   * [8:channel-id]
+   * [32:funding-txid]
+   * [1:funding-output-index]
    * [2:len]
    * [len:data]
 
@@ -127,16 +128,19 @@ The 2-byte `len` field indicates the number of bytes in the immediately followin
 
 #### Requirements
 
+The channel is referred to by `funding-txid` and `funding-output-index`,
+unless `funding-txid` is 0, in which case `funding-output-index` of 0 refers
+to the currently opening channel (ie. before `funding_signed` or `funding_created` are sent), and a `funding-output-index` of 255 refers to all channels.
+
 A node SHOULD send `error` for protocol violations or internal
 errors which make channels unusable or further communication unusable.
 A node MAY send an empty [data] field.  A node sending `error` MUST
-fail the channel referred to by the `channel-id`, or if `channel-id`
-is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
+fail the channel referred to by the error message, or if all channels, MUST
+fail all channels and MUST close the connection.
 A node MUST set `len` equal to the length of `data`.  A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitment_signed` message when failure was caused by an invalid signature check.
 
-A node receiving `error` MUST fail the channel referred to by
-`channel-id`, or if `channel-id` is `0xFFFFFFFFFFFFFFFF` it MUST fail
-all channels and MUST close the connection.  A receiving node MUST truncate
+A node receiving `error` MUST fail the channel referred to by the message,
+or if all channels, it MUST fail all channels and MUST close the connection.  If no existing channel is referred to by the message, the receiver MAY fail all channels and close the connection. A receiving node MUST truncate
 `len` to the remainder of the packet if it is larger.
 
 A receiving node SHOULD only print out `data` verbatim if the string is composed solely of printable ASCII characters.
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -127,17 +127,17 @@ The 2-byte `len` field indicates the number of bytes in the immediately followin
 
 #### Requirements
 
-The channel is referred to by `channel-id` unless `channel-id` is all-zero, in which case it refers to all channels.
+The channel is referred to by `channel-id` unless `channel-id` is zero (ie. all bytes zero), in which case it refers to all channels.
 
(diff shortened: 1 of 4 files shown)


01-messaging.md
```diff
@@ -119,24 +119,28 @@ For simplicity of diagnosis, it is often useful to tell the peer that something
...
+   * [1:funding-output-index]
    * [2:len]
    * [len:data]
 
 The 2-byte `len` field indicates the number of bytes in the immediately following field.
 
 #### Requirements
 
+The channel is referred to by `funding-txid` and `funding-output-index`,
+unless `funding-txid` is 0, in which case `funding-output-index` of 0 refers
```
Roasbeef: Nice addition. As a minor nit, I think we should specify we mean "all zeroes" as in `'0'*32` for the `funding-txid` field to have a connection-wide scope.

02-peer-protocol.md
```diff
@@ -209,20 +208,19 @@ signature, it will broadcast the funding transaction.
...
+The sender MUST set `funding-txid` to the transaction ID of a non-malleable transaction, which it MUST NOT broadcast, and MUST set `funding-output-index` to the output number of that transaction which corresponds the funding transaction output as defined in [BOLT #3](03-transactions.md#funding-transaction-output), and MUST set `signature` to the valid signature using its `funding-pubkey` for the initial commitment transaction as defined in [BOLT #3](03-transactions.md#commitment-transaction).
 
 The recipient MUST fail the channel if `signature` is incorrect.
 
 #### Rationale
 
-The `output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the protocol.  The limit of 255 outputs should not be overly burdensome.
+The `funding-output-index` can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the gossip protocol.  The limit of 255 outputs should not be overly burdensome.
```
Roasbeef: > can only be 1 byte, since that's how we'll pack it into the channel-id used throughout the gossip protocol

We actually currently use `2-bytes` in the gossip protocol for `channel-id`s IIRC. However as you note, I think `1-byte` is sufficient for de-multiplexing channel updates.

02-peer-protocol.md
```diff
@@ -70,6 +70,9 @@ If this fails at any stage, or a node decides that the channel terms
 offered by the other node are not suitable, the channel establishment
 fails.
 
+Note that multiple channels can operate in parallel, but only after
```
Roasbeef: Excellent. IMO the allowance of parallel channels was made implicit by using chan ID's (before this PR), but as always it's better to make things explicit when possible 😎

02-peer-protocol.md
cdecker: The extraction tool is also complaining about the message not aligning to 2 bytes, so that'd be another vote for 2 bytes here as well.

02-peer-protocol.md
Roasbeef: > introduces a state in the peer

State in the peer already exists? 

Good point about concurrent channel funding, that somehow slipped my mind. Here's a suggestion (pretty much was `lnd` does today): 
   * Allow concurrent channel funding. 
   * Each node has a policy which dictates the max allowed number of in-flight channel funding workflows. If a peer attempts to initiate a new funding workflow that puts the total number of active in-flight funding workflows over this limit, then that attempt is rejected. 
   * Don't trigger retransmission semantics before the `funding_signed` message is sent. This saves nodes from having to persist pending channel state until they've actually committed funds. 
   * Re-introduce temp channel ID's. Dictate a loose schema governing how they're generated: the connecting peer starts from `1` and increments with each new channel, and listening peer starts from `1 << 63` and also increments with each channel. 

The above retains using `utxo`'s for de-multiplexing, allows nodes to only allow a single in-flight funding workflow (if they wish), solves issues with concurrent channel fundings (there's essentially two funding "streams"), and makes "temp chanID's" more predictable (along with avoiding the possibility of collisions).

07-routing-gossip.md
```diff
@@ -19,16 +19,31 @@ It contains the necessary signatures by the sender to construct the `channel_ann
...
     * [64:bitcoin-signature]
 
 The willingness of the endpoints to announce the channel is signaled during the connection setup by setting a `channels_public` bit in the `localfeatures` field.
-Should both endpoints have signaled that they'd like to publish the channel then the `announcement_signatures` message MUST directly sent following the `funding_locked` message that established the corresponding channel.
+
+### Requirements
+
+If both endpoints have signaled that they'd like to publish the channel then the `announcement_signatures` message MUST be sent, otherwise they MUST NOT be sent.
+
+If sent, `announcement_signatures` messages MUST NOT be sent until `funding_locked` has been sent, and the funding transaction is has at least 6 confirmations.
```
Roasbeef: Seems like this has been the case since they were introduced, no?

pm47: There has been some back and forth on this, but it wasn't the case in the last iteration:

> the announcement_signatures message MUST directly [be] sent following the funding_locked message

02-peer-protocol.md
rustyrussell: Oh, yeah, don't know why I thought 1 byte...

01-messaging.md
rustyrussell: Good, fixed.

02-peer-protocol.md
rustyrussell: Concurrent funding from both sides should Just Work.  But we would need to re-add temp channel ids (easy now we have 34 bytes, though, we can just use random value).

Hmm, OK, you convinced me :)  I didn't add a limit for now, but see latest push.

I've thought about forgetting channel if we disconnect before receipt of funding_signed (or send of funding_signed for the accepter).  The corner case is that B has sent funding_signed and A didn't get it.  B retransmits, and then it gets an error, which I think is works fine.  I'll create a separate patch for this once we're done here (I've filed #120 for this).

07-routing-gossip.md
rustyrussell: Yes, I think we need to detach this now we accept that the short channel-id can change on an active channel.

I didn't want to do it before because we didn't know when to re-transmit it, but I think the new language fixes this.

02-peer-protocol.md
```diff
@@ -209,35 +212,37 @@ signature, it will broadcast the funding transaction.
 
 1. type: 34 (`funding_created`)
 2. data:
-    * [8:temporary-channel-id]
-    * [32:txid]
-    * [1:output-index]
+    * [32:temporary-funding-id]
```
pm47: `temporary-funding-id` should be `temporary-channel-id`

01-messaging.md
```diff
@@ -119,24 +119,25 @@ For simplicity of diagnosis, it is often useful to tell the peer that something
...
 #### Requirements
 
+The channel is referred to by `channel-id` unless `channel-id` is all-zero, in which case it refers to all channels.
+
 A node SHOULD send `error` for protocol violations or internal
 errors which make channels unusable or further communication unusable.
 A node MAY send an empty [data] field.  A node sending `error` MUST
-fail the channel referred to by the `channel-id`, or if `channel-id`
-is `0xFFFFFFFFFFFFFFFF` it MUST fail all channels and MUST close the connection.
+fail the channel referred to by the error message, or if `channel-id` is all-zero, MUST
```
cdecker: missing "it" before the MUST

01-messaging.md
```diff
@@ -119,24 +119,25 @@ For simplicity of diagnosis, it is often useful to tell the peer that something
...
+fail all channels and MUST close the connection.
 A node MUST set `len` equal to the length of `data`.  A node SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed` or `commitment_signed` message when failure was caused by an invalid signature check.
 
-A node receiving `error` MUST fail the channel referred to by
-`channel-id`, or if `channel-id` is `0xFFFFFFFFFFFFFFFF` it MUST fail
-all channels and MUST close the connection.  A receiving node MUST truncate
+A node receiving `error` MUST fail the channel referred to by the message,
+or if `channel-id` is all-zero, it MUST fail all channels and MUST close the connection.  If no existing channel is referred to by the message, the receiver MUST ignore the message. A receiving node MUST truncate
```
cdecker: Maybe we can extract the "all-zero" part into a definition and call it "zero"

02-peer-protocol.md
```diff
@@ -209,35 +212,37 @@ signature, it will broadcast the funding transaction.
...
 1. type: 35 (`funding_signed`)
 2. data:
-    * [8:temporary-channel-id]
+    * [32:channel-id]
     * [64:signature]
 
 #### Requirements
 
-The sender MUST set `temporary-channel-id` the same as the `temporary-channel-id` in the `open_channel` message, and MUST set `signature` to the valid signature using its `funding-pubkey` for the initial commitment transaction as defined in [BOLT #3](03-transactions.md#commitment-transaction).
+The sender MUST set `channel-id` by exclusive-OR of the `funding-txid` and the `funding-output-index` from the `funding_created` message, and MUST set `signature` to the valid signature using its `funding-pubkey` for the initial commitment transaction as defined in [BOLT #3](03-transactions.md#commitment-transaction).
```
cdecker: Not sure whether this has some impact on our ability to retrieve the funding transaction. If we use `bitcoind` for example we cannot use the built-in transaction index anymore. I'd rather have a 34 byte channel-id.

cdecker:
Some minor comments inline, other than that LGTM :+1:

02-peer-protocol.md
rustyrussell: Ack, fixed.

01-messaging.md
rustyrussell: Added.  Eliding the "it" is allowed, but hurts clarity.

01-messaging.md
rustyrussell: Ack. I've made that clear, by using "zero" instead of "all-zero" with the first usage followed by " (ie. all bytes zero)".  That's explicit, though it's hard to think of any other meaning of zero :)

02-peer-protocol.md
rustyrussell: Well, we can grind it out :)  But in practice, when we receive a message we *must* know about the channel; we're not going to be asking bitcoind.

34 is really ugly for alignment...

02-peer-protocol.md
```diff
@@ -79,7 +83,7 @@ desire to set up a new channel.
 
 1. type: 32 (`open_channel`)
 2. data:
-   * [8:temporary-channel-id]
+   * [32:temporary-channel-id]
```
Roasbeef: How is this to be created? As it's large enough I guess nodes could just use some `chacha20` based CSPRNG...

07-routing-gossip.md
```diff
@@ -19,16 +19,30 @@ It contains the necessary signatures by the sender to construct the `channel_ann
...
+height, the next 3 bytes indicating the transaction index within the
+block, and the least significant two bytes indicating the output
+index which pays to the channel.
+
 The `announcement_signatures` message is created by constructing a `channel_announcement` message corresponding to the newly established channel, and sign it with the secrets matching their `node-id` and `bitcoin-key`, and send them using an `announcement_signatures`.
 The recipient MAY fail the channel if the `node-signature` or `bitcoin-signature` is incorrect.
 The recipient SHOULD queue the `channel_announcement` message for its peers if it has sent and received a valid `announcement_signatures` message.
-If either endpoints does not signal `channels_public` then `announcement_signatures` MUST NOT be sent.
+
+On reconnection, a node SHOULD retransmit the `announcement_signatures` message if it has not received an `announcement_signatures` message, and MUST respond to the first `announcement_signatures` message after reconnection with its own `announcement_signatures` message.
```
Roasbeef: So we send `announcement_signatures` unconditionally now? 

This block was removed in the latest diff: 
> If either endpoints does not signal `channels_public` then `announcement_signatures` MUST NOT be sent.

rustyrussell: No, it's just moved to the first paragraph:

> If both endpoints have signaled that they'd like to publish the channel then the `announcement_signatures` message MUST be sent, otherwise they MUST NOT be sent.

Roasbeef: Ahh, I missed that in my initial pass, nevermind!

PR #121, opened 2017-03-05
BOLT 2,3: SHOULD NOT create malleable funding tx.

rustyrussell:
But note that our funding transaction example is.


cdecker:
Not a huge problem since we actually wait for confirmations, but nice to have :+1:

PR #123, opened 2017-03-06
BOLT 3: Use revocation key hash rather than revocation key

rustyrussell:
This means more bytes if we need to create a penalty tx, but less for normal operation:

Witness script for offered htlcs: 139 bytes -> 133 bytes.
Witness script for accepted htlcs: 156 bytes -> 139 bytes.

It's also a little simpler; it's just an OP_IF around the old scripts to test for the revocation key being the top arg.

UNFINISHED: Fix weight calculations, test vectors.


rustyrussell:
Yeah, the 2-of-3 trick was *brilliant*; that's why this was only 6 bytes smaller.  But for accepted HTLCs, it's a bigger win.

Issue #128, opened 2017-03-15
Value of dust-limit-satoshis for the closing transaction


pm47:
Since both nodes may choose a different value for `dust-limit-satoshis` (say `dust-limit-a` and `dust-limit-b`), I think we need to specify which one is to be used for the closing transaction in a mutual close scenario.

This is definitely a corner case, but it still can happen with `channel-reserve-satoshis` >> `dust-limit-satoshis` if we close a channel shortly after creation.

I think using `dust-limit-a` for the output going to A and `dust-limit-b` for the output going to B would make sense, but it might be easier to just use min( `dust-limit-a`, `dust-limit-b`) or max( `dust-limit-a`, `dust-limit-b`).

What do you guys think?

Cheers,

Pierre

rustyrussell:
Pierre-Marie Padiou <notifications@github.com> writes:
> Since both nodes may choose a different value for `dust-limit-satoshis` (say `dust-limit-a` and `dust-limit-b`), I think we need to specify which one is to be used for the closing transaction in a mutual close scenario.
>
> This is definitely a corner case, but it still can happen with `channel-reserve-satoshis` >> `dust-limit-satoshis` if we close a channel shortly after creation.
>
> I think using `dust-limit-a` for the output going to A and `dust-limit-b` for the output going to B would make sense, but it might be easier to just use min( `dust-limit-a`, `dust-limit-b`) or max( `dust-limit-a`, `dust-limit-b`).
>
> What do you guys think?

Great question!

If A lets B choose the dust limit, there's a potential for an
unrelayable transaction (too low), or funds wasted on miners (too high).

At its root, this is a fundamental disagreement: if A considers its
output above the dust limit (ie. "Give me my money!") and B considers
A's output below the dust limit (ie. "No, that won't propagate!"), they
*cannot* mutual close.

Thus the simplest solution is to allow breakdown naturally:

        The sender MUST set `signature` to the Bitcoin signature of the
        close transaction with the node responsible for paying the
        bitcoin fee paying `fee-satoshis`, without populating any output
        which is below its own `dust-limit-satoshis`.  The sender MAY also
        eliminate its own output from the mutual close transaction.

        The receiver MUST check `signature` is valid for either the close
        transaction with the given `fee-satoshis` as detailed above and
        its own `dust-limit-satoshis` OR that same transaction with the
        sender's output eliminated, and MUST fail the connection if it
        is not.

Note that we might as well let a node eliminate its own transaction
[...]

PR #130, opened 2017-03-17
BOLT 2: change dust limit for closing to each use their own.

rustyrussell:
Using the receiver's dust limit implies a naive node can be fooled
(by a peer with massive dust limit) into not getting its own output.

Using our own opens the possibility of creating different transactions,
so we explicitly allow the tractable case, while accepting failure on
the case where disagreement is real.

Closes: #128
Reported-by: Pierre-Marie Padiou

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -377,11 +377,13 @@ estimate of cost of inclusion in a block.
 The sender MUST set `signature` to the Bitcoin signature of the close
 transaction with the node responsible for paying the bitcoin fee
 paying `fee-satoshis`, without populating any output which is below
-the receiver's `dust-limit-satoshis`.
+its own `dust-limit-satoshis`. The sender MAY also eliminate its own
+output from the mutual close transaction.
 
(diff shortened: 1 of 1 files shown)


pm47:
I just have one remark: it seems that the `MUST fail the connection if it is not` part could be done on the sender's side, because it already has all the information to conclude that there won't be an agreement. Wouldn't that be more straightforward?

rustyrussell:
> I just have one remark: it seems that the `MUST fail the connection if it is not` part could be done on the sender's side, because it already has all the information to conclude that there won't be an agreement. Wouldn't that be more straightforward?

I opted for simplicity of implementation, rather than simplicity of
discovering what happened if they disagree.

We could add:
        
        The sender MAY check if the receiver will eliminate it the
        sender's output as dust when it did not, and fail the
        connection.

I'm tempted to see if it becomes a problem before adding anything,
however.

pm47:
> I'm tempted to see if it becomes a problem before adding anything, however.

Agreed. I am fine with the PR as-is 👍 .

PR #132, opened 2017-03-22
BOLT 4: add test vectors for reply messages

sstone:


cdecker:
Finally getting started on this, sorry for the delay. I can't reproduce the `filler`, which should probably be called `cipher_stream`. My code produces `e9c975b07c9a374ba64fd9be3aae955e917d34d1fa33f2e90f53bbf4394713c6a8c9b16ab5f12fd45edd73c1b0c8b33002df376801ff58aaa94000bf8a86f92620f343baef38a580102395ae3abf9128d1047a0736ff9b83d456740ebbb4aeb3aa9737f18fb4afb4aa074fb26c4d702f42968888550a3bded8c05247e045b866baef0499f079fdaeef6538f31d44deafffdfd3afa2fb4ca9` instead of your test vectors. The code that produces this in c-lightning is:

	u8 key[KEY_LEN];
	size_t streamlen = tal_len(reply);
	u8 stream[streamlen];
	u8 *result = tal_arr(ctx, u8, streamlen);
	generate_key(key, "ammag", 5, shared_secret);
	generate_cipher_stream(stream, key, streamlen);
	xorbytes(result, stream, reply, streamlen);

Could you maybe share the `ammag` key @sstone?

sstone:
You're right, I was reusing my filler generation function instead of the stream generation function as specified. I've changed it, it now matches your cipher stream, and I will generate new test vectors. 
Thanks!

PR #134, opened 2017-03-22
BOLT 1: introduce `ping` and `pong` control messages

Roasbeef:
Connections between nodes within the network may be very long lived as
payment channels have an indefinite lifetime. However, it’s likely that
for a significant portion of the life-time of a connection, no new data
will be exchanged. Additionally, on several platforms it’s possible that
Lightning clients will be put to sleep without prior warning. As a
result, we use a distinct ping message in order to probe for the
liveness of the connection on the other side, and also to keep the
established connection active.

This commit adds two new control messages to the protocol: `ping` and
`pong`. Their usage within the network is similar to the usage of such
message within other established protocols: `ping` messages specify a
number of bytes to be contained in the payload of a `pong` message, and
`pong` messages are to be sent in response to receiving a `ping` message.

Additionally, the ability for a sender to request that the receiver send
a response with a particular number of bytes enables nodes on the
network to create synthetic traffic. Such traffic can be used to
partially defend against packet and timing analysis as nodes can fake
the traffic patterns of typical exchanges, without applying any true
updates to their respective channels.

When combined with the onion routing protocol defined in BOlT#4, careful
statistically driven synthetic traffic can serve to further bolster the
privacy of participants within the network.

A future extension would be to allow the `ping` messages to also specify a 
particular time-delay to be observed before responding.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -210,7 +210,7 @@ the other side, and also to keep the established connection active.
 
 Additionally, the ability for a sender to request that the receiver send a
 response with a particular number of bytes enables nodes on the network to
-create _synthentic_ traffic. Such traffic can be used to partially defend
+create _synthetic_ traffic. Such traffic can be used to partially defend
 against packet and timing analysis as nodes can fake the traffic patterns of
 typical exchanges, without applying any true updates to their respective
 channels. 
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -15,8 +15,7 @@ All data fields are big-endian unless otherwise specified.
     * [The `init` message](#the-init-message)
     * [The `error` message](#the-error-message)
   * [Control Messages](#control-messages)
-    * [The `ping` message](#the-ping-message)
-    * [The `pong` message](#the-pong-message)
+    * [The `ping` and `pong` messages](#the-ping-and-pong-messages)
   * [Acknowledgements](#acknowledgements)
   * [References](#references)
   * [Authors](#authors)
@@ -158,44 +157,46 @@ it leak information, thus the optional data field.
 
 ## Control Messages
 
-### The `ping` message
+### The `ping` and `pong` messages
 
 In order to allow for the existence of very long-lived TCP connections, at
 times it may be required that both ends keep alive the TCP connection at the
-application level.
+application level.  Such messages also allow obsfusation of traffic patterns.
 
 1. type: 18 (`ping`)
 2. data: 
     * [2:num_pong_bytes]
-
-#### Requirements
-
-A node receiving a `ping` message MUST respond with a `pong` message that
-contains the specified `num_pong_bytes` number of bytes within it's data
-payload. These `ping` messages may be sent any time at the sender's discretion,
-however `ping` messages MUST NOT be sent with a frequency lower than one every
-30 seconds.
-
-### The `pong` message
+    * [2:byteslen]
+    * [byteslen:ignored]
 
 The `pong` message is to be sent whenever a `ping` message is received. It
 serves as a reply, and also serves to keep the connection alive while
-explicitly notifying the other end that the receiver is still active.
+explicitly notifying the other end that the receiver is still active. Within
+the received `ping` message, the sender will specify the number of bytes to be
+included within the data payload of the `pong` message
 
 1. type: 19 (`pong`)
 2. data:
     * [2:byteslen]
-    * [byteslen:randombytes]
+    * [byteslen:ignored]
 
 #### Requirements
 
-A `pong` message MUST only be sent in response to receiving a `ping` message.
-Within the received `ping` message, the sender will specify the number of bytes
-to be included within the data payload of the `pong` message. Senders of the
-`pong` message may retrieve the payload bytes from any source as the bytes are
-opaque and convey no meaning. Senders of a `pong` message SHOULD take care to
-ensure that the bytes being sent don't contain sensitive data such as secrets,
-or portions of initialized memory.
+A node sending `pong` or `ping` SHOULD set `ignored` to zeroes, but MUST NOT
+set `ignored` to sensitive data such as secrets, or portions of initialized
+memory.
+
+A node SHOULD NOT send `ping` messages more often than once every 30 seconds,
+and MAY terminate the network connection if it does not receive a corresponding
+`pong`: it MUST NOT fail the channels in this case.
+
+A node receiving a `ping` message SHOULD fail the channels if it has received
+significantly in excess of one `ping` per 30 seconds, otherwise if
+`num_pong_bytes` is less than 65534 it MUST respond by sending a `pong` message
+with `byteslen` equal to `num_pong_bytes`, otherwise it MUST ignore the `ping`.
+
+A node receiving a `pong` message MAY fail the channels if `byteslen` does not
+correspond to any `ping` `num_pong_bytes` value it has sent.
 
 ### Rationale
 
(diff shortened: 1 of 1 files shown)


01-messaging.md
```diff
@@ -153,6 +156,73 @@ diagnosis, as it indicates that one peer has a bug.
...
+    * [2:num_pong_bytes]
+
+#### Requirements
+
+A node receiving a `ping` message MUST respond with a `pong` message that
+contains the specified `num_pong_bytes` number of bytes within it's data
+payload. These `ping` messages may be sent any time at the sender's discretion,
+however `ping` messages MUST NOT be sent with a frequency lower than one every
+30 seconds.
+
```
rustyrussell: Usually clearest to separate sender and receiver requirements (eg. what happens if the sender violates spec and flood pings?).  We should also specify padding on send is allowed, and MBZ for future extensions.

The sender MAY send `ping` messages MAY be sent any time at the sender's discretion, but MUST NOT send `ping` more often than than once every 30 seconds, and MUST NOT set `num-pong-bytes` greater than 65533.  The sender MAY pad the `ping` message after the `num-pong-bytes` field with zeroes, and MUST NOT use non-zero padding.

The receiver SHOULD fail the connection if `num-pong-bytes` is greater than 65535.

01-messaging.md
```diff
@@ -153,6 +156,73 @@ diagnosis, as it indicates that one peer has a bug.
...
+#### Requirements
+
+A `pong` message MUST only be sent in response to receiving a `ping` message.
+Within the received `ping` message, the sender will specify the number of bytes
+to be included within the data payload of the `pong` message. Senders of the
+`pong` message may retrieve the payload bytes from any source as the bytes are
+opaque and convey no meaning. Senders of a `pong` message SHOULD take care to
+ensure that the bytes being sent don't contain sensitive data such as secrets,
+or portions of initialized memory.
+
```
rustyrussell: Split into sender/receiver.  Allow ignoring of pings, for case of ping-flood.

The sender SHOULD send a `pong` message in response to a `ping` message with non-zero `num-pong-bytes`, and MUST NOT send a `pong` message otherwise.

The receiver MAY fail the connection if the `pong` message length does not correspond to any `ping` it has sent, and MAY fail the connection if no `pong` message is received in reply to its `ping`.

rustyrussell:
A. Pings SHOULD be green.
B. No, blue.
A. Pings SHOULD be green or blue.
B. What about chartreuse?
A. Pings MUST NOT be red.
B. What about color-blind users?

01-messaging.md
```diff
@@ -153,6 +156,73 @@ diagnosis, as it indicates that one peer has a bug.
...
+application level.
+
+1. type: 18 (`ping`)
+2. data: 
+    * [2:num_pong_bytes]
+
+#### Requirements
+
+A node receiving a `ping` message MUST respond with a `pong` message that
+contains the specified `num_pong_bytes` number of bytes within it's data
```
pm47: Would it be a good idea to also put padding data in the `ping` message for further obfuscation ? Otherwise eavesdroppers might guess a `ping` has been sent?

01-messaging.md
pm47: > Usually clearest to separate sender and receiver requirements (eg. what happens if the sender violates spec and flood pings?) 

In that case we also need to add:

The receiver SHOULD fail the connection if it receives more `ping` with a frequency higher one every 30 seconds.

pm47:
@rustyrussell I like your tastes in 🍷 🍸

01-messaging.md
cdecker: Can @Roasbeef confirm that padding with 0s does not risk exposing encryption keys for the channel?

01-messaging.md
cdecker: The non-zero requirement allows for uni-directional pings, which is kind of neat, but do we want that special case in here?

Also with the 30 second min-delta and 65k limit this allows us to continually waste about 2kB/s on the responder's upstream, which may cause problems on some link types. We might want to restrict that later.

We also do not have a timeout period in which the `ping` recipient MUST reply, allowing the sender to eventually close the channel.

01-messaging.md
rustyrussell: Yep, see below...

01-messaging.md
rustyrussell: Re: suggesting cutoff.  If packets are delayed, you might get them all at once in a burst.  So it's a tricky threshold.  Safest to ignore them if you think there are too many.

01-messaging.md
rustyrussell: > The non-zero requirement allows for uni-directional pings, which is kind of neat, but do we want that special case in here?

I do worry about the special case, but for real obfuscation we need it I think.  If our normal traffic pattern is -> add  -> add -> commit <- revoke_and_ack, we need "no-reply" pings to simulate it.

> this allows us to continually waste about 2kB/s on the responder's upstream,

There are other ways to flood, so I don't think this is a huge issue. In particular, you can send an unknown odd packet.

> We also do not have a timeout period in which the ping recipient MUST reply, allowing the sender to eventually close the channel.

Figuring out how long to timeout is really hard; I don't think we can spec it here, since it may vary widely.

Also, this should not say "fail the connection" (which has a specific meaning involving closing the channel(s) and cleaning up): that should only happen due to a real timeout), it should say "close the network connection":

The receiver MAY fail the connection if the pong message length does not correspond to any ping it has sent, and MAY close the network connection if no pong message is received in reply to its ping.

rustyrussell:
OK, how about this:

https://gist.github.com/rustyrussell/ffff7cf17300ea08954b6b9bcc0ac601

cdecker:
You're right, flooding is probably the least of our concerns. I'm happy with the PR, either with you diff applied or not. Though I prefer grouping `ping` and `pong` into a single heading and having the random data trailer explicit like you propose.

Roasbeef:
I've just push a _highly_ anticipated review fixup commit. I've incorporated a bit of Rusty's gist (mostly combining the ping and pong sections) in the latest revision. 

RE flooding: I agree with Rusty that we shouldn't really be concerned above ping flooding as a annoying node can just propose HTLC's that'll automatically be failed (set you as final hop in the sphinx payload, bogus payment hash).

01-messaging.md
Roasbeef: Good idea! Added in the latest revision.

01-messaging.md
```diff
@@ -153,6 +156,73 @@ diagnosis, as it indicates that one peer has a bug.
...
+payload. These `ping` messages may be sent any time at the sender's discretion,
```
Roasbeef: Done.

01-messaging.md
```diff
@@ -153,6 +156,73 @@ diagnosis, as it indicates that one peer has a bug.
...
+however `ping` messages MUST NOT be sent with a frequency lower than one every
+30 seconds.
```
Roasbeef: Heh, I dun goofed...fixed in the latest revision.

01-messaging.md
Roasbeef: @cdecker yes padding with all 0s doesn't risk leaking our encryption keys. On the wire, it'll just be seen as our stream of random bytes from the chacha stream.

01-messaging.md
```diff
@@ -153,6 +155,80 @@ diagnosis, as it indicates that one peer has a bug.
...
+channels have an indefinite lifetime. However, it's likely that for a
+significant portion of the life-time of a connection, no new data will be
+exchanged. Additionally, on several platforms it's possible that Lightning
+clients will be put to sleep without prior warning.  As a result, we use a
+distinct ping message in order to probe for the liveness of the connection on
+the other side, and also to keep the established connection active. 
+
+Additionally, the ability for a sender to request that the receiver send a
+response with a particular number of bytes enables nodes on the network to
+create _synthentic_ traffic. Such traffic can be used to partially defend
```
pm47: Isn't that _synthetic_?

PR #135, opened 2017-03-22
BOLT 02: add a `chain-hash` field to the `open_channel` msg

Roasbeef:
This commit adds a `chain-hash` field to message that commences a
funding workflow. This field is used to specify a _target_ chain for the
proposed channel. In order to uniquely identity blockchains in a manner
that doesn't require strict coordination between developers, the genesis
hash of the target chain is used. For channels opened on the Bitcoin
blockchain, the `chain-hash` field should _always_ be set to:
000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f.

Introducing this new field _immediately_ allows nodes within the network
to open a channel within any Satoshi-derived blockchain. Nodes can have
channels open across different blockchains globally, but also have many
channels open across distinct blockchains with the same peer.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -101,13 +101,11 @@ desire to set up a new channel.
    * [33:first-per-commitment-point]
 
 
-The `chain-hash` value denotes the exact blockchain the opened channel will
-reside within. In order to unique identify a given blockchain, the `chain-hash`
-value is to be sent to the _genesis hash_ of the respective blockchain. The
-genesis hash is the block hash of the _first_ block (the genesis) block of the
-blockchain. The existence of the `chain-hash` allows nodes to open channel
+The chain-hash value denotes the exact blockchain the opened channel will
+reside within. This is usually the genesis hash of the respective blockchain.
+The existence of the chain-hash allows nodes to open channel
 across many distinct blockchains as well as have channels within multiple
-blockchains opened to the same peer (if they support the target chains). 
+blockchains opened to the same peer (if they support the target chains).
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -100,6 +101,14 @@ desire to set up a new channel.
    * [33:first-per-commitment-point]
 
 
+The `chain-hash` value denotes the exact blockchain the opened channel will
```
rustyrussell: Too blockchain-specific.  We should make the language more flexible, eg:

The `chain-hash` value denotes the exact blockchain the opened channel will
reside within. This is usually the _genesis hash_ of the respective blockchain.
The existence of the `chain-hash` allows nodes to open channel
across many distinct blockchains as well as have channels within multiple
blockchains opened to the same peer (if they support the target chains).

02-peer-protocol.md
```diff
@@ -198,6 +211,9 @@ acceptance of the new channel.
 #### Requirements
 
 
+The receiving MAY reject the channel if the `chain-hash` value within the
```
rustyrussell: MUST.  You can't really do anything here, and you certainly can't ignore the field (otherwise you'll get weird failures later).

02-peer-protocol.md
```diff
@@ -113,6 +122,10 @@ FIXME: Describe Dangerous feature bit for larger channel amounts.
 
 #### Requirements
 
+A sending node MUST ensure that the `chain-hash` value be a valid genesis block
```
rustyrussell: A sending node MUST ensure that the `chain-hash` value identifies the chain they they wish to open the channel within.  For the Bitcoin blockchain, the `chain-hash` value MUST be (encoded in hex):
000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f.

02-peer-protocol.md
Roasbeef: Yeah, it really should be mandatory as the channel would be useless to you.

02-peer-protocol.md
Roasbeef: Fixed.

02-peer-protocol.md
Roasbeef: Good point, I've adopted your more general language here.

PR #136, opened 2017-03-22
BOLT 2: require that we don't fulfill close to timeout either.

rustyrussell:
It was never mentioned, but fulfilling a timedout HTLC means a race
between timing out and fulfilling, which is bad.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -487,6 +487,9 @@ so we should fulfill the incoming HTLC as soon as we can to reduce latency.
 ### Risks With HTLC Timeouts
 
 
+Once an HTLC has timed out where it could either be fulfilled or timed-out;
+care must be taken around this transition both for offered and received HTLCs.
+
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -487,6 +487,9 @@ so we should fulfill the incoming HTLC as soon as we can to reduce latency.
 ### Risks With HTLC Timeouts
 
 
+Once an HTLC has timed out where it could either be fulfilled or timed-out;
```
pm47: I don't understand the first part of the sentence, but that's probably because my english is not good enough.

cdecker: I think it refers to the fact that a-posteriori both states are possible, e.g., due to a late reveal of the payment key, and we must force a unique outcome by dropping to the chain early enough.

pm47: Danke schön!

PR #139, opened 2017-03-29
BOLT 4: allow more overpaying.

rustyrussell:
The idea of the "SHOULD fail if amount is too much" was courtesy against
overpaying, but that's a bad idea of you value privacy and some vendor has
well-known prices.  Allow a factor of two, at least.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -592,8 +592,10 @@ If the payment hash is unknown, the final node MUST fail the HTLC:
 1. type: PERM|15 (`unknown_payment_hash`)
 
 If the amount paid is less than the amount expected, the final node
-MUST fail the HTLC.  If the amount paid is more than the amount
-expected, the final node SHOULD fail the HTLC:
+MUST fail the HTLC.  If the amount paid is more than twice the amount
+expected, the final node SHOULD fail the HTLC.  This allows the sender
+to reduce information leakage by altering the amount, without allowing
+accidental gross overpayment:
 
 1. type: PERM|16 (`incorrect_payment_amount`)
 


pm47:
On a related note, I think the randomness of `payment-hashes` is so critical, I wonder if it would be appropriate to generate it using a similar trick as what we have with `revocation-key` generation? Meaning that we would generate it with data from both the sender and the receiver, but only the receiver would be able to compute the preimage?

Let's say a big merchant website is compromised, then currently the attacker could harm the whole network by purposefully reusing the same `payment-hashes` for all payments (thus allowing intermediate nodes to cheat against each other), whereas with a co-generated `payment-hash`, only the actual customers of the website would be harmed (which is inevitable).

Also, what are the risk of a merchant just using a bad random generator, or even re-using payment hashes by mistake?

Cheers,
Pierre

rustyrussell:
> Should the excess amount be returned to the sender by the vendor (A), or is it something the sender is willing to pay to improve its privacy (B)?

B.  This simply allows overpayment for privacy reasons.  It's going to be required, I think, but I have the same nervousness you do, hence the doubling not removing the requirement altogether.

> Let's say a big merchant website is compromised, then currently the attacker could harm the whole network by purposefully reusing the same payment-hashes for all payments (thus allowing intermediate nodes to cheat against each other),

Not really a threat to the network, AFAICT.  Consider A->B->C.  If B knows the preimage already, it just steals A's money.  C says it wasn't paid, but A proves it got the receipt.  It's better than C getting compromised and sending out the wrong bitcoin addresses: in that case A has no proof that it received the wrong address.

> Also, what are the risk of a merchant just using a bad random generator, or even re-using payment hashes by mistake?

Pretty bad; they don't get their funds when someone catches them.  But it's hard to give good advice for generating a deterministic secret, too...

However, this problem somewhat goes away when we switch to keypairs, or (better) Schnorr signatures (see https://medium.com/@rusty_lightning/decorrelation-of-lightning-payments-7b6579db96b0 ).  Since the payer chooses the blinding factors, even reuse is only detectable for the final hop; we should blind this too, so we need payer and payee to both reuse to cause problems.

Note that nothing protects the payee from the payer if the payee reuses, but at least an honest payer can still work.

pm47:
>>Also, what are the risk of a merchant just using a bad random generator, or even re-using payment hashes by mistake?

>Pretty bad; they don't get their funds when someone catches them. But it's hard to give good advice for generating a deterministic secret, too...

If intermediate nodes find out about the vulnerability, they will be able to steal from upstream nodes and the merchant will just not see any incoming payments I think. I find it disturbing that technical incompetence from a merchant may cause customers to loose funds. But on second thought, the customer will be able to prove it sent a payment to a hash signed by the merchant by providing the preimage and make a claim, so we are probably fine.

cdecker:
Well, maybe we can use @pm47's solution A, by routing "through" the intended recipient, back to the sender. That'd mean that I want to pay 1k satoshi to the vendor. I compute a route back to me that goes through the vendor, and delivers a 1k fee there (maybe flagging in the per-hop payload that this is not just generosity), and route the remainder back to me. This'd double the fees, but we are not restricted to arbitrary overpay ranges.

pm47:
That's clever, and it reminds me of an idea from CJP for atomically paying multiple parties using fees.

The issue with this I think, is that you loose the ability to prove that you made a payment to the merchant corresponding to a payment request signed by the merchant, which is a serious drawback?

cdecker:
We can always mix in a secret by the intended recipient, with some crypto-magic :wink: But I'd like that much more than leaving up to the intended payment as cost of privacy.

rustyrussell:
The "loop payment" idea is an interesting one, and can be done today: you just pay a massive fee to the actual merchant hop.  It's not clear how well this works in practice at the moment, since if the route goes back the same way everyone can see the correlation.

With decorrelation this works better, and I think we can even get a receipt back with a little effort...

cdecker:
I think we probably need some more time to discuss the details of the loop payment, since we also need to signal to the intended recipient, and where to forward, both of which is currently done in the next-hop address.

I'm fine with the changes in here: ACK a54fa2baeb0908cae66c64cfc473eeb758c406f9

PR #140, opened 2017-03-30
BOLT 4: error if the final outgoing_ctlv_value is unexpected.

rustyrussell:
Not doing this check means an inconsistency in behaviour, which could
theoretically allow a hop to probe: if the next hop is the last, it might
not care, but if it's not it will return an error.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -174,9 +174,13 @@ Field Description:
      Inclusion of this field allows a node to both authenticate the information
      specified by the original sender and the paramaters of the HTLC forwarded,
 	 and ensure the original sender is using the current `cltv-expiry-delta` value.
+     If there is no next hop, `cltv-expiry-delta` is zero.
      If the values don't correspond, then the HTLC should be failed+rejected as
      this indicates the incoming node has tampered with the intended HTLC
      values, or the origin has an obsolete `cltv-expiry-delta` value.
+     The node MUST be consistent in responding to an unexpected
+     `outgoing_cltv_value` whether it is the final hop or not, to avoid
+     leaking that information.
 
 
 Nodes forwarding HTLC's MUST construct the outgoing HTLC as specified within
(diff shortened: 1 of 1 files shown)


cdecker:
Great catch, I did indeed specify this in the next-hop address, but missed it here.

04-onion-routing.md
```diff
@@ -601,6 +606,20 @@ If the `cltv-expiry` is too low, the final node MUST fail the HTLC:
...
 1. type: 17 (`final_expiry_too_soon`)
 
+If the `outgoing_cltv_value` does not match the `ctlv-expiry` of the
+HTLC at the final hop:
+
+1. type: 18 (`final_incorrect_cltv_expiry`)
+2. data:
+   * [4:cltv-expiry]
+
+If the `amt_to_forward` does not match the `incoming_htlc_amt` of
```
cdecker: Hm, this is introducing special handling for the last hop when it comes to the `amt_to_forward`, since this is the amount the node is allowed to keep, maybe we can flip this with the `fee`? That way validation is always identical:

- keep `fee`, i.e., full amount on last hop
- forward `amt_to_forward`, i.e., 0 on the last hop

rustyrussell: That doesn't work if we allow non-precise payments, though.  The second-to-final node can take more fee than expected, and the last node won't detect it if it allows anything other than a specific amount  (eg. donation address, and even before #139 overpaying detection was a SHOULD not a MUST)...

cdecker: Huh? The sender does know how much should be left for the last hop, so she can just set the fee to that value. Really it should always be `incoming_htlc_amt = fee + amt_to_forward`, so it is indifferent whether we have 0 `amt_to_forward` or 0 `fee`.

rustyrussell: Oh, I see your suggestion: that we put 'fee' in the onion instead of 'amt_to_forward'?

Consider A->B->C->D.

The advantage of `amt_to_forward` is that if node B takes more fee than expected, node C detects it immediately, and returns an error, blaming B.  If C only knows fee, it subtracts it and then D returns an error (insufficient payment), and A has no idea which node took too much...

cdecker: Oh, you're right, I got it mixed up, we're only saying how much to forward, not the fee. So fee shaving in the hopes of being the second to last hop is indeed a problem, since it would result in failed payments if we're not the second to last hop. Happy to merge your version as is :wink:

PR #143, opened 2017-04-03
bolt07: Add flag to disable channels temporarily.

cdecker:
This is useful to signal either permanent or temporary channel unavailability. Temporary unavailability may include anything from a connection loss for mobile nodes, or prolonged lack of capacity, though I'd like to discourage the latter. In the case of permanent unavailability this can frontrun the actual closure and tell nodes not to use this channel (-direction) even before they see the on-chain close.

It's more of a heads up, and on-chain settlements should be used for permanent removal.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -276,11 +276,6 @@ it wants to change fees.
     * [4:fee-base-msat]
     * [4:fee-proportional-millionths]
 
-### Requirements
-
-The creating node MUST set `signature` to the signature of the
-double-SHA256 of the entire remaining packet after `signature` using its own `node-id`.
-
 The flags bitfield is used to indicate the direction of the channel this update concerns, i.e., it identifies the node that this update originated from, and signal various options concerning the channel.
 The following table specifies the meaning of the individual bits:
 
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -281,8 +281,19 @@ it wants to change fees.
...
-The creating node MUST set `short-channel-id` to
-match those in the already-sent `channel_announcement` message, and MUST set the least-significant bit of `flags` to 0 if the creating node is `node-id-1` in that message, otherwise 1.  It MUST set other bits of `flags` to zero.
+The flags bitfield is used to indicate the direction of the channel this update concerns, i.e., it identifies the node that this update originated from, and signal various options concerning the channel.
+The following table specifies the meaning of the individual bits:
+
+| Bit Position  | Name        | Meaning                          |
+| ------------- | ----------- | -------------------------------- |
+| 0             | `direction` | Direction this update refers to. |
+| 1             | `disable`   | Disable the channel.             |
+
```
rustyrussell: This table belongs above the "Requirements" section, as it's explanatory.  But looks good.

Is that the third style of table we have so far? :)

cdecker: The table style matches the one in bolt 09-features, and also matches the Github Flavored markdown :wink: 

I moved the table and associated text out of the requirements section.

07-routing-gossip.md
```diff
@@ -276,13 +276,24 @@ it wants to change fees.
...
 ### Requirements
 
 The creating node MUST set `signature` to the signature of the
 double-SHA256 of the entire remaining packet after `signature` using its own `node-id`.
 
-The creating node MUST set `short-channel-id` to
-match those in the already-sent `channel_announcement` message, and MUST set the least-significant bit of `flags` to 0 if the creating node is `node-id-1` in that message, otherwise 1.  It MUST set other bits of `flags` to zero.
+The creating node MUST set `short-channel-id` to match those in the already-sent `channel_announcement` message, and MUST set the least-significant bit of `flags` to 0 if the creating node is `node-id-1` in that message, otherwise 1.
```
pm47: Nitpicking:
> The creating node MUST set `short-channel-id` to match those in the already-sent `channel_announcement` message, and MUST set the `direction` bit of `flags` to 0 if the creating node is `node-id-1` in that message, otherwise 1.

cdecker: Fixed in the latest commit, good catch :+1:

PR #144, opened 2017-04-03
bolt10: Initial draft of the subdomain structure.

cdecker:
Thought I should finally write this up, and see if it people like it. The DNS seeds give us an easy way to bootstrap new nodes, encourage random connections and help nodes find previous peers again, should they get disconnected or change network.

This PR is still a bit thin on the policies side, but it has pointers to the Bitcoin DNS seed policy :wink: 
Also querying on features and realm could be interesting to add.

The risk in relying on the DNS seeds should be minimal. The queries get cached by the ISP, i.e., we do not see the client's IP or even how many in the ISP's network, and the authenticated encryption on the transport layer should prevent MitM attacks.

I'll polish my reference implementation a bit and link to it later.

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -291,3 +291,16 @@ UTC
 inline
 fundee's
 BOLTs
+DNS
+subdomain
+subdomains
+wildcard
+tuple
+tuples
+resolvers
+hostnames
+prepending
+A
+AAAA
+SRV
+TTL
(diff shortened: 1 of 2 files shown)


cdecker:
Ok, finally got around to rework the BOLT, it now uses dot-separated conditions that allow to restrict what the client would want to receive. These conditions include:

 - realm
 - address class
 - `node_id`
 - numer of desired results

and may be extended later. The reference implementation is currently being reworked, but should be working soon :wink:

cdecker:
Forgot to mention this during the last call, but it should be ready to merge. The reference implementation is up and running (though still very hacky due to the pivot halfway through).

What do you guys (@rustyrussell @Roasbeef @sstone) think?

cdecker:
Great, thanks guys ^^

Issue #98, opened 2017-01-31
BOLT 7: identify next hop using short-channel-id instead of node-address


pm47:
In order to identify the next hop, we currently use a `node-address` defined as `hash160(node-id)`. Using a `channel-id` would work the same way, and use 8 bytes instead of 20 bytes.

Not only is `channel-id` smaller than `node-id`, but it also carries more information because it identifies the next node and the exact channel to use. If we allow several channels between the same two nodes (I don't remember what we decided in Milan), we need to know what channel to use when forwarding the htlc because the fees/expiry-delta might not be the same. So even if we need to use a hash for some other security reason, using hash160(channel-id) instead of hash160(node-id) would still be useful.

cdecker:
Using the `channel-id` certainly is a possibility, and it would have the potential of reducing the size of the packet. My first concern was that the node processing the packet may be forced to use a specific channel, even though it has a cheaper alternative, however it can always ignore the instruction in the `address` field. On the other hand I don't think that specifying the exact channel changes too much, even though it conveys more information, after all either we have a channel that requires an equal or smaller fee or we don't. The same is true for all parameters.

Whatever we put in the `address` field, just has to make sense to the origin of the packet, since they'll have to pick an alias for a channel, and the node processing the packet, since they'll have to forward it. We could even add a new alias, that is node specific and just enumerates the node's channels, but adding aliases probably is a bad idea.

rustyrussell:
Christian Decker <notifications@github.com> writes:
> Using the `channel-id` certainly is a possibility, and it would have the potential of reducing the size of the packet. My first concern was that the node processing the packet may be forced to use a specific channel, even though it has a cheaper alternative, however it can always ignore the instruction in the `address` field. On the other hand I don't think that specifying the exact channel changes too much, even though it conveys more information, after all either we have a channel that requires an equal or smaller fee or we don't. The same is true for all parameters.

Well, if peers agree to send via some other means, that's a private
arrangement.

I like using channel-id:
1) It's smaller.
2) It simplfies error handling; we know which channel information to
   send in response.

Cheers,
Rusty.

cdecker:
Ok, thinking about it some more `channel-id` seems to me. The more important question is whether we want to adjust the routing info size as well to use `channel-id`s as well. It means we save `20 * (20 - 8) = 240` bytes on each packet. The downside is that we're pretty much locked to just use `channel-id`s since this is a parameter that must be the same for all hops on a path.

[...]

PR #145, opened 2017-04-04
First attempt (unimplemented!) at simplifying the onion:

rustyrussell:
1. Only one per-hop thing, called `per-hop`, or `hops_data` when in aggregate.
2. Move HMAC to the end of stuff it covers, both of the packet itself, and the per-hop.
3. Use `channel-id` instead of RIPEMD(nodepubkey).
4. Use 4 byte amounts.
5. This is all for realm "0", we can have future realms.  We also have 16
   bytes of unused padding.
6. No longer need the `gamma` key, but document the `_um_` key used for
   errors.

Q: should `associated data` be the version byte?  AFAICT we don't commit to
   that.


cdecker:
The associated data is the payment hash this onion is associated with. This is our replay protection as specified in [02-peer-protocol.md](02-peer-protocol.md) :wink: 

It is appended to the payload and considered for the next hop's HMAC both when creating the onion as well as verifying its integrity at the hop.

rustyrussell:
> The associated data is the payment hash this onion is associated with. This is our replay protection as specified in [02-peer-protocol.md](02-peer-protocol.md) :wink: 
>
> It is appended to the payload and considered for the next hop's HMAC both when creating the onion as well as verifying its integrity at the hop.

Ah, right, thanks for the reminder.

It still concerns me a little that we don't commit to the version field
though I can't see an immediate flaw.

pm47:
I like this proposal, unsurprisingly! Especially the fact that we route to a channel, and also I never really understood the two-parts payload. I guess Laolu will be the one to convince?

(linking to #98)

cdecker:
I implemented this proposal in a feature branch, it's working nicely and saves us quite a few complications :-) So for me this is a definite :+1:

04-onion-routing.md
```diff
@@ -76,81 +70,59 @@ The use of a fixed nonce is safe since the keys are never reused.
...
-The per-hop payloads has a similar structure:
+The `realm` byte determines the format of the `per-hop`; so far
+only `realm` 0 is defined, and for that, the `per-hop` format is:
 
 ~~~~
-+-------------+-------------+-------------+-----+------------+
-| n_1 payload | n_2 payload | n_3 payload | ... | hop filler |
-+-------------+-------------+-------------+-----+------------+
++----------------------+--------------------------+-------------------------------+--------------------+
+| channel_id (8 bytes) | amt_to_forward (4 bytes) | outgoing_cltv_value (4 bytes) | padding (16 bytes) |
```
Roasbeef: Including the `channel_id` here is a bit of an abstraction leak. Why should the payment routing layer dictate exactly _which_ channel should be utilized by the next node when forwarding a payment? In the case of a node with multiple same-realm links (channels) to another node, the node accepting the HTLC to forward has the most up-to-date contextual information which allows it to forward the payment in an optimal manner. If we allow the sender to dictate which _channel_ rather than which _node_ to forward the HTLC to, then we rob intermediate nodes of HTLC forwarding flexibility. 

From another angle: ideally, the routing layer in implementations treats multiple same-realm links between two nodes as a _single_ link, and performs path finding taking this simplification into account. Treating these as a single link simplifies path finding, and also allows intermediate forwarding nodes to freely fragment payments at the link level or, in a simpler case, have complete control as to which channels they wish to use to satisfy the forwarding requirements. 

In my opinion, either a pubkey, or a pubkey hash (a before) should be used in-place of the `channel_id`. I still stand by my opinion that `channel_id`'s are only really useful for light clients within the network when verifying channel authentication proofs. Once Bitcoin gains UTXO commitments, I'd be open to making a push to eliminate them from the protocol all together as plain old outpoints let us avoid all the re-org edge cases we currently need to account for due to `channel_id`s. 

Also why 4 bytes for the `amt_to_forward`? This isn't enough to express satoshis, lets alone milli satoshis.

04-onion-routing.md
```diff
@@ -76,81 +70,59 @@ The use of a fixed nonce is safe since the keys are never reused.
...
+ - A 20-byte `HMAC` used to verify the packet's integrity.
 
 The overall structure of the packet is depicted below. The network format of the packet consists of the individual parts being serialized into one continguous byte-stream and then transferred to the recipient of the packet. Due to the fixed size of the packet it does not need to be prefixed by its length when transferred over a connection.
 
 ~~~~
-+--------+-----------------+
-| header | per-hop payload |
-+--------+-----------------+
++------------------+-----------------------+------------------------+-----------------+
+| version (1 byte) | public key (33 bytes) | hops_data (20x53 bytes) | HMAC (20 bytes) |
```
Roasbeef: In this proposal, since the 20-byte `next_hop` field has been eliminated, the MAC's no longer need to be 20 bytes. If we move forward with this, we can instead modify the packet-wide MAC to be 32-bytes.

04-onion-routing.md
pm47: > Including the channel_id here is a bit of an abstraction leak. Why should the payment routing layer dictate exactly which channel should be utilized by the next node when forwarding a payment?

To me, the reason why the payment routing layer should use channels is that it actually only deals with channels. My opinion is that using `channel_id` is the natural thing to do, and using node identifier would be breaking the separation of layers.

> If [so], we allow the sender to dictate which channel rather than which node to forward the HTLC to, then we rob intermediate nodes of HTLC forwarding flexibility.

We don't, the forwarding node still can use whatever channel it prefers, right?

> From another angle: ideally, the routing layer in implementations treats multiple same-realm links between two nodes as a single link, and performs path finding taking this simplification into account.

But it doesn't seem very consistent with channel-centric failure messages (BOLT 4). Why do we return `0x1000` (UPDATE) errors if we only care about node-node links? What if the sender asks a route based on channel X and received an error/update for channel Y?

>Treating these as a single link simplifies path finding, and also allows intermediate forwarding nodes to freely fragment payments at the link level or, in a simpler case, have complete control as to which channels they wish to use to satisfy the forwarding requirements.

I have never understood how payment fragmentation works. How can the forwarding node split the onion?

> Also why 4 bytes for the amt_to_forward? This isn't enough to express satoshis, lets alone milli satoshis.

FWIW, this is consistent with the 4 bytes used in the `update_add_htlc` message. I think @rustyrussell wanted to put a hard upper limit on amounts.

04-onion-routing.md
cdecker: :+1:

04-onion-routing.md
cdecker: Indeed the amount uses 4 bytes to limit the per-transfer capacity to 0.04 BTC initially, but it is rather straightforward to increase this (using a new realm) once we feel confident that the implementations are safe to use for larger values.

As to the use of `short_channel_id`, I'm rather indifferent, both sides have points, the `id` is shorter and can be ignored by the processing hop if it has a parallel channel that is better suited, but the pubkey matches the network topology better and simplifies some of the computations.

BTW I also don't think we can currently do ad-hoc multipath with the onion routing.

04-onion-routing.md
rustyrussell: OK, will fix mac.

04-onion-routing.md
rustyrussell: 1. Our gossip infrastructure is in terms of channels: *not* using it here is an abstraction violation!  In particular, we are asking to use the channel at its advertized capacity and rate.  If you fail it, you have to provide a reason based on that channel (you can succeed any way you want of course, nobody can stop you).

2. 4 byte amounts are a logical corollary of 4 byte HTLC amounts.  If a path is upgraded later to allow 8 bytes, it'll use a new realm byte or the extra space (conveniently, this is 0 right now!).

Roasbeef: Hah, yeah I overlooked the whole onion packet thing for the link-layer fragmentation idea. 

All the other messages use 8 bytes for HTLC amounts, and generally for any amounts dealing with Bitcoin. I thought the idea was to implement this as "policy" meaning that nodes would reject any amount violating the 8 byte constraint, rather than modify any wire messages or other encapsulated protocol message like the sphinx packet.

Roasbeef:
A meta-question: if we're open to completely changing the sphinx packet format, which has been itself complete for 5+ months, then what else is free to change within the spec at this point? 

Personally, I've been holding back on making some changes for the sake of implementation convergence, and also to not waste our collective development time. In my opinion, at this point (almost 6 months since we've started to draft) changes to the spec should be constrained to: fixing things that are broken, adding additional details, eliminating ambiguities or patching holes in the spec. FWIW, this change doesn't fall into either of those categories (that I just made up 5 mins ago).

rustyrussell:
I was just having the same discussion with @cdecker.  That's the reason I was so tentative on this proposal; I didn't think we'd see anything at this stage of this magnitude :(  I agree with @Roasbeef 's list BTW, though perhaps some wiggle-room for trivial stuff.

On the other hand, I confess that it's the first time I really digested that part of the spec, and I was unable to avoid mentally simplifying it; Christian implemented it and concurred.

I think we might need to drag in @pm47 and/or @sstone to see what they think?  They, too, would need to reimplement.

pm47:
I too agree with @Roasbeef's statement, but to be fair we opened #98 as soon as we started implementing this part of the spec so I am not sure the 5 months duration should really be taken litterally. Furthermore, as stated above I do think node addressing is kind of inconsistent with current failure messages, to some extent. However if we consider it sets a bad precedent wrt to finalization of the spec, I would happily keep the onion as it is -- OTOH reimplementing is not an issue.

04-onion-routing.md
```diff
@@ -210,41 +183,35 @@ This recursive algorithm is initialized by setting the first hop's (`k=1`) ephem
 The sender then iteratively computes the ephemeral public keys, shared secrets and blinding factors for nodes `{n_2, ..., n_r}`.
 
 Once the sender has all the required information it can construct the packet.
-Constructing a packet routed over `r` hops requires `r` 32 byte ephemeral public keys, `r` 32 byte shared secrets, `r` 32 byte blinding factors and `r` 20 byte per-hop payloads.
-The construction returns one 1254 byte packet and the first hop's address.
+Constructing a packet routed over `r` hops requires `r` 32 byte ephemeral public keys, `r` 32 byte shared secrets, `r` 32 byte blinding factors and `r` 65 byte `per-hop` payloads.
+The construction returns one 1114 byte packet and the first hop's address.
```
Roasbeef: How was this `1114` value calculated? From my breakdown we get: `1 + 33 + (20*65) + 32 = 1366`

04-onion-routing.md
```diff
@@ -76,81 +69,59 @@ The use of a fixed nonce is safe since the keys are never reused.
...
 
 The overall structure of the packet is depicted below. The network format of the packet consists of the individual parts being serialized into one continguous byte-stream and then transferred to the recipient of the packet. Due to the fixed size of the packet it does not need to be prefixed by its length when transferred over a connection.
 
 ~~~~
-+--------+-----------------+
-| header | per-hop payload |
-+--------+-----------------+
++------------------+-----------------------+------------------------+-----------------+
+| version (1 byte) | public key (33 bytes) | hops_data (20x65 bytes) | HMAC (32 bytes) |
++------------------+-----------------------+------------------------+-----------------+
```
Roasbeef: Looking at this again, I don't think we actually need this final HMAC over the _entire_ packet, as each of the per-hop HMAC's themselves will ensure integrity of the entire packet (along with committing to the associated data).

cdecker: This is just the HMAC extracted by hop `i` form `hops_data` and passed to hop `i+1` so that it can check the integrity. So it's not a separate HMAC, but one of the `hops_data` HMACs decrypted and exposed.

This also means that the HMACs are shifted by one, leaving the last HMAC blank, which is what we currently use to identify the terminal. We could rework it so that every hop decrypts its own HMAC, but that may be a bigger rework in the handling logic.

cdecker:
I took the liberty of adding a commit that updates the test vectors for the new format as well. They match the C version of `lightningd` (cdecker/lightning@dd1c46c3a6eeb97d2688d167f36007c9dc6e8c9e) and the go version (cdecker/lightning-onion@5b9cf349222bed4d6d3453f390119f53cf300f1c). I hope I did not introduce any copy-paste errors, these blobs are a bit unwieldy :wink:

Hopefully Eclair comes to the same results, let me know if I can assist in any way.

sstone:
Yes, we do find the same results. Thanks!

cdecker:
Great, thanks @sstone, let's leave this open till the next scheduled meeting, so we can collect any remaining feedback, but it seems so far we all agree on merging it :+1:

04-onion-routing.md
rustyrussell: Oops, that was calculated before I restored the truncated HMACs.  Should be 1366 (1 + 33 + (20*65) + 32).  Pushing fixup.

rustyrussell:
Hey, @cdecker can you paste in the new example Go code too? func ConstructPacket is now wrong. Thanks!

04-onion-routing.md
rustyrussell: We could leave the full hop on the front, like:

    | version (1 byte) | public key (33 bytes) | [n_0 realm | n_0 per-hop | n_0 HMAC ] | hops_data (20x65 bytes) |

Since per-hop is (for realm 0):

    | channel_id (8 bytes) | amt_to_forward (4 bytes) | outgoing_cltv_value (4 bytes) | padding (16 bytes) |

We could then remove `amount-msat` and `cltv-expiry` from the `update_add_htlc` message.  The channel-id and padding are redundant, but OTOH it probably makes adding future realms easier.

cdecker:
Good point, pushed a new commit with the code change.

04-onion-routing.md
cdecker: That's the format we forward from one hop to the next, so we definitely do not want to have the `channel_id`, `forward_amt` and `outgoing_cltv` in there since those are instructions for the receiving hop and would be transferred in cleartext, hence hop `n` would see the destination of the next forwarding instruction, i.e., who hop `n+2` is.

This is the shift I was referring to, we decrypt our own instructions, and the HMAC for the _next_ hop, which is kinda weird but that's how it works.

rustyrussell:
Squashed the fixes, and applied.

PR #148, opened 2017-04-08
BOLT 4: clarify record keeping requirements.

rustyrussell:
We didn't note the actual requirements: we MUST reject replays we have forwarded
or paid to avoid replay attacks.  The details are difficult however; we have
to clean them out at some stage, and restrict the size somehow.  Suggest some
ways we could do that.


04-onion-routing.md
```diff
@@ -327,8 +327,8 @@ The node MUST check that the ephemeral public key is on the `secp256k1` curve.
 Should this not be the case the node MUST abort processing the packet and report a route failure to the sender.
 
 The node then computes the shared secret as described below, using the private key corresponding to its public key and the ephemeral key from the packet.
-The node MUST keep a log of previously used shared secrets.
-Should the shared secret already be in the log it MUST abort processing the packet and report a route failure, since this is likely a replay attack, otherwise the shared secret is added to the log.
+
+The node MUST abort processing and report a route failure if it receives a duplicated routing info which it has already forwarded or redeemed locally.  This prevents a node on the route from retrying a payment multiple times and attempting to track its progress by traffic analysis.  Note that this could be done using a log of previous shared secrets or HMACs, which can be forgotten once that HTLC would not be accepted anyway (eg. once `outgoing_cltv_value` has passed).  Such a log may use a probabilist data structure, but MUST rate-limit commitments as necessary to constrain the worst-case storage requirements or false positives of this log.
```
Roasbeef: probabilist -> probabilistic

Roasbeef: Alternatively, since we currently commit to the payment hash within the sphinx packet-wide MAC, if a node encounters a duplicate payment hash along with an identical sphinx packet, then they can simply directly redeem the HTLC rather than continuing to forward it.

cdecker: We were unsure whether we want to make this the default action, instead of being nice and not taking the money. If you take the money then the network becomes fingerprintable, i.e., I saw that there is a payment going through and I can just try the same payment key at ever node in the network. This will no longer be a problem once we have the disassociation between payment keys.

rustyrussell: True.  In this case it's not a privacy compromise to immediately redeem.

It's probably extra work for no actual gain, but let's allow it:

    The node MUST detect a duplicated routing info which it has already forwarded or redeemed 
     locally; it MAY immediately redeem the HTLC using the preimage (if known), otherwise it MUST
     abort processing and report a route failure.

PR #149, opened 2017-04-12
[Trivial] Include a channel_update with the disable on temporary channel failure

cdecker:
Should be simple enough: if you fail a channel because it is temporarily unavailable, be nice enough to include the `channel_update` so we can spread the word

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -533,6 +533,9 @@ If an otherwise unspecified transient error occurs for the outgoing
 channel (eg. peer unresponsive, channel capacity reached):
 
 1. type: 7 (`temporary_channel_failure`)
+2. data:
+   * [2:len]
+   * [len:channel_update]
 
 If an otherwise unspecified permanent error occurs for the outgoing
 channel (eg. channel (recently) closed):


rustyrussell:
I was going to suggest that the Requirements be updated to insist that the update message have the 'disabled' bit, but that's already covered:

  * The origin node MAY treat the `channel_update` as invalid if it
    should not have caused the failure.

So, ack from me!

pm47:
Sorry for not having spoken up in time, but I am not convinced by this change.

To me, `temporary_channel_failure` messages are reserved for errors for which:
- either the receiver don't need to know the details (eg: in flight htlc value too high, too many accepted htlcs) 
- or the sender don't want to give the details (eg: insufficient funds because probing attack).

This error means "please try again in a few seconds": this is not the case for a channel that is disabled a can stay that way for days.

Also, there is a dedicated `0x1000 (UPDATE)` type for error messages that enclose an update which `temporary_channel_failure` is missing.

I'd rather introduce a new message:

    type: UPDATE|XX (channel_disabled)
    data:
        [2:len]
        [len:channel_update]

and keep temporary_channel_failure as-is.

rustyrussell:
> Sorry for not having spoken up in time, but I am not convinced by this change.
>
> To me, `temporary_channel_failure` messages are reserved for errors for which:
> - either the receiver don't need to know the details (eg: in flight htlc value too high, too many accepted htlcs) 
> - or the sender don't want to give the details (eg: insufficient funds because probing attack).
>
> This error means "please try again in a few seconds": this is not the case for a channel that is disabled a can stay that way for days.

If we allow failure without an update, we open the door to advertizing a
route and not actually forwarding anything, which is why every failure
on a well-formed offer should be accompanied by an update.  Eventually
(1.1?)  we'll want to sign and wrap failures in an onion on the way
back, such that the originator can prove any bad behaviour.

Meanwhile, in theory you could broadcast another update once the channel
is clear, or if you see that disabled update on the network.  The
alternative is to include a timeout field in the `channel_update` when
the disabled flag is set, but that's fraught with clock issues...

Cheers,
Rusty.

pm47:
Hey Rusty,

> If we allow failure without an update, we open the door to advertizing a route and not actually forwarding anything

But this is always possible, for example in an adversarial scenario we can image that a node will "never be happy" and always provide a new update with more fees or a greater expiry_delta or whatever? In the end I think it really depends on the sender deciding that this particular channel isn't reliable, if we want to stay trustless.
> Eventually (1.1?) we'll want to sign and wrap failures in an onion on the way back, such that the originator can prove any bad behaviour.

It was Fabrice who worked on this so I am not 100 % sure but I'll go out on a limb and say that it is already the case? We can already prove that an error message originates from a given node I think. Maybe Christian could confirm?

> Meanwhile, in theory you could broadcast another update once the channel is clear, or if you see that disabled update on the network. The alternative is to include a timeout field in the `channel_update` when the disabled flag is set, but that's fraught with clock issues...

I don't get this part. I completely agree that `channel_disabled` should be accompanied by a `channel_update`, I merely advocated that it would better be done in a dedicated message, and that we should keep the possibility to not provide a `channel_update` for some kind of transient errors. I think there will always be errors (eg: too many in-flight htlcs) that will never be explained/justified by a `channel_update`.

Cheers and happy Workers' Day :tada: (is that a thing outside France?)

Pierre

cdecker:
> But this is always possible, for example in an adversarial scenario we can image that a node will "never be happy" and always provide a new update with more fees or a greater expiry_delta or whatever? In the end I think it really depends on the sender deciding that this particular channel isn't reliable, if we want to stay trustless.

Yes, an unhappy node that just rejects everything can always happen. The update essentially gives us a way to tell other nodes about it, putting a bit of a cost on such behavior, and reducing its impact. If the rejection happened due to a real reason we should be able to describe it in an update, and the recipient of that update can forward it to other nodes so they don't attempt the same failing path. If the node just doesn't like to forward we can tell our neighborhood about it and they will try to avoid that node as well.

> It was Fabrice who worked on this so I am not 100 % sure but I'll go out on a limb and say that it is already the case? We can already prove that an error message originates from a given node I think. Maybe Christian could confirm?

Yes, we have an onion wrapped reply, that is sort of authentic (HMAC), but we are missing the ability to prove failures to third parties. Basically the ability to broadcast a failure telling the world that there is something fishy along the failed route, and putting a cost on this sort of misbehavior, since others will route around that as well. Currently we only have the reply going to the sender, but others cannot verify its authenticity.

> I merely advocated that it would better be done in a dedicated message, and that we should keep the possibility to not provide a channel_update for some kind of transient errors. I think there will always be errors (eg: too many in-flight htlcs) that will never be explained/justified by a channel_update.

The hope is that these transient errors can be somehow mapped to a `channel_update` that is useful for others as well, the sledgehammer approach being the `channel_disabled` flag if nothing else fits. For example too many in-flight HTLCs might be a good case to ramp up fees, to discourage others putting more load on that channel. That being said I can live without the `channel_update` being mandatory :wink: 

> Cheers and happy Workers' Day :tada: (is that a thing outside France?)
Thanks, you too :fireworks:

pm47:
> Yes, we have an onion wrapped reply, that is sort of authentic (HMAC), but we are missing the ability to prove failures to third parties.

Got it, thanks.

> The hope is that these transient errors can be somehow mapped to a channel_update that is useful for others as well, the sledgehammer approach being the channel_disabled flag if nothing else fits. For example too many in-flight HTLCs might be a good case to ramp up fees, to discourage others putting more load on that channel. 

In understand the idea, that makes sense. For version 1.0 though I think we would have been fine with a more basic "just try again later" semantic.

The specific case of a disconnected peer still deserves its own error message, because it is not really a transient error: the origin node wouldn't have sent the payment in the first place, had it received the `channel_update` with disable=1 in time via the normal broadcast mechanism. In this regard, it is very similar to `amount_below_minimum`/`fee_insufficient`/`incorrect_cltv_expiry`/`expiry_too_soon`, so treating it differently would not be consistent. FWIW I believe a lot of channels will regularly go on and off, so this will happen a lot.

> That being said I can live without the channel_update being mandatory 😉

And I can live with it being mandatory, but only if its type has the `UPDATE` flag :grin:

Cheers,

Pierre

rustyrussell:
> Hey Rusty,
>
>> If we allow failure without an update, we open the door to advertizing a route and not actually forwarding anything
>
> But this is always possible, for example in an adversarial scenario we can image that a node will "never be happy" and always provide a new update with more fees or a greater expiry_delta or whatever? In the end I think it really depends on the sender deciding that this particular channel isn't reliable, if we want to stay trustless.

True, but such an attack doesn't scale; the first node broadcasts the
first update reply, etc.  It's a little complex, because we also need to
get reliable timestamps on the updates which are currently unrestricted.

(I have a scheme for this, BTW.  Only allow (say) 16 updates per block,
and use a block count, the lower 4 bits of the latest blockhash a 4 bit
counter.  That prevents you from producing future timestamps, and could
be combined with a rule that you only accept new updates which are
within the last few blocks.)

>> Eventually (1.1?) we'll want to sign and wrap failures in an onion on the way back, such that the originator can prove any bad behaviour.
>
> It was Fabrice who worked on this so I am not 100 % sure but I'll go out on a limb and say that it is already the case? We can already prove that an error message originates from a given node I think. Maybe Christian could confirm?

No, you can send a corrupt or invalid update, and nobody can prove
anything :(  You don't sign it with your key, and the recipients don't
sign that they received it intact, so they could corrupt it too.

>> Meanwhile, in theory you could broadcast another update once the channel is clear, or if you see that disabled update on the network. The alternative is to include a timeout field in the `channel_update` when the disabled flag is set, but that's fraught with clock issues...
>
> I don't get this part. I completely agree that `channel_disabled` should be accompanied by a `channel_update`, I merely advocated that it would better be done in a dedicated message, and that we should keep the possibility to not provide a `channel_update` for some kind of transient errors. I think there will always be errors (eg: too many in-flight htlcs) that will never be explained/justified by a `channel_update`.

Then a node could advertise a free channel, then send "transient
failure" every time.  I'd rather have transient failure provable.

I'd prefer to see if this is a common problem: for now, nodes won't
broadcast the update they get anyway.

> Cheers and happy Workers' Day :tada: (is that a thing outside France?)

Yep, it's May Day here...

Thanks!
Rusty.

PR #156, opened 2017-05-02
Clarify init-message description

EmelyanenkoK:
Add link to BOLT-09 and some explanations.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -84,7 +84,7 @@ Once authentication is complete, the first message reveals the features supporte
 |`0b01` | Party emitted message requires this feature|
 |`0b00` | Ignore: it is padding.|
 
-However each known feature can define it's own semantic for bits (see [BOLT #9](09-features.md)). ⚠️ Note, this part of specification is under development yet.
+However each known feature can define it's own semantic for bits (see [BOLT #9](09-features.md)). Note, this part of specification is under development yet.
 
 Both fields `globalfeatures` and `localfeatures` MUST be padded to bytes with zeros.
 
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -75,16 +75,18 @@ a buffer with 6 bytes of pre-padding.
 ### The `init` message
 Once authentication is complete, the first message reveals the features supported or required by this node, even if this is a reconnection.
 
-[BOLT #9](09-features.md) specifies lists of global and local features. Each feature is represented in `globalfeatures` or `localfeatures` by 2 bits. First bit (Most Significant Bit First notation) indicates if the feature is supported or not, second one indicates if the feature is required by party emitted message or not. Thus 2-bits field can indicate:
+[BOLT #9](09-features.md) specifies lists of global and local features. Each feature is represented in `globalfeatures` or `localfeatures` by 2 bits. Default meaning for bits:
 
 | | |
 |---|---|
-|`0b11` | Party emitted message supports this feature but doesn't require it|
-|`0b10` | Party emitted message supports this feature and requires it|
-|`0b01` | Party emitted message doesn't support this feature and doesn't require it|
+|`0b11` | Ignore: Undefined behviour |
+|`0b10` | Party emitted message supports this feature but it is optional|
+|`0b01` | Party emitted message requires this feature|
 |`0b00` | Ignore: it is padding.|
 
-Both fields `globalfeatures` and `localfeatures` should be padded to bytes.
+However each known feature can define it's own semantic for bits (see [BOLT #9](09-features.md)). ⚠️ Note, this part of specification is under development yet.
+
+Both fields `globalfeatures` and `localfeatures` MUST be padded to bytes with zeros.
 
 1. type: 16 (`init`)
 2. data:
@@ -102,19 +104,24 @@ connection.
 The sending node SHOULD use the minimum lengths required to represent
 the feature fields.  
 
-The sending node SHOULD set first feature bit to `0b1` for supported features, and MUST set second feature bit to `0b1` for required features.
+For known features node should follow this feature rules.
 
-The receiving node MUST fail the connection if it receives a message in which unsupported or unknown feature has second bit set to `0b0`.
+For unknown features and features without it's own semantic, node should follow default set of rules:
 
-The receiving node MUST fail the connection if it receives a message in which required (by receiving node) feature has first bit set to `0b0`.
+1. The sending node SHOULD set feature bits to `0b10` for optional features, and MUST set feature bits to `0b01` for required features.
 
-The receiving node MUST NOT use feature during session if it receives a message in which this feature has first bit set to `0b0`.
+2. The receiving node MUST fail the connection if it receives an `init` message in which unsupported or unknown feature has bits set to `0b01`.
+
+3. The receiving node MUST fail the connection if it receives a message in which required (by receiving node) feature has bits set to `0b00`.
+
+4. The receiving node MUST NOT use feature during session if it receives a message in which this feature has bits set to `0b00`.
+
 
 Each node MUST wait to receive `init` before sending any other messages.
 
 #### Rationale
 
-The even/odd semantic allows future incompatible changes, or backward
+This semantic allows future incompatible changes, or backward
 compatible changes.  Bits should generally be assigned in pairs, so
 that optional features can later become compulsory.
 
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -75,8 +75,9 @@ a buffer with 6 bytes of pre-padding.
 ### The `init` message
 
 Once authentication is complete, the first message reveals the features supported or required by this node, even if this is a reconnection.
-Odd features are optional, even features are compulsory (_it's OK to be odd_).
-The meaning of these bits will be defined in the future.
+
+[BOLT #9](09-features.md) specifies lists of global and local features. Each feature is represented in `globalfeatures` or `localfeatures` by 2 bits: first one indicates is it feature mandatory (`0b1`) or not(`0b0`), and second one indicates is it feature supported or not. Odd features are optional, even features are compulsory (_it's OK to be odd_).
+`globalfeatures` and `localfeatures` fields should be padded to ceil number of bytes.
 
 1. type: 16 (`init`)
 2. data:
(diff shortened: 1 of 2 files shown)


rustyrussell:
I like the reference to BOLT 9, but the rest is not quite right.

*Even bits* mean you can't use a node unless you understand them, and if you do understand them, you must support that feature.  *Odd bits* don't.

pm47:
Even if @EmelyanenkoK didn't get it right on his first pass, I think the explanation effort is worthwhile. @sstone and I always had trouble wrapping our heads around the exact semantic of _it's OK to be odd_.

cdecker:
Yeah, I think _it's ok to be odd_ should be specified somehow like this: if `has_even_bits(nodefeatures xor peerfeatures)` then you may not talk to this peer, ever. `has_odd_bits(nodefeatures xor peerfeatures)` you may talk to the peer, but some features may need to be turned off.

01-messaging.md
```diff
@@ -75,8 +75,9 @@ a buffer with 6 bytes of pre-padding.
 ### The `init` message
 
 Once authentication is complete, the first message reveals the features supported or required by this node, even if this is a reconnection.
-Odd features are optional, even features are compulsory (_it's OK to be odd_).
-The meaning of these bits will be defined in the future.
+
+[BOLT #9](09-features.md) specifies lists of global and local features. Each feature is represented in `globalfeatures` or `localfeatures` by 2 bits: first one indicates is it feature mandatory (`0b1`) or not(`0b0`), and second one indicates is it feature supported or not. Odd features are optional, even features are compulsory (_it's OK to be odd_).
```
Roasbeef: > second one indicates _if the_ feature _is_ supported or not

Roasbeef: > first one indicates _if the_ feature _is_ mandatory

Roasbeef: Also what's "first" and "second" here? MSB, LSB?

Roasbeef:
Honestly, I don't think new description helps clarify the confusion around the "it's ok to be odd". I personally have trouble parsing if the "evenness" applies to the evaluated digit, or the _position_ of the bit in the dual-bit field. 

`lnd` implements feature parsing as follows:
   * Each pair of bits are compared (local vs remote):
      * If we don't know of a feature for this index, and they have the required bit set, then we're incompatible (disconnect). 
       * If we know of a feature for this index, and the bit is optional, then this feature isn't applied to the feature set of the connection. 
        * Otherwise, this feature is recognized, and applies to the current session. 
    * Each pair of bits are compared (remote vs local): 
        * If they have a feature index set to required, and we don't know of that bit, then we're incompatible (disconnect).
        * If they know of a feature and have it set to optional, and we don't know of this feature, then it isn't applied to the feature set of the connection.  
        * Otherwise, this feature is recognized, and applies to the current session. 

I propose we remove the wording related to ok-ness with even/odd (since there's been much confusion since it was originally proposed) and instead replace it with an explicit parsing methodology to figure out which features should be applied to a session, or if the feature sets are incompatible.

EmelyanenkoK:
Sorry for long delay.
Unfortunately, I still dont get it. Lets consider one feature:
If we receive init message in which feature _f_ has bits:
`0b11` second bit is odd so we can work with this node even if we do not understand _f_, counterparty supports this feature
`0b01` second bit is odd so we can work with this node even if we do not understand _f_, counterparty doesn't support this feature
`0b10` second bit is even so if we do not understand _f_ disconnect immidiately, counterparty supports this feature
`0b00` __???__ Explanation below has no sense: second bit is even so if we do not understand _f_ disconnect immidiately, counterparty doesnt support this feature.

I think that we should make table for all 4 possible bits with description.

01-messaging.md
EmelyanenkoK: I assume Most Significant Bit First rule here. I will update this place as soon as I understand it the rest.

AndrewSamokhvalov:
@EmelyanenkoK I think it might be like this:
`0b01` - required
`0b10` - optional
`0b11` - not determined
`0b00` - not determined

`active` - feature accepted by both sides and working.
`non-active` - feature not accepted, and is not working.
`disconnect` - sides unable to operate with each other properly.

local feature | remote feature | outcome
--|--|--
required  | required | active
required | optional | active
required | not determined | disconnect
optional| required | active
optional| optional | active
optional| not determined | non-active

AndrewSamokhvalov:
Meanwhile the result of decoding:
```
// TestFeaturesRemoteRequireError checks that we throw an error if remote peer
// has required feature which we don't support.
func TestFeaturesRemoteRequireError(t *testing.T) {
	const (
		first  = "first"
		second = "second"
	)

	localFeatures := NewFeatureVector([]Feature{
		{first, OptionalFlag},
	})

	remoteFeatures := NewFeatureVector([]Feature{
		{first, OptionalFlag},
		{second, RequiredFlag},
	})

	var b1 bytes.Buffer
	localFeatures.Encode(&b1)
	fmt.Println(b1.Bytes()) // [0 1 2]; 2 -> 0b|00|00|00|10


	var b2 bytes.Buffer
	remoteFeatures.Encode(&b2)
	fmt.Println(b2.Bytes()) // [0 1 6] 6 -> 0b|00|00|01|10

	if _, err := localFeatures.Compare(remoteFeatures); err == nil {
		t.Fatal("error wasn't received")
	}
}
```

Roasbeef:
Ahh, OK scratch my last comment, I see what you're getting at now @EmelyanenkoK. In the current draft we only specify what's optional or required. We _don't_ specify how a node signals if they have _implemented_ that feature or not. I think your latest break down is accurate. For the case of `0b00`, I think that's meant to be interpreted as padding and therefore should be ignored.

AndrewSamokhvalov:
It seems that misconception here lies in the fact that some of us think that _even/odd bits_ it is even numbers, thereby `0 - 0b00 (even)`, `1 - 0b01 (odd)`, `2 - 0b10 (event)`,  `3 - 0b10 (odd)`, `4 - 0b11 (even)`, and some of us think of `even/odd bits` as  the positions of bit inside the byte, thereby `0 - 0b00 - (unknown)`, `1 - 0b01 - (even)`, `0 - 0b10 - (odd)`, `0 - 0b11 - (unknown)`

EmelyanenkoK:
Sorry for mess with commits and new pull request. I'm not familiar with this stuff.

My private conversation with @AndrewSamokhvalov shows that we understand _it's ok to be odd_ very differently. I think that each two bits are treated separately as number (from 0 to 3) and oddness or evenness of this number determines optionality, while Andrew think that number of bit in sequence determines is it even or odd.
I updated my clarification according to discussion above, however current version (https://github.com/lightningnetwork/lightning-rfc/pull/156/commits/6655a6e55c3aca4f2f7e3a532e250f8a73cc89ce) contradicts with lnd code. @rustyrussell and @pm47, does it contradicts with your code too?

rustyrussell:
EmelyanenkoK <notifications@github.com> writes:
> Sorry for long delay.
> Unfortunately, I still dont get it. Lets consider one feature:
> If we receive init message in which feature _f_ has bits:
> `0b11` second bit is odd so we can work with this node even if we do not understand _f_, counterparty supports this feature

No, if you see an even bit you don't understand, you have to stop.  So
you have to terminate here.

> `0b01` second bit is odd so we can work with this node even if we do not understand _f_, counterparty doesn't support this feature

Yep.

> `0b10` second bit is even so if we do not understand _f_ disconnect immidiately, counterparty supports this feature

Yes.

> `0b00` __???__ Explanation below has no sense: second bit is even so if we do not understand _f_ disconnect immidiately, counterparty doesnt support this feature.

That's the normal case; neither bit is signalled.

So, A publishes the features it supports, B publishes the features it
supports.  It's up to A to decide if it doesn't like B's features, and
vice versa.

Now on to the more subtle points: features are assigned in pairs because
if we want to make a feature compulsory in (far!) future, we just use
its even bit instead of its odd bit.

Also, the SEMANTIC of feature bits is up to the feature itself: the
above rules only apply if you *don't* know what the feature is.  Perhaps
having both X bits set means something weird for a particular feature?
Perhaps X is optional and you don't have to do anything, perhaps X is
only activated if both sides offer it, etc.

Note that these semantics are not yet defined for the existing bits :(
So even and odd are (presumably) equivalent.

However, we could say for channels_public: Even means "you must offer to
announce the channel otherwise I won't let you open_channel", odd means
"I'll announce the channel iff you do".  Setting both is the same effect
as setting the even bit only.

initial_routing_sync could be the similar: if you can't send an initial
routing update to me, terminate.  Maybe if your node is not up-to-date
itself?  Seems a bit harsh though, so I'd say leave this alone.

Hope that clarifies?
Rusty.

EmelyanenkoK:
PR updated according to https://github.com/lightningnetwork/lightning-rfc/pull/156#issuecomment-301359666 .

Everything about odd/even semantic removed (as suggested https://github.com/lightningnetwork/lightning-rfc/pull/156#issuecomment-301208150) and substituted with plain table with value of bits and this value meaning .

It is also noted, that known features can define it's own meaning of bits.

`Feature bits semantic` column was added to table in BOLT-09.

@rustyrussell please review

pm47:
> Now on to the more subtle points: features are assigned in pairs because if we want to make a feature compulsory in (far!) future, we just use its even bit instead of its odd bit.

I had missed this subtelty, it all makes more sense with that in mind!

> Also, the SEMANTIC of feature bits is up to the feature itself: the above rules only apply if you *don't* know what the feature is.

This was another source of confusion for me.

Here is how it is currently implemented in eclair:
1) when we receive their init, if one even bit is unknown we close the connection
2) for `channels_public`, we will send sigs only if both have bit 0 set
3) for `initial_routing_sync` we are quite accommodating:
- if they ask for `initial_routing_sync` we comply
- if we ask for `initial_routing_sync` and they don't comply, we don't disconnect

AFAICT we have the same interpretation as @rusty above.

rustyrussell:
Thanks to everyone for this discussion, it's made me understand how much confusion I caused :(

I reworked the last commit to try to separate it more cleanly: the semantics of specific bits in BOLT9, and the treatment of unknown bits in BOLT1.  I was pretty happy with it, but we're likely to lose a feature bit due to #169.

The result is here: I rebased it, then added my proposed rewording on top, then finally merged #169 so you can see what it finally looks like:

https://github.com/rustyrussell/lightning-rfc/commits/pr-156-reword-rebase-add-pr-169

Issue #158, opened 2017-05-03
HMAC length in onion reply mismatch between spec and test vectors


cdecker:
I'm currently integrating the onion reply into our client and I stumbled across an inconsistency between the spec and the test vectors. The spec says we ought to use 32 byte untruncated HMACs in the reply, while the test vectors are using 20 byte truncated HMACs.

How do we resolve this inconsistency? Shall we use the truncated ones or the full HMACs? I'd prefer using the non-truncated ones since that matches all the other HMACs in the spec, and we removed the truncated ones everywhere else.

sstone:
Ah yes sorry I generated the test vectors with 20 bytes MACs and did not update them. I agree that it makes sense to use 32 bytes HMAC there and I've opened a pull request. to fix them.

cdecker:
Perfect, thanks @sstone, funny that we implemented it the same way :wink:

cdecker:
Seems that was rather easy to fix, @Roasbeef I'd also like some feedback from you guys on this so that all participating implementations have signed this off before merging ^^

Roasbeef:
+1 for using 32 bytes HMACs uniformly.

cdecker:
Perfect, the PR is already merged so closing this, thanks gentlemen :wink:

PR #159, opened 2017-05-03
BOLT4: fix onion reply test vectors (see #158)

sstone:


cdecker:
Thanks @sstone, just checked the test vectors and they match. :+1:

PR #163, opened 2017-05-09
Backticks underscores and typos

### LINKED ISSUE #141: Underscores, dashes and backticks.
We should make all fields consistent: I initially used - instead of _ because _ causes _emphasis_ in markdown.  But since we're supposed to surround all references with `` we should be OK.

Currently we have a nasty mix; I prefer _ because I can use those names exactly in the source code.

I've not made a PR, as I don't want to interfere with Real Work.

rustyrussell:
Converts all fields to use _ instead of -, and always surround by ``.  Also fixes up field names where they weren't correct (name had changed, or simply typos).

Closes #141


rustyrussell:
Tried with no `` around numbers, then rendering looked weird.  So reverted.

rustyrussell:
I checked the `./tools/extract-formats.py --message-types --message-fields *.md` output, and I'd actually missed the sha256-of-onion field (my replace regex didn't expect numbers in field names).  Fixed and rebased.

Issue #120, opened 2017-03-01
Forget channels on disconnect if we haven't sent/received funding_signed


rustyrussell:
(As discussed in PR #119 )

This nicely reduces the requirements if nodes vanish before getting far into channel setup.  The only issue is that if one side has sent funding_signed and the other hasn't received it: in this case the re-transmission of funding_signed will cause an error.

rustyrussell:
OK @pm47 I think we can treat the opening dialog as an atomic block:
1.  Fundee remembers nothing UNTIL they send funding_signed.  Then they assume it was received.
2.  Funder remembers nothing UNTIL they receive funding_signed (they they broadcast funding tx).

If the funding_signed is sent but not received, then funder will reconnect and create a new channel; fundee will eventually time out the old one anyway.

This makes the implementation much simpler, AFAICT.  @pm47 ?

pm47:
I think it works. It better does, because this is how we have implemented it in eclair 😜.

From a "security-of-funds" perspective, what matters most is that the funder remembers the channel (whatever that means) **before** it publishes the funding tx. This should be emphasized I think.

As an aside: in eclair we are using bitcoin core both as a gateway to watch the blockchain, and as a wallet. But if we ask bitcoind to directly provide inputs to the funding tx, the resulting tx may (and will as we discovered on testnet) be malleated, which will break the channel setup. After trying several workarounds, we ended up using an intermediate P2WPKH address (which corresponding key we import in bitcoind). Once this intermediate tx has one confirmation, we make our funding tx spend it. It fixes the malleability issue, and allow the funder to not have to remember anything if the fundee disconnects itself before the funding tx is published (because bitcoind sees the intermediate tx as a pay-to-itself). But there are downsides: it makes the channel creation process longer and costlier (x2 fees), and peers must stay connected for ~10 minutes before the funding tx is published. As a consequence, we often find ourselves forgetting channels under the conditions described in this ticket.

Cheers,

Pierre

rustyrussell:
This was actually closed by #174.

matsjj:
I was wondering, in

https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#requirements-14

it now says

A non-funding node:

 - upon disconnection:
    - if it has sent the funding_signed message:
        - MUST remember the channel for reconnection.

Doesn't this allow another peer to DoS, creating thousands of channels, but never actually broadcasting the funding transaction?

[...]

PR #164, opened 2017-05-11
BOLT 2: detail retransmission, simplify opening retransmission.

rustyrussell:
1. Point out explcitly when to broadcast the funding tx
   (Suggested by pm47).
2. Explictly add timeouts for not receiving funding_locked.
3. Allow repeated update_fail/fulfill for retransmissions.
4. We forget channels, not nodes (can have more than 1 per node).
5. Require forgetting before `funding_signed` so opening is atomic (apparently
   Eclair already does this).
6. Note that we should retransmit any error if we get a reconnect, and move
   closing_signed retransmission into the same paragraph.
7. Explicitly note that updates which were followed by a commitment_signed
   MUST be a superset of what was sent before.
8. Rationale section to describe the reasons for the retransmission rules.

Closes: #120

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -282,13 +282,12 @@ per-commitment point to be used for the following commitment
 transaction, derived as specified in
 [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
 
-A funder SHOULD perform unilateral close if it does not receive
-`funding_locked` after a reasonable timeout.
- 
 A non-funding node SHOULD forget the channel if it does not see the
-funding transaction after a reasonable timeout, and MAY forget the
-channel or perform unilateral close if it does not receive
-`funding_locked` after a reasonable timeout.
+funding transaction after a reasonable timeout.
+
+From the point of waiting for `funding_locked` onwards, a node MAY
+fail the channel if it does not receive a required response from the
+other node after a reasonable timeout.
 
 #### Rationale
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -948,11 +948,9 @@ lists the acknowledgment conditions for each message:
 * `revoke_and_ack`: acknowledged by `commitment_signed` or `closing_signed`
 * `shutdown`: acknowledged by `closing_signed`.
 
-Before retransmitting `commitment_signed`, the node MUST retransmit
-all `update_` messages (the other node will have forgotten
-them, as required above), with the exception that `update_fee` messages
-MAY be retransmitted with different values.  The node MAY
-transmit additional updates before `commitment_signed`.
+On reconnection, the node MUST retransmit all `update_` messages, and
+MAY transmit additional updates before `commitment_signed`, with the
+exception that `update_fee` MAY be retransmitted with a different value.
 
 A node MUST not assume that
 previously-transmitted messages were lost: in particular, if it has
@@ -974,21 +972,19 @@ this case, the funder will forget the channel and presumably open
 a new one on reconnect; the other node will eventually forget the
 original channel due to never receiving `funding_locked`.
 
-For normal operation and during shutdown, the updates are similarly
-forgotten, though in this case the loss of `revoke_and_ack` can cause
+For normal operation and during shutdown, the loss of `revoke_and_ack` can cause
 an entire retransmission of the previous updates and commitment:
 this is OK, as the recipient will simply ignore duplicates (as
 documented in the requirements for [`update_add_htlc`](#adding-an-htlc-update_add_htlc) and [Removing an HTLC](#removing-an-htlc-update_fulfill_htlc-update_fail_htlc-and-update_fail_malformed_htlc).
 
-This retransmission is required to generally be a super-set of the
-original, to give the same result whether the first transmission was
-committed or not.  The exception for `update_fee` applies because
-`update_fee` simply overrides previous `update_fee` messages, though a
-retransmission has to send at least one `update_fee` message if one was
-sent originally.
+Retransmission is required to be a super-set of the original, to give
+the same result whether the first transmission was committed or
+not. But note that technically `update_fee` is always a super-set of
+any previous `update_fee` because it simply replaces it.
 
-Note that if no `commitment_signed` was sent in the first place,
-there's no requirement to retransmit those (uncommitted) updates.
+This means that the receiving node may or may not forget `update_`
+until it sends out `revoke_and_ack`; it does not have to commit them
+to permanent storage.
 
 Once an acceptable `closing_signed` has been received, the protocol
 guarantees that it will be better to use than any previous
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -261,7 +261,7 @@ The sender MUST set `channel_id` by exclusive-OR of the `funding_txid` and the `
 
 The recipient MUST fail the channel if `signature` is incorrect.
 
-The recipient it SHOULD broadcast the funding transaction on receipt of a valid `funding_signed` and MUST NOT broadcast the funding transaction earlier.
+The recipient SHOULD broadcast the funding transaction on receipt of a valid `funding_signed` and MUST NOT broadcast the funding transaction earlier.
 
 ### The `funding_locked` message
 
@@ -284,15 +284,20 @@ transaction, derived as specified in
 
 A funder SHOULD perform unilateral close if it does not receive
 `funding_locked` after a reasonable timeout.
-
-A non-funding node MAY forget the channel if it does not receive
-`funding_locked` after a reasonable timeout.
+ 
+A non-funding node SHOULD forget the channel if it does not see the
+funding transaction after a reasonable timeout.
 
 #### Rationale
 
-The funder needs to unilateral close to get its funds back if the
-other node doesn't respond; the non-funder can simply forget the
-channel ever existed, since no funds are at risk.
+The funder normally needs to unilateral close to get its funds back if
+the other node doesn't respond, though a large `push_msat` may make
+this uneconomical.
+
+The non-funder can simply forget the channel ever existed, since no
+funds are at risk; even if `push_msat` is significant, if it remembers
+the channel forever on the promise of the funding transaction finally
+appearing, there is a denial of service risk.
 
 #### Future
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -261,6 +261,8 @@ The sender MUST set `channel_id` by exclusive-OR of the `funding_txid` and the `
 
 The recipient MUST fail the channel if `signature` is incorrect.
 
+The recipient it SHOULD broadcast the funding transaction on receipt of a valid `funding_signed` and MUST NOT broadcast the funding transaction earlier.
+
 ### The `funding_locked` message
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -280,6 +282,18 @@ per-commitment point to be used for the following commitment
...
+A funder SHOULD perform unilateral close if it does not receive
+`funding_locked` after a reasonable timeout.
+
+A non-funding node MAY forget the channel if it does not receive
+`funding_locked` after a reasonable timeout.
+
+#### Rationale
+
+The funder needs to unilateral close to get its funds back if the
+other node doesn't respond; the non-funder can simply forget the
```
pm47: If `pushMsat`>0 the non-funder does have funds at stake though.

02-peer-protocol.md
```diff
@@ -894,40 +911,45 @@ any message), they are independent of requirements here.
...
 
-The last `closing_signed` (if any) must always be retransmitted, as there
-is no explicit acknowledgment.
-
-Before retransmitting `commitment_signed`, the node MUST send
-appropriate `update_` messages (the other node will have forgotten
-them, as required above).
+Before retransmitting `commitment_signed`, the node MUST retransmit
+all `update_` messages (the other node will have forgotten
+them, as required above), with the exception that `update_fee` messages
```
pm47: I am curious why you allow a special case for `update_fee`? Wouldn't that be simpler to just send messages as there were and send an additional `update_fee` if necessary?

02-peer-protocol.md
```diff
@@ -938,6 +960,38 @@ previously sent.
...
 fail the channel if they occur.
 
+### Rationale
+
+The effect of requirements above are that the opening phase is almost
+atomic: if it doesn't complete, it starts again.  The only exception
+is where the `funding_signed` message is sent and not received: in
+this case, the funder will forget the channel and presumably open
+a new one on reconnect; the other node will eventually forget the
+original channel due to never receiving `funding_locked`.
```
pm47: "due to never receiving `funding_locked`" -> "due to never detecting the funding tx"?

The fundee doesn't really care about receiving a funding_locked, if it sees the funding tx on the blockchain (which maybe pays him `pushMsat`).

02-peer-protocol.md
```diff
@@ -938,6 +960,38 @@ previously sent.
...
+For normal operation and during shutdown, the updates are similarly
+forgotten, though in this case the loss of `revoke_and_ack` can cause
+an entire retransmission of the previous updates and commitment:
+this is OK, as the recipient will simply ignore duplicates (as
+documented in the requirements for [`update_add_htlc`](#adding-an-htlc-update_add_htlc) and [Removing an HTLC](#removing-an-htlc-update_fulfill_htlc-update_fail_htlc-and-update_fail_malformed_htlc).
+
+This retransmission is required to generally be a super-set of the
+original, to give the same result whether the first transmission was
+committed or not.  The exception for `update_fee` applies because
+`update_fee` simply overrides previous `update_fee` messages, though a
```
pm47: This answers my comment above, but I still find that overly complicated.

02-peer-protocol.md
```diff
@@ -938,6 +960,38 @@ previously sent.
...
+retransmission has to send at least one `update_fee` message if one was
+sent originally.
+
+Note that if no `commitment_signed` was sent in the first place,
+there's no requirement to retransmit those (uncommitted) updates.
```
pm47: Should we really give this flexibility? Enforcing a single chain of updates simplifies disconnection handling on the receiver's side.

02-peer-protocol.md
rustyrussell: Good point!  Will reword.

02-peer-protocol.md
rustyrussell: It would be simpler, to specify, but an implementation *may* simply choose to always send an `update_fee` on reconnect in this case, and not record them?

02-peer-protocol.md
rustyrussell: Doesn't make a difference, since the receiver discards them anyway...

Once you've send commitment_signed, *then* you need to start a timer for HTLCs you may be committed do. .  If you don't, you don't need to.

02-peer-protocol.md
rustyrussell: If you never receive funding_locked, you can't use the channel, and you'll presumably eventually time out.  You're right though, I keep forgetting about the pushMsat case; you do need to care here.

I'm starting to hate pushMsat :(  It means I can pushMsat to you, get the funding tx then abort, and you'll have to remember it just in case I decide to pay you...

02-peer-protocol.md
```diff
@@ -261,6 +261,8 @@ The sender MUST set `channel_id` by exclusive-OR of the `funding_txid` and the `
 
 The recipient MUST fail the channel if `signature` is incorrect.
 
+The recipient it SHOULD broadcast the funding transaction on receipt of a valid `funding_signed` and MUST NOT broadcast the funding transaction earlier.
```
rustyrussell: Thanks, fixed.

02-peer-protocol.md
rustyrussell: (And the receiving side doesn't care, of course)

02-peer-protocol.md
pm47: Ah! I am not a fan either, I've always seen it as a _trusted_ workaround to setup balanced channels.

02-peer-protocol.md
pm47: My point was that it would indeed allow the receiver to not discard changes, and just make sure that whatever it receives afterwards is either new or already known. It is more error prone than rollbacking IMO, and conceptually I find it easier to see a reconnection as a simple replay.

AndrewSamokhvalov: I shall agree with @pm47 on this topic, because it seems that by applying the rule above we would deviate from the simple "acknowledged by" approach.

rustyrussell: I grabbed another coffee to mull over this idea; it has appeal, though it seems weird to be committed before sending "commit".

My only concern is that this increases the window for HTLC-in-flight failures in practice.

 eg. An HTLC comes in to A, it forwards it to B (who has vanished/failed).  In my implementation, there's a 10ms timer before sending commitment_signed: during that time it may get a TCP reset (or other error) from B.  As written now, A can simply fail the HTLC if it can't reconnect in reasonable time.

If B doesn't (always) forget, A can't fail the HTLC until it gives up on the channel with B entirely, which presumably will take longer.

02-peer-protocol.md
rustyrussell: Perhaps we should say "Retransmission is required to be a super-set of the original, to give the same result whether the first transmission was committed or not.  But note that technically update_fee is always a super-set of any previous update_fee because it simply replaces it"?

02-peer-protocol.md
rustyrussell: Actually, after more contemplation, I agree with you.  Simplicity wins.  I will push an update.

We can restore the fast-fail behavior in 1.1 with update_unadd_htlc if we want, which has other benefits.

pm47: If you agree on that, I don't think we can allow the mutation of previously sent `update_fee`, because a fee that the funder cannot afford may invalidate subsequent `update_add_htlc`, right? Even if the sender can afford a new higher fee, the fact that we break the "simple replay" scenario kind of defeats the purpose in my opinion.

To be clear, I see two options, either:
(1) we only allow a single flow of updates (add, fail, fulfill, fee), and reconnection is a simple replay: simpler and maybe best for now?
or 
(2) we allow forking the update flow in case of a disconnection, and reconnection is a rollback: slightly more complex, but maybe more optimal?

I think all updates should be treated the same way, and if we already know we will introduce an `update_unadd_htlc`, then I think (1) might be the way to go. I may be influenced by having worked on database wal!

02-peer-protocol.md
```diff
@@ -894,40 +916,43 @@ any message), they are independent of requirements here.
...
 
 A node MUST handle continuing a previous channel on a new encrypted
-transport.  On disconnection, a node MAY forget nodes which have not
-sent or received an `accept_channel` message, and MAY forget nodes
-which have not sent `funding_locked` after a reasonable timeout.
+transport.
+
+On disconnection, the funder MUST remember the channel for
+reconnection if it has broadcast the funding transaction, otherwise it
+MUST NOT.
```
cdecker: This should probably allow for unilateral closes after a timeout as well, i.e., the situation where funder broadcasts, disconnect, timeout -> unilateral close, reconnect. In that case the reconnect must be aborted.

02-peer-protocol.md
```diff
@@ -894,40 +916,43 @@ any message), they are independent of requirements here.
...
 * `revoke_and_ack`: acknowledged by `commitment_signed` or `closing_signed`
 * `shutdown`: acknowledged by `closing_signed`.
 
-The last `closing_signed` (if any) must always be retransmitted, as there
-is no explicit acknowledgment.
-
-Before retransmitting `commitment_signed`, the node MUST send
-appropriate `update_` messages (the other node will have forgotten
-them, as required above).
+On reconnection, the node MUST retransmit all `update_` messages, and
```
cdecker: Shall we say that on reconnect both parties have to roll back any changes since the last committed state, and then replay them? Instead of allowing states to diverge and then trying to reconcile it by discarding duplicates. Would allow for a brief inconsistency if nodes replay different updates, but they may only act on them once they were committed to anyway. May be a bit simpler to grok, since it aligns with the transactional model.

Roasbeef: I agree with cdecker here. If specification complaint nodes simply replay the last commitment exchange leading up to the current state, then it seems that they only need to remember the delta between the two states (the `update_*` messages). This seems much simpler to implement correctly compared to allow desynchronization, and attempting to selectively reconcile the divergence. 

With this in mind, from my PoV, the storage requirements are:
   * A delta package of state `#N-1` where the current state number if `N` for the _remote_ peer
   * A delta package is created _locally_ before sending the `commit_sig` message to extend the remote peer's commitment chain. 
 
Upon restart, the local peer retransmits the following:

  * The delta package for state `N-1` for the commitment transaction of the _remote_ peer

I think it's as simple as that? I think specifying retransmission for the state machine distinct from this "acknowledge by" table will make it easer to new comers to quickly digest the details of how retransmission is handled. If the above works, then it seems that the counters themselves actually don't need to be persistent? As we don't care about filtering out duplicates. 

When processing a delta package, you roll back your state to the prior state, apply the package, then if the resulting state matches, send the _prior_ revoke message, and you're done. Otherwise, you missed some messages and should accept this new state.

02-peer-protocol.md
Roasbeef: `push_msat` was never intended to be a trusted workaround to set up balanced channels. It simply attempts to address the case wherein the channel initiator wishes to _immediately_ pay the responder.

rustyrussell:
Note that this has opened a bigger can of worms, so I've set up discussion on #172 

I will close this and resubmit just the minor fixes and change to the opening retransmission rules which seem non-controversial.

PR #165, opened 2017-05-11
Even more underscores

rustyrussell:
More trivial things I missed with my simple regex search, plus a name consistency update.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -110,7 +110,7 @@ This output sends funds to the other peer, thus is a simple P2WPKH to `remotekey
 This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or to the remote peer using the payment preimage or the revocation key.  The output is a P2WSH, with a witness script:
 
     # To you with revocation key
-    OP_DUP OP_HASH160 <revocationkey-hash> OP_EQUAL
+    OP_DUP OP_HASH160 <RIPEMD160(revocationkey)> OP_EQUAL
     OP_IF
         OP_CHECKSIG
     OP_ELSE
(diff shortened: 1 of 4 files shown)


03-transactions.md
```diff
@@ -86,14 +86,14 @@ This output sends funds back to the owner of this commitment transaction, thus m
         # Penalty transaction
         <revocation-pubkey>
```
pm47: Just to be sure, are you keeping dashes in `<revocation-pubkey>`, `<local-delayedkey>`, `<ripemd-of-payment_hash>` on purpose? 

Should we also fix inconsistencies between pubkey and key? We have both `<revocation-key>` and `<revocation-pubkey>`.

Ther are also inconsistencies between `<revocation-sig>` and `<revocationkey-hash>`. If that's outside the scope of this PR I'll submit another one.

03-transactions.md
```diff
@@ -120,14 +120,14 @@ This output sends funds to a HTLC-timeout transaction after the HTLC timeout, or
             OP_DROP 2 OP_SWAP <localkey> 2 OP_CHECKMULTISIG
         OP_ELSE
             # To you with preimage.
-            OP_HASH160 <ripemd-of-payment-hash> OP_EQUALVERIFY
+            OP_HASH160 <ripemd-of-payment_hash> OP_EQUALVERIFY
```
Roasbeef: Dashes here should actually be underscores?

03-transactions.md
```diff
@@ -148,11 +148,11 @@ This output sends funds to the remote peer after the HTLC timeout or using the r
             OP_SIZE 32 OP_EQUAL
         OP_IF
             # To me via HTLC-success transaction.
-            OP_HASH160 <ripemd-of-payment-hash> OP_EQUALVERIFY
+            OP_HASH160 <ripemd-of-payment_hash> OP_EQUALVERIFY
```
Roasbeef: Missed an `underscore -> dashes` substitution here.

03-transactions.md
rustyrussell: Hmm, I'll fix those too.  Was just aiming for actual defined field names, but that was kind of silly.

03-transactions.md
rustyrussell: Technically, I was trying to replace all field names (payment_hash is a field of update_add_htlc).  But I think we should actually put <RIPEMD160(payment_hash)> here.  Will add a new commit...

rustyrussell:
OK, feedback incorporated, more conversions.  I'm tempted to leave any remaining fixups for a separate typo PR later :)

rustyrussell:
Assigned to @Roasbeef: if you're happy, please apply...

PR #167, opened 2017-05-15
BOLT 4: Fix inconsistencies in error messages

pm47:
This follows a discussion started in #149. I am only proposing minimal changes, and individual commits are independant:
- https://github.com/lightningnetwork/lightning-rfc/commit/c07b276f7d0d87fc1efbd8f7ea3640a80793b7c1: since #149 added a mandatory `channel_update` message to `temporary_channel_failure`, then this error must have  the `UPDATE` flag
- https://github.com/lightningnetwork/lightning-rfc/commit/142e20769c828e93e7f540c61515cfb218216e91: since `channel_update` is fixed-size, we don't really need to provide its length (credits @cdecker)
- https://github.com/lightningnetwork/lightning-rfc/pull/167/commits/1736ae54960917890e33174838031fd87c79a335: since we already have `amount_below_minimum`, `fee_insufficient`, `incorrect_cltv_expiry` and `expiry_too_soon` errors when incoming htlcs do not comply with a specific field in `temporary_channel_failure`, we should also have a `channel_disabled` error for consistency reasons.

NB:
- in order to reduce the number of error messages, @cdecker also proposed to use a single type for `channel_update`-related errors, containing a bitfield to provide hints for the reason. It would also allow composing of errors. But unfortunately this wouldn't be compatible with providing fine-grain details, like what we do with `fee_insufficient`.`htlc_msat` or `incorrect_cltv_expiry`.`cltv_expiry`
- for the record I would have preferred only providing a `channel_update` when it actually explains the error (with specific error messages like we have currently), but @rustyrussell prefers always attaching one to `temporary_channel_failure` to [make failures provable](https://github.com/lightningnetwork/lightning-rfc/pull/149#issuecomment-298283452). With that in mind, I think a single `UPDATE` error message with a bitfield (that may be all zeroes) would make more sense, but I left that outside of this PR.

(I also tried [putting error messages in a table](https://github.com/pm47/lightning-rfc/commit/e6f61140b90c43f1bf34b5610c1925640979fb90), WDYT?)

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -489,7 +489,7 @@ If the ephemeral key in the onion is unparsable:
 If an otherwise unspecified transient error occurs for the outgoing
 channel (eg. peer unresponsive, channel capacity reached):
 
-1. type: 7 (`temporary_channel_failure`)
+1. type: UPDATE|7 (`temporary_channel_failure`)
 2. data:
    * [`2`:`len`]
    * [`len`:`channel_update`]
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -487,7 +487,7 @@ If the ephemeral key in the onion is unparsable:
    * [`32`:`sha256_of_onion`]
 
 If an otherwise unspecified transient error occurs for the outgoing
-channel (eg. peer unresponsive, channel capacity reached):
+channel (eg. channel capacity reached, too many in-flight htlc):
 
 1. type: UPDATE|7 (`temporary_channel_failure`)
 2. data:
@@ -544,6 +544,15 @@ setting for the outgoing channel:
 
 1. type: UPDATE|14 (`expiry_too_soon`)
 2. data:
+   * [`2`:`len`]
+   * [`len`:`channel_update`]
+   
+If the channel is disabled, we tell them the the current channel
+setting for the outgoing channel:
+   
+1. type: UPDATE|20 (`channel_disabled`)
+2. data:
+   * [`2`: `flags`]
    * [`2`:`len`]
    * [`len`:`channel_update`]
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -166,6 +166,19 @@ Nodes forwarding HTLCs MUST construct the outgoing HTLC as specified within
 `per_hop`.  Otherwise, deviation from the specified HTLC parameters
 may lead to extraneous routing failure.
 
+### Payload for the last node
+
+The last node in the route could just discard its payload since it will not forward payments. However, when building the route, the original
+sender must use a payload for the last node with the following values:
+* `outgoing_cltv_value` is set to the final expiry specified by the recipient
+* `amt_to_forward` is set to the final amount specified by the recipient
+
+This way, the final node can check these values and return errors if needed, which will defeat probing attacks by the next to last node which could
+try to find out if the next node is the last one (by re-sending HTLCs with different amounts/expiries):
+
+The last node will extract its onion payload from the HTLC it has received and compare its values to the HTLC values.
+See the [Returning Errors](#returning-errors) section below for more details.
+
 ## Packet Construction
 
 Assuming a _sender node_ `n_0` wants to route a packet to a _final recipient_ `n_r`.
@@ -584,7 +597,7 @@ HTLC at the final hop:
 2. data:
    * [`4`:`cltv_expiry`]
 
-If the `amt_to_forward` does not match the `incoming_htlc_amt` of
+If the `amt_to_forward` is higher than `incoming_htlc_amt` of
 the HTLC at the final hop:
 
 1. type: 19 (`final_incorrect_htlc_amount`)
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -275,3 +275,4 @@ metadata
 Bitcoin's
 Versioning
 checksum
+expiries


04-onion-routing.md
```diff
@@ -536,16 +533,22 @@ outgoing channel:
 1. type: UPDATE|13 (`incorrect_cltv_expiry`)
 2. data:
    * [`4`:`cltv_expiry`]
-   * [`2`:`len`]
-   * [`len`:`channel_update`]
+   * [`92`:`channel_update`]
```
Roasbeef: Hmm, by moving to a fixed size message here, aren't we locking ourselves in a bit by assuming the `channel_update` message will _always_ be 92 bytes?

Roasbeef: Also the fact that message can have additional data appended to them.

rustyrussell: Yep, this part is bogus.

rustyrussell:
Oh, you previously queried final_incorrect_cltv_expiry and final_incorrect_htlc_amount: they exist because the other versions require an update message, and obviously the final hop doesn't have one.

This should be noted in the document :)

04-onion-routing.md
pm47: Fair enough!

04-onion-routing.md
```diff
@@ -541,6 +541,14 @@ setting for the outgoing channel:
...
 2. data:
    * [`92`:`channel_update`]
+   
+If the channel is disabled, we tell them the the current channel
+setting for the outgoing channel:
+   
+1. type: UPDATE|20 (`channel_disabled`)
+2. data:
+   * [`2`: `flags`]
+   * [`92`:`channel_update`]
```
pm47: I don't see it as a temporary failure: disabling a channel (e.g. when the peer is disconnected) is similar to updating the fee: we publish a `channel_update`, and from now on we will send `channel_disabled` in response to htlcs attempting to use the channel.

pm47:
> Oh, you previously queried final_incorrect_cltv_expiry and final_incorrect_htlc_amount: they exist because the other versions require an update message, and obviously the final hop doesn't have one.

We currently have ` 	incorrect_payment_amount` and `final_incorrect_htlc_amount` which are both final-node errors and seem to overlap.

I don't get the `final_incorrect_cltv_expiry`, since there is no `outgoing_cltv_value` at the final node (per-hop payload is all zeroes)?

cdecker:
Sorry @pm47 for the length confusion, I had forgotten about the possibility to add new fields to the end. I like the `channel_disabled` error, it does not have the connotation that the channel will be right back that `temporary_channel_failure` has :+1:

rustyrussell:
>> Oh, you previously queried final_incorrect_cltv_expiry and final_incorrect_htlc_amount: they exist because the other versions require an update message, and obviously the final hop doesn't have one.
>
> We currently have ` 	incorrect_payment_amount` and `final_incorrect_htlc_amount` which are both final-node errors and seem to overlap.

incorrect_payment_amount is "I asked for $2, you sent me $1".

> I don't get the `final_incorrect_cltv_expiry`, since there is no `outgoing_cltv_value` at the final node (per-hop payload is all zeroes)?

No, the HMAC is zero, the other field are still there.

We had to fix this because otherwise it provides a method for the prior
node to find out whether this node is the final hop, either by reducing
the cltv_expiry by one, or reducing amount_msat by one (and hoping this
is an unspecified amount payment or they have fuzzed the amount a
little).

Quoting BOLT 4:

   * `amt_to_forward` - The amount in milli-satoshi to forward to the next
     (outgoing) hop specified within the routing information. 
...
        incoming_htlc_amt - fee >= amt_to_forward

     Where `fee` is calculated according to the receiving node's advertised fee
     schema as described in [BOLT 7](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#htlc-fees), or 0 if this node is the
	 final hop.

   * `outgoing_cltv_value` - The CLTV value that the _outgoing_ HTLC carrying
     the packet should have. 

        cltv_expiry - cltv_expiry_delta = outgoing_cltv_value

...
        If there is no next hop, `cltv_expiry_delta` is zero.

Cheers,
Rusty.

sstone:
@rustyrussell we've added a section that explains how to build the payload for the last node (we had missed this part)

04-onion-routing.md
```diff
@@ -575,7 +597,7 @@ HTLC at the final hop:
 2. data:
    * [`4`:`cltv_expiry`]
 
-If the `amt_to_forward` does not match the `incoming_htlc_amt` of
+If the `amt_to_forward` is higher than `incoming_htlc_amt` of
```
rustyrussell: No, this just enables probing the other way, where the second-to-final node takes one millisatoshi *less* than the fee it's told to.

cdecker: But we're telling the next hop how much to forward so he'll just swallow that extra satoshi, mimicking the behavior of the last hop. The intermediate nodes will not fail with fee-too-high (otherwise the whole thing is way too brittle) so why should the last node?

sstone: Ok, I've restored the original description

sstone: @rustyrussell Actually now I think it is rather the opposite:the specs above say that a HTLC should be rejected if amt_to_forward > incoming_htlc_amt - fee. For the last hop, the specs also says that the fee is set to 0 so the rejection criteria becomes incoming_htlc_amt > amt_to_forward which is exactly what I wrote. If the last node tests for strict equality then it behaves differently from the other nodes

rustyrussell: Yes, you're right, my mistake!  To avoid further delay, I'll force pushed back the version prior to your restoration, then merge right now.

rustyrussell:
OK, I had to fix the spellcheck then rebase on master anyway (spellcheck words caused trivial merge conflict), so I folded in the update-len-removal/reversion patches to make them disappear.

PR #169, opened 2017-05-15
Allow a per-channel rather than per-session policy for announcing opened channels 

Roasbeef:
This commit gives peers the ability to signal their intent to make a
channel private in the `open_channel` message. This differs from the
current method as now peers are able to create multiple channels with
heterogeneous announcement policies _without_ disconnecting and
re-connecting in-between each channel funding. The prior requirement
for the nodes to re-connect (in order to create a new channel with 
opposite policy of the active) was burdensome and unnecessary.

Additionally, it may not always be the case that upon _connection_ (but 
not yet funding workflow initiation), that a peer knows whether or not it 
wishes to advertise a connected channels. A peer may have originally 
connected just to receive gossip data, but then wishes to establish a 
channel. In this case, the node may need to _reconnect_ to apply their 
desired channel visibility to the to-be-opened channel. This PR 
eliminates this case, as the node is able to decide _when_ it starts 
a funding workflow, and isn't forced to decide before hand.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -99,6 +99,7 @@ desire to set up a new channel.
    * [`33`:`payment_basepoint`]
    * [`33`:`delayed_payment_basepoint`]
    * [`33`:`first_per_commitment_point`]
+   * [`1`:`channel_flags`]
 
 
 The `chain_hash` value denotes the exact blockchain the opened channel will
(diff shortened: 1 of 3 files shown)


02-peer-protocol.md
```diff
@@ -94,6 +94,7 @@ desire to set up a new channel.
    * [`4`:`feerate_per_kw`]
    * [`2`:`to_self_delay`]
    * [`2`:`max_accepted_htlcs`]
+   * [`1`:`announce_channel`]
```
rustyrussell: Let's put it at the end to preserve alignment.

02-peer-protocol.md
```diff
@@ -113,13 +114,22 @@ The `temporary_channel_id` is used to identify this channel until the funding tr
...
 
 FIXME: Describe Dangerous feature bit for larger channel amounts.
 
 
 #### Requirements
 
+If `announce_channel` is set to `false` (`0x00`), then both sides MUST
+NOT send `announcement_signatures` messages after the `funding_locked`
+message is sent.
+
```
rustyrussell: This is specified already in BOLT 7.  Plus, it's not in terms of sender or receiver :)

02-peer-protocol.md
```diff
@@ -142,6 +152,9 @@ allow commitment transactions to propagate through the Bitcoin
 network.  It SHOULD set `htlc_minimum_msat` to the minimum
 amount HTLC it is willing to accept from this peer.
 
+The receiving not MUST failed the channel if `announce_channel` is set
+to `false` (`0x00`), yet they wish to publicly announce the channel.
+
```
rustyrussell: s/not//node/.  It's a bit vague, though.  I mean, they might wish to announce it, but it might not be a dealbreaker?

How about:

    The receiving node MAY fail the channel if `announce_channel` is `0x00` and it wants to publicly announce the channel.

02-peer-protocol.md
pm47: Could yet another features bitfield be more future-proof? I am fine with a single byte.

rustyrussell: Yes, lets switch:

* announce_channel -> flags
* "`true` (`0x01`)" to "least significant bit is set"
* "`false` (`0x00`)" to "least significant bit is clear".

rustyrussell:
Was rebasing this and applying my suggested changes, and noticed the semantic change: now the funder decides unilaterally whether the channel is public, and the fundee accepts that or fails the channel.  I think that's OK, and simpler, but note that it *is* a change, and requires changing the announcement_signatures sentence too.

Roasbeef:
@rustyrussell your fixup commit looks good to me! I dig more conservatively defining a `channel_flags` byte, and using a single bit rather than allocating the entire byte to this specific purpose.

rustyrussell:
Folded in fixup commit, and will rebase now.

PR #175, opened 2017-05-16
BOLT 2,4,7: use 8 bytes for amounts, restrict add_htlc for bitcoin only.

rustyrussell:
We had 4 byte fields for amounts because people have no ability to assess
risk, and this limited the damage to $70 at a time.

But then that means $1 maximum HTLCs on Litecoin, which isn't enough
for a cup of (decent) coffee.

Rather than have boutique hacks for Litecoin we enlarge the fields now,
and simply have a bitcoin-specific restriction that the upper 4 bytes be 0.

Suggested-by: Olaoluwa Osuntokun <laolu32@gmail.com>

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -90,7 +90,7 @@ desire to set up a new channel.
    * [`8`:`dust_limit_satoshis`]
    * [`8`:`max_htlc_value_in_flight_msat`]
    * [`8`:`channel_reserve_satoshis`]
-   * [`4`:`htlc_minimum_msat`]
+   * [`8`:`htlc_minimum_msat`]
    * [`4`:`feerate_per_kw`]
    * [`2`:`to_self_delay`]
    * [`2`:`max_accepted_htlcs`]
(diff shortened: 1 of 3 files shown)


EmelyanenkoK:
As I understand in the future LN will support cross-chain swap. Thus we should be aware of ethereum-based chains. In ethereum indivisible unit is 1e-18 of one eth.
Thus I propose to extend `funding_satoshis` field too, from 8bytes to 16. It allows us to lock up sufficient amount of ether in the channel (18.5 eth for now)

rustyrussell:
@EmelyanenkoK I understand what you're saying, but we're going to need more changes for ethereum, and as no hardware supports 128-bit numbers natively, I'm reluctant to change this.  If people want to send more than 18.5 ethers(?) which seems like $1500 right now, we can add a field for them.

cdecker:
That branch name... LGTM

Roasbeef:
@EmelyanenkoK well as is written now ETH, can't simply copy-pasta the spec into the domain of their blockchain. In fact, I'm not aware of any specification efforts on their part (nor of the design of their LN emulation), so for now I think it make sense to just focus on BTC (which allows BTC derived coins like LTC adopt the spec verbatim).

PR #181, opened 2017-05-24
Retransmission with state counters

rustyrussell:
OK, so if we keep rollback and add per-channel reconnect counters, this is what it looks like.   I know @adiabat wanted simple retransmit, but I think this is actually simpler.

Caveat: I haven't actually *implemented* this yet, but I'm about to try to do so...


02-peer-protocol.md
```diff
@@ -919,18 +919,24 @@ independent of particular channels; their transmission requirements
 are covered there, and other than being transmitted after `init` (like
 any message), they are independent of requirements here.
 
+1. type: 136 (`channel_reestablish`)
```
Roasbeef: Still reading this diff, but so far: I dig this! Such an explicit message helps to sidestep the "The Uncertain Signature Problem" all together.

02-peer-protocol.md
```diff
@@ -940,35 +946,38 @@ so the effects of `update_fulfill_htlc` is not completely reversed.
...
+that channel.  The sending node MUST set `commitments_received` to the
+number of `commitment_signed` messages received, and MUST set
+`revocations_received` to the number of `revoke_and_ack` messages
+received.
+
+If `commitments_received` is one less than the number of
+`commitment_signed` messages the receiving node has sent, it MUST send
+a set of `update_` messages against the prior commitment transaction
+followed by `commitment_signed`, otherwise if `commitments_received`
+is not equal to the number of `commitment_signed` messages the
```
Roasbeef: As this indicates that the channel states have desynchronized, or the transmitting party had persistence corruption?

02-peer-protocol.md
```diff
@@ -940,35 +946,38 @@ so the effects of `update_fulfill_htlc` is not completely reversed.
...
+for each channel, and MUST wait for to receive the other node's
+`channel_reestablish` message before sending any other messages for
+that channel.  The sending node MUST set `commitments_received` to the
+number of `commitment_signed` messages received, and MUST set
+`revocations_received` to the number of `revoke_and_ack` messages
+received.
+
+If `commitments_received` is one less than the number of
+`commitment_signed` messages the receiving node has sent, it MUST send
+a set of `update_` messages against the prior commitment transaction
```
Roasbeef: Does this mean only the _outbound_ send of `update_*` messages, or both inbound (updates I've made) and outbound (updates you've made)?

02-peer-protocol.md
```diff
@@ -940,35 +946,38 @@ so the effects of `update_fulfill_htlc` is not completely reversed.
...
-Otherwise, on reconnection, a node MUST retransmit old messages after `funding_signed` which may not
-have been received, and MUST NOT retransmit old messages which have
-been explicitly or implicitly acknowledged.  The following table
-lists the acknowledgment conditions for each message:
-
-* `funding_locked`: acknowledged by `update_` messages, `commitment_signed`, `revoke_and_ack` or `shutdown` messages.
-* `update_` messages: acknowledged by `revoke_and_ack`.
-* `commitment_signed`: acknowledged by `revoke_and_ack`.
-* `revoke_and_ack`: acknowledged by `commitment_signed` or `closing_signed`
-* `shutdown`: acknowledged by `closing_signed`.
```
rustyrussell: Good point, this should be explicit.  We *could* add it in to the commit counter, but I prefer the simplicity of "always retransmit" and associated "ignore duplicates".

02-peer-protocol.md
rustyrussell: Yeah, something's badly wrong.  It's also a big hint that you only have to handle the two cases, and only then until you've received the revoke_and_ack.

If we add multiple commits-in-flight, this "one less" becomes "N less".

02-peer-protocol.md
rustyrussell: Outbound.  If you've missed inbound, the other node will re-xmit because the `commitments_received` *you* send is one too low.

02-peer-protocol.md
```diff
@@ -952,22 +952,23 @@ Otherwise, on reconnection, a node MUST transmit `channel_reestablish`
...
+`revoke_and_ack` message received.
+
+If `commitments_received` is one less than the commitment number of
+the last `commitment_signed` message the receiving node has sent, it
+MUST send a set of `update_` messages against the prior commitment
+transaction followed by `commitment_signed`, otherwise if
+`commitments_received` is not equal to the commitment number of the
+last `commitment_signed` message the receiving node has sent, it
+SHOULD fail the channel.
+
```
sstone: What happens if the receiving node does not remember the updates it has sent ? sending a `commitment_signed` would be illegal I think

rustyrussell: Always resend updates then commitment signed. And counters in reestablish msg make it clear whether it received previous or not.

sstone:
How are forwarded HTLC resent when a node restarts ?

Suppose we're Bob, with the following setup:
Alice <-- channel #ab --> Bob <-- channel #bc --> Carol

Alice and Bob have both signed HTLC #1 and #2.
Bob has forwarded HTLC #1 and #2 but has not signed anything yet. 
Bob's sub-system that handles channel #bc restarts.
iiuc Bob and Carol forgets both HTLCs in  channel #bc. Upon reconnection, Bob's counters will match Carol's.
-> How are HTLCs #1 and #2 sent to Carol ?

Bob can probably tell than its #ab channel has HTLCs going to #bc that have not been processed by its #bc channel yet, but this is probably more complex to implement than the 'retransmit'' model (where Bob's #bc channel would just resend unacked messages).

rustyrussell:
Fabrice Drouin <notifications@github.com> writes:
> How are forwarded HTLC resent when a node restarts ?
>
> Suppose we're Bob, with the following setup:
> Alice <-- channel #ab --> Bob <-- channel #bc --> Carol
>
> Alice and Bob have both signed HTLC #1 and #2.
> Bob has forwarded HTLC #1 and #2 but has not signed anything yet. 

> Bob's sub-system that handles channel #bc restarts.
> iiuc Bob and Carol forgets both HTLCs in  channel #bc. Upon reconnection, Bob's counters will match Carol's.
> -> How are HTLCs #1 and #2 sent to Carol ?

Bob will probably retransmit, but certainly has the option of failing
the incoming HTLCs instead.

> Bob can probably tell than its #ab channel has HTLCs going to #bc that have not been processed by its #bc channel yet, but this is probably more complex to implement than the 'retransmit'' model (where Bob's #bc channel would just resend unacked messages).

That's certainly a valid implementtion, though
scan-and-retransmit-as-required is how I implemented reconnection in our
old daemon.  Otherwise you need to persist that information, which means
hitting the database on ever send...

Cheers,
Rusty.

sstone:
Yes, the "rollback" option mimics the batch protocol: one batch "write" call, vs one write call per send with the "retransmit" option. Same amount of data to write in both cases, but with less calls (though it's probably much less of a win in the case of database calls).

02-peer-protocol.md
pm47: Do you mean that:
"it MUST send a set of `update_` messages against the prior commitment transaction followed by `commitment_signed`"
is actually:
"it MUST re-send the same `update_` messages that lead to the prior commitment, followed by the same `commitment_signed`?

rustyrussell: You're right.  I'll fix this using your wording.

Now, in practice, you have to remember the state when you send commitment_signed, so you probably just retransmit (that's what I do).

Re: expired HTLCs.  The receiver will fail with expiry_too_soon, yes.  But note that there's only a narrow window where sender would send them: they had to set their deadline when sending the first commitment_signed message, and that would have expired and forced them to go on-chain.  Otherwise, they might have received the commitment_signed, and published the commitment transaction containing the expired HTLC then fulfilled.

02-peer-protocol.md
```diff
@@ -940,35 +946,45 @@ so the effects of `update_fulfill_htlc` is not completely reversed.
...
-fail the channel if they occur.
+channel, and the following requirements do not apply.
+
+On reconnection, a node MUST retransmit `funding_locked` unless it has
+received an `update_` or `revoke_and_ack` for that channel, otherwise
+it MAY retransmit `funding_locked`.  On reconnection, a node MUST ignore
+a redundant `funding_locked` if it receives one.
+
+On reconnection, a node MUST transmit `channel_reestablish`
+for each channel, and MUST wait for to receive the other node's
```
rustyrussell: Yes.  Although I had to fix the definition of the revocations_received; it was defined to be the commit index of last-received revocation, which is undefined.  So I added one, and added a revocations_received explanation paragraph.

02-peer-protocol.md
```diff
@@ -958,8 +958,8 @@ for each channel, and MUST wait for to receive the other node's
 `channel_reestablish` message before sending any other messages for
 that channel.  The sending node MUST set `commitments_received` to the
 commitment number of the last `commitment_signed` it received, and
-MUST set `revocations_received` to the commitment number of the last
-`revoke_and_ack` message received.
+MUST set `revocations_received` to one greater than the commitment number of the last
```
pm47: How about calling the fields `next_commitment_number` and `next_revocation_number`? It would solve the initialization problem, and would also fit nicely with the new phrasing in 3f9748b7b63071296a05502c418fe6dd1c14683f.

rustyrussell: I actually changed it to explicitly be the next indices, then read your updated comment before pushing. It makes some sense, and it's consistent between revocations and commitments: this means the commitment value is 1 greater than in the previous definition though (effectively still the total number received, but counting the signature from open/accept too).

rustyrussell:
>
>
>
>> @@ -958,8 +958,8 @@ for each channel, and MUST wait for to receive the other node's
>  `channel_reestablish` message before sending any other messages for
>  that channel.  The sending node MUST set `commitments_received` to the
>  commitment number of the last `commitment_signed` it received, and
> -MUST set `revocations_received` to the commitment number of the last
> -`revoke_and_ack` message received.
> +MUST set `revocations_received` to one greater than the commitment number of the last
>
> How about calling the fields `next_commitment_number` and `next_revocation_number`? It would solve the initialization problem, and would also fit nicely with the new phrasing in 3f9748b7b63071296a05502c418fe6dd1c14683f.

Yes, agreed.

Here's the new text:

On reconnection, a node MUST transmit `channel_reestablish`
for each channel, and MUST wait for to receive the other node's
`channel_reestablish` message before sending any other messages for
that channel.  The sending node MUST set `next_local_commitment_number` to the
commitment number of the next `commitment_signed` it expects to receive, and
MUST set `next_remote_revocation_number` to the commitment number of the

If `next_local_commitment_number` is equal to the commitment number of
the last `commitment_signed` message the receiving node has sent, it
MUST reuse the same commitment number for its next `commitment_signed`,
otherwise if `next_local_commitment_number` is not one greater than the commitment number of the
last `commitment_signed` message the receiving node has sent, it
SHOULD fail the channel.

If `next_remote_revocation_number` is equal to the commitment number of
the last `revoke_and_ack` the receiving node has sent, it MUST re-send
the `revoke_and_ack`, otherwise if `next_remote_revocation_number` is not
equal to one greater than the commitment number of the last `revoke_and_ack` the
receiving node has sent (or equal to zero if none have been sent), it SHOULD fail the channel.

...

Note that the `next_local_commitment_number` starts at 1 since
commitment number 0 is created during opening.
`next_remote_revocation_number` will be 0 until the
`commitment_signed` for commitment number 1 is received, at which
point the revocation for commitment number 0 is sent.

02-peer-protocol.md
```diff
@@ -940,35 +946,44 @@ so the effects of `update_fulfill_htlc` is not completely reversed.
...
-case where the corresponding commitment transaction is broadcast by
-the other side at any time.  This is particularly important if a node
-does not simply retransmit the exact same `update_` messages as
-previously sent.
-
-A receiving node MAY ignore spurious message retransmission, or MAY
-fail the channel if they occur.
+channel, and the following requirements do not apply.
+
+On reconnection, a node MUST retransmit `funding_locked` unless it has
```
pm47: How about moving this paragraph below the next one, and rephrase it to:
"If `next_local_commitment_number` is zero in both the `channel_reestablish` it sent and received, then the node MUST retransmit `funding_locked`, otherwise it MUST NOT. On reconnection, a node MUST ignore a redundant `funding_locked` if it receives one."

Would that work? I find it more straightforward to understand and implement, and I think there is an ambiguity in case of discarded `update_*` in the current version.

rustyrussell: Brilliant!  Patched exactly like that.

rustyrussell:
I squashed fixme commits, for final review phase.

02-peer-protocol.md
pm47: Oops, right

PR #183, opened 2017-05-31
Bolt 11: Invoice Protocol for Lightning Payments

rustyrussell:
For those who reviewed the previous lightning-payencode draft, the only recent change (modulo bugfixes) is that the timestamp is 35 bits, so it's partitioned nicely in the bech32 encoding.

Also, the payment hash is now a tagged field (in case it changes format), and version has been eliminated (it would be almost impossible to transition to a new version).

### DIFF
--- a/00-introduction.md
+++ b/00-introduction.md
@@ -21,6 +21,7 @@ This is version 0.
 7. [BOLT #7](07-routing-gossip.md): P2P Node and Channel Discovery
 8. [BOLT #8](08-transport.md): Encrypted and Authenticated Transport
 9. [BOLT #9](09-features.md): Assigned Feature Flags
+11. [BOLT #11](11-payment-encoding.md): Invoice Protocol for Lightning Payments
 
 ## Glossary and Terminology Guide
 
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -76,6 +76,9 @@ Human Readable Part concatenated with the Data Part and zero bits
 appended to the next byte boundary, with a trailing byte containing
 the recovery ID (0, 1, 2 or 3).
 
+A reader MUST check that the `signature` is valid (see the `n` tagged
+field specified below).
+
 ## Rationale
 
 `signature` covers an exact number of bytes because although the SHA-2
@@ -95,6 +98,7 @@ Currently defined Tagged Fields are:
 
 * `p` (1): `data_length` 52.  256-bit SHA256 payment_hash: preimage of this provides proof of payment.
 * `d` (13): `data_length` variable.  short description of purpose of payment (ASCII),  e.g. '1 cup of coffee'
+* `n` (19): `data_length` 53.  The 33-byte public key of the payee node.
 * `h` (23): `data_length` 52.  256-bit description of purpose of payment (SHA256).  This is used to commit to an associated description which is too long to fit, such as may be contained in a web page.
 * `x` (6): `data_length` variable.  `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
 * `f` (9): `data_length` variable, depending in version. Fallback on-chain address: for bitcoin, this starts with a 5 bit `version`; a witness program or P2PKH or P2SH address.
@@ -122,6 +126,9 @@ which SHOULD be a complete description of the purpose of the payment.
 
 A writer SHOULD use the minimum `x` `data_length` possible.
 
+A writer MAY include an `n` field, which MUST be set to the public key
+used to create the `signature`.
+
 If a writer offers more than one of any field type, it MUST specify
 the most-preferred field first, followed by less-preferred fields in
 order.
@@ -142,11 +149,14 @@ indicate a sequence of non-public channels to traverse.
 A writer MUST pad field data to a multiple of 5 bits, using zeroes.
 
 A reader MUST skip over unknown fields, an `f` field with unknown
-`version`, or a `p`, `h` or `r` field which does not have ``data_length`` 52,
-52 or 79 respectively.
+`version`, or a `p`, `h`, `n` or `r` field which does not have ``data_length`` 52,
+52, 53 or 79 respectively.
 
 A reader MUST check that the SHA-2 256 in the `h` field exactly
 matches the hashed description.
+
+A reader MUST use the `n` field to validate the signature instead of
+performing signature recovery if a valid `n` field is provided.
 
 ### Rationale
 
@@ -165,6 +175,9 @@ complex orders; thus the `h` field allows a summary, though the method
 by which the description is served is as-yet unspecified, and will
 probably be transport-dependent.  The `h` format could change in future
 by changing the length, so readers ignore it if not 256 bits.
+
+The `n` field can be used to explicitly specify the destination node ID,
+instead of requiring signature recovery.
 
 The `x` field gives advance warning as to when a payment will be
 refused; this is mainly to avoid confusion.  The default was chosen
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -62,31 +62,26 @@ whatever is being offered in return.
 
 The data part consists of multiple sections:
 
-1. `version`: 0 (5 bits)
 1. `timestamp`: seconds-since-1970 (35 bits, big-endian)
-1. `payment_hash`: (256 bits + 4 trailing zero bits)
 1. Zero or more tagged parts.
 1. `signature`: bitcoin-style signature of above. (520 bits)
 
 ## Requirements
 
-A writer MUST set `version` to 0, MUST set `timestamp` to the time to
+A writer MUST set `timestamp` to the time to
 the number of seconds since Midnight 1 January 1970, UTC in
-big-endian, and MUST set `payment_hash` to the SHA-2 256-bit hash of
-the `payment_preimage` which will be given in return for payment,
-followed by four zero bits.  A writer MUST set `signature` to a valid
+big-endian.  A writer MUST set `signature` to a valid
 512-bit secp256k1 signature of the double SHA2 256-bit hash of the
-Human Readable Part concatenated with a bytes for each 5 bits of the
+Human Readable Part concatenated with a byte for each 5 bits of the
 Data Part, with a trailing byte containing the recovery ID (0, 1, 2 or
 3).
 
 ## Rationale
 
-The 5-bit `version` fits neatly into base32 encoding (0 == `q`), and
-allows for future incompatible upgrades.  `signature` covers the
-pre-base32 encoded data for simplicity, especially since the data part
-may not end on an 8-bit boundary.  It also allows public key recovery,
-so the identity of the payee node is implied.
+`signature` covers the pre-base32 encoded data for simplicity,
+especially since the data part may not end on an 8-bit boundary.  It
+also allows public key recovery, so the identity of the payee node is
+implied.
 
 ## Tagged Fields
 
@@ -98,12 +93,11 @@ Each Tagged Field is of format:
 
 Currently defined Tagged Fields are:
 
-* `d` (13): short description of purpose of payment (ASCII),  e.g. '1 cup of coffee'
-* `h` (23): 256-bit description of purpose of payment (SHA256).  This is used to commit to an associated description which is too long to fit, such as may be contained in a web page.
-* `x` (6): `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
-* `f` (9): fallback on-chain address.  For bitcoin, this is 5 bits of
-      witness version followed by a witness program, with witness
-      version `3` (17) meaning P2PKH, `j` (18) meaning [P2SH](https://github.com/bitcoin/bips/blob/master/bip-0016.mediawiki); both are followed by a 20-byte hash value.
+* `p` (1): length 52.  256-bit SHA256 payment_hash: preimage of this provides proof of payment.
+* `d` (13): length variable.  short description of purpose of payment (ASCII),  e.g. '1 cup of coffee'
+* `h` (23): length 52.  256-bit description of purpose of payment (SHA256).  This is used to commit to an associated description which is too long to fit, such as may be contained in a web page.
+* `x` (6): length variable.  `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
+* `f` (9): length variable, depending in version. Fallback on-chain address: for bitcoin, this starts with a 5 bit `version`; a witness program or p2pkh or p2sh address.
 * `r` (3): extra routing information.  This should be appended to the route
       to allow routing to non-public nodes; there may be more than one of these.
    * `pubkey` (264 bits)
@@ -112,6 +106,10 @@ Currently defined Tagged Fields are:
    * `cltv_expiry_delta` (16 bits, big-endian)
 
 ### Requirements
+
+A writer MUST include exactly one `p` field, and set `payment_hash` to
+the SHA-2 256-bit hash of the `payment_preimage` which will be given
+in return for payment.
 
 A writer MUST NOT include more than one `d`, `h`, or `x` fields, and
 MAY include more then one `f` field.
@@ -124,8 +122,11 @@ which SHOULD be a complete description of the purpose of the payment.
 
 A writer SHOULD use the minimum `x` field length possible.
 
-If a writer offers more than one `f` field, it SHOULD specify
-preferred addresses first.  For bitcoin payments, a writer MUST set an
+If a writer offers more than one of any field type, it MUST specify
+the most-preferred field first, followed by less-preferred fields in
+order.
+
+For bitcoin payments, a writer MUST set an
 `f` field to a valid witness version and program, or `17` followed by
 a public key hash, or `18` followed by a script hash.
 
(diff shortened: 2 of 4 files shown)


11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+      to allow routing to non-public nodes; there may be more than one of these.
+   * `pubkey` (264 bits)
+   * `channel_id` (64 bits)
+   * `fee` (64 bits, big-endian)
+   * `cltv_expiry_delta` (16 bits, big-endian)
+
+### Requirements
+
+A writer MUST NOT include more than one `d`, `h`, or `x` fields, and
+MAY include more then one `f` field.
```
kallewoof: Typo: than

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+payment.
+
+## Payer / Payee Requirements
+
+A payer SHOULD NOT attempt a payment after the `timestamp` plus
+`expiry` has passed.  Otherwise, if a lightning payment fails, a payer
+MAY attempt to use the address given the first `f` field it
+understands for payment.  A payer MAY use the sequence of channels
+specified by `r` to route to the payee.  A payee SHOULD consider the
+fee amount and payment timeout before initiating payment.
```
kallewoof: Unless I'm confused, you meant "payer SHOULD consider the fee amount and payment timeout before initiating payment".

00-introduction.md
```diff
@@ -21,6 +21,7 @@ This is version 0.
 7. [BOLT #7](07-routing-gossip.md): P2P Node and Channel Discovery
 8. [BOLT #8](08-transport.md): Encrypted and Authenticated Transport
 9. [BOLT #9](09-features.md): Assigned Feature Flags
+11. [BOLT #11](11-payment-encoding.md): Invoice Protocol for Lightning Payments
```
luke-jr: What happened to 10? Can you give me some hint as to how many BIPs to set aside for when we merge BOLTs into the BIP repo?

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+> lntb20m1qpvj6chqqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqfpp3x9et2e20v6pu37c5d9vax37wxq72un98hp58yjmdan79s6qqdhdzgynm4zwqd5d7xmw5fk98klysy043l2ahrqsqsj75rjv443nrh8gu5xutlyyqyx6ul76m2rx87yxr5gdfagyywn3s6wtpfrl6elncce7rmh6kndvr5nur76w9u7z0k3gq93fyfpu9zqq3jp9as
+
+Breakdown:
+
+* `lntb`: prefix, lightning on bitcoin testnet
+* `20m`: amount (20 milli-bitcoin)
+* `1`: Bech32 separator
+* `q`: version (0)
+* `pvj6chq`: timestamp (1496146656)
+* `f`: tagged field: fallback address
```
Saicere: Not a big deal, but the breakdown for this example is missing the payment hash.

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+  * `pp`: field length (`p` = 1. 1 * 32 + 1 == 33)
+  * `3x9et2e20v6pu37c5d9vax37wxq72un98`: `3` = 17, so P2PKH address
+* `h`: tagged field: hash of description...
+* `qsj75rjv443nrh8gu5xutlyyqyx6ul76m2rx87yxr5gdfagyywn3s6wtpfrl6elncce7rmh6kndvr5nur76w9u7z0k3gq93fyfpu9zqq`: signature
+* `3jp9as`: Bech32 checksum
+
```
Saicere: Not a big deal, but the breakdown for this example is missing the payment hash.

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+* `f` (9): fallback on-chain address.  For bitcoin, this is 5 bits of
+      witness version followed by a witness program, with witness
+      version `3` (17) meaning P2PKH, `j` (18) meaning [P2SH](https://github.com/bitcoin/bips/blob/master/bip-0016.mediawiki); both are followed by a 20-byte hash value.
+* `r` (3): extra routing information.  This should be appended to the route
+      to allow routing to non-public nodes; there may be more than one of these.
+   * `pubkey` (264 bits)
+   * `channel_id` (64 bits)
+   * `fee` (64 bits, big-endian)
+   * `cltv_expiry_delta` (16 bits, big-endian)
+
```
Saicere: This is not consistent with the description for the "r" tagged field description in the [simplified description](https://github.com/rustyrussell/lightning-payencode) which says 32-bit fee and 32-bit cltv. Which is correct?

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+* `pvj6chq`: timestamp (1496146656)
+* `qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypq`: payment hash.
+* `r`: tagged field: route information
+  * `z0`: field length (`z` = 2, `0` = 15.  2 * 32 + 15 = 79)
+    `q20q82gphp2nflc7jtzrcazrra7wwgzxqc8u7754cdlpfrmccae92qgzqvzq2ps8pqqqqqq5qqqqqqc`: pubkey `029e03a901b85534ff1e92c43c74431f7ce72046060fcf7a95c37e148f78c77255`, `channel_id` 0102030405060708, `fee` 20 millisatoshi, `cltv_expiry_delta` 3.
+* `f`: tagged field: fallback address...
+* `h`: tagged field: hash of description...
+* `t7xnzx2sl9rkrc5lgzkcdqs57sj0s0vz8z9g2wk4hucfdrtvupkx93wqjcj8lpejzc95k4p4hw0qrfay5x36def3ret6yd9s0vqwtysq`: signature
+* `m5z5dw`: Bech32 checksum
+
```
Saicere: Likely related to the inconsistency mentioned above, the "r" field here is encoded with 32-bit (padded to 35-bit) fee and cltv which does not match the definition under Tagged Fields. And unless my manual bech32 decoding is wrong, the cltv in the example reads 24 rather than 3?

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+* `lnbc`: prefix, lightning on bitcoin mainnet
+* `20m`: amount (20 milli-bitcoin)
+* `1`: Bech32 separator
+* `q`: version (0)
+* `pvj6chq`: timestamp (1496146656)
+* `qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypq`: payment hash.
+* `r`: tagged field: route information
+  * `z0`: field length (`z` = 2, `0` = 15.  2 * 32 + 15 = 79)
+    `q20q82gphp2nflc7jtzrcazrra7wwgzxqc8u7754cdlpfrmccae92qgzqvzq2ps8pqqqqqq5qqqqqqc`: pubkey `029e03a901b85534ff1e92c43c74431f7ce72046060fcf7a95c37e148f78c77255`, `channel_id` 0102030405060708, `fee` 20 millisatoshi, `cltv_expiry_delta` 3.
```
Saicere: Likely related to the inconsistency mentioned above, the "r" field here is encoded with 32-bit (padded to 35-bit) fee and cltv which does not match the definition under Tagged Fields. And unless my manual bech32 decoding is wrong, either the fee or the cltv is padded inconsistently (left-padding would give incorrect cltv, right-padding would give incorrect fee.)

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+A writer MUST include at least one `r` field if it does not have a
+public channel associated with its public key.  The `pubkey` is the
+node ID of the start of the channel, `channel_id` is the channel ID
+field to identify the channel, `fee` is the total fee required to use
+that channel to send `amount` to the final node, specified in 10^-11
+currency units, and `cltv_expiry_delta` is the block delta required
+by the channel.  A writer MAY include more than one `r` field to
+indicate a sequence of non-public channels to traverse.
+
+A writer MUST pad fields to a multiple of 5 bits, using zeroes.
```
Saicere: This seems to be consistently padding with trailing bits, but should probably be made explicit. Another thing that confused me initially was that the term "field" is used to refer both to the d/h/x/f/r "tagged field" as a whole, as well as individual "fields" in the "tagged field". 

> A writer MUST include at least one `r` field (...) `channel_id` is the channel ID field to identify the channel

The individual "fields" are not padded as far as I can tell in the examples, while the "tagged field" as a whole is, and the distinction is important for "tagged fields" with multiple "fields".

00-introduction.md
cdecker: 10 is the DNS seed protocol, which I haven't finalized yet. I don't think that merging the bolts into BIPs is actually desirable since they are not directly Bitcoin related, rather they describe a protocol built on top of Bitcoin, or other cryptocurrency.

luke-jr: Lightning is part of Bitcoin, therefore should be using the BIP process.

11-payment-encoding.md
```diff
@@ -0,0 +1,345 @@
...
+1. `type` (5 bits)
+1. `length` (10 bits, big-endian)
+1. `data` (length x 5 bits)
+
+Currently defined Tagged Fields are:
+
+* `d` (13): short description of purpose of payment (ASCII),  e.g. '1 cup of coffee'
+* `h` (23): 256-bit description of purpose of payment (SHA256).  This is used to commit to an associated description which is too long to fit, such as may be contained in a web page.
+* `x` (6): `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
+* `f` (9): fallback on-chain address.  For bitcoin, this is 5 bits of
```
luke-jr: Maybe the invoice format should be compatible with BIP 21, so existing clients can use the fallback?

11-payment-encoding.md
rustyrussell: Thanks, fixed!

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+
+* `m` (milli): multiply by 0.001
+* `u` (micro): multiply by 0.000001
+* `n` (nano): multiply by 0.000000001
+* `p` (pico): multiply by 0.000000000001
+
+## Requirements
+
+A writer MUST include `amount` if payments will be refused if less
+than than that.  A writer MUST encode `amount` as a positive decimal
```
rustyrussell: Thanks, fixed.

11-payment-encoding.md
```diff
@@ -0,0 +1,339 @@
...
+   * `channel_id` (64 bits)
+   * `fee` (64 bits, big-endian)
+   * `cltv_expiry_delta` (16 bits, big-endian)
+
+### Requirements
+
+A writer MUST NOT include more than one `d`, `h`, or `x` fields, and
+MAY include more then one `f` field.
+
+A writer MUST include either a `d` or `h` field, and MUST NOT include
```
rustyrussell: Thanks.  I added the word ASCII, and expanded the example where it was used.

11-payment-encoding.md
rustyrussell: Fixed, thanks.

11-payment-encoding.md
rustyrussell: Spec is canonical, good catch.  I've updated spec.

11-payment-encoding.md
rustyrussell: Indeed.  Fixed.

00-introduction.md
JaredR26: > Lightning is part of Bitcoin, therefore should be using the BIP process.

Additional features that happen to take advantage of Bitcoin's blockchain are not Bitcoin, not a part of Bitcoin, and should not be BIP's.

Otherwise, Blockchain.info, coinbase, BU, classic, and other companies/interests have a legitimate right to submit BIPs for their own systems developed on top of Bitcoin.  BIP is for things that exclusively affect and relate to Bitcoin's core architecture and consensus.

ftrader: @JaredR26 : Please do not spread misinformation. BU is not a system developed on top of Bitcoin, it is a Bitcoin client, as is Classic. Just because these clients propose alterations to the consensus rules for blocks > 1MB does not mean they are not part of Bitcoin.

JaredR26: So just because Lightning proposes things built on top of Bitcoin, it therefore is a part of Bitcoin, but BU's proposed consensus changes to Bitcoin itself don't get to count?  What about coinbase account-to-account transfers or coinbase-account-to-bitstamp-account transfers, can we get a BIP for those?  What about the miner fast relay network that is already in use today, why wouldn't that have a BIP?

If it isn't related directly to Bitcoin core / consensus, it doesn't need a BIP.  The only fundamental difference here is that @luke-jr is attempt to have BOLT's and lightning appear to carry the same weight that are currently associated with BIP's.  This inadvertently waters down the meaning of what a BIP is and what qualifies, and turns it into a subjective discussion of what gets to be called a BIP and what doesn't.  There's no technical benefit for the two to be combined, and it doesn't help Bitcoin either.

luke-jr: BIPs have never been limited to Core or Bitcoin consensus. They encompass anything within the scope of Bitcoin that needs standardisation. BU trolls may as part of their agenda to hijack Bitcoin, try to spin BIPs as limited in scope to a specific part of Bitcoin, but this is simply not true.

tools/spellcheck.sh
```diff
@@ -54,6 +54,8 @@ do
 	# quoted identifers eg. `htlc_id`
 	# field descriptions, eg. `* [`num_htlcs*64`:`htlc_signature]'
 	# indented field names, eg. '    `num_htlcs`: 0'
+	# lightning addresses, eg. `lnbc1qpvj6chq...`
+	# BIP 174 addresses, eg. `bc1qpvj6chq...`
```
halseth: 173

cdecker:
At the risk of ending up in bikeshed world: there are a few minor things I dislike about the proposal.

1. The fact that tags are allowed to repeat (route hops currently being the only use of this) complicates the tag handling, I would have preferred an aggregation step before serializing as tag such that the tag keys are guaranteed to be unique.
2. Having fields in cleartext in the human readable part (currency and amount), though nice for the user, is rather strange as it mixes encoding with content. It just feels like the "there's a PGP header, it must be ok". Users should never decide whether to send a transfer or not based on the encoded request, but rather have their client decode it completely and then decide based on the full information.
3. We are mixing encoding and payload when converting parts of the payload individually (to adhere to character boundaries in the encoded format) and then signing it. A cleaner implementation would serialize the payload (similar to what we already do with the network messages), then sign that serialized payload (appending the signature to the end), and only then encoding it in bech32. That'd get rid of the extremely awkward partial conversions we currently do for `int`s and strings.

Sorry for the late comments, but it took me some time to rewrite the python implementation :wink:

rustyrussell:
> At the risk of ending up in bikeshed world: there are a few minor things I dislike about the proposal.
>
> 1. The fact that tags are allowed to repeat (route hops currently being the only use of this) complicates the tag handling, I would have preferred an aggregation step before serializing as tag such that the tag keys are guaranteed to be unique.

We *could* encode multiple steps in a single `r` and avoid this.  But I
like that fixed-length fields are ignored if wrong length, so we won't
run out of letters with future enhancements.

> 2. Having fields in cleartext in the human readable part (currency and amount), though nice for the user, is rather strange as it mixes encoding with content. It just feels like the "there's a PGP header, it must be ok". Users should never decide whether to send a transfer or not based on the encoded request, but rather have their client decode it completely and then decide based on the full information.

It's a minor preference, but I like being able to glance an parse a
payment; I'd prefer the description in HRP as well by that gets a bit
silly with spaces...

When looking through a series of payments, the amount and currency are
nice to have at a glance.

And being able to parse fields by pulling apart makes it quite easy to
explain.

> 3. We are mixing encoding and payload when converting parts of the payload individually (to adhere to character boundaries in the encoded format) and then signing it. A cleaner implementation would serialize the payload (similar to what we already do with the network messages), then sign that serialized payload (appending the signature to the end), and only then encoding it in bech32. That'd get rid of the extremely awkward partial conversions we currently do for `int`s and strings.

Agreed, I really dislike the signature scheme too.

We should assemble the HRP, and the data as a bitstream, then sign that
(with zero bits appended to an 8-bit binary), then bech32 the result.

I'm sure there's a way to do a bitstream neatly in python, I just don't
know what it is...

> Sorry for the late comments, but it took me some time to rewrite the python implementation :wink: 

I look forward to the PR :)

Thanks,
Rusty.

rustyrussell:
@pm47 I've got a rework which makes the signature sign the bytestring (rather than the weird 5-bits-per-byte thing it does now).  The weird signature encoding fell out from the python bech32 code, but I wanted to check with you before I "fix" it in case it makes things harder...

pm47:
@rustyrussell no worries, we are still working on our implementation. If it makes it simpler all the better!

11-payment-encoding.md
```diff
@@ -0,0 +1,355 @@
...
+1. `signature`: bitcoin-style signature of above. (520 bits)
+
+## Requirements
+
+A writer MUST set `timestamp` to the time to
+the number of seconds since Midnight 1 January 1970, UTC in
+big-endian.  A writer MUST set `signature` to a valid
+512-bit secp256k1 signature of the double SHA2 256-bit hash of the
+Human Readable Part concatenated with a byte for each 5 bits of the
+Data Part, with a trailing byte containing the recovery ID (0, 1, 2 or
```
rustyrussell: This is a very good q!  It's currently defined by the secp256k1 implementation, and I can't find a description of it anywhere else.  I think we need to define the values here in this case.

tools/spellcheck.sh
rustyrussell: Thanks, fixed BTW.

PR #184, opened 2017-06-08
BOLT 02: opt-in dual-funding

sstone:
The goal of this PR is to enable dual-funding, which we believe is an important feature because it lets nodes create channels that can be used to pay on both sides from the start, which improves user experience. Also, from a routing point of view it may be easier to bootstrap a network of balanced channels.
 
The protocol has been changed as little as possible and existing messages are reused, so even though dual-funding was supposed to be a 1.1 feature it may makes its way into 1.0.
 
This is partially inspired by what lnd was doing some time ago (maybe it was a feature branch or an option ?) but I’ve chosen to keep the number of messages to a minimum and use serialized tx instead of breaking them up into specific fields: some of the opening messages have been modified to include serialized transactions. This is not optimal, but I believe that it is not significant since these messages will only be exchanged once, and all implementations include tools that understand the bitcoin protocol so we don't have to reimplement serialization of tx inputs, outputs, …
 
To minimize the number of messages, the funder now provides a partial funding tx in their open message. The fundee will then be able to add their own inputs and change output to the funding tx, as well as the multisig funding output, compute the txid of the funding tx, and compute a signature for the funder’s commit tx.
 
Another option would be to keep the messages that we have now and add new specific messages for the dual funding process, which would create 2 clearly identified workflows for opening channels (with or without dual funding).
 
 
The gist of it is:
 
A --- open-channel ----> B 
// open_channel includes a partial funding tx with A's inputs and change outputs
 
A <-- accept_channel --- B 
// accept channel includes a complete but unsigned funding tx, as well as A's commit sig
// A and B now have an unsigned funding tx that cannot be malleated
 
A --- funding_created --> B 
// funding_created includes B's commit sig
// A and B now have signed commit txs
 
A <-- funding_signed --- B 
// funding signed includes witness scripts for every input added by B
// A has now a fully signed funding tx
 
On the plus side:
* users can start with channels that can be used to pay on both sides
 
On the minus side:
* opening messages are larger than they could be
* funder must create and send a partial funding tx in the first message
* fees are still paid by the funder, even if the other side also funds the channel

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -31,22 +31,17 @@ operation, and closing.
 
 
 Channel establishment begins immediately after authentication, and
-consists of the funding node sending an `open_channel` message,
-followed by the responding node sending `accept_channel`. With the
-channel parameters locked in, the funder is able to create the funding
-transaction and both versions of the commitment transaction as described in
-[BOLT
-03](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#bolt-3-bitcoin-transaction-and-script-formats).
-The funder then sends the outpoint of the funding output along with a
-signature for the responder's version of the commitment transaction
-with the `funding_created` message. Once the responder learns the
-funding outpoint, she is able to generate the initiator's commitment
-for the commitment transaction, and send it over using the
-`funding_signed` message.
-
-Once the channel funder receives the `funding_signed` message, they
-must broadcast the funding transaction to the Bitcoin network. After
-the `funding_signed` message is sent/received, both sides should wait
+consists of the funding node sending an `open_channel` message, which
+includes a partial funding transaction.
+The responding node replies with `accept_channel` which may include additional 
+inputs for the funding transaction ("dual funding"), and a signature for the
+funder's first commitment transaction as described in [BOLT 03](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#bolt-3-bitcoin-transaction-and-script-formats).
+The funding node then sends a `funding_created` message that includes a signature for the responder's first
+commitment transaction, and the responder replies with a `funding_signed` message that includes
+signatures for all the funding inputs that they added in their `accept_channel` message.
+Both nodes now have a signed commitment transaction, and the funding node has a fully signed funding transaction that
+they can broadcast.  
+After the `funding_signed` message is sent/received, both sides should wait
 for the funding transaction to enter the blockchain and reach their
 specified depth (number of confirmations). After both sides have sent
 the `funding_locked` message, the channel is established and can begin
@@ -100,6 +95,8 @@ desire to set up a new channel.
    * [`33`:`delayed_payment_basepoint`]
    * [`33`:`first_per_commitment_point`]
    * [`1`:`channel_flags`]
+   * [`2`: `len`]
+   * [`len`: `partial_funding_tx`]
 
 
 The `chain_hash` value denotes the exact blockchain the opened channel will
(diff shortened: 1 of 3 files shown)


02-peer-protocol.md
```diff
@@ -174,13 +172,19 @@ are not valid DER-encoded compressed secp256k1 pubkeys.
 
 The receiver MUST NOT consider funds received using `push_msat` to be received until the funding transaction has reached sufficient depth.
 
+The sender must set `len` to the length of the serialized, unsigned, partial funding transaction, which MUST be followed by the transaction itself.
+
+`partial_funding_tx` is an unsigned, partial funding transaction that includes:
+ * inputs that the sender wants to use to fund the channel. These inputs MUST be either P2WPKH or P2WSH and must include the witness script in their signature script field.
+ * an optional change output
+ 
+Please note that this partial transaction does not include the multisig funding output, since the sender does not know the receiver's public key yet.
```
sstone: yes

02-peer-protocol.md
```diff
@@ -215,65 +219,96 @@ acceptance of the new channel.
...
+`funding_tx` is an unsigned funding transaction which includes:
+ * optional additional inputs that the sender wants to use to fund the channel. These inputs MUST be either P2WPKH or P2WSH and must include the witness script in their signature script field.
+ * an optional change output
+ * a funding transaction output as defined in [BOLT #3](03-transactions.md#funding-transaction-output)
+The sender MUST then set `signature` to the valid signature using its funding_pubkey for the initial commitment transaction as defined in BOLT #3
+ 
+The amount that the sender will start with, i.e. that will be included in his first commit transaction, is the difference between the additional inputs
+and the change output.
+
+If the sender does not wish to fund the channel, they MAY set `len` to 0 and not include the funding transaction since the
```
sstone: yes it would be a bit more consistent to always return a tx.

if the channel is funded by both nodes, then any one them could choose to double-spend their funding inputs before the funding tx is published. The funding tx would never confirm, and the channel would never reach the "open" state. It's annoying but easy to detect with tools that LN nodes must already have implemented, nobody looses money and there's nothing to do so I don't find it scary but maybe I've missed something ?

cdecker: Well, it forces us to actively doublespend if we fail to create the channel: let's say A proposes a channel to be opened to B. A funds the channel, sends it over to B, which adds its own funds and sends back a signed funding TX. Now A just sits there and does nothing. B must now double spend the funds after it notices that the channel will not be created, otherwise A could dig up the funding TX when B wants to spend its funds for something completely unrelated, i.e., interfering with a later payment, and locking the funds into the unilateral closure procedure, potentially hurting B. The only way for B to safeguard against this is to doublespend right away.

It's not dangerous as in "we can lose funds", but A can interfere and be annoying at a later point in time. It's not currently an issue because the funding party is the one starting the confirmation, and is the only one which has all the signatures ready.

sstone: yes having the spend the funding outputs when channel opening times out is annoying. I think that it can be mitigated by adding time-outs to the funding output scripts, I'll post an update asap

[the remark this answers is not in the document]
joemphilips: Is it possible to add time-outs without introducing new BIP? It seems impossible to me.

02-peer-protocol.md
```diff
@@ -29,24 +29,18 @@ operation, and closing.
...
-`funding_signed` message.
-
-Once the channel funder receives the `funding_signed` message, they
-must broadcast the funding transaction to the Bitcoin network. After
-the `funding_signed` message is sent/received, both sides should wait
+consists of the funding node sending an `open_channel` message, which
+includes a partial funding transaction.
+The responding node replies with `accept_channel` which may include additional 
+inputs for the funding transaction ("dual funding"), and a signature for the
+funder's fist commitment transaction as described in [BOLT 03](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#bolt-3-bitcoin-transaction-and-script-formats).
```
jonathancross: "fist" => "first"

02-peer-protocol.md
```diff
@@ -215,65 +219,96 @@ acceptance of the new channel.
...
+ 
+The amount that the sender will start with, i.e. that will be included in his first commit transaction, is the difference between the additional inputs
+and the change output.
+
+If the sender does not wish to fund the channel, they MAY set `len` to 0 and not include the funding transaction since the
+receiver will have both public keys and will be able to build it.
+
+#### Rationale
+
+Once `accept_channel` has been received, both node have an unsigned funding transaction that cannot be malleated (all
```
jonathancross: "node" =>"nodes"

t-bast:
We can probably close this in favor of #524 ?

michaelfolkson:
I think this can be closed as it was superseded by #524 which has since been superseded by #851.

PR #191, opened 2017-06-27
BOLT7: Add network view pruning

### LINKED ISSUE #186: Forget channels that have not received an update for x hours
We currently have no good way of forgetting really old channels/nodes. By having nodes forget channels that they've not heard about in the last `n` hours (or similar) we can prune the channel graph easily, by putting a little bit more load on the network. I'd suggest having nodes issue a `channel_update` `n-1` hours after the last update, which should provide ample time for the update to propagate. If nothing changed the update should just resend the previous update, with the current timestamp.

cdecker:
Explicitly mentions that nodes SHOULD monitor the chain for channel
closes, and that a node MAY be removed if no open channels for that
node remain open.

Also mentions the 2 week lazy pruning we discussed on the call.

Closes #186

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -379,6 +379,20 @@ The node creating `channel_update` SHOULD accept HTLCs which pay an
 older fee for some time after sending `channel_update` to allow for
 propagation delay.
 
+## Pruning the Network View
+
+Nodes SHOULD monitor the funding transactions in the blockchain to identify channels that are being closed.
+If the funding output of a channel is being spent, then the channel is to be considered closed and SHOULD be removed from the local network view.
+
+Nodes MAY prune nodes added through `node_announcement` messages from their local view if the announced node no longer has any open channels associated.
+This is a direct result from the dependency of a `node_announcement` being preceeded by a `channel_announcement`,
+
+Several scenarios may result in channels becoming unusable and the endpoints unable to send updates for these channels.
+This happens for example in case that both endpoints lose access to their private keys, and cannot sign a `channel_update` or close the channel on-chain.
+These channels are unlikely to be part of a computed route since they would be partitioned off from the rest of the network, however they would remain in the local network view, and be forwarded to other nodes forever.
+For this reason nodes MAY prune channels should the timestamp of the latest `channel_update` be older than 2 weeks (1209600 seconds).
+In addition nodes MAY ignore channels with a timestamp older than 2 weeks.
+
 ## Recommendations for Routing
 
 As the fee is proportional, it must be calculated backwards from the
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -385,7 +385,7 @@ Nodes SHOULD monitor the funding transactions in the blockchain to identify chan
 If the funding output of a channel is being spent, then the channel is to be considered closed and SHOULD be removed from the local network view.
 
 Nodes MAY prune nodes added through `node_announcement` messages from their local view if the announced node no longer has any open channels associated.
-This is a direct result from the dependency of a `node_announcement` being preceeded by a `channel_announcement`,
+This is a direct result from the dependency of a `node_announcement` being preceded by a `channel_announcement`,
 
 Several scenarios may result in channels becoming unusable and the endpoints unable to send updates for these channels.
 This happens for example in case that both endpoints lose access to their private keys, and cannot sign a `channel_update` or close the channel on-chain.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -385,13 +385,16 @@ Nodes SHOULD monitor the funding transactions in the blockchain to identify chan
 If the funding output of a channel is being spent, then the channel is to be considered closed and SHOULD be removed from the local network view.
 
 Nodes MAY prune nodes added through `node_announcement` messages from their local view if the announced node no longer has any open channels associated.
-This is a direct result from the dependency of a `node_announcement` being preceded by a `channel_announcement`,
+This is a direct result from the dependency of a `node_announcement` being preceded by a `channel_announcement`.
+
+### Recommendation on pruning stale entries
 
 Several scenarios may result in channels becoming unusable and the endpoints unable to send updates for these channels.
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -379,6 +379,20 @@ The node creating `channel_update` SHOULD accept HTLCs which pay an
...
+
+Nodes SHOULD monitor the funding transactions in the blockchain to identify channels that are being closed.
+If the funding output of a channel is being spent, then the channel is to be considered closed and SHOULD be removed from the local network view.
+
+Nodes MAY prune nodes added through `node_announcement` messages from their local view if the announced node no longer has any open channels associated.
+This is a direct result from the dependency of a `node_announcement` being preceded by a `channel_announcement`,
+
+Several scenarios may result in channels becoming unusable and the endpoints unable to send updates for these channels.
+This happens for example in case that both endpoints lose access to their private keys, and cannot sign a `channel_update` or close the channel on-chain.
+These channels are unlikely to be part of a computed route since they would be partitioned off from the rest of the network, however they would remain in the local network view, and be forwarded to other nodes forever.
```
pm47: Why would they be partitioned off the network?

07-routing-gossip.md
```diff
@@ -379,6 +379,20 @@ The node creating `channel_update` SHOULD accept HTLCs which pay an
...
+For this reason nodes MAY prune channels should the timestamp of the latest `channel_update` be older than 2 weeks (1209600 seconds).
```
pm47: Shouldn't we set the time in blocks instead, since a channel is really created in a block? (nitpicking, I know!). We could round it to 2000 blocks.

07-routing-gossip.md
cdecker: For a channel to become garbage we both endpoints have to die, and lose all their information (settlement tx and/or private key), otherwise they could close normally. For that channel to become reachable there must be a path leading to that channel, and someone along that channel is the first person that doesn't suffer that loss, so hopefully they'd close the channel that goes into the cluster of failed channels, and partition them off the network. It might be a bit wishful thinking that this'd happen right away but hopefully we could isolate this dysfunctional cluster just by this.

07-routing-gossip.md
cdecker: I was playing with that thought as well, however we are currently using a timestamp to determine replacements in the updates. We aren't specifying that these should be wall-clock timestamps, but my guess is that this'd how things turn out. We can of course switch to calling them blocktime, but then we'd only be able to do a single update each 10 minute interval which may be a bit coarse grained. Everybody is free to implement the time counting how they'd like, this just gives them permission to forget.

pm47: Fair enough!

rustyrussell: Hmm, I'd prefer to timestamp locally rather than relying on timestamp field.  If we end up needing more sophisticated ratelimiting (the current scheme is bypassable by connecting to every node), I think we'll end up using block number + counter instead of timestamp (with a limit on how many updates per block).

As stands, this effectively introduces a requirement that the timestamp field be epoch time.

cdecker: I wasn't suggesting that we trust the timestamp, I just don't want to split terminology. timestamps in updates are basically sequence numbers that can be freely incremented, but my suggestion would be to just use timestamps since that's an easy enough way of ensuring we have monotonically increasing sequences. Keeping the prune timeout in seconds aligns nicely with this and we don't need assumptions on the blockrate for later support of other chains.

07-routing-gossip.md
```diff
@@ -379,6 +379,20 @@ The node creating `channel_update` SHOULD accept HTLCs which pay an
 older fee for some time after sending `channel_update` to allow for
 propagation delay.
 
+## Pruning the Network View
+
+Nodes SHOULD monitor the funding transactions in the blockchain to identify channels that are being closed.
+If the funding output of a channel is being spent, then the channel is to be considered closed and SHOULD be removed from the local network view.
+
+Nodes MAY prune nodes added through `node_announcement` messages from their local view if the announced node no longer has any open channels associated.
+This is a direct result from the dependency of a `node_announcement` being preceded by a `channel_announcement`,
```
Roasbeef: The comma at the end of this line should actually be a period.

07-routing-gossip.md
```diff
@@ -379,6 +379,20 @@ The node creating `channel_update` SHOULD accept HTLCs which pay an
...
+
+Several scenarios may result in channels becoming unusable and the endpoints unable to send updates for these channels.
+This happens for example in case that both endpoints lose access to their private keys, and cannot sign a `channel_update` or close the channel on-chain.
+These channels are unlikely to be part of a computed route since they would be partitioned off from the rest of the network, however they would remain in the local network view, and be forwarded to other nodes forever.
+For this reason nodes MAY prune channels should the timestamp of the latest `channel_update` be older than 2 weeks (1209600 seconds).
+In addition nodes MAY ignore channels with a timestamp older than 2 weeks.
```
Roasbeef: I'm not sure we should be strongly specifying such a timeout value. Pruning channels that haven't received updates in a period of time `T` is simply node policy. I'd say the wording here should be relaxed to entail a _recommendation_ rather than specify a value that nodes _must_ follow.

cdecker: Ok, moved the passive pruning out into its own subsection, calling it a recommendation, and explicitly forbidding any form of peer enforcement. I'd like to keep the 2 weeks ball park figure, even though it's not enforced.

rustyrussell:
> Ok, moved the passive pruning out into its own subsection, calling it a recommendation, and explicitly forbidding any form of peer enforcement. I'd like to keep the 2 weeks ball park figure, even though it's not enforced.

Well, you need to know how often to rebroadcast your channel.  Let's say
you SHOULD do that every two weeks.  Then say that you MAY prune > 2
weeks?

Cheers,
Rusty.

cdecker:
That's how it is currently formulated. My point is merely that it should not be allowed to punish someone for relaying older information. We also cannot prevent nodes from forgetting earlier, hence the 2 weeks threshold.

Issue #187, opened 2017-06-14
Remove public keys from `channel_announcement` message


pm47:
Instead, use key recovery based on the signatures (like in #183) to save space.

We would save 4*33 - 1 byte (4 * two "recovery bits") = 131 bytes.

cdecker:
Can we squash the recovery bits into a single byte? That way we would have the nice 128 byte alignment :wink:

rustyrussell:
Defer to 1.1, but @cdecker points out we should move features before keys in case we want to eliminate it later.

cdecker:
Feature bitmap has been moved, giving us the flexibility once we decide that we want to do this.

pm47:
Will switch to Schnorr instead.

PR #192, opened 2017-06-27
BOLT7: Reorder feature bitmaps in order to allow future changes

cdecker:
Appending new fields to the end of the messages allows us to add new
fields to an existing message, however it does not allow removing
existing fields, e.g., dropping the pubkeys like #187 proposes. Moving
the features bitmap at the beginning of the signed payload allows
this type of change in the future. Nodes verify the integrity of the
message and then check whether there are any even bits they don't
implement. These even bits being required features would then result
in the message being discarded.

In addition to what we discussed during the call I also went ahead and
did the same reordering on `node_announcement`, which I think has the
same issue.

There is a subtle change in semantics, i.e., previously we would
add channels with unknown bits to our local view, but then ignore them
when computing a route. Now we no longer add them to our view, and may
discard the announcement altogether, stopping the broadcast. This is
safe I think since otherwise we'd be forwarding things we can only
verify the signatures of, but nothing else.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -80,13 +80,13 @@ announcement message; that is done by having a signature from each
     * [`64`:`node_signature_2`]
     * [`64`:`bitcoin_signature_1`]
     * [`64`:`bitcoin_signature_2`]
+    * [`2`:`len`]
+    * [`len`:`features`]
     * [`8`:`short_channel_id`]
     * [`33`:`node_id_1`]
     * [`33`:`node_id_2`]
     * [`33`:`bitcoin_key_1`]
     * [`33`:`bitcoin_key_2`]
-    * [`2`:`len`]
-    * [`len`:`features`]
 
 ### Requirements
 
@@ -108,6 +108,9 @@ Thus the hash skips the 4 signatures, but hashes the rest of the message, includ
 
 The creating node SHOULD set `len` to the minimum length required to
 hold the `features` bits it sets.
+
+The receiving node MUST verify the integrity and authenticity of the message by verifying the signatures.
+If there is an unknown even bit in the `features` field the receiving node MUST NOT parse the remainder of the message and MUST NOT add the channel to its local network view, and SHOULD NOT forward the announcement.
 
 The receiving node MUST ignore the message if the output specified
 by `short_channel_id` does not
@@ -132,9 +135,6 @@ previous message's `node_id_1` and `node_id_2` as well as this
 `node_id_1` and `node_id_2` and forget channels connected to them,
 otherwise it SHOULD store this `channel_announcement`.
 
-The receiving node SHOULD NOT route through a channel which has an
-unknown `features` bit set which is even.
-
 The receiving node SHOULD forget a channel once its funding output has
 been spent or reorganized out.
 
@@ -158,8 +158,8 @@ rebroadcasting (perhaps statistically).
 
 New channel features are possible in future; backwards compatible (or
 optional) ones will have odd feature bits, incompatible ones will have
-even feature bits (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide).  These will be propagated by nodes even if they
-can't use the announcements themselves.
+even feature bits (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide).
+Incompatible features will result in the announcement not being forwarded by nodes that don't understand them.
 
 ## The `node_announcement` message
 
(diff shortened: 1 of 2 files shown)


cdecker:
And it also breaks the alignment police since the features break the 8 byte alignment...

Alternatively we could just byte the bullet and add a version number or only allow these kinds of changes with new message types...

07-routing-gossip.md
```diff
@@ -158,8 +160,8 @@ rebroadcasting (perhaps statistically).
 
 New channel features are possible in future; backwards compatible (or
 optional) ones will have odd feature bits, incompatible ones will have
-even feature bits (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide).  These will be propagated by nodes even if they
-can't use the announcements themselves.
+even feature bits (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide).
+Incompatible features will result in the announcement not being forwarder by nodes that don't understand them.
```
rustyrussell: typo: s/forwarder/forwarded/

07-routing-gossip.md
```diff
@@ -109,6 +109,11 @@ Thus the hash skips the 4 signatures, but hashes the rest of the message, includ
 The creating node SHOULD set `len` to the minimum length required to
 hold the `features` bits it sets.
 
+The receiving node MUST verify the integrity and authenticity of the message by verifying the signatures.
+The receiving node MUST check that the `features` field does not contain any unknown even bits.
```
rustyrussell: I try to avoid saying "MUST check", because it doesn't say what to do if the check fails.  In fact, this sentence can be eliminated.

07-routing-gossip.md
```diff
@@ -233,6 +235,10 @@ valid signature using `node_id` of the double-SHA256 of the entire
 message following the `signature` field (including unknown fields
 following `alias`), and MUST NOT further process the message.
 
+The receiving node SHOULD verify that the `features` field does not contain any unknown even bits.
```
rustyrussell: Eliminate this sentence too.

07-routing-gossip.md
```diff
@@ -80,13 +80,13 @@ announcement message; that is done by having a signature from each
     * [`64`:`node_signature_2`]
     * [`64`:`bitcoin_signature_1`]
     * [`64`:`bitcoin_signature_2`]
+    * [`2`:`len`]
```
Roasbeef: As pointed out in my top-level review comment, this assumes that channels are still using the regular 2-of-2 construction. In my mind, this change should be a first step that allows us to be more flexible w.r.t the way channel validation (on both the signature and output checking level) is performed. 

For example, it's possible to (today w/o any further changes to Bitcoin) create funding output that's actually just a regular `p2wkh` output. Both sides are able to in a distributed fashion: conjointly generate multiplicative shares of a private key, and then conduct a 2-party protocol to sign arbitrary messages under the full public key. 

In order to allow things like the above (and other much more exotic constructions), then feature bits should instead be moved to lie either right _below_ the node signatures, or at the very _top_ of the channel announcement message. Moving it to the very top will also let us signal changes to the signature algorithm used in the future. 

Two such possible signature upgrades include: schnorr signatures (to let us aggregate sigs of the same message and save cycles using batch validation _across_ diff messages) and even further pairing based signatures (which would let us aggregate signatures across _batches_ of messages. Both of these signatures schemes could be adopted immediately, the only lagging factor being the maturity of implementations available to the various lightning daemon implementations.

07-routing-gossip.md
```diff
@@ -170,12 +172,12 @@ nodes for which a channel is not already known are ignored.
 1. type: 257 (`node_announcement`)
 2. data:
    * [`64`:`signature`]
+   * [`2`:`flen`]
+   * [`flen`:`features`]
```
Roasbeef: Similar comment here about moving the feature bits to the _very top_ of the message.

07-routing-gossip.md
cdecker: The problem with moving the features above the signatures, which indeed would allow us to swap out the signatures as well, was that then we'd need to have two serializations: one to create the signature and one for the on-wire transmission. We could either do that by skipping the signatures while signing, or mandating that the signatures should be zero-filled. I think both options are rather ugly.

If we really want to switch the messages this much we should probably just define a new message type altogether.

07-routing-gossip.md
cdecker: Similar objection :wink:

cdecker:
> Also, why wasn't the `channel_update` message modified as well?

We don't have a feature bitmap in `channel_update`, but we could use the `flags` field, which is limited to 8 bits, one of which is already assigned though.

cdecker:
> We should disable the alignment check, at least where a variable-len field has been found.

I disabled the alignment check if the offset is not a simple integer, this is usually an indication that we have a variable length field.

07-routing-gossip.md
Roasbeef: Fair point. We have a good bit of space in the type bytes that we can use to add additional messages in the future. In a sense, I guess it's also a bit more explicit to add in a new message type (keeping in the even/odd semantics) vs modifying some feature bits and creating a new message entirely.

Roasbeef:
Scratch the suggestion about doing the same to the `channel_update` message. I now see that the original rationale for this change doesn't apply there 😆.

rustyrussell:
@cdecker Pushed fixup which addresses my original wording issues complaints and typo.  If seems OK, please merge...

cdecker:
LGTM, I rebased and merged the fixups before merging with `master`

PR #193, opened 2017-06-27
BOLT7: Allow channel_updates for non-public channels

### LINKED ISSUE #188: Minimal routing sync feature bit
Opening a non-public channel without initial routing sync seems to be an optimal option for a bandwidth and computation constrained Lightning node primairly focused on payments rather than payment routing (a lightweight mobile wallet for example). Let's suppose such a minimal node `A` is connected to a public node `B` via a non-public channel, this implies that node `A` needs to add a `B -> A` routing tag to it's payment requests which means it needs to have an access to `B`'s latest `fee` and `cltv_expiry_delta`.

Currently `localfeatures` allows for all or nothing initial routing sync options but what if a feature bit could be added which would instruct a node `B` to only send an info sufficient for `A` to add a routing tag to it's payment requests and nothing else?

cdecker:
This was pointed out by @btcontract in #188: we need to communicate
our forwarding parameters even for private channels since otherwise
the other endpoint cannot use the private channel for incoming
routes. So we also accept `channel_update`s for our own channels even
for channels that were not announced publicly. Adds a bit of special
handling for our own channels in the gossip, but it is needed since
private channels would be completely unusable otherwise.

Fixes #188

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -276,7 +276,7 @@ in the `flags` field
 to indicate which end this is.  It can do this multiple times, if
 it wants to change fees.
 
-A node MAY still create a `channel_update` to communicate the channel parameters to the other endpoint, even though the channel has not been announced, e.g., because the `channels_public` bit was not set.
+A node MAY still create a `channel_update` to communicate the channel parameters to the other endpoint, even though the channel has not been announced, e.g., because the `announce_channel` bit was not set.
 For further privacy such a `channel_update` MUST NOT be forwarded to other peers.
 Note that such a `channel_update` that is not preceded by a `channel_announcement` is invalid to any other peer and would be discarded.
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -276,6 +276,10 @@ in the `flags` field
 to indicate which end this is.  It can do this multiple times, if
 it wants to change fees.
 
+A node MAY still create a `channel_update` to communicate the channel parameters to the other endpoint, even though the channel has not been announced, e.g., because the `channels_public` bit was not set.
+For further privacy such a `channel_update` MUST NOT be forwarded to other peers.
+Note that such a `channel_update` that is not preceded by a `channel_announcement` is invalid to any other peer and would be discarded.
+
 1. type: 258 (`channel_update`)
 2. data:
     * [`64`:`signature`]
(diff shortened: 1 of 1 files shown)


btcontract:
Should a receiving node enable initial routing sync if it wants to receive these special `channel_update`s or will a sending node be transmitting them anyway?

07-routing-gossip.md
```diff
@@ -276,6 +276,10 @@ in the `flags` field
 to indicate which end this is.  It can do this multiple times, if
 it wants to change fees.
 
+A node MAY still create a `channel_update` to communicate the channel parameters to the other endpoint, even though the channel has not been announced, e.g., because the `channels_public` bit was not set.
```
rustyrussell: MUST.  Otherwise, the poor node is unable to receive payments :(

rustyrussell:
@btcontract Yes, as written.

For post 1.0, we'll have a more sophisticated filtering mechanism for route updates, but for the moment it's all or nothing.  We could make an exception here, but I'd prefer to get the filtering mechanism in place instead....

btcontract:
@rustyrussell my worry is that handling these initial syncs may become infeasible very fast for a lite client. Guess it's irrelevant if improved update mechanisms are expected to come soon after 1.0 but if not then having such an exception would be nice.

cdecker:
The initial sync only relates to the initial dump of old announcements, not the updates that come in while the connection is up and running. This `channel_update` being generated after the connection establishment means that we should send it in either case, as it is an online update, not a batch update of older updates.

07-routing-gossip.md
cdecker: Well, it's only in that direction, isn't it? If the node is a deadend it can also just not send the `channel_update` since the other side cannot possibly route a payment through it :wink:

07-routing-gossip.md
```diff
@@ -310,11 +314,10 @@ The creating node MUST set `timestamp` to greater than zero, and MUST set it to
 
 It MUST set `cltv_expiry_delta` to the number of blocks it will subtract from an incoming HTLCs `cltv_expiry`.  It MUST set `htlc_minimum_msat` to the minimum HTLC value it will accept, in millisatoshi.  It MUST set `fee_base_msat` to the base fee it will charge for any HTLC, in millisatoshi, and `fee_proportional_millionths` to the amount it will charge per millionth of a satoshi.
 
-The receiving node MUST ignore `flags` other than the least significant bit.
-It SHOULD ignore the message if `short_channel_id` does
-not correspond to a previously
-known, unspent channel from `channel_announcement`, otherwise the `node_id`
-is taken from the `channel_announcement`: `node_id_1` if least-significant bit of flags is 0 or `node_id_2` otherwise.
+The receiving nodes MUST ignore the `channel_update` if it does not correspond to one of its own channels, if it the `short_channel_id` does not match a previous `channel_announcement`, or the channel has been closed in the meantime.
```
Roasbeef: if it the -> if the

rustyrussell:
Force-pushed rebase and typo fix.

PR #194, opened 2017-06-28
BOLT 2: closing tx fee clarifications

rustyrussell:
1. You can't eliminate an output and also guarantee a certain fee, so
    we need to define exactly how to do this.

    Since the output is (presumably) dust, we might as well just discard it
    (effectively increasing the fee).  This avoids the peer directly benefiting
   from the elimination as well.

2. There are two final commitment transactions, which can be different if each side has a different
    dust level.  Make it clear that the max fee must be less or equal to the *base* fee, not the actual
    (post-output-elimination) fee.  This means it could be marginally higher than the final fee in theory,
    but in practice it's a tiny effect at worst.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -415,8 +415,8 @@ estimate of cost of inclusion in a block.
 
 The sender MUST set `signature` to the Bitcoin signature of the close
 transaction with the node responsible for paying the bitcoin fee
-paying `fee_satoshis`, without populating any output which is below
-its own `dust_limit_satoshis`. The sender MAY also eliminate its own
+paying `fee_satoshis`, then removing any output which is below
+its own `dust_limit_satoshis`. The sender MAY then also eliminate its own
 output from the mutual close transaction.
 
 The receiver MUST check `signature` is valid for either the close
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -408,7 +408,7 @@ Nodes SHOULD send a `closing_signed` message after `shutdown` has
 been received and no HTLCs remain in either commitment transaction.
 
 A sending node MUST set `fee_satoshis` lower than or equal to the
-fee of the final commitment transaction.
+base fee of the final commitment transaction as calculated in [BOLT #3](03-transactions.md#fee-calculation).
 
 The sender SHOULD set the initial `fee_satoshis` according to its
 estimate of cost of inclusion in a block.


02-peer-protocol.md
```diff
@@ -408,15 +408,15 @@ Nodes SHOULD send a `closing_signed` message after `shutdown` has
...
 
 The sender SHOULD set the initial `fee_satoshis` according to its
 estimate of cost of inclusion in a block.
 
 The sender MUST set `signature` to the Bitcoin signature of the close
 transaction with the node responsible for paying the bitcoin fee
-paying `fee_satoshis`, without populating any output which is below
-its own `dust_limit_satoshis`. The sender MAY also eliminate its own
+paying `fee_satoshis`, then removing any output which is below
+its own `dust_limit_satoshis`. The sender MAY then also eliminate its own
```
pm47: My brain hurts... :persevere:

Doesn't this invalidate the second either clause of following paragraph?

rustyrussell:
>
>
>
>>  
>  The sender SHOULD set the initial `fee_satoshis` according to its
>  estimate of cost of inclusion in a block.
>  
>  The sender MUST set `signature` to the Bitcoin signature of the close
>  transaction with the node responsible for paying the bitcoin fee
> -paying `fee_satoshis`, without populating any output which is below
> -its own `dust_limit_satoshis`. The sender MAY also eliminate its own
> +paying `fee_satoshis`, then removing any output which is below
> +its own `dust_limit_satoshis`. The sender MAY then also eliminate its own
>
> My brain hurts... :persevere:
>
> Doesn't this invalidate the second either clause of following paragraph?

No, this just clarifies that we calculate fee then remove outputs, not
remove outputs then calculate fee.  We can still disagree with the other
side on what outputs should be removed...

We should probably move the construction of the closing tx to BOLT #3
with the other transactionsm even though it's pretty trivial.  This was
a lesser change, however.

Hope that helps,
Rusty.

pm47:
As I understand it, we are now explicitely saying that the signature sent should have outputs above the remote's `dust_limit_satoshis`. Hence when the receiver wants to verify it, it only has to check against its own `dust_limit_satoshis` right?

I agree that this should be done in BOLT 3

02-peer-protocol.md
halseth: To me this reads as "create signature for transaction before removing output" ... "then remove the output". 

Wouldn't it be better to add something similar to your comment, @rustyrussell: "we calculate fee then remove outputs, not remove outputs then calculate fee." That is much less ambiguous :)

Roasbeef: Agreed that this paragraph is pretty confusing. Here's a slightly re-worked version that I think we'll all find mouc more explicit: 

> The sender MUST set `signature` to the Bitcoin signature of the channel
closure transaction. The set `fee_satoshi` fee should be subtracted from
the output of the node that was responsible for paying the bitcoin fees
during the lifetime of the channel. When calculating the amount of
satoshis to subtract from the fee payer's output the absolute fee should
be calculated first, then any trimmed outputs removed. Each trimmed output
should contribute to (reduce the) the fee value explicitly subtracted from
the fee payer's output. The sender MAY also eliminate its own output 
from the mutual close transaction.

rustyrussell: @Roasbeef Hmm, that's backwards from what I meant.  Trimming dust doesn't change inputs, so yeah, it effectively ups the fee paid.  This is the same logic we use for commit txs.

But as agreed, will merge, and then the move to BOLT 3, and re-open another PR for clarification if we still need it...

PR #197, opened 2017-07-02
BOLT #2: Make onion packet vary size

AndrewSamokhvalov:
My apology if I have missed the context of the conversation, but it seems for me that in order to update the sphinx onion packet version in the future we have to have onion packet size decoupled from the `update_add_htlc` message, so that we could make smooth transitions, by simultaneously maintaining several types of the onion packet.


rustyrussell:
Yes, the fixed size is deliberate.  We'd add a new `update_htlc_add2` message if we needed to change the onion representation (either size or interpretation).

PR #199, opened 2017-07-03
Shutdown retransmission clarifications

rustyrussell:
Fairly self-explanatory, I think.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -992,9 +992,9 @@ transaction is broadcast by the other side at any time.  This is
 particularly important if a node does not simply retransmit the exact
 same `update_` messages as previously sent.
 
-On reconnection if the node has sent a previous `shutdown` it MUST
-retransmit it, and if the node has sent a previous `closing_signed` it
-MUST then retransmit the last `closing_signed`.
+On reconnection if the node has sent a previous `closing_signed` it
+MUST then retransmit the last `closing_signed`, otherwise if the node
+has sent a previous `shutdown` it MUST retransmit it.
 
 ### Rationale
 
@@ -1011,8 +1011,8 @@ polite to retransmit before disconnecting again, but it's not a MUST
 because there are also occasions where a node can simply forget the
 channel altogether.
 
-There is similarly no acknowledgment for `closing_signed`, or
-`shutdown`, so they are also retransmitted on reconnection.
+`closing_signed` can only be sent after `shutdown`, so only one needs
+to be retransmitted on reconnection.
 
 The handling of updates is similarly atomic: if the commit is not
 acknowledged (or wasn't sent) the updates are re-sent.  However, we
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -980,7 +980,7 @@ last `commitment_signed` message the receiving node has sent, it
 SHOULD fail the channel.
 
 If `next_remote_revocation_number` is equal to the commitment number of
-the last `revoke_and_ack` the receiving node has sent, it MUST re-send
+the last `revoke_and_ack` the receiving node has sent and the receiving node has not already received a `closing_signed`, it MUST re-send
 the `revoke_and_ack`, otherwise if `next_remote_revocation_number` is not
 equal to one greater than the commitment number of the last `revoke_and_ack` the
 receiving node has sent (or equal to zero if none have been sent), it SHOULD fail the channel.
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -980,7 +980,7 @@ last `commitment_signed` message the receiving node has sent, it
 SHOULD fail the channel.
 
 If `next_remote_revocation_number` is equal to the commitment number of
-the last `revoke_and_ack` the receiving node has sent, it MUST re-send
+the last `revoke_and_ack` the receiving node has sent and the receiving node has not already received a `closing_signed`, it MUST re-send
```
pm47: As you state in the rationale below, the sender won't send a `closing_signed` until it has sent the last `revoke_and_ack`.

Thus I don't see how this scenario is possible?

rustyrussell: Yes, exactly.  It *shouldn't* happen, thus you don't need to handle it.

pm47: Should we mention it though? There are an infinity of other scenarii that shouldn't happen, that we don't specify. Why is this specific one more important than the others? I might be missing something but to me it isn't really different from a node sending, say, two `funding_locked` in a row. That sounds silly I know but my concern is consistency.

How about stating somewhere else more generally what a node should do when the counterparty violates the order of messages of the protocol, eg. always fail the channel? I think that would be a reasonable thing, since AFAICT we were careful to specify when to ignore messages.

rustyrussell:
>
>
>
>> @@ -980,7 +980,7 @@ last `commitment_signed` message the receiving node has sent, it
>  SHOULD fail the channel.
>  
>  If `next_remote_revocation_number` is equal to the commitment number of
> -the last `revoke_and_ack` the receiving node has sent, it MUST re-send
> +the last `revoke_and_ack` the receiving node has sent and the receiving node has not already received a `closing_signed`, it MUST re-send
>
> Should we mention it though? There are an infinity of other scenarii that shouldn't happen, that we don't specify. Why is this specific one more important than the others? I might be missing something but to me it isn't really different from a node sending, say, two `funding_locked` in a row. That sounds silly I know but my concern is consistency.

We should be specifying the requirements in *every* case, not relying on
the reader to know that things shouldn't happen.

Here, we say you MUST re-send, but my implementation won't if we're
already in closing_signed, which I think is reasonable, but violated the
spec before this change.

> How about stating somewhere else more generally what a node should do when the counterparty violates the order of messages of the protocol, eg. always fail the channel? I think that would be a reasonable thing, since AFAICT we were careful to specify when to ignore messages.

I'd prefer each requirement to be spelled out in each case.  eg. you
should receive X or Y, if you receive anything else for that channel,
you SHOULD fail the channel.

That's much easier to implement, and test!

We should fix places which don't do this...

Thanks,
Rusty.

rustyrussell:
> Roasbeef requested changes on this pull request.
>
>
>
>> @@ -992,9 +992,9 @@ transaction is broadcast by the other side at any time.  This is
>  particularly important if a node does not simply retransmit the exact
>  same `update_` messages as previously sent.
>  
> -On reconnection if the node has sent a previous `shutdown` it MUST
> -retransmit it, and if the node has sent a previous `closing_signed` it
> -MUST then retransmit the last `closing_signed`.
> +On reconnection if the node has sent a previous `closing_signed` it
>
> Why do we even require sending an _identical_ `closing_signed` message? I don't see a reason to make nodes worry about an _additional_ layer of persistence. 
>
> It seems that upon re-connect, the closing negotiation should just start from scratch (similar to a failed funding negotiation). It also could be the case the the fee preferences of either side have changed significantly since the last connection. 

Agreed.  Let me create a patch.

Should we mark the closing tx RBF?  I don't expect any implementation to
do this any time soon, but it would allow us to use a prior closing tx
with higher fee if the current one gets stuck (or even, the final
commitment tx)?

02-peer-protocol.md
```diff
@@ -992,9 +992,9 @@ transaction is broadcast by the other side at any time.  This is
 particularly important if a node does not simply retransmit the exact
 same `update_` messages as previously sent.
 
-On reconnection if the node has sent a previous `shutdown` it MUST
-retransmit it, and if the node has sent a previous `closing_signed` it
-MUST then retransmit the last `closing_signed`.
+On reconnection if the node has sent a previous `closing_signed` it
```
rustyrussell: Done, added new patch.

PR #200, opened 2017-07-03
BOLT 11: Clarify that providing a payment description or description hash is mandatory

pm47:
The spec is a little ambiguous because it currently says "A writer MUST include either a `d` or `h` field", like if we must have one or the other.

Some test vectors have neither a `d` nor a `h` so I guess your intent was to keep it optional @rustyrussell ?


rustyrussell:
> The spec is a little ambiguous because it currently says "A writer MUST include either a `d` or `h` field", like if we must have one or the other.
>
> Some test vectors have neither a `d` nor a `h` so I guess your intent was to keep it optional @rustyrussell ?
> You can view, comment on, or merge this pull request online at:

Yes, we should fix the test vectors!

I don't think an optional payment description is a good idea; there
has to be some description of what the payment is for.  Of course,
we can't actually make it a *good* description, but the payer should
be able to count on there being one!

pm47:
Having a description/decription-hash mandatory makes sense, that was also the opinion of @cdecker.

I tried to rephrase the requirements a little bit, and updated the test vectors accordingly

11-payment-encoding.md
```diff
@@ -228,7 +228,7 @@ https://github.com/rustyrussell/lightning-payencode
 # Examples
 
 > ### Please make a donation of any amount using payment_hash 0001020304050607080900010203040506070809000102030405060708090102 to me @03e7156ae33b0a208d0744199163177e909e80176e55d97a2f221ede0f934dd9ad
-> lnbc1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqq7fshvguvjs864g4yj47aedw4y402hdl9g2tqqhyed3nuhr7c908g6uhq9llj7w3s58k3sej3tcg4weqxrxmp3cwxuvy9kfr0uzy8jgpy6uzal
+> lnbc1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdpl2pkx2ctnv5sxxmmwwd5kgetjypeh2ursdae8g6twvus8g6rfwvs8qun0dfjkxaq32vjcgqxyuj7nqphl3xmmhls2rkl3t97uan4j0xa87gj5779czc8p0z58zf5wpt9ggem6adl64cvawcxlef9djqwp2jzzfvs272504sp0lkg3c
```
rustyrussell: Hmm, I get a different signature: lnbc1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdpl2pkx2ctnv5sxxmmwwd5kgetjypeh2ursdae8g6twvus8g6rfwvs8qun0dfjkxaq8rkx3yf5tcsyz3d73gafnh3cax9rn449d9p5uxz9ezhhypd0elx87sjle52x86fux2ypatgddc6k63n7erqz25le42c4u4ecky03ylcqca784w

When I decode your examples, the pubkey I recover is 0392e8ececed7f419e5d15f3b2ba4f86e0fc8e5ff533c3445332bc29d36cf0aaed instead of the documented 03e7156ae33b0a208d0744199163177e909e80176e55d97a2f221ede0f934dd9ad.

I've updated https://github.com/rustyrussell/lightning-payencode with these examples (which are otherwise correct) as well as tightening the encoder to enforce the 'd' or 'h' requirement (and many others!), so you should be able to debug there if you have a different result.

pm47: Right, there was a change in how data bits are split that we missed. I'll fix the test vectors tomorrow.

@martindale what's with all the 👍 ❤️ ? :stuck_out_tongue_winking_eye:

pm47:
@rustyrussell that should be better, test vectors now match your implementation

11-payment-encoding.md
```diff
@@ -115,25 +115,19 @@ A writer MUST include exactly one `p` field, and set `payment_hash` to
...
-both.  If included, a writer SHOULD make `d` a complete description of
+A writer MUST include either exactly one `d` or exactly one `h` field.  If included, a 
+writer SHOULD make `d` a complete description of
 the purpose of the payment.  If included, a writer MUST make the preimage
 of the hashed description in `h` available through some unspecified means,
 which SHOULD be a complete description of the purpose of the payment.
 
-A writer SHOULD use the minimum `x` `data_length` possible.
+A writer MAY include an `x` field, which SHOULD use the minimum `data_length` 
+possible.
```
rustyrussell: ... and MUST NOT include more than one.

This requirement got dropped when the first sentence got reworked.

pm47: Well I used the same terminology as for the `n` field, "MAY include an `..` field", if that is not clear we should change both.

How about changing both to "MAY include *one* `..` field", it may be less ambiguous and is consistent with the wording of other paragraphs?

PR #201, opened 2017-07-04
Closing tx detailed specification

rustyrussell:
Spec changes:
1. Insist recipient actually check received fee is valid.
2. Make sender MUST (not SHOULD) send a better offer on every update.
3. Actually specify what closing tx should look like!  And move rest of that detail to BOLT 3.

The "forward progress" check is a bit nastier than it should be due to reconnect, where we have to allow re-transmissions.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -424,14 +424,13 @@ If `fee_satoshis` is equal to its previously sent `fee_satoshis`, the receiver
 SHOULD close the connection and SHOULD sign and broadcast the final closing
 transaction.
 
-Otherwise, if this is the first `closing_signed` received, the
+Otherwise, the
 recipient MUST fail the connection if `fee_satoshis` is greater than
 the base fee of the final commitment transaction as calculated in
-[BOLT #3](03-transactions.md#fee-calculation).  Otherwise the
+[BOLT #3](03-transactions.md#fee-calculation), and the
 recipient SHOULD fail the connection if `fee_satoshis` is not strictly
 between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, unless it is equal to the previously-received
-`fee_satoshis` and it has reconnected since then.
+`fee_satoshis`, unless it has reconnected since then.
 
 If the receiver agrees with the fee, it SHOULD reply with a
 `closing_signed` with the same `fee_satoshis` value, otherwise it
@@ -441,8 +440,9 @@ and its previously-sent `fee_satoshis`.
 #### Rationale
 
 The "strictly between" requirements ensure that we make forward
-progress, even if only by a single satoshi at a time, but we have
-to allow for retransmission on reconnection.
+progress, even if only by a single satoshi at a time.  To avoid
+keeping state and handle the corner case where fees have shifted
+between reconnections, negotiation restarts on reconnection.
 
 Note that there is limited risk if the closing transaction is
 delayed, and it will be broadcast very soon, so there is usually no
@@ -991,7 +991,7 @@ same `update_` messages as previously sent.
 
 On reconnection if the node has sent a previous `shutdown` it MUST
 retransmit it, and if the node has sent a previous `closing_signed` it
-MUST then retransmit the last `closing_signed`.
+MUST send another `closing_signed`.
 
 ### Rationale
 
@@ -1009,7 +1009,9 @@ because there are also occasions where a node can simply forget the
 channel altogether.
 
 There is similarly no acknowledgment for `closing_signed`, or
-`shutdown`, so they are also retransmitted on reconnection.
+`shutdown`, so they are also retransmitted on reconnection.  In the
+case of `closing_signed`, negotation restarts on reconnection, so it
+need not be an exact retransmission.
 
 The handling of updates is similarly atomic: if the commit is not
 acknowledged (or wasn't sent) the updates are re-sent.  However, we
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -414,38 +414,35 @@ The sender SHOULD set the initial `fee_satoshis` according to its
 estimate of cost of inclusion in a block.
 
 The sender MUST set `signature` to the Bitcoin signature of the close
-transaction with the node responsible for paying the bitcoin fee
-paying `fee_satoshis`, then removing any output which is below
-its own `dust_limit_satoshis`. The sender MAY then also eliminate its own
-output from the mutual close transaction.
+transaction as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
-The receiver MUST check `signature` is valid for either the close
-transaction with the given `fee_satoshis` as detailed above and its
-own `dust_limit_satoshis` OR that same transaction with the sender's
-output eliminated, and MUST fail the connection if it is not.
+The receiver MUST check `signature` is valid for either variant of close
+transaction specified in [BOLT #3](03-transactions.md#closing-transaction),
+and MUST fail the connection if it is not.
+
+If `fee_satoshis` is equal to its previously sent `fee_satoshis`, the receiver
+SHOULD close the connection and SHOULD sign and broadcast the final closing
+transaction.
+
+Otherwise, if this is the first `closing_signed` received, the
+recipient MUST fail the connection if `fee_satoshis` is greater than
+the base fee of the final commitment transaction as calculated in
+[BOLT #3](03-transactions.md#fee-calculation).  Otherwise the
+recipient SHOULD fail the connection if `fee_satoshis` is not strictly
+between its last-sent `fee_satoshis` and its previously-received
+`fee_satoshis`, unless it is equal to the previously-received
+`fee_satoshis` and it has reconnected since then.
 
 If the receiver agrees with the fee, it SHOULD reply with a
 `closing_signed` with the same `fee_satoshis` value, otherwise it
-SHOULD propose a value strictly between the received `fee_satoshis`
+MUST propose a value strictly between the received `fee_satoshis`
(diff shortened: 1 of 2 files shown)


03-transactions.md
```diff
@@ -234,6 +235,45 @@ The witness script for the output is:
 
 To spend this via penalty, the remote node uses a witness stack `<revocationsig> 1` and to collect the output the local node uses an input with nSequence `to_self_delay` and a witness stack `<local_delayedsig> 0`
 
+## Closing Transaction
+
+Note that there are two possible variants for each node.
+
+* version: 1
```
pm47: Isn't that version 2?

02-peer-protocol.md
```diff
@@ -414,38 +414,35 @@ The sender SHOULD set the initial `fee_satoshis` according to its
...
+
+If `fee_satoshis` is equal to its previously sent `fee_satoshis`, the receiver
+SHOULD close the connection and SHOULD sign and broadcast the final closing
+transaction.
+
+Otherwise, if this is the first `closing_signed` received, the
+recipient MUST fail the connection if `fee_satoshis` is greater than
+the base fee of the final commitment transaction as calculated in
+[BOLT #3](03-transactions.md#fee-calculation).  Otherwise the
+recipient SHOULD fail the connection if `fee_satoshis` is not strictly
```
pm47: I think this doesn't work in case of a reconnection, where both parties re-send their last `closing_signed`?

03-transactions.md
```diff
@@ -234,6 +235,45 @@ The witness script for the output is:
...
+   * `txin[0]` sequence: 0xFFFFFFFF
+   * `txin[0]` script bytes: 0
+   * `txin[0]` witness: `0 <signature_for_key1> <signature_for_key2>`
+* txout count: 0, 1 or 2
+   * `txout` amount: final balance to be paid to one node (minus `fee_satoshis` from `closing_signed` if this peer funded the channel)
+   * `txout` script: as specified in that node's `scriptpubkey` in its `shutdown` message.
+
+### Requirements
+
+Each node offering MUST subtract the fee given by `fee_satoshis` from
```
pm47: How about:
> Each node offering a signature MUST subtract the fee given by `fee_satoshis` from
the output to the funder; it MUST then remove any output below its own `dust_limit_satoshis`, and MAY also eliminate its own output.

03-transactions.md
```diff
@@ -234,6 +235,45 @@ The witness script for the output is:
...
+side uses its own `dust_limit_satoshis`, and the result can be a
+signature validation failure, if they disagree on what the closing
+transaction should look like.
+
+However, if one side chooses to eliminate its own output, there's no
+reason for the other side to fail the closing protocol, so this is
+explicitly allowed.  The signature will indicate which variant
+has been used.
+
+It is assumed that `dust_limit_satoshis` is greater than the funding
```
pm47: Something is missing here?

03-transactions.md
rustyrussell: @roasbeef said version 1, since version 2 only needed for CLTV, which we don't need here.  I don't really mind...

02-peer-protocol.md
rustyrussell: That's why the sentence ends:
... unless it is equal to the previously-received `fee_satoshis` and it has reconnected since then.

It's a bit awkward, I agree :(

03-transactions.md
rustyrussell: Yes, that's better,

03-transactions.md
rustyrussell: Oops!  yep....

02-peer-protocol.md
```diff
@@ -414,38 +414,35 @@ The sender SHOULD set the initial `fee_satoshis` according to its
...
+between its last-sent `fee_satoshis` and its previously-received
+`fee_satoshis`, unless it is equal to the previously-received
+`fee_satoshis` and it has reconnected since then.
```
Roasbeef: In line with my comment on #199 (which is related to this PR as well), why are we requiring nodes remember the _prior_ close negotiation state _between_ re-connects. It may be the case that fees have gone down significantly since the prior negotiation, and both nodes would be massively overpaying (or the other way around). If we instead allow nodes to forget any negotiation upon re-connect then this allows them to have more completely control over the fees that they're paying. 

From another angle, it seems that auto-attempt-to-re-close-on-re-connect may not even be a behavior that we desire. Days (even weeks) may pass between re-connects, it may not necessarily be the case that if a user wished to cooperatively close out the channel 3 days ago, that they _still_ wish to do so upon re-connect (whenever that is).

03-transactions.md
Roasbeef: Using version 1 here allow us to blend in a bit more with regular (non contract) traffic (assuming an eventual saturation of transactions spending segwit inputs).

03-transactions.md
```diff
@@ -234,6 +235,44 @@ The witness script for the output is:
...
+Note that there are two possible variants for each node.
+
+* version: 1
+* locktime: 0
+* txin count: 1
+   * `txin[0]` outpoint: `txid` and `output_index` from `funding_created` message
+   * `txin[0]` sequence: 0xFFFFFFFF
+   * `txin[0]` script bytes: 0
+   * `txin[0]` witness: `0 <signature_for_key1> <signature_for_key2>`
+* txout count: 0, 1 or 2
```
Roasbeef: In what case would having _no_ transaction outputs be warranted or valid?

03-transactions.md
```diff
@@ -234,6 +235,44 @@ The witness script for the output is:
...
+   * `txout` amount: final balance to be paid to one node (minus `fee_satoshis` from `closing_signed` if this peer funded the channel)
+   * `txout` script: as specified in that node's `scriptpubkey` in its `shutdown` message.
+
```
Roasbeef: BIP 69 should be used here as well. I'm assuming y'all already do so?

pm47: yes

rustyrussell:
> Roasbeef requested changes on this pull request.
>
>
>
>> @@ -234,6 +235,45 @@ The witness script for the output is:
>  
>  To spend this via penalty, the remote node uses a witness stack `<revocationsig> 1` and to collect the output the local node uses an input with nSequence `to_self_delay` and a witness stack `<local_delayedsig> 0`
>  
> +## Closing Transaction
> +
> +Note that there are two possible variants for each node.
> +
> +* version: 2
>
> Why a version of 2? We're just spending a plain multi-sig. In the future we can use more "covert" outputs with functionality equivalent to the multi-sig output. With such outputs (and if this is just a version 1), then channel opening and closing transactions gain a greater anonymity set, as they look like regular bitcoin transactions. 

Agreed, it was fixed in the fixup to v1.

>> +The receiver MUST check `signature` is valid for either variant of close
> +transaction specified in [BOLT #3](03-transactions.md#closing-transaction),
> +and MUST fail the connection if it is not.
> +
> +If `fee_satoshis` is equal to its previously sent `fee_satoshis`, the receiver
> +SHOULD close the connection and SHOULD sign and broadcast the final closing
> +transaction.
> +
> +Otherwise, if this is the first `closing_signed` received, the
> +recipient MUST fail the connection if `fee_satoshis` is greater than
> +the base fee of the final commitment transaction as calculated in
> +[BOLT #3](03-transactions.md#fee-calculation).  Otherwise the
> +recipient SHOULD fail the connection if `fee_satoshis` is not strictly
> +between its last-sent `fee_satoshis` and its previously-received
> +`fee_satoshis`, unless it is equal to the previously-received
> +`fee_satoshis` and it has reconnected since then.
>
> In line with my comment on #199 (which is related to this PR as well), why are we requiring nodes remember the _prior_ close negotiation state _between_ re-connects. It may be the case that fees have gone down significantly since the prior negotiation, and both nodes would be massively overpaying (or the other way around). If we instead allow nodes to forget any negotiation upon re-connect then this allows them to have more completely control over the fees that they're paying. 

Hmm, I actually like this, if only because it's stateless.  If there's
consensus on this, let's loosen the spec.

>>From another angle, it seems that auto-attempt-to-re-close-on-re-connect may not even be a behavior that we desire. Days (even weeks) may pass between re-connects, it may not necessarily be the case that if a user wished to cooperatively close out the channel 3 days ago, that they _still_ wish to do so upon re-connect (whenever that is). 

Someone wants to close, they're probably not going to wait 3 days for
it.  You can't *unoffer*, since the close isn't revocable.

> +Note that there are two possible variants for each node.
> +
> +* version: 1
> +* locktime: 0
> +* txin count: 1
> +   * `txin[0]` outpoint: `txid` and `output_index` from `funding_created` message
> +   * `txin[0]` sequence: 0xFFFFFFFF
> +   * `txin[0]` script bytes: 0
> +   * `txin[0]` witness: `0 <signature_for_key1> <signature_for_key2>`
> +* txout count: 0, 1 or 2
>
> In what case would having _no_ transaction outputs be warranted or valid? 

It's theoretically possible if your dust limits and fees are large
enough.  I'd put it under *won't happen*, except if it does, it's better
than leaving unspent UTXOs.

> +   * `txin[0]` sequence: 0xFFFFFFFF
> +   * `txin[0]` script bytes: 0
> +   * `txin[0]` witness: `0 <signature_for_key1> <signature_for_key2>`
> +* txout count: 0, 1 or 2
> +   * `txout` amount: final balance to be paid to one node (minus `fee_satoshis` from `closing_signed` if this peer funded the channel)
> +   * `txout` script: as specified in that node's `scriptpubkey` in its `shutdown` message.
> +
>
> BIP 69 should be used here as well. I'm assuming y'all already do so?

Yep, and I just checked: that's specified in the top of BOLT #3 for all
transactions.

PR #203, opened 2017-07-10
BOLT 7: add chain_hash to channel_update and channel_announcement

Roasbeef:
This commit adds a 32-byte `chain_hash` value to both the
`channel_update` and `channel_announcement` messages. The rationale for
this change is that this value is already present within the
`open_channel` for identifying _which_ chain to open the channel
within. As is now, if a pair of peers had channels open on two chains
which somehow are encoded using the same `short_channel_id`, then the
announcements would be ambitious. We resolve this by explicitly
including the `chain_hash` is all channel related announcement
messages.

Note that with this change, we now require 40-bytes to uniquely
identify a channel globally.

Additionally, this modification of the channel announcement messages
allows peers to start building up a heterogenous network graph.

### DIFF
--- a/00-introduction.md
+++ b/00-introduction.md
@@ -110,6 +110,17 @@ This is version 0.
      MUST support the feature in question, while odd numbers indicate
      that the feature MAY be disregarded by the other endpoint.
 
+
+* `chain_hash`:
+   * Used in several of the BOLT documents, and denotes the genesis hash of a
+     target blockchain. This allows nodes to create and reference channels on
+     several blockchains. Nodes are to ignore any messages which reference a
+     `chain_hash` that are unknown to them.
+
+     For the main chain Bitcoin blockchain, the `chain_hash` value MUST be
+     (encoded in hex):
+     `000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f`.
+
 ## Theme Song
 
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -82,6 +82,7 @@ announcement message; that is done by having a signature from each
     * [`64`:`bitcoin_signature_2`]
     * [`2`:`len`]
     * [`len`:`features`]
+    * [`32`:`chain_hash`]
     * [`8`:`short_channel_id`]
     * [`33`:`node_id_1`]
     * [`33`:`node_id_2`]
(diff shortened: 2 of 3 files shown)


07-routing-gossip.md
```diff
@@ -300,6 +310,11 @@ The following table specifies the meaning of the individual bits:
 The creating node MUST set `signature` to the signature of the
 double-SHA256 of the entire remaining packet after `signature` using its own `node_id`.
 
+The creating node MUST set `chain_hash` to match 32-byte hash that
+uniquely identifies the channel within the `channel_announcement` message.
```
cdecker: I guess this should be the tuple (`chain_hash`, `short_channel_id`) which uniquely identifies the channel.

07-routing-gossip.md
```diff
@@ -300,6 +310,11 @@ The following table specifies the meaning of the individual bits:
...
+For the Bitcoin blockchain, the `chain_hash` value MUST be (encoded in
+hex): `000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f`.
```
cdecker: We should specify the hash once and reference it thereon.

Roasbeef: Good point. Where does everyone think this definition should go? In the introduction, perhaps? 

I ended up sticking it in the glossary, but there may be a better spot for it.

07-routing-gossip.md
Roasbeef: Modified to point out that channels are uniquely identified by that two tuple.

00-introduction.md
```diff
@@ -110,6 +110,17 @@ This is version 0.
...
      that the feature MAY be disregarded by the other endpoint.
 
+
+* `chain_hash`:
+   * Used in several of the BOLT documents, and denotes the genesis hash of a
+     target blockchain. This allows nodes to create and reference channels on
+     several blockchains. Nodes are to ignore any messages which reference a
+     `chain_hash` that are unknown to them.
+
+     For the mahin chain Bitcoin blockchain, the `chain_hash` value MUST be
```
pm47: typo mahin -> main

rustyrussell:
> +* `chain_hash`:
> +   * Used in several of the BOLT documents, and denotes the genesis hash of a
> +     target blockchain. This allows nodes to create and reference channels on
> +     several blockchains. Nodes are to ignore any messages which reference a
> +     `chain_hash` that are unknown to them.
> +
> +     For the mahin chain Bitcoin blockchain, the `chain_hash` value MUST be
>
> typo mahin -> main

He's obviously using the American pronunciation :)

PR #198, opened 2017-07-02
Script for extracting structured protocol messages

ysangkok:
This changes extract-formats.py so that other scripts can use it, but retains normal functionality.
The new script (structured.py) parses the CSV variant and shows a representation of an OrderedMap.
This could be used to write parsers.

PR #205, opened 2017-07-11
Script for extracting structured protocol messages (replaces #198)

cdecker:
This PR replaces #198 since I was unable to rebase and push onto the original branch (Allow edits from maintainers is off or pushing to someone else's `master` is not allowed).

It includes the structured format from @ysangkok and I took the liberty of cleaning both scripts up a bit.


cdecker:
I checked that the CSV format output is identical to the version before, so this should not break anything

cdecker:
Merged as discussed during the meeting on the 2017/07/10.

PR #208, opened 2017-07-11
BOLT7: Add shadow route extension in the recommendations

### LINKED ISSUE #185: Protect against last hop detection via timelocks
Currently the recipient can be guessed by the second to last hop due to the short timelock on the HTLC, i.e., it is not plausible that this is just an intermediate hop since the timelock cannot further be reduced. We discussed adding a random offset to the CLTV timelocks in order to avoid this sort of detection. This can further be strengthened by computing an appendix/shadow route to the actual route of random length, and offsetting all CLTVs by the sum of CLTV values. It is still unclear whether the sender or the recipient should be the one computing the shadow route.

cdecker:
This is a recommendation to fuzz the CLTV on the HTLCs such that nodes
along the route have a harder time identifying the intended
recipient. We can either add a random offset or we can start a random
walk from the intended recipient and create a shadow route extension.

Closes #185

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -205,6 +205,8 @@ blindGroupElement
 tx
 pubkeys
 CHECKLOCKTIMEVERIFY
+CLTV
+CLTVs
 generateCipherStream
 XK
 lflen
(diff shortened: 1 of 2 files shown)


cdecker:
Rebased and merged as per meeting notes.

PR #211, opened 2017-07-21
BOLT09: Further specify how bits are specified.

cdecker:
Switched from an ambiguous representation as _bits_ to a bitmask that
is used to check for presence. This was a misunderstanding that Eclair
and c-lightning had. Eclair was interpreting bits=3 as setting the 3rd
bit, whereas c-lightning was interpreting bits=3 as the bitmask to
check for the bits that may be set to indicate mandatory and optional
support.

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -288,3 +288,4 @@ lollypop
 UTC
 inline
 fundee's
+bitmask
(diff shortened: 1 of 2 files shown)


cdecker:
@sstone sorry for picking our variant as default, but it seems kind of nice being able to express both mandatory and optional fields with just one entry, also we reset to use the first bits :-)

PR #213, opened 2017-07-28
Correct channel_id length to be consistent with other structure membe…

grisha:
…rs, whose length is expressed in bytes, not bits.


grisha:
Nevermind, channel_id is indeed 32 bytes, somehow I didn't catch that, sorry!

PR #214, opened 2017-07-29
Closing tx rebase

rustyrussell:
This is the rebase and rollup of the changes in #199 and #201 which got horribly inter-twined.

1. Change descriptions of closing tx construction to references to BOLT 3. 
2. Recipient *should* check the fee offer has improved in closing_signed.
3. Therefore, sender *must* improve closing offer.
4. Offers do not persist across reconnection, so no state req'd, and
    also helps if fee has changed.
5. You don't need to re-send `shutdown` if you received `closing_signed`
    (implicit acknowledgement).
6. You don't have to accept a `channel_reestablish` which requests the last
    revoke_and_ack be retransmitted if you've already received `closing_signed`
    (which is an implicit acknowledgement).

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -414,38 +414,34 @@ The sender SHOULD set the initial `fee_satoshis` according to its
 estimate of cost of inclusion in a block.
 
 The sender MUST set `signature` to the Bitcoin signature of the close
-transaction with the node responsible for paying the bitcoin fee
-paying `fee_satoshis`, then removing any output which is below
-its own `dust_limit_satoshis`. The sender MAY then also eliminate its own
-output from the mutual close transaction.
+transaction as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
-The receiver MUST check `signature` is valid for either the close
-transaction with the given `fee_satoshis` as detailed above and its
-own `dust_limit_satoshis` OR that same transaction with the sender's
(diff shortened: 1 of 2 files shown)


02-peer-protocol.md
```diff
@@ -414,38 +414,35 @@ The sender SHOULD set the initial `fee_satoshis` according to its
...
-The receiver MUST check `signature` is valid for either the close
-transaction with the given `fee_satoshis` as detailed above and its
-own `dust_limit_satoshis` OR that same transaction with the sender's
-output eliminated, and MUST fail the connection if it is not.
+The receiver MUST check `signature` is valid for either variant of close
+transaction specified in [BOLT #3](03-transactions.md#closing-transaction),
+and MUST fail the connection if it is not.
+
+If `fee_satoshis` is equal to its previously sent `fee_satoshis`, the receiver
+SHOULD close the connection and SHOULD sign and broadcast the final closing
```
Roasbeef: From my reading, this is the normal case in which both sides eventually decide on a valid fee for the cooperative closure transaction. 

In this case, why should the connection be closed all together? It's possible that either side wishes to open a new channel, or that the parties had more than one channel amongst each other. Additionally, this may have been Alice's only connection within the network, therefor if the connection is closed upon channel closure, then she won't receive any graph updates,

PR #217, opened 2017-08-06
Reword proportional fee explanation.

kek-coin:
The old wording was the wrong way around - implying one would charge millions of satoshis per transferred satoshi.


rustyrussell:
Applied, thanks for this!

Cheers,
Rusty.

PR #218, opened 2017-08-07
Add a feature bit to allow nodes to opt-out of channel updates

Roasbeef:
This PR adds a new section to the 7th BOLT document detailing a
mechanism that nodes can use to opt-out of dynamic channel updates.
Doing so is primarily a bandwidth optimization, as the node is able to
only receive announcements for new channels and nodes joining the
network, rather than announcements for each channel update within the
network.

Additionally, we propose the additional ability of a node to signal that 
they don’t wish to receive any dynamic channel updates using an 
unused feature bit in the init message.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -423,6 +423,25 @@ Other more advanced considerations involve diversity of routes to
 avoid single points of failure and detection, and channel balance
 of local channels.
 
+## Opting out of `channel_update` messages
+
+A peer may wish to update out of receiving any `channel_update` messages as a
+bandwidth and processing optimization. Note that opting out of `channel_update`
+messages doesn't entirely affect a node's ability to interact with the network.
+In the case of a node that wishes to send semi-frequent payments, if a payment
+fails due to using an outdated routing policy for a particular channel link,
+then the error response will contain the latest `channel_update` message for
+that link. For nodes which are primarily focused on payment forwarding, the
+extra traffic of `channel_update` messages are unnecessary.
+
+In order to opt-out of `channel_update` messages the `no_channel_updates` flag
+should be set within the `init` message that the node advertises to other peers
+upon initial connection. 
+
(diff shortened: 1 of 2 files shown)


rustyrussell:
Main points from discussion:
1. This is a premature optimization.
2. It's not something we want to encourage, as it's antisocial (such a node isn't useful for other nodes).
3. We know we have to revisit syncing in 1.1 anyway ( @roasbeef hates the term "dump"! )

So deferring until 1.1...

cdecker:
I see the need to optimize bandwidth, however, if we go down this road, I'd like to broaden the bit to suppress all gossip messages, not just the updates. This could be used by slow/low-bandwidth nodes to designate a single node as source for its gossip data (by turning the bit off for that node) and suppress all other nodes.

@rustyrussell this concerns the gossip background noise during operation, not the dump-on-connect, for which we already have bit 3.

07-routing-gossip.md
```diff
@@ -423,6 +423,25 @@ Other more advanced considerations involve diversity of routes to
...
+then the error response will contain the latest `channel_update` message for
+that link. For nodes which are primarily focused on payment forwarding, the
+extra traffic of `channel_update` messages are unnecessary.
+
+In order to opt-out of `channel_update` messages the `no_channel_updates` flag
+should be set within the `init` message that the node advertises to other peers
+upon initial connection. 
+
+If a node receives an `init` message with the `no_channel_updates` flag set,
+then the receiving not MUST NOT relay any channel update messages to the
```
jonathancross: Miner typo: "...then the receiving **_node_** MUST NOT relay any..."

t-bast:
Can we close this PR now that we have `GossipTimestampFilters` that achieve this (unless I'm missing something)?

ysangkok:
@roasbeef, is this PR superseded by #641 ?

PR #224, opened 2017-08-21
Onchain htlc clarifications

rustyrussell:
This is what happens when you actually implement on-chain HTLC handling; I found some gaps in the spec.  No real changes here, though.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -512,10 +512,12 @@ A node MUST NOT offer an HTLC (`update_add_htlc`) in response to an incoming HTL
 the incoming HTLC has been irrevocably committed.
 
 A node MUST NOT fail an incoming HTLC (`update_fail_htlc`) for which it has committed 
-to an outgoing HTLC, until the removal of the outgoing HTLC is irrevocably committed.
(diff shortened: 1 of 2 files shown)


02-peer-protocol.md
```diff
@@ -512,10 +512,12 @@ A node MUST NOT offer an HTLC (`update_add_htlc`) in response to an incoming HTL
 the incoming HTLC has been irrevocably committed.
 
 A node MUST NOT fail an incoming HTLC (`update_fail_htlc`) for which it has committed 
-to an outgoing HTLC, until the removal of the outgoing HTLC is irrevocably committed.
+to an outgoing HTLC, until the removal of the outgoing HTLC is irrevocably committed, or the outgoing on-chain HTLC output has been spent with sufficient depth.
```
Roasbeef: Why would a node fail an incoming HTLC if the outgoing HTLC they send has been claimed on-chain? At that point (as soon as the pre-image is revealed), they can send a fulfill message backwards in the route to fully complete the payment.

rustyrussell: "on-chain HTLC has been spent via the HTLC-timeout transaction"

PR #229, opened 2017-08-22
BOLT 5: clarifies penalty txn weight calculation

cfromknecht:
Attempts to clarify the weight calculation of penalty transactions, and makes sweeping the `to_remote` output optional without breaking any existing constraints. Assuming these figures are correct, the decision to sweep the `to_remote` _does not_ change the current unidirectional limit of 483 HTLCs.  Thus, the option to do so can be made independently by either party/implementation. I've currently specified this as MAY, but we could introduce a stronger modifier if we think that's appropriate.

The previous equation used to calculate `max_num_htlcs` slightly underestimated the theoretical maximum weight, since non-witness data was treated as 1:1 with witness data.  Ultimately, this had no effect on the computed results, but figured we should be more specific here for the purpose of properly estimating transaction fees.

This commit also modifies the `to_local_script` to use the latest construction; the derived weights have been updated accordingly.

Thanks in advance for any help with review!

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -374,29 +374,39 @@ should cover this. [FIXME: May have to divide and conquer here, since they may b
 
 There are three different scripts for penalty transactions, with the following witnesses weight (details of the computation in [Appendix A](#appendix-a-expected-weights)):
 
-    to_local_penalty_witness: 154 bytes
+    to_local_penalty_witness: 160 bytes
     offered_htlc_penalty_witness: 243 bytes
     accepted_htlc_penalty_witness: 249 bytes
 
 The penalty txinput itself takes 41 bytes, thus has a weight of 164, making the weight of each input:
 
-    to_local_penalty_input_weight: 318 bytes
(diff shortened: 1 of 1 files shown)


05-onchain.md
```diff
@@ -374,29 +374,39 @@ should cover this. [FIXME: May have to divide and conquer here, since they may b
...
  
-    max_num_htlcs = (400000 - 318 - 55) / 413  = 967
+    max_num_htlcs = (400000 - 324 - 272 - 4*53 - 2) / 413 = 966
  
-Thus we could allow 483 HTLCs in each direction (with one `to_local` output) and still resolve it with a single penalty
-transaction.
+Thus we could allow 483 HTLCs in each direction (with both `to_local` and
+`to_remote` outputs) and still resolve it with a single penalty transaction.
+Note that even if the `to_remote` output is not swept, the resulting
+`max_num_htlcs` is 967, which yields the same unidirectional limit of 483 HTLCs.
```
junderw: 967 >>> 966

PR #231, opened 2017-08-25
BOLT 04: update test  vectors

sstone:
maximum error payload size is now 256 bytes, see #227


rustyrussell:
You mean 256?

sstone:
Ah yes sorry I'll edit the comment above

PR #233, opened 2017-08-31
BOLT 5: requirement to fail HTLCs which don't have outputs in the commit tx

rustyrussell:
BOLT 5 only talks in terms out HTLC outputs, but not all HTLCs have outputs.

HTLCs which are dust for both sides are easy, but others require the
commit tx to be buried before we can consider the HTLC failed.

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -208,6 +208,9 @@ B's *commitment transaction*, hence the requirement to handle both.
 Each HTLC output can only be spent by us after it's timed out,
 or them if they have the payment preimage.
 
+There can be HTLCs which are not represented by an output: either
+because they were trimmed as dust, or in the case where B has two
+valid commitment transactions, and the HTLCs differ in each.
 
 The HTLC has *timed out* once the depth of the latest block is equal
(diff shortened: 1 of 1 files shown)


05-onchain.md
```diff
@@ -208,6 +208,9 @@ B's *commitment transaction*, hence the requirement to handle both.
 Each HTLC output can only be spent by us after it's timed out,
 or them if they have the payment preimage.
 
+There can be HTLCs which are not represented by an output: either
+because they were trimmed as dust, or in the case where B has two
+valid commitment transactions, and the HTLCs differ in each.
```
pm47: > or in the case where B has two valid commitment transactions, and the HTLCs differ in each

Just to be sure: You mean when an htlc has already been fulfilled/failed, we just sent a new `commitment_signed`, and didn't yet received a `revoke_and_ack`. The counterparty has then two valid unrevoked tx, one with the htlc, one without.

rustyrussell: Exactly, and also it can happen the other way.  You could have added a new HTLC, sent commitment_signed and not received revoke_and_ack...

Issue #209, opened 2017-07-15
Include current `perCommitmentPoint` in `ChannelReestablish`


btcontract:
Current local channel state may be completely lost at any moment, this is most relevant to mobile devices (destruction, loss, etc) but can happen on any setup. It would be nice to have a way to claim at least some of the remaining channel funds in such a situation.

Having current `perCommitmentPoint` included in `ChannelReestablish` would allow for the following scheme:

- save static channel parameters once a new channel is created (on a backup server for example, linked to a hash of BIP 32 master key so user can retrieve it later using a mnemonic code)
- retrieve this info on a new device after an old one is lost
- reestablish a connection, wait for incoming `ChannelReestablish`, then send and `Error` and wait for a current commit to be published, then use `perCommitmentPoint` and saved static channel parameters to claim a P2WPKH output.

HTLC outputs can't be claimed but this may be okay in a majority of cases as mobile device is much more likely to get lost when it's in a steady state (has no in-flight HTLCs) anyway.

ysangkok:
Why not just store the current `per_commitment_point` on that server? If the spec starts catering to malfunctioning clients (surely it is malfunctioning if it loses the current `per_commitment_point`, right?), where do we stop? And how should a node react if it observes an incorrect `per_commitment_point`, having NOT forgotten the correct `per_commitment_point` ifself?

btcontract:
> Why not just store the current per_commitment_point on that server? 

it will require an additional server call on each commit (which occasionally may not be recorded due to network conditions and so on). Also, unlike "revoked tx watchdog" server case, this will require some form of accounting i.e. which _last_ `per_commitment_point` belongs to whom, this is bad for privacy. 

> And how should a node react if it observes an incorrect per_commitment_point, having NOT forgotten the correct per_commitment_point ifself

my guess it client should close a channel unilaterally in this case to discourage sending outdated or irrelevant points. 

> If the spec starts catering to malfunctioning clients (surely it is malfunctioning if it loses the current per_commitment_point, right?), where do we stop?

spec catering to malfunctioning clients does not seem bad to me. LN on mobile devices would be nearly unusable otherwise.

rustyrussell:
OK, so you would send your *own* perCommitmentPoint.  I think this would be sufficient for non-HTLC recovery, but I'd need to double-check.  I don't hate the idea, however.

Roasbeef:
[...]

PR #240, opened 2017-09-06
Send secret on reestablish

rustyrussell:
Our first optional field.  I didn't bother specifying a feature bit, but we could.

After some consideration, I think this is better than the original #209 proposal.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -50,6 +50,7 @@ The messages are grouped logically into 4 groups by their most significant set b
 The size of the message is required to fit into a 2 byte unsigned int by the transport layer, therefore the maximum possible size is 65535 bytes.
 A node MUST ignore any additional data within a message, beyond the length it expects for that type.
 A node MUST fail the channels if it receives a known message with insufficient length for the contents.
+A node which understands an option (each is numbered) MUST include all the fields annotated with that option.
 
 ### Rationale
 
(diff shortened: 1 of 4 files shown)


btcontract:
Sorry, I don't get it. When `A` connects after data loss and sends an `Error`, `B` (ideally) spends it's current local commit tx so `A` should know `B`'s current `remotePerCommitmentPoint` to be able to claim it's P2WPKH output. 

But as currently proposed `B` only sends `A`'s own last per commitment secret in `ChannelReestablish`, how can it help with `A`'s claiming task?

rustyrussell:
Anton Kumaigorodski <notifications@github.com> writes:
> Sorry, I don't get it. When `A` connects after data loss and sends an `Error`, `B` (ideally) spends it's current local commit tx so `A` should know `B`'s current `remotePerCommitmentPoint` to be able to claim it's P2WPKH output. 
>
> But as currently proposed `B` only sends `A`'s own last per commitment secret in `ChannelReestablish`, how can it help with `A`'s claiming task?

Indeed, this was pointed out to me on the call.

We need to send our own per-commitment-point here too.  I have not got
around to fixing the pull request, I'll do it now.

Thanks!
Rusty.

rustyrussell:
Approved, assigned to @Roasbeef as per call: https://docs.google.com/document/d/1NFvEC704RxRu_XIaYhXWgcttYXVCpIV3EmTslEO4qnY/edit?usp=sharing

PR #243, opened 2017-09-18
BOLT2, BOLT3: reduce attack surface

sstone:
add specific htlc payment keys, let node specify their final scriptpubkey in open/accept

The goal of this PR is to reduce the attack surface for LN nodes by minimizing
the set of private keys that they need to have access to. Currently, nodes
need to know the funding private key and the payment key because it is used
for their main output as well as their HTLC outputs. Using separate keys, only
HTLC outputs (which are typically much smaller than the main output) would be
at risk.

This PR does not address attacks where the attacker is the remote party and
can just steal the funding privatekey, or publish a revoke commit tx...

It also does not prevent attacks where the attacker sends HTLCs to
their own nodes (but there are strategies to mitigate this)

In more details:

Letting nodes specify their final scriptpubkey in their open/accept
messages mitigates an attack where an attacker could trick a node into
closing a channel with their scriptpubkey, in which case they could
spend the closing tx right away.

Having separate keys for pending HTLCs means that nodes don't
need to know the `tolocal` private key and mitigates an attack
where the attacker would steal the `tolocal` private key, have the
channel close unilaterally (by sending garbage, make the node unresponsive
...) and then spend the `toremote` output in the commitment tx published by
the remote party.


tools/extract-formats.py
```diff
@@ -32,6 +32,9 @@ def guess_alignment(message, name, sizestr):
     if 'signature' in name:
         return 1
 
+    if message == 'open_channel' and name == 'len':
+        return 1
+
```
rustyrussell: We should just drop the alignment checks altogether, if this is needed.

rustyrussell:
Closing, as this was split into compulsory (htlckey, #272) and optional (scriptpubkey lockin, #266) both of which have now been merged.

PR #247, opened 2017-09-22
Multiple fixes to BOLT 3

jimpo:
1. Fix pseudcode for shachain algorithm: when deriving a secret from a seed and index, the bits of the index must be iterated in reverse order, as indicated in [reference code](https://github.com/ElementsProject/lightning/blob/master/ccan/ccan/crypto/shachain/shachain.c#L55).
2. Fix appendix numbers and TOC links.
3. Remove extra 0x01 byte from private keys in test vectors: an 0x01 byte is appended to the end of private keys in the test vectors to mark them as using compressed serialization to derive the pubkeys. Two of the private keys have two 0x01 bytes appended, presumably by accident.
4. Correct txid of funding transaction in test vector: the txid has the wrong endianness since Bitcoin reverses the byte order of double-SHA256 hashes.


03-transactions.md
```diff
@@ -472,7 +472,7 @@ The I'th secret P MUST match the output of this algorithm:
 
     generate_from_seed(seed, I):
         P = seed
-        for B in 0 to 47:
+        for B in 47 down to 0:
```
rustyrussell: Yes, indeed.  This is because the index is the negative of the commitment number, which we (um) flip-flopped on.  Good catch!

03-transactions.md
```diff
@@ -802,7 +802,7 @@ The funding transaction is paid to the following keys:
 The funding transaction has a single input, and a change output (order
 determined by BIP69 in this case):
 
-    input txid: adbb20ea41a8423ea937e76e8151636bf6093b70eaff942930d20576600521fd
+    input txid: fd2105607605d2302994ffea703b09f66b6351816ee737a93e42a841ea20bbad
```
rustyrussell: Indeed, the other txid in BOLT 3 is also reversed.  Annoying as it is, we should be consistent with bitcoind. Gah!

PR #252, opened 2017-10-02
BOLT #2: Send cltv_expiry_delta values on open/accept for channel.

rustyrussell:
See https://github.com/cdecker/lightning-integration/issues/3

There's a mistake in the spec, where we were using the outgoing side of
a channel to control the CTLV delta.  But it's the receipient which is
vulnerable if it's too low, so the recipient should set it.

This exchanges values at channel open, and relies on the counterparty
to advertize it correctly in its `channel_update` messages.

There's another patch which changes the "Risks With HTLC Timeouts"
section to cover the setting of cltv_expiry_delta in detail, but
that's not ready yet.


cdecker:
I'm still a bit confused on how a node would verify that this delta was respected, since it does not have the previous expiry. The only thing it can check it, if he is the last hop, the expiry is not smaller than `current_blockheight + min_expiry_delta`, so this is only verifiable on the last hop in any case. Notice that for someone to abuse this we'd need the origin to build an onion that specifies an incorrect delta and the node applying that delta does not enforce that the delta is correct, so not a huge problem :wink:

pm47:
NACK, I think I understand part of your point and the rationale behind it, but I still disagree on the conclusion. 

First I don't think the spec currently says that we should be "using the outgoing side of a channel to control the CTLV delta". To me it says that we should be using the outgoing side of the *next* channel, which means that the recipient *does* currently set it:
> the number of blocks it will subtract from an incoming HTLCs `cltv_expiry`

This works well except for the last segment in the route, so we do have an initialization problem.

IIUC, your reasoning is as follows:
1) the requirement for the last htlc's minimum `cltv_expiry` is similar to the one for `cltv_expiry_delta`
2) therefore we should extend `cltv_expiry_delta`'s definition to cover both cases
3) this implies changes in the channel opening messages

To me, we should only use `cltv_expiry_delta` for intermediate nodes. To compute the last `cltv_expiry` we should:
- recommend a minimum base value (say 9?)
- use a [shadow route](https://github.com/lightningnetwork/lightning-rfc/pull/208) on top of that

The final node is free to return a `final_expiry_too_soon` if it is not happy!

Cheers,

Pierre

rustyrussell:
> I'm still a bit confused on how a node would verify that this delta was respected, since it does not have the previous expiry.

A node subtracts the number of blocks it wants from the expiry, as it has always done.  I think it works, see #253 example.   I'm coding it up right now to be sure though...

04-onion-routing.md
```diff
@@ -553,7 +553,7 @@ outgoing channel:
    * [`2`:`len`]
    * [`len`:`channel_update`]
 
-If the ctlv-expiry is too near, we tell them the the current channel
+If the cltv-expiry is too near, we tell them the the current channel
```
cdecker: cltv-expiry should be marked up as `cltv_expiry`

rustyrussell:
@pm47 c-lighting currently when routing A->B->C uses the B->C channel_update cltv value (+1 in case a block comes through immediately) for the final CLTV, and the A->B channel_update cltv value for the delta across node B.  This is definitely wrong!

You're suggesting we use 9 for the final CLTV, then B->C channel update for the delta across B?   Which makes sense.

But if we can use 9 on the final hop, why not just use 10 everywhere and eliminate this field for v1?  That would make things simpler, at least...

pm47:
> @pm47 c-lighting currently when routing A->B->C uses the B->C channel_update cltv value (+1 in case a block comes through immediately) for the final CLTV, and the A->B channel_update cltv value for the delta across node B. This is definitely wrong!

Yes that's wrong. I was previously arguing that this in an implementation issue, not a spec issue, but on 2nd thought, it *is* a spec issue since we have a different understanding.

> You're suggesting we use 9 for the final CLTV, then B->C channel update for the delta across B? Which makes sense.

It does, right? ;-)

I think using the B->C channel for the delta across B is the *only* way that makes sense. A weakness of your proposal is that if B has another channel B->D, it can't tell anymore A to use a different `cltv_expiry_delta` when paying D, instead A will use a greater `final_cltv_expiry`, which is not what we want. Now if you consider that being able to require a different `cltv_expiry_delta` depending on next hop isn't useful, then it becomes a node attribute, as opposed to a channel attribute, and we should remove it from the `channel_update` message altogether. 

Note that I am only suggesting using 9 as a base/minimum value for the `final_cltv_expiry`, we don't even have to explicitely require it. To me, conflating `cltv_expiry_delta` and `final_cltv_expiry` is wrong because the *actual* `final_cltv_expiry` will be 1-20 times the value of a typical `cltv_expiry_delta`. The reason why I keep mentioning shadow routes is because I think they make the whole reasoning `cltv_expiry_delta` ~ `final_cltv_expiry` invalid.

I am actually not suggesting anything new, that is just my understanding of the current spec and I think it makes perfect sense.

> But if we can use 9 on the final hop, why not just use 10 everywhere and eliminate this field for v1? That would make things simpler, at least...

We could, but I view it the other way around: if we define `cltv_expiry_delta`=10 then `final_cltv_expiry` should be defined as N * 10, with 1 <= N < 20 chosen at random.

Roasbeef:
> c-lighting currently when routing A->B->C uses the B->C channel_update cltv value (+1 in case a block comes through immediately) for the final CLTV, and the A->B channel_update cltv value for the delta across node B. This is definitely wrong!

Yep, this is wrong. It should be using the C <- B `channel_udpdate` cltv value plus some change for the payload that travels over the `B -> C` route same for the `A -> B` hop. 

> To me, we should only use cltv_expiry_delta for intermediate nodes. To compute the last cltv_expiry we should:
>   * recommend a minimum base value (say 9?)

If we don't allow the destination to specify the _exact_ CLTV expiry that they're expecting (computed as a relative offset from the current height), then if the sender randomly selects a value, we increase the number of routing failures. This is due to the fact that if the destination continually sends a `final_expiry_too_soon`, w/o giving any further hint as to the expected value, then the sender can only continually ratchet the expiry up (or down) until the payment succeeds. 

By allowing the destination to specify their _minimum_ value, then shadow routes work just fine as they'll result in a _higher_ expiry than expected, and we minimize extraneous HTLC routing failures within the network. As was pointed out before, this value is only critical to the _receiver_ as they tell us how much time they need in order to safely handle all on-chain edge cases. 

Specifying a global value that all nodes should use doesn't take into account the dynamic conditions of the chain itself. You CANNOT assume that your transaction will ALWAYS get into blocks N+C, where `N` is the current height and `C` is what your fee estimation concludes will be the block tha includes your transaction. We've all seen in the past chain congestion seemingly come out of no where, and the effect that the occasionally hour delta block has on the confirmation times and fee rates of the subsequent blocks. We should be conservative rather than prescriptive, taking off our rose-tinted testnet chain glasses (though the 30 second blocks the past day or two have been nice!), and factor in the worst case scenarios we've all seen on mainnet.

Roasbeef:
Using a global value everywhere doesn't increase the efficacy of shadow routes as if all the outgoing values are multiples of 10 (from the PoV) of the second-to-last hop, then they know that the payment is destined for that terminal node. This is due to the fact that all the cltv expiry deltas advertised by each nodes won't necessarily be a multiple of 10, so using the shadow route in this manner doesn't provide any plausible deniability. 

Instead, when crafting a shadow route, one should examine the known (or privately known) sub-graph surrounding the terminal node, and use an additional delta s.t it can plausibly be used as a valid timeout yet still extend to the "shadow" destination.

pm47:
> if the sender randomly selects a value, we increase the number of routing failures.

That's not what I said... And I am not even against allowing the destination to specify their minimum value some way or another (why not even in the payment request, that could make sense). What I am saying is that whatever minimum value we announce, the actual value will probably be factors greater due to shadow routes, hence I doubt the scenario you describe would occur anyway.

I am opposed to adding `cltv_expiry_delta` to `open`/`accept` channels, which is what this PR is about. I tried to explain why I think this fixes something that didn't need fixing, and creates at least one problem.

> As was pointed out before, this value is only critical to the receiver as they tell us how much time they need in order to safely handle all on-chain edge cases.

I think no-one is challenging that fact.

> Specifying a global value that all nodes should use doesn't take into account the dynamic conditions of the chain itself.  You CANNOT assume that your transaction will ALWAYS get into blocks N+C, where N is the current height and C is what your fee estimation concludes will be the block tha includes your transaction. 

I am not assuming that (and I'm definitely not wearing rose-tainted glasses :-s). In case of sudden contention the final node is always free to return a `final_expiry_too_soon`.

> Using a global value everywhere doesn't increase the efficacy of shadow routes as if all the outgoing values are multiples of 10 (from the PoV) of the second-to-last hop, then they know that the payment is destined for that terminal node. This is due to the fact that all the cltv expiry deltas advertised by each nodes won't necessarily be a multiple of 10, so using the shadow route in this manner doesn't provide any plausible deniability.

You might want to re-read what I wrote: setting a global `cltv_expiry_delta`=10 was the assumption here. Otherwise indeed it would be silly.

rustyrussell:
> Specifying a global value that all nodes should use doesn't take into account the dynamic conditions of the chain itself. You CANNOT assume that your transaction will ALWAYS get into blocks N+C, where `N` is the current height and `C` is what your fee estimation concludes will be the block tha includes your transaction. We've all seen in the past chain congestion seemingly come out of no where, and the effect that the occasionally hour delta block has on the confirmation times and fee rates of the subsequent blocks. We should be conservative rather than prescriptive, taking off our rose-tinted testnet chain glasses (though the 30 second blocks the past day or two have been nice!), and factor in the worst case scenarios we've all seen on mainnet. 

No, your response to congestion should be to increase fees with your
peer.  You can't alter HTLCs already committed anyway, so increasing
cltv_expiry_delta is the wrong approach AFAICT.

I am coming to believe best off coming up with our best possible
"conservative" value, and making that the constant for
`cltv_expiry_delta` for v1.  Maybe that's 9.  Maybe that's 20.  We can
make it changable in future (though for backwards compat, we can only
really reduce it).

I vote for 10.  More than that and I have to take off my socks.

rustyrussell:
>> You're suggesting we use 9 for the final CLTV, then B->C channel update for the delta across B? Which makes sense.
>
> It does, right? ;-)
>
> I think using the B->C channel for the delta across B is the *only*
> way that makes sense. A weakness of your proposal is that if B has
> another channel B->D, it can't tell anymore A to use a different
> `cltv_expiry_delta` when paying D, instead A will use a greater
> `final_cltv_expiry`, which is not what we want.

It does though.  My proposal has two parts:

1. Peers exchange cltv_expiry_delta on channel estalbishment.
2. Peers advertise each other's cltv_expiry_delta in `channel_update`
   instead of their own.

So, if we have A->B, B->C, B->D, if A wants to pay D it looks at B->D
channel_update (contains the `cltv_expiry_delta` D chose), and A->B
(contains the `cltv_expiry_delta` B chose).

> Note that I am only suggesting using 9 as a base/minimum value for the
> `final_cltv_expiry`, we don't even have to explicitely require it. To
> me, conflating `cltv_expiry_delta` and `final_cltv_expiry` is wrong
> because the *actual* `final_cltv_expiry` will be 1-20 times the value
> of a typical `cltv_expiry_delta`.

Well, specifying a minimum still makes sense, and using channel_update
which already exists makes sense.

> The reason why I keep mentioning shadow routes is because I think they make the whole reasoning `cltv_expiry_delta` ~ `final_cltv_expiry` invalid.
>
> I am actually not suggesting anything new, that is just my understanding of the current spec and I think it makes perfect sense.

Oh yes, it makes sense, *if* we add a minimum requirement for the final
hop.  But if we know a sensible minimum, we should just make that
universal.

>> But if we can use 9 on the final hop, why not just use 10 everywhere and eliminate this field for v1? That would make things simpler, at least...
>
> We could, but I view it the other way around: if we define `cltv_expiry_delta`=10 then `final_cltv_expiry` should be defined as N * 10, with 1 <= N < 20 chosen at random.

If we allow actual deltas >= cltv_expiry_delta, then it would might not
always be a multiple of 10, if the sender chose to add some extra delta.

Ignoring that, yes, it becomes trivial.  10 per hop, plus a 10*N "shadow
route".

pm47:
> So, if we have A->B, B->C, B->D, if A wants to pay D it looks at B->D channel_update (contains the `cltv_expiry_delta` D chose), and A->B (contains the `cltv_expiry_delta` B chose).

But you are proving my point I think? The `cltv_expiry_delta` (between the outgoing and incoming htlc at node B) will be the same for A->C and A->D: B cannot anymore tell A to use a different `cltv_expiry_delta` depending on the next hop. What am I missing?  :dizzy_face:

Another question I have is: how can nodes update their `cltv_expiry_delta` now they are announced by their counterparties?

rustyrussell:
> >    So, if we have A->B, B->C, B->D, if A wants to pay D it looks at B->D channel_update (contains the cltv_expiry_delta D chose), and A->B (contains the cltv_expiry_delta B chose).

> But you are proving my point I think? The cltv_expiry_delta (between the outgoing and incoming htlc at node B) will be the same for A->B->C and A->B->D: B cannot anymore tell A to use a different cltv_expiry_delta depending on the next hop. What am I missing? :dizzy_face:

Oh yes, true.  But I don't think it makes sense to differentiate like that, though: it's really a node property disguised as a channel property :)

> Another question I have is: how can nodes update their cltv_expiry_delta now that they are announced by their counterparties?

We'd have to add a new msg for that, yes.

Roasbeef:
>  What I am saying is that whatever minimum value we announce, the actual value will probably be factors greater due to shadow routes

Agreed. The value is to be treated as a lower bound. 

> No, your response to congestion should be to increase fees with your
peer. 

This only works up to a certain point... 

Nodes should be able to select a value that works best for them, their fee estimation policy, and their sensitivity to time locks. We can't assume that implementations will always target then _next_ block. In fact, I believe Bitcoin Core has disabled the 1-2 block estimates by default. As we've seen in the past, the widely used fee estimation within the network can be easily manipulated by outliers, causing users to pay exorbitant fees in response to someone's fat finger.

Roasbeef:
>> Another question I have is: how can nodes update their cltv_expiry_delta now that they are announced by their counterparties?

> We'd have to add a new msg for that, yes.

Perhaps this should just be a node attribute after all (as pm47 suggested)?

rustyrussell:
> This only works up to a certain point...
> 
> Nodes should be able to select a value that works best for them, their fee estimation policy, and their sensitivity to time locks. We can't assume that implementations will always target then next block. In fact, I believe Bitcoin Core has disabled the 1-2 block estimates by default. As we've seen in the past, the widely used fee estimation within the network can be easily manipulated by outliers, causing users to pay exorbitant fees in response to someone's fat finger.

Sure.  But I don't think an algorithm exists to get this dance right, and leaving it to implementations just means everyone will get it wrong.  We're better off picking a conservative value for cltv delta and using it everywhere for v1, then getting clever in the next spec version.

pm47:
What do you think about putting an optional `min_final_cltv_expiry` field in the payment request? It may sound strange at first, but we are already putting optional routing information there right?

That would solve everything, no need to add a new field to `open`/`accept` and no new message for updating the value. And we keep the "specify-a-delta-per-channel" feature (which usefulness is admittedly debatable).

Roasbeef:
> What do you think about putting an optional min_final_cltv_expiry field in the payment request? It may sound strange at first, but we are already putting optional routing information there right?

Seems kinda of hackish....if it isn't provided, then what does the sender do? Also I'm not sure we want to tie information critical for routing through the network to the current payment request format, which will certainly change before we do any major overhauls w.r.t payment routing. 

Circling back on your earlier suggestion: perhaps we should just lift the CLTV expiry into the `NodeAnnouncement` message (as from my PoV this conversation has made clear to me that this is really a node level parameter). I can't think of a case where a node would actually desire having _distinct_ expiry deltas on different channels. This would also achieve the original intent of this change: allow one to use the CLTV delta of the _outgoing_ node when crafting a route. 

However, this _still_ may require adding a new field to `open/accept` so nodes can send single-hop payments, or use the channel itself before the 6 confirmation window has come to pass. 

Seems like the only option that _doesn't_ require adding new fields to `open/accept` is opting to use a global value...which I have qualms about due to the uncertainty of confirmation times in response to fee estimates

pm47:
> Seems kinda of hackish....if it isn't provided, then what does the sender do?

It is hackish yes. If not provided, then the sender could use a default "conservative global value"?

I'm not a fan of putting it in the `NodeAnnouncement` for the reasons you state. Besides, if we pay a node which only has unannounced channels, in addition to the private route info we should provide the `NodeAnnouncement` in... the payment request ;-)

rustyrussell:
This conversation has convinced me that simplicity wins; there are no *great* solutions and the ones we have are complex and hard to understand.

However, if I can't convince you all of a single global value of 15 for v1, I think the second simplest is a 'c' tag in BOLT11 which specifies the final minimum HTLC delta, defaulting to 9 (?).

Roasbeef:
Assuming the `c` tag in BOLT11, this would still mean that one can't route over a channel unless _both_ edges have been advertised (if we agree that we should be using the expiry of the outgoing (to) edge)?

rustyrussell:
No.  A->B->C, you'd use the c= value for the final CLTV, and the B->C `update_channel` for the delta across B.

pm47:
> However, if I can't convince you all of a single global value of 15 for v1, I think the second simplest is a 'c' tag in BOLT11 which specifies the final minimum HTLC delta, defaulting to 9 (?).

I'm slightly in favor of the 'c' tag. The more I think about it, the more I find it consistent with other aspects of the spec. Eg if you look at BOLT 4 errors, `unknown_payment_hash` (15) / `incorrect_payment_amount` (16) / `final_expiry_too_soon` (17) would all be related to information available only in the payment request.

But I wouldn't oppose settling on a global value of 15.

pm47:
Note: I ate my own dog food and submitted #258.

rustyrussell:
@pm47 has convinced us.  I'm closing this, and am rewriting #253 assuming #258.

PR #256, opened 2017-10-10
BOLT 4: `final_incorrect_*` are `PERM` errors

pm47:
The `PERM` flag should be used when the related request will *always* fail if sent again without modification.

That is the case with `final_incorrect_cltv_expiry` and `final_incorrect_htlc_amount`.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -594,14 +594,14 @@ If the `cltv_expiry` is too low, the final node MUST fail the HTLC:
 If the `outgoing_cltv_value` does not match the `cltv_expiry` of the
 HTLC at the final hop:
 
-1. type: 18 (`final_incorrect_cltv_expiry`)
+1. type: PERM|18 (`final_incorrect_cltv_expiry`)
 2. data:
    * [`4`:`cltv_expiry`]
 
 If the `amt_to_forward` is higher than `incoming_htlc_amt` of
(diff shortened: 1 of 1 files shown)


pm47:
Note: this has nothing to do with the discussion in #252, I think `final_expiry_too_soon` remains a temporary error.

rustyrussell:
> The `PERM` flag should be used when the related request will *always* fail if sent again without modification.
>
>
> That is the case with `final_incorrect_cltv_expiry` and `final_incorrect_htlc_amount`.

It's a bit of a corner case, because both these should "never happen",
and they are actually an error in the penultimate node, not the final
node.  We said:

    If node sending the error is the final node:
    
    * If the PERM bit is set, the origin node SHOULD fail the payment,
      otherwise it MAY retry the payment if the error code is understood
      and valid.  (In particular, `final_expiry_too_soon` can occur if the
      block height has changed since sending, `temporary_node_failure`
      could resolve within a few seconds).

Currently, that you *can* retry using a different route.  If we set PERM
you're not supposed to...

pm47:
Right, it can either be an error made by (a) the origin sender when building the onion, or by (b) the penultimate node when building the last htlc I think.

I was focusing on (a) and didn' think about (b). I agree that in case (b) we want to retry. Closing this...

PR #253, opened 2017-10-03
Deriving cltv and example

rustyrussell:
Sooo...many...words...

PR #258, opened 2017-10-10
BOLT 7,11: Added an optional `min_final_cltv_expiry` field in BOLT 11

pm47:
This PR competes with #252.

It clarifies how `channel_update` messages are only to be used in the context of relaying payments, and how both htlc amounts and expiries are to be calculated backwards from the values provided in the payment request.

This means that we don't need the `channel_update` for the first channel in a route, which allows making a payment through a channel which hasn't had any announcements yet.

Note that the clever calculation leading to the default value of 9 for `min_final_cltv_expiry`
has been made in #253.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -282,11 +282,22 @@ padding within `addresses` if they require certain alignment.
 ## The `channel_update` message
 
 After a channel has been initially announced, each side independently
-announces its fees and minimum expiry for HTLCs.  It uses the 8-byte
+announces the fees and minimum expiry delta it requires to relay HTLCs
+through this channel.  It uses the 8-byte
 channel shortid which matches the `channel_announcement` and one bit
 in the `flags` field
 to indicate which end this is.  It can do this multiple times, if
 it wants to change fees.
+
+Note that the `channel_update` message is only useful in the context 
+of *relaying* payments, not *sending* payments. When making a payment
+ `A` -> `B` -> `C` -> `D`, only the `channel_update`s related to channels 
+ `B` -> `C` (announced by `B`) and `C` -> `D` (announced by `C`) will 
+ come into play. When building the route, htlc's amounts and expiries need
+ to be calculated backwards from the destination to the source. The initial
+ exact value for `amount_msat` and minimal value for `htlc_expiry`, which are
+  to be used for the last htlc in the route, are provided in the payment request
+ (see [BOLT #11](11-payment-encoding.md#tagged-fields)).
 
 A node MAY still create a `channel_update` to communicate the channel parameters to the other endpoint, even though the channel has not been announced, e.g., because the `announce_channel` bit was not set.
 For further privacy such a `channel_update` MUST NOT be forwarded to other peers.
@@ -421,10 +432,6 @@ In addition nodes MAY ignore channels with a timestamp older than 2 weeks.
 Notice that this is a node policy and MUST NOT be enforced by peers, e.g., by closing channels when receiving outdated gossip messages.
 
 ## Recommendations for Routing
-
-As the fee is proportional, it must be calculated backwards from the
-destination to the source: only the amount required at the final
-destination is known initially.
 
 When calculating a route for an HTLC, the `cltv_expiry_delta` and the fee both
 need to be considered: the `cltv_expiry_delta` contributes to the time that funds
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -293,7 +293,7 @@ Note that the `channel_update` message is only useful in the context
 of *relaying* payments, not *sending* payments. When making a payment
  `A` -> `B` -> `C` -> `D`, only the `channel_update`s related to channels 
  `B` -> `C` (announced by `B`) and `C` -> `D` (announced by `C`) will 
- come into play. When building the route, htlcs' amounts and expiries need
+ come into play. When building the route, amounts and expiries for htlcs need
  to be calculated backwards from the destination to the source. The initial
  exact value for `amount_msat` and minimal value for `htlc_expiry`, which are
   to be used for the last htlc in the route, are provided in the payment request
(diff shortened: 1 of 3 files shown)


07-routing-gossip.md
```diff
@@ -422,10 +433,6 @@ Notice that this is a node policy and MUST NOT be enforced by peers, e.g., by cl
 
 ## Recommendations for Routing
 
-As the fee is proportional, it must be calculated backwards from the
-destination to the source: only the amount required at the final
```
Roasbeef: Why was this removed?

11-payment-encoding.md
```diff
@@ -162,6 +166,11 @@ A reader MUST skip over unknown fields, an `f` field with unknown
 A reader MUST check that the SHA-2 256 in the `h` field exactly
 matches the hashed description.
 
+A reader MUST use a greater value for the `cltv_expiry` of the last
```
Roasbeef: Feels like this section should instead be placed somewhere in #253 as this section currently defines what implementers should note when _parsing_.

07-routing-gossip.md
Roasbeef: Ah ok I see it was moved above!

11-payment-encoding.md
```diff
@@ -109,6 +109,7 @@ Currently defined Tagged Fields are:
 * `n` (19): `data_length` 53.  The 33-byte public key of the payee node.
 * `h` (23): `data_length` 52.  256-bit description of purpose of payment (SHA256).  This is used to commit to an associated description which is too long to fit, such as may be contained in a web page.
 * `x` (6): `data_length` variable.  `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
+* `c` (24): `data_length` (16 bits, big-endian). Minimum `cltv_expiry` to use for the last htlc. Default is 9 if not specified.
```
Roasbeef: I think the `data_length` section here should be modified to be 4 (4 * 5 = 20, so 4 unused bits).

rustyrussell: Make it variable, like x.  If you can fit it in 5 bits, great.  We should enforce a maximum of 6 (ie. fits in 32 bits) though, that's a separate PR.

rustyrussell: Can we rename it to `final_cltv` or `min_final_cltv` because we already have a `cltv_expiry` field in the onion...

jimpo: Should the data length be specified as a multiple of 5 bits? 10 bits seems like plenty.

Roasbeef: In the `update_channel` messages the delta is expressed using 16 bits. Seem like we should maintain uniformity of the allotted size of "delta" values throughput the spec.

Agreed on using a distinct name.

rustyrussell: Ack...

07-routing-gossip.md
pm47: Yes I thought it made more sense there and it wasn't really a recommendation anyway.

11-payment-encoding.md
pm47: Yes, I chose 16 bits to be consistent with `update_channel` as laolu points out. Will update the name.

11-payment-encoding.md
pm47: Agreed, I removed it

pm47: I also added a section in the rationale and took inspiration in [this comment of yours](https://github.com/lightningnetwork/lightning-rfc/pull/252#issuecomment-335260301)

rustyrussell:
Please change to `data_length` variable like the `x` field above.

And change:

    A writer MAY include one `x` field, which SHOULD use the minimum `data_length`  possible.

To:
    A writer MAY include one `x` field, and MAY include one `c` field.  It SHOULD use  the minimum `data_length`  possible for these fields.

I will add a separate PR which lets the reader reject x and c fields > 60 bits (handling things > 64 bits is completely unnecessary complexity, since they're supposed to be minimum length and 60 bits for either is insane).

pm47:
Well I explained why I didn't make this change, it is because it wouldn't be consistent with how the expiry is handled in the `r` field?

rustyrussell:
Hmm, sorry, I must have missed that :(

In the 'r' field we have other values which are already 8 bit aligned.  I guess this only saves 2 bytes, and I'm not sure about whether `c` will be common; depends on how well 9 works out?  But `x` is already variable, so the logic is trivial...

pm47:
I don't really mind. Let's see what @Roasbeef says? He seemed more in favor of consistency.

Roasbeef:
I'm in favor of the current state, so allowing `x` to be variable size as well (ensuring it's minimally packed). This lets us re-use the existing code already used for the expiry value.

Roasbeef:
Ah, actually as the expiry delta (an values with different names but similar purposes across the spec) are _always_ 16-bits, I'd argue that it should be consistent and use a _fixed_ length 16-bit value. 

FWIW, for now `lnd` will _always_ set it's own `c` value 😉

rustyrussell:
Well, the default value c=9 was basically chosen because that's what  ECLAIR did and it seems reasonable.  If people want a different default, please speak up now! @Roasbeef @pm47

pm47:
Default c=9 is fine with me!

rustyrussell:
Assigned to @pm47 for him to decide whether to commit or revert c varlen patch before merge.

pm47:
I'm keeping it!

rustyrussell:
Soo... I think we have consensus?  If so, please apply, so I can apply #253 (or prompt @Roasbeef for final review...)

pm47:
@sstone told me there wasn't any objection during monday's call so I merged as per our guidelines. 

@Roasbeef please let me know if I misunderstood.

07-routing-gossip.md
zinmar222269-cloud: aed3d28d5a202a0dd7b037fca4c04f58242fa909

PR #259, opened 2017-10-11
Delete unused 06-irc-announcements.md

romanz:


rustyrussell:
Can we just delete this instead?

cdecker:
Happy to delete it as well, nobody is using it anyway :-)

romanz:
Just to be sure, you mean that we can delete the whole [06-irc-announcements.md](https://github.com/lightningnetwork/lightning-rfc/blob/master/06-irc-announcements.md) file?

PR #264, opened 2017-10-17
And lets never mention it again

rustyrussell:
Clean up after #259


rustyrussell:
I'm going to use it provisorially until I google whether it's a word or not.

Issue #261, opened 2017-10-12
Nodes should disallow giant CLTV values when forwarding HTLCs.


rustyrussell:
... for some definition of "giant".  This needs a new WIRE_EXPIRY_TOO_FAR error, plus we need to decide on values. 

Adding a `max_cltv_expiry` to channel_update seems way overkill.  We could leave it implementation specific and say "make it reasonable"?

PR #265, opened 2017-10-30
BOLT 2,4: allow an error for HTLCs which expire too far away.

rustyrussell:


pm47:
I wonder if this error should have the UPDATE flag as well? Even if the `channel_update` message doesn't have information on the maximum allowed expiry, we do provide it in `incorrect_cltv_expiry` error where it is not more helpful.

rustyrussell:
Well, incorrect_cltv_expiry does help, since we might have changed it for the outgoing channel?  I guess we could have a max_cltv field in future updates, though I think this corner case isn't worth the bytes.

pm47:
> Well, incorrect_cltv_expiry does help, since we might have changed it for the outgoing channel?

Oops, you're right!

pm47:
Linking to #261

02-peer-protocol.md
```diff
@@ -519,6 +521,9 @@ In general, we need to complete one side of the exchange before dealing with the
 Fulfilling an HTLC is different: knowledge of the preimage is by definition irrevocable, 
 so we should fulfill the incoming HTLC as soon as we can to reduce latency.
 
+An HTLC with an extremely long expiry is a denial-of-service vector,
+so is not allowed: the exact value of reasonable is currently unclear
```
cdecker: "so it is"

rustyrussell: It's actually correct in British English, but slightly old-fashioned/formal. I often get picked up on it by Americans, for whom it's not idiomatic.  Will fix...

PR #266, opened 2017-10-31
Commit to scriptpubkey as an option

rustyrussell:
The first commit is taken from #240 the rest are new based on #243

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,5 +9,5 @@ python:
   - "3.5"
   - "3.6"
 script:
-  - (set -e; for i in 0?-*.md; do echo "Extracting $i"; python3 tools/extract-formats.py --message-types --message-fields --check-alignment $i; done)
+  - (set -e; for i in 0?-*.md; do echo "Extracting $i"; python3 tools/extract-formats.py --message-types --message-fields $i; done)
   - tools/spellcheck.sh --check [0-9][0-9]-*.md
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -98,7 +98,8 @@ desire to set up a new channel.
    * [`33`:`htlc_basepoint`]
    * [`33`:`first_per_commitment_point`]
    * [`1`:`channel_flags`]
-
+   * [`2`:`shutdown_len`] (`option_upfront_shutdown_script`)
+   * [`shutdown_len`: `shutdown_scriptpubkey`] (`option_upfront_shutdown_script`)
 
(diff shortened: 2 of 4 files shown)


pm47:
I think having the `shutdown_scriptpubkey` as an option makes perfect sense, but I'd rather have the `htlc_basepoint` mandatory, even if that means pushing it to 1.1.

Here are my arguments:
- I think the fact that we use the `payment_basepoint` to spend htlc outputs was a mistake and I can't see any scenario where it has advantages. Therefore I see this more as a bugfix than a new feature.
- This is low-level stuff, that shouldn't be up to implementers IMO, and so far we have always been careful to remove that kind of decision from them.
- Does any one of eclair/lightningd/lnd *not* plan to support this? If we all will implement it, then I think having the field mandatory actually makes it easier.

rustyrussell:
I agree.  But using the even feature bit makes it mandatory, and both bits are assigned: so implementers can choose.

pm47:
By mandatory I meant making it the only possibility (no feature bit). But I know we agreed on no more breaking changes...

rustyrussell:
But 1.1 isn't going to break compatibility, so either it's in 1.0 or it's a feature bit...

pm47:
You are right... Then I'd rather have it in 1.0. :-)

Le 1 nov. 2017 23:04, "Rusty Russell" <notifications@github.com> a écrit :

But 1.1 isn't going to break compatibility, so either it's in 1.0 or it's a
feature bit...

—
You are receiving this because you commented.

Reply to this email directly, view it on GitHub
<https://github.com/lightningnetwork/lightning-rfc/pull/266#issuecomment-341257207>,
or mute the thread

rustyrussell:
As a gesture of professional courtesy, I dropped everything else today and implemented it (hacked in as a new field, no option).  It took a day, and was not too bad (but subdaemon still has the payment secret, but that's an implementation detail).

https://github.com/ElementsProject/lightning/compare/master...rustyrussell:htlckey

I leave it to others to decide whether this is something we break freeze for...

pm47:
> As a gesture of professional courtesy, I dropped everything else today and implemented it

Much obliged! A fair level of reciprocity was in order, hence I too implemented it [here](https://github.com/ACINQ/eclair/commit/95ae0e915b3e1eebb308bbab3e601fae04961d09).

It would seem the burden of making the decision now lies in the hands of @Roasbeef :-)

rustyrussell:
OK, rebased and now only adds scriptpubkey is an option (htlckey is already added to master as a compulsory change).

PR #271, opened 2017-11-13
Added IP Address field for payreqs to help opening channels

junderw:
I like the fallback on-chain address field... but why not also include the IP address and allow the wallet to fallback to attempting a channel open message?

I would much rather have the wallet attempt to route a payment to me and if it didn't work, it should attempt to open a channel with me and push the funds in the opening commit.

There should be a way for the user to send the payment request hash in the openchannel message as well. That way I can connect the payreq I displayed to X user and the openchannel message I received as a fallback when payment routing failed.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -116,6 +116,7 @@ Currently defined Tagged Fields are:
    * `short_channel_id` (64 bits)
    * `fee` (64 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `a` (29): `data_length` 29.  IP (v4/v6) address information encoded in the same manner as the p2p protocol in Bitcoin. First 128 bits are the IP address encoded using  IPv4-mapped IPv6 address format. The last 16 bits are the port number. This can be used to fallback to opening a channel with the payee node. (used with n tag)
 
 ### Requirements
 


Roasbeef:
Why is the IP address needed? The payment request include the destination. With that information, you can lookup if the node has an authenticated `NodeAnnouncement` in you channel graph. If they do, then it _should_ include their current authenticated IPv4/IPv6/onion addresses.

junderw:
The use case I am thinking of is an early Lightning where there is fragmentation and the possibility that a user's wallet can not find our node on the graph at all is much higher... (assuming that if you can't route payments to me there is also a non-zero chance that we are not on each other's channel graphs...)

If wallet client devs will implement this feature (fallback on opening channel when routing payment fail by checking for IP in NodeAnnouncement.) we can close this.

rustyrussell:
Jonathan Underwood <notifications@github.com> writes:
> The use case I am thinking of is an early Lightning where there is fragmentation and the possibility that a user's wallet can not find our node on the graph at all is much higher... (assuming that if you can't route payments to me there is also a non-zero chance that we are not on each other's channel graphs...)

The gossip broadcast is universal, so if we're a node they can connect
to, they'll know about it (or can find out).  Even if they don't have
a channel route to the node.

It's possible in future that we might want to include the gossip
message here, but it shouldn't be necessary for now.

rustyrussell:
I'm closing this: note that for bootstrap, DNS is available too.

PR #273, opened 2017-11-15
BOLT #5: Added permalink as reference for one of FIXMEs

renlord:
added permalink to reference implementation code for coinbase maturity check.


rustyrussell:
Thanks!  Unless someone obejcts, I'll apply this under the "typo/spelling" rule in about 24 hours.

Issue #281, opened 2017-11-27
We do care about their dust limit


rustyrussell:
In case we're relying on them due to dataloss protect

PR #283, opened 2017-11-27
Add a check on counterparty's `dust_limit` value in `open`/`accept`

pm47:
Usually the counterparty would only hurt itself if it chooses too low a `dust_limit`, but in the specific scenario of a data loss, we want the counterparty's commitment tx to be relayed and confirmed on the network so we better check it has a sane value.


pm47:
This fixes #281.

PR #287, opened 2017-11-28
Use remote's `htlc_minimum_msat` in `channel_update`

pm47:
This is a consequence of commit d1fbfd30f8ad1463e4583466f0ff12fc4b582483: we now only use *outgoing* `channel_update` messages to build a route.

Basically the node publishing the `channel_update` says: I can only send htlcs > `htlc_minimum_msat` through this channel.

I think this is consistent with the current definition of the `amount_below_minimum` error (BOLT 4):
> If the HTLC does not reach the current minimum amount, we tell them the amount of the incoming HTLC and the current channel setting for the outgoing channel:


cdecker:
Reopening because it was auto-closed by ElementsProject/lightning#388

Issue #286, opened 2017-11-28
revoked transaction handling


nayuta-ueno:
question : [Revoked Transaction Close Handling - Requirements](https://github.com/lightningnetwork/lightning-rfc/blob/963b10311383f97e2a72f94e867c1111dd8e8e77/05-onchain.md#requirements-6)

```
A node MUST resolve all unresolved outputs as follows:
  3. A's offered HTLCs: The node MUST resolve this in one of three ways by spending:
    - ...
    - ...
    - the HTLC-timeout tx if B publishes them
```

`A's offered HTLCs` is `Received HTLC outputs` at B's commitment transaction.
B publish not `HTLC-success tx` but `HTLC-timeout tx` ?

pm47:
There may be two issues here:
- A's offered and B's offered seem to be inverted
- "the HTLC-timeout tx if B publishes them" in invalid since #105

nayuta-ueno:
Thank you reply !

>    A's offered and B's offered seem to be inverted

Is BOLT5 description invert?
I'm reading BOLT5(revoked transaction) as follow sequence:

```
A           B
|           |
+---------->| A's offer(case 3)
|           |
|<----------+ B's offer(case 4)
|           |
~~~~~~~~~~~~~~
~~~~~~~~~~~~~~
|           | broadcast old commit_tx
|           +---------->
|           |
```

>    "the HTLC-timeout tx if B publishes them" in invalid since #105

oh...
Is all [`Revoked Transaction Close Handling`](https://github.com/lightningnetwork/lightning-rfc/blob/master/05-onchain.md#revoked-transaction-close-handling) old ?

pm47:
>> A's offered and B's offered seem to be inverted

> Is BOLT5 description invert?

"A's offered HTLCs" should be "HTLCs B -> A", and "B's offered HTLCs" should be "HTLCs A -> B". Not being a native speaker I'm not sure if the spec is inverted or I'm just lost in translation.

>Is all Revoked Transaction Close Handling old ?

AFAICT, no. Before #105, we couldn't spend htlc outputs of the commitment tx directly with a revocation, that's why we had to publish the counterparty's HTLC-timeout tx. But it's not the case anymore. Apart from that, I think BOLT 5 is okay.

nayuta-ueno:
Thank you, I think I understood what you wrote.

PR #288, opened 2017-11-28
Removed HTLC-timeout for revoked tx handling

pm47:
Publishing the remote's HTLC-timeout transactions is not needed anymore since #105.

Also, I wonder if "A's offered HTLCs"/"B's offered HTLCs" are inverted (see #286)?


pm47:
OK, I found it redundant with items 5. and 6., but #293 is far better. Closing this

PR #293, opened 2017-11-29
BOLT 5: fix reversed penalty HTLC handling clauses, make it clearer.

rustyrussell:
See #288 and #286 for the confusion these clauses caused.

HTLC outputs can be resolved by:
1. using revocation key,
2. timeout/preimage use if that's possible, OR
3. the cheating party's HTLC-success/HTLC-timeout tx (which we already specify that you have to spend using revocation key).

Hopefully this is now clearer.


dcousens:
@junderw

PR #298, opened 2017-12-01
BOLT3: the closing transaction now uses version 2

Roasbeef:
In this commit, we modify the cooperative closing transaction to use
version 2. Currently eclair and lnd already use version 2, while
c-lightning uses version 1. The commitment transaction already uses
version 2, so making this additional transaction (which spends the
funding output) also use version 2 would be consistent. Additionally,
as a best practice, we should be using the highest currently
defined/use transaction version.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -247,7 +247,7 @@ To spend this via penalty, the remote node uses a witness stack `<revocationsig>
 
 Note that there are two possible variants for each node.
 
-* version: 1
+* version: 2
 * locktime: 0
 * txin count: 1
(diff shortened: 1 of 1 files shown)


hirokuma:
Why current BOLT select version 1 ?
https://github.com/ElementsProject/lightning/issues/311
(Our developing node uses version 1.)

rustyrussell:
Yes, technically this proposal can't be merged for a week, as we make spec decisions on our two-weekly meeting.  However, @cdecker merged the change into c-lightning, so all three implementations now use version 2.  Would recommend updating.

PR #299, opened 2017-12-04
Copy edit bolt4

toadlyBroodle:
-substantial overhaul of BOLT 4 to bring it in line with stylesheet guidelines
-have noted required clarifications with [FIXME: ] tags
-for the sake of clarity, have strictly enforced node terminology for 'origin node', 'final node', 'processing node', 'sending peer', and 'receiving peer'
  -please double check all usages are correct, as it was often opaque which was being referred to in original


rustyrussell:
OK, this being on top of previous changes has deeply confused github.  I've rebased on current master, and pushed it to my own tree as 'pr-299-rebased'.  If you're happy with that, you or I can force push that over this branch.

https://github.com/rustyrussell/lightning-rfc/tree/pr-299-rebased

Generally, when creating new changes you should:
1. Make sure master is up-to-date (ie. git checkout master; git pull).
2. Start a new branch from there (ie. git checkout -b bolt4-copy-edit).
3. After commits, push that branch to github (ie. git push --set-upstream origin bolt3-copy-edit).

If you want to do unrelated cleanups (which don't conflict), it's nicer in its own branch, too.  That splits the reviewing burden as well as any merging burden.

toadlyBroodle:
Ok duly noted. What I did was update my local master branch to the upstream master to keep it current and then rebased copy-edit-bolt4 branch onto updated master. I was hoping this would make it easier to merge, but I guess it did the opposite. Sorry about that.

toadlyBroodle:
And in the future I will also do my best to keep branches more focused so they're easier to merge and review. Thanks for the feedback.

toadlyBroodle:
There are still a number of [FIXME:] tags that I need further guidance on.

04-onion-routing.md
```diff
@@ -630,7 +631,7 @@ The origin node:
   - once the return message has been decrypted:
     - SHOULD store a copy of the message.
     - SHOULD continue decrypting, until the loop has been repeated 20 times.
-    - SHOULD use constant `ammag` and `um` keys to deobfuscate [FIXME: is this correct? was 'obfuscate'] the route length.
+    - SHOULD use constant `ammag` and `um` keys to obfuscate the route length. [FIXME: should this be indented under 'once return message has been decrypted?']
```
rustyrussell: No, this is correct.  If we have a route length 7, we only have 7 ammag and um keys.  But we really need to do the decryption 20 times (to avoid timing attacks which may detect that we finished early), so we use dummy ones.

04-onion-routing.md
```diff
@@ -661,121 +662,141 @@ The top byte of `failure_code` can be read as a set of flags:
...
+report an error:
+1. type: 21 (`expiry_too_far`)
+
+If the channel is disabled, the current channel setting for the outgoing
+channel are reported:
+1. type: UPDATE|20 (`channel_disabled`)
+2. data:
+   * [`2`: `flags`]
+   * [`2`:`len`]
+   * [`len`:`channel_update`]
```
rustyrussell: For future work, I'd restore the text you had, just move the definitions of the structures into its own section, then turn the words describing each one into requirements.  But let's leave for now!

04-onion-routing.md
toadlyBroodle: Ah tricky ok, thanks.

04-onion-routing.md
toadlyBroodle: Ah I get your meaning now, I will tackle this in next PR

toadlyBroodle: Left myself a couple [TODO] tags

rustyrussell:
OK, this addressess all of @shannona feedback in #303 as well, so I'm taking that as second approval.  I'm applying now, knowing there are a few more cleanup rounds to go.

Thanks!

PR #305, opened 2017-12-06
BOLT 5: spell out each HTLC case (as a basis for more proofing)

rustyrussell:
The proof-readers rightly noted how confusing the current treatment of
HTLCs is.  There are four different cases, but I tried to address them
in two separate sections, with conditionals.

This expands it out, separating sections for Our Commitment Tx and
Their Commitment Tx, then subsections for our HTLCs and their HTLCs
in each one.

It means some duplicated requirements and rationales, but it should now
be very clear.

As a side effect, we no longer refer to A and B at all: it's all US and THEM.
This needs further clearing up, but for now makes it clear what *we* need to do
for all cases.

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -10,17 +10,19 @@ There are three ways a channel can end:
 2. The bad way (*unilateral close*): something goes wrong, without necessarily any evil intent on either side (maybe one party crashed, for instance). Anyway, one side publishes its latest *commitment transaction*.
 3. The ugly way (*revoked transaction close*): one of the parties deliberately tries to cheat by publishing an outdated version of its *commitment transaction* (presumably one that was more in her favor).
 
-Because Lightning is designed to be trustless, there is no risk of loss of funds in any of these 3 cases, provided that the situation is properly handled. The goal of this document is to explain exactly how node A should react to seeing any of these on-chain.
+Because Lightning is designed to be trustless, there is no risk of loss of funds in any of these 3 cases, provided that the situation is properly handled. The goal of this document is to explain exactly how a node should react to seeing any of these on-chain.
 
 # Table of Contents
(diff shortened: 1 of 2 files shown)


05-onchain.md
```diff
@@ -2,106 +2,94 @@
 
 ## Abstract
 
-Lightning allows for two parties (A and B) to make transactions off-chain, by both holding a cross-signed *commitment transaction*, which describes the current state of the channel (basically the current balance). This *commitment transaction* is updated every time a new payment is made, and is spendable at all times.
+Lightning allows for two parties (nodes A and B) to conduct transactions off-chain by giving each of them a cross-signed *commitment transaction*, which describes the current state of the channel (basically, the current balance). This *commitment transaction* is updated every time a new payment is made, and is spendable at all times.
```
dcousens: "basically" is probably unnecessary here

rustyrussell: Well, there are in-flight payments, hence the basically..

05-onchain.md
```diff
@@ -424,74 +441,68 @@ A node MUST resolve all unresolved outputs as follows:
...
     to_local_penalty_input_weight: 324 bytes
     offered_htlc_penalty_input_weight: 407 bytes
     accepted_htlc_penalty_input_weight: 413 bytes
 
-
 The rest of the penalty transaction takes 4+1+1+8+1+34+4=53 bytes of non-witness
-data, assuming it has a pay-to-witness-script-hash (the largest standard output
-script), in addition to a 2 byte witness header.
+data, assuming that it has a pay-to-witness-script-hash (the largest standard output
+script), in addition to a 2-byte witness header.
```
dcousens: another nit,  but "2-byte" here,  but "2 bytes" everywhere else?

PR #312, opened 2017-12-08
BOLT#11 Explicitly mention signature removal + Add signature related data in examples

junderw:
This would have saved me a lot of time and pain. Sure, I should have realized that the document says something paradoxical (signature of the data part.... but the data part includes the signature..... so I should have guessed it meant "excluding the signature"...)

Explicit mention of the signature part might help people. Might not. But helps to be explicit.


rustyrussell:
This is *great*; feedback from implementers like this is exactly what we need, as we're usually too close to it to see with fresh eyes.

Thanks!

rustyrussell:
It was noted in the meeting that the change to UTF-8 is technically a narrowing (it's unspecified, at the moment you might have zero bytes in there), but it's also heavily implied by "human readable", so we agreed that's a win.

I'm going to rebase since one example changed, force push to your branch, and then apply...

Issue #309, opened 2017-12-07
Use of Tor hidden service to identify LN nodes?


AmkG:
Currently, by my reading of the spec, it seems that knowledge about "how to connect" to an LN node is shared by sharing IP addresses (IPv6 or IPv4) to peers.

However, it may be useful to also support using a Tor "hidden" service (.onion domain) to identify how to connect to a peer.  My understanding is that Bitcoin can do this: https://en.bitcoin.it/wiki/Setting_up_a_Tor_hidden_service

------

The reason why I ask is because in some countries (e.g. mine), ISPs will give you dynamic IP allocation by default and not a static public IP.  The static public IP is more expensive.

There's also the issue of NATs.  I can't find it now but I came across a post somewhere where someone's wireless ISP had a whopping 12 layers of NATs using private addresses before reaching a public IP address while researching about these issues.  Since a lot of those NATs are private infrastructure of the ISP you can't really ask them to forward the LN port to the homeowner.

Tor "hidden" services can work despite dynamic IP and automatically has NAT punch-through; the privacy is a good bonus too.

I want to be able to build a single-board computer with a lightweight LN node software that requires as little setup as possible (i.e. as close to "plug and play" as feasible).  The user story would be to put this single-board computer in my home, then contact it by mobile data on my mobile device to spend LN funds.  The node would be permanently on and therefore could safely route, and since it's in my home it's a bit safer than my mobile device that anyone can mug me to get.  However, typical residences often do not have static IP (because many ISPs apparently charge extra for static IP) and ISPs may impose additional NATs (as per the example I gave but can't find again).  Tor hidden services work around those by providing NAT punch-through as part of the onion network.  The Bitcoind fullnode can use Tor, but (at least as of 1.0) it seems LN can't?  Or can I use some other technique?

pm47:
Hi,

Nodes can already advertise V2 and V3 tor service addresses in their `node_announcement`, see [BOLT 7](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md#the-node_announcement-message).

AmkG:
oh, good!  I suppose I should have thought to search for "onion" rather than "hidden", ha.  Thanks!

PR #316, opened 2017-12-11
BOLT 7: mention Tor hidden service.

rustyrussell:
This is a common term to search for, rather than onion address (which is what Tor hidden services use).

COPY-EDIT: a.k.a. or AKA or aka?  We seem to use various....

See: #309 @amkg

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -225,7 +225,7 @@ The following `address descriptor` types are defined:
    * `2`: ipv6. data = `[16:ipv6_addr][2:port]` (length 18)
    * `3`: tor v2 onion service. data = `[10:onion_addr][2:port]` (length 12)
        * Version 2 onion service addresses. Encodes an 80-bit truncated `SHA-1` hash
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -225,7 +225,7 @@ The following `address descriptor` types are defined:
    * `2`: ipv6. data = `[16:ipv6_addr][2:port]` (length 18)
    * `3`: tor v2 onion service. data = `[10:onion_addr][2:port]` (length 12)
        * Version 2 onion service addresses. Encodes an 80-bit truncated `SHA-1` hash
-         of a 1024-bit `RSA` public key for the onion service.
+         of a 1024-bit `RSA` public key for the onion service (aka Tor hidden service).
```
toadlyBroodle: a.k.a.

toadlyBroodle:
I suggest we use a.k.a. for consistency with e.g. and i.e. I'll add these to the stylesheet.

PR #319, opened 2017-12-11
BOLT#11 Add breaking changes as per mailing list

junderw:
From the mailing list:

```
Hello all,

Recently I have implemented BOLT11 in node JS. You can find it on github at
bitcoinjs/bolt11 (check the wip branch, I am still working on tests and
looking at maybe splitting encode up into two functions, encode and sign,
if anyone wants to help)

Here are some comments:

1. There should be a tags_length integer before the first tag. This will
get rid of the awkward dance of parsing from the front and end at the same
time and make it easier to detect poorly encoded data quickly.
2. Why is a description or description commit hash a MUST include tag... it
should be SHOULD.
3. Descriptions say they can encode ASCII only. Sorry, but this is
nonsense. Full unicode support via UTF8 should be supported.
4. There should be a maximum length (in bytes) for description. (Otherwise
why have a commit hash tag at all?)
5. There should be a recommendation for ways for the parsing client to
fetch the commited purpose description data. (Maybe encode a URL after the
32 byte commit hash?)
6. Lightning payment requests can get pretty long, but perhaps there should
be a limit. Apparently bech32 with the given parameters can only handle one
error correction once it is longer than 1023 words.


Now for some things on the BOLT document itself:

1. I made a pull request for adding signature preimage hex data etc. to the
examples to help people writing decoders. Some of the language was unclear
to me and lead to some goof ups.
2. Would be nice to have a fuller set of test cases on the RFC. I will work
on some cases to increase coverage on the bitcoinjs repo. Would be happy
for any help.
```

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -71,7 +71,8 @@ whatever is being offered in return.
 The data part consists of multiple sections:
 
 1. `timestamp`: seconds-since-1970 (35 bits, big-endian)
-1. Zero or more tagged parts.
+1. `tagged_parts_data_length`: length (in 5 bit words) of the entire tagged parts section (15 bits, big-endian)
+1. One\* or more tagged parts. \*(Since payment_hash is required)
 1. `signature`: bitcoin-style signature of above. (520 bits)
 
 ## Requirements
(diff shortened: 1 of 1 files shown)


junderw:
"multibyte" is not a word...??????????

Yeah I guess not... *sigh*

rustyrussell:
OK, see excellent on-list discussion here: https://lists.linuxfoundation.org/pipermail/lightning-dev/2017-December/thread.html#818 and the resulting changes in #338 

We didn't, however, break the format; it would require a horrible bug at this stage to get consensus on that.

junderw:
ok, I rescind my suggestions on breaking the format.

Deferred to #338

PR #339, opened 2017-12-21
BOLT-4 Re-Edits

shannona:
Minor edits, clarifications, and standardizations.


pm47:
I've noticed that the numbering in section "Failure Messages" is broken. Should I go ahead and fix it in another PR or would that conflict with all the (great) copy-edit work that you are doing?

rustyrussell:
Please patch on top... I'm about to apply this (sorry, way too busy with other stuff).

PR #346, opened 2018-01-04
Bolt7 copy edit 1

toadlyBroodle:
First pass copy edit of first half of BOLT7 to bring in line with stylesheet guidelines

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -4,19 +4,19 @@ This specification describes simple node discovery, channel discovery, and chann
 
 Node and channel discovery serve two different purposes:
 
- - Channel discovery allows the creation and maintenance of a local view of the network's topology, so that the node can discover routes to the desired destination.
- - Node discovery allows nodes to broadcast their ID, host, and port, so that other nodes can open connections and establish payment channels.
- 
+ - Channel discovery allows the creation and maintenance of a local view of the network's topology, so that a node can discover routes to desired destinations.
+ - Node discovery allows nodes to broadcast their ID, host, and port, so that other nodes can open connections and establish payment channels with them.
+
 To support channel discovery, peers in the network exchange
-`channel_announcement` messages, which contain information about new
-channels between two nodes. They can also exchange `channel_update`
+`channel_announcement` messages containing information regarding new
+channels between the two nodes. They can also exchange `channel_update`
 messages, which update information about a channel. There can only be
 one valid `channel_announcement` for any channel, but at least two
 `channel_update` messages are expected.
 
 To support node discovery, peers exchange `node_announcement`
-messages, which supply additional information about nodes. There can be
-multiple `node_announcement` messages, to update node information.
+messages, which supply additional information about the nodes. There may be
+multiple `node_announcement` messages, in order to update the node information.
 
 # Table of Contents
 
@@ -33,8 +33,8 @@ multiple `node_announcement` messages, to update node information.
 
 ## The `announcement_signatures` Message
 
-This is a direct message between two endpoints of a channel and serves as an opt-in mechanism to allow the announcement of the channel to the rest of the network.
-It contains the necessary signatures by the sender to construct the `channel_announcement` message.
+This is a direct message between the two endpoints of a channel and serves as an opt-in mechanism to allow the announcement of the channel to the rest of the network.
+It contains the necessary signatures, by the sender, to construct the `channel_announcement` message.
 
 1. type: 259 (`announcement_signatures`)
 2. data:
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -118,84 +118,93 @@ announcement message: this is accomplished by having a signature from each
...
+  - MUST compute the double-SHA256 hash `h` of the message, beginning at offset
+  256, up to the end of the message.
+    - Note: the hash skips the 4 signatures but hashes the rest of the message,
+    including any future fields appended to the end.
+  - MUST verify that `node_signature_1` and `node_signature_2` are valid
+    signatures of the hash `h`, using `node_id_1` and `node_id_2`'s respective
+    secrets.
+  - MUST verify that `bitcoin_signature_1` and `bitcoin_signature_2` are valid
+  signatures of the hash `h`, using `bitcoin_key_1` and `bitcoin_key_2`'s
+  respective secrets.
```
rustyrussell: Instead of "MUST verify" this should be "MUST set", same with the node_signature above. One of them will be the one the node generates itself, the other will be the one the peer sends in the `announcement_signatures` message.

toadlyBroodle: fixed

07-routing-gossip.md
```diff
@@ -47,52 +47,59 @@ The willingness of the initiating node to announce the channel is signaled durin
 
 ### Requirements
 
-The `announcement_signatures` message is created by constructing a `channel_announcement` message, corresponding to the newly established channel, and signing it with the secrets matching an endpoint's `node_id` and `bitcoin_key`. The message is then sent using an `announcement_signatures`.
+The `announcement_signatures` message is created by constructing a `channel_announcement` message, corresponding to the newly established channel, and signing it with the secrets matching an endpoint's `node_id` and `bitcoin_key`. The message is then sent using an `announcement_signatures` message.
```
shannona: This is a little awkward with the multiple uses of message, especially in the updated final sentence. Maybe something like. "This `channel_announcement` message is then sent ..." or if it's accurate perhaps "This packet is then sent ..."

07-routing-gossip.md
```diff
@@ -111,90 +118,99 @@ announcement message; this is accomplished by having a signature from each
...
-previous message's `node_id_1` and `node_id_2` as well as this
-`node_id_1` and `node_id_2` and forget channels connected to them,
-otherwise it SHOULD store this `channel_announcement`.
-
-The receiving node SHOULD forget a channel once its funding output has
-been spent or reorganized out.
+The creating node:
+  - MUST set `chain_hash` to the 32-byte hash that uniquely identifies the chain
+  that the channel was opened within:
+    - MUST set the _Bitcoin blockchain_ `chain_hash` value (encoded in hex) to: `000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f`.
```
shannona: It would probably be better to say in line 122 "MUST set its `chain_hash`" to better differentiate it from line 124, and then I'd suggest not indenting line 124, which suggests a dependency rather than a parallelism.

07-routing-gossip.md
```diff
@@ -111,90 +118,99 @@ announcement message; this is accomplished by having a signature from each
...
-deep, the requirement against rebroadcasting only applies if the
-transaction hasn't moved to a different block.
+While channels should not be advertised before they are sufficiently deep, the
+requirement against rebroadcasting only applies if the transaction has not moved
+to a different block.
 
-To avoid having to store excessive-sized messages, yet allow
-reasonable expansion in future, nodes are allowed to restrict
-rebroadcasting (perhaps statistically).
+In order to avoid storing excessively-large messages, yet still allow for
```
shannona: "excessively large" (no hyphens for compounds with adverbs)

07-routing-gossip.md
```diff
@@ -208,84 +224,92 @@ nodes for which a channel is not already known are ignored.
...
+  - if `node_id` is NOT previously known from a `channel_announcement` message,
+  OR if `timestamp` is NOT greater than the last-received `node_announcement`
+  from this `node_id`:
+    - SHOULD ignore the message.
+  - otherwise:
+    - if `timestamp` is greater than the last-received `node_announcement` from
+    this `node_id`:
+      - SHOULD queue the message for rebroadcasting.
+      - MAY choose NOT to queue messages longer than the minimum expected length.
+  - SHOULD NOT connect to a node which has an unknown, _even_ `features` bit set
```
shannona: Should this be up in "if `features` field contains unknown _even bits_:" (e.g. line 288) as "SHOULD NOT connect to that node".

07-routing-gossip.md
toadlyBroodle: fixed

07-routing-gossip.md
toadlyBroodle: changed

07-routing-gossip.md
toadlyBroodle: fixed

07-routing-gossip.md
toadlyBroodle: Good question @rustyrussell?

Issue #302, opened 2017-12-04
2-week timestamp expiry on `channel_update` implies timestamp is epoch time


rustyrussell:
Section 7 says "nodes MAY prune channels should the timestamp of the latest `channel_update` be older than 2 weeks (1209600 seconds)"

Yet timestamps are only required to be sequential, not an actual  timestamp ("The creating node MUST set `timestamp` to be greater than that for any previous `node_announcement` it has created. It MAY base it on a UNIX timestamp.", so there's no way to know the age of a timestamp, at least not by looking at the timestamp.

Reported-by: @shannona

pm47:
What are the downsides of requiring timestamps to be based on epoch time?

rustyrussell:
Pierre-Marie Padiou <notifications@github.com> writes:
> What are the downsides of requiring timestamps to be based on epoch time?

We don't have a good way of enforcing it.  If I put a timestamp of 2020
in the announcement, will it never time out?  Do you reject it as
stupid?

It would be fairly simple to say "can't be greater than getmediantime()"
of past 11 blocks; a conservative implementation would use the previous
block, to avoid the update getting refused by peers which don't know the
lastest block.  That's approximately an hour in the past.  Heck, we
could add 3600 seconds to make it look like 'current time'.

Then the requirements are well-defined.

PR #350, opened 2018-01-08
BOLT 7: timestamp must be a UNIX timestamp.

rustyrussell:
As agreed at the previous meeting, we use the timestamp to prune,
so it does have to be a valid timestamp.  We also suggest ignoring
if it's in the far future, too.

The minutes suggest we can prune for any reason, and that's really
true anyway; the requirements to keep it around are only SHOULD.

Closes: #302

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -359,7 +359,7 @@ Bits which are not assigned a meaning must be set to 0.
 A node MAY create and send a `channel_update` with the `disable` bit set to signal the temporary unavailability of a channel, e.g., due to loss of connectivity, or the permanent unavailability, e.g., ahead of an on-chain settlement.
 A subsequent `channel_update` with the `disable` bit unset MAY re-enable the channel.
 
-The creating node MUST set `timestamp` to greater than zero, and MUST set it to greater than any previously-sent `channel_update` for this `short_channel_id`.
+The creating node MUST set `timestamp` to greater than zero, and MUST set it to greater than any previously-sent `channel_update` for this `short_channel_id`, and SHOULD base it on a UNIX timestamp.
 It MUST set `cltv_expiry_delta` to the number of blocks it will subtract from an incoming HTLCs `cltv_expiry`. It MUST set `htlc_minimum_msat` to the minimum HTLC value the other end of the channel will accept, in millisatoshi. It MUST set `fee_base_msat` to the base fee it will charge for any HTLC, in millisatoshi, and `fee_proportional_millionths` to the amount it will charge per transferred satoshi in millionths of a satoshi.
 
 The receiving nodes MUST ignore the `channel_update` if it does not correspond to one of its own channels, if the `short_channel_id` does not match a previous `channel_announcement`, or if the channel has been closed in the meantime.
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -289,6 +289,12 @@ The receiving node MAY use `rgb_color` and `alias` to reference nodes in interfa
 
 ### Rationale
 
+The `timestamp` field is used by nodes for pruning (either if it's too
+far in the future, or if it's been two weeks with no update), so it
+makes sense to have it be a UNIX timestamp (ie. seconds since UTC
+1970-01-01).  It need only be exact, however, given the unusual case
```
pm47: > It need only be exact, however, given the unusual case of two `channel_update`s within a second.

I don't understand this, what do you mean?

Roasbeef: Same, can't really parse this section.

rustyrussell: Oops, should be "We can't require that it be a timestamp, however..."

rustyrussell:
@Roasbeef @pm47 OK, unscrewed it, please ack.

07-routing-gossip.md
```diff
@@ -289,6 +289,12 @@ The receiving node MAY use `rgb_color` and `alias` to reference nodes in interfa
 
 ### Rationale
 
+The `timestamp` field is used by nodes for pruning (either if it's too
+far in the future, or if it's been two weeks with no update), so it
+makes sense to have it be a UNIX timestamp (ie. seconds since UTC
+1970-01-01).  It can't be a hard requirement, however, given the possible case
```
pm47: Shouldn't this last sentence be in the `channel_update` section? Also, it is not clear that the first requirement applies to both `node_announcement` and `channel_update`.

Maybe either duplicate/reference the requirement, or move this whole paragraph to the "Pruning the Network View" section?

07-routing-gossip.md
```diff
@@ -289,6 +289,12 @@ The receiving node MAY use `rgb_color` and `alias` to reference nodes in interfa
...
+of two `channel_update`s within a second.
```
halseth: I don't think this explains well enough why this can't be made a hard requirement; the reason is because we want nodes to be allowed to broadcast more than one `node_announcement`/`channel_update` a second, into "the far future"? 

Seems like that would be malicious behaviour where we could just ignore the announcements.

07-routing-gossip.md
rustyrussell: Oops, yes, I edited the wrong section!  I fixed it up...  Good catch!

07-routing-gossip.md
rustyrussell: If you need to send out a temporary disable, you don't have much choice.  It will propagate slowly, of course (each node batches), but it could happen in theory.

halseth: I guess a node could use frequent `channel_updates` to do some high frequency payments with its immediate channel counter parties. 

But if we allow this (by saying it is not a hard requirement for it to be a unix timestamp), then maybe we shouldn't say that nodes prune based on it being a unix timestamp? 

You might end up with pruning a channel that has issued a lot of updates and now is considered "far into the future", or not pruning a channel that has been inactive for a very long time, but issued a channel update with a future timestamp before it left.

cdecker: Well, we can't prevent some of the foot-gun situations, this is best effort, and we explicitly mention that there is the possibility that frequent updates resulting in far in the future timestamps result in the channel being dropped altogether.

halseth: Fair enough, thanks!

PR #355, opened 2018-01-10
BOLT 2: clarify that nodes may support early mutual close

### LINKED ISSUE #335: Allow mutual close before `min_depth`
The spec is a little vague on this, but I don't see why we shouldn't allow it?

Currently eclair and lnd only allow an unilateral close before `min_depth`, not sure about c-lightning.

pm47:
Clarify that nodes MAY initiate/handle a mutual close before the funding transaction has reached `min_depth`.

If it doesn't support it, the receiving node will just ignore the `shutdown`. Alternative was to fail the channel, which could also make sense. A feature bit was probably overkill.

I may have been a little bit verbose, let me know what you think.

This fixes #335.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -412,6 +412,9 @@ along with the `scriptpubkey` it wants to be paid to.
 #### Requirements
 
 A sending node:
+  - if it hasn't sent a `funding_created` (if it is a funder) or a `funding_signed` (if it is a fundee):
+    - MUST NOT send a `shutdown`
+  - MAY send a `shutdown` before a `funding_locked`, i.e. before the funding transaction has reached `min_depth`.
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -412,6 +412,9 @@ along with the `scriptpubkey` it wants to be paid to.
 #### Requirements
 
 A sending node:
+  - if it hasn't sent a `funding_created` (if it is a funder) or a `funding_signed` (if it is a fundee):
+    - MUST NOT send a `shutdown`
+  - MAY send a `shutdown` before a `funding_locked`, i.e. before the funding transaction has reached `min_depth`.
```
cdecker: missing comma after `i.e.`

PR #372, opened 2018-02-03
BOLT 4: update sphinx packet test vector

Roasbeef:
In this commit, we update the test vector for the final onion packet. In
commit 068b0bccf94e8cdaf5f298dade0fcc8cc8421ef6, the per-hop payloads
were updated to use 8 byte amounts everywhere. However, the test vectors
were not updated. In 578573f92fc04d9390fdd376ad050702a0c59f1e the test
vectors were updated to use the proper version prefix. However, this
assumed that the state of the vectors as is was correct.

To remedy this, we've updated the test vectors to reflect the final
result using the current format for encoding the per-hop payloads. This
final test vector was generated using the original tool that we used to
confirm compatibility between the C and Go versions.


cdecker:
utACK 3e2d1f7

sstone:
We would also need to update the payloads to
```
0x000000000000000000000000000000000000000000000000000000000000000000
0x000101010101010101000000000000000100000001000000000000000000000000
0x000202020202020202000000000000000200000002000000000000000000000000
0x000303030303030303000000000000000300000003000000000000000000000000
0x000404040404040404000000000000000400000004000000000000000000000000
```
and the hmacs to:

```
0x9b122c79c8aee73ea2cdbc22eca15bbcc9409a4cdd73d2b3fcd4fe26a492d376
0x548e58057ab0a0e6c2d8ad8e855d89f9224279a5652895ea14f60bffb81590eb
0x0daed5f832ef34ea8d0d2cc0699134287a2739c77152d9edc8fe5ccce7ec838f
0x62cc962876e734e089e79eda497077fb411fac5f36afd43329040ecd1e16c6d9
```

PR #374, opened 2018-02-04
BOLT 4: Clarify the onion construction and suggest optimization.

jimpo:
As was noted in by @cfromknecht [on the mailing list](https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-February/000981.html), the current spec suggests a quadratic time algorithm for onion packet construction, while a linear time one is possible.

I have reworded the relevant section to introduce the concept of an ephemeral private key, which I find easier to reason about and suggests the linear time algorithm.

As a separate change that may be worth putting in a separate pull request instead, the instructions for deriving the shared secret is incomplete in that it only says to perform ECDH, but not to take the SHA256 hash of the serialized point. All implementations do this, as does the reference code in the spec. Also, the blinding factor calculation instructions had a mistake.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -265,38 +265,33 @@ packet commits to but that is not included in the packet itself. Associated
 data will be included in the HMACs and must match the associated data provided
 during integrity verification at each hop.
 
-Next, for each hop along the route, the sender computes an
-_ephemeral public key_, a _shared secret_, and a _blinding factor_. The blinding
-factor is used at each hop to blind the ephemeral public key for the next hop.
-The hop receiving the header will perform ECDH with the ephemeral public key
-and its own private key in order to derive the same shared secret.
-However, when generating the packet, the sending node doesn't have access to the
-other nodes' private keys. So instead, it uses the commutative property of
-multiplication to blind each hop's public key with all previous blinding
-factors and performs ECDH, using each hop's blinded public key and the
-`sessionkey`.
+Each hop along the route sees a blinded _ephemeral public key_ belonging to the
+sender, from which it derives a _shared secret_ and a _blinding factor_, used to
+reblind the ephemeral public key for the next hop. The shared secret is a hash
+of the ECDH output element of the hop's node key and the sender's ephemeral key.
+The blinding factor is then derived from the secret key and multiplied with the
+ephemeral key to generate the ephemeral key for the next hop.  Because the
+blinding of a public key is a scalar multiplication with a group element, by
+virtue of the commutative property, the sender can generate the corresponding
+private key for the next hop by multiplying the blinding factor by its ephemeral
+private key.
 
-The transformations at hop `k` are as follows:
+To construct the onion, the sender initializes the ephemeral private key for the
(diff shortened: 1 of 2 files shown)


04-onion-routing.md
```diff
@@ -280,7 +280,9 @@ The transformations at hop `k` are as follows:
 
  - The shared secret `ss_k` is computed by first blinding the hop's public key
  `nodepk_k` with all previous blinding factors `{b_1, ..., b_{k-1}}` (if any),
- and second, executing ECDH with the blinded public key and the `sessionkey`.
+ and second, executing ECDH with the blinded public key and the `sessionkey` to
+ obtain a group element. The shared secret `ss_k` is then calculated as the
+ `SHA256` hash of the ECDH output, serialized in the compressed format.
```
cdecker: This is explained further down in the [Shared Secret](https://github.com/jimpo/lightning-rfc/blob/ac143c18533752124b09956a02cfb4563cb08051/04-onion-routing.md#shared-secret) section. I'd prefer having a reference to that part rather than duplicating the instructions.

jimpo:
@cdecker Thanks for the suggestion. I'm going to try reordering some of the sections to reduce the number of forward references (moving Shared Secret up for example).

04-onion-routing.md
```diff
@@ -252,6 +257,41 @@ compare its values against those of the HTLC. See the
...
+the hop's node ID key. The resulting curve point is serialized to the
+DER-compressed representation and hashed using `SHA256`. The hash output is used
+as the 32-byte shared secret.
+
+Elliptic-curve Diffie-Hellman (ECDH) is an operation on an EC private key and
+an EC public key that outputs a curve point. For this protocol, the ECDH
+variant implemented in `libsecp256k1` is used, which is defined over the
+`secp256k1` elliptic curve. During packet construction, the sender uses the
+ephemeral private key and the hop's public key as inputs to ECDH, whereas
+during packet forwarding, the hop uses the ephemeral public key and it's own
```
rustyrussell: s/it's/its/

jimpo: Fixed.

04-onion-routing.md
```diff
@@ -265,36 +305,21 @@ packet commits to but that is not included in the packet itself. Associated
...
 
-The origin node then iteratively computes the ephemeral public keys, shared
-secrets, and blinding factors for hops `{n_2, ..., n_r}`.
+ - The sender executes ECDH with the hop's public key and the ephemeral private
+ key to obtain a curve point, which is hashed using `SHA256` to produce the
+ shared secret `ss_k`.
+ - The blinding factor is the `SHA256` hash of the concatenation between the
+ ephemeral public key `epk_k` and the shared secret `ss_k`.
+ - The ephemeral private key for the next hop `ek_{k+1}` is computed by
+ multiplying the current ephemeral private key `ek_k` by the blinding factor.
```
cfromknecht: would add that this multiplication is taken modulo the curve order, so that it isn't confused with taking it modulo the secp prime

PR #375, opened 2018-02-05
BOLT02: Add rationale for 2^24 satoshi per channel limit

### LINKED ISSUE #353: Rationale for max value of `funding_satoshis` in `open_channel` message
Hello I noticed in the [requirements](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#requirements) of the peer protocol the max value of funding_satoshis MUST be set to less than 2^24, although the field in the message is capable of 2^64. Is it possible to add the rationale for this choice to the specification [rationale](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md#rationale) ?

cdecker:
Fixes #353

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -215,6 +215,10 @@ The receiving node MUST NOT:
 
 #### Rationale
 
+The requirement for `funding_satoshi` to be less than 2^24 satoshi is a temporary self-imposed limit while implementations are not yet considered stable.
+It can be lifted at any point in time, or adjusted for other currencies, since it is solely enforced by the endpoints of a channel.
+Specifically, [the routing gossip protocol](07-routing-gossip.md) does not discard channels that have a larger capacity.
+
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -215,6 +215,10 @@ The receiving node MUST NOT:
 
 #### Rationale
 
+The requirement for `funding_satoshi` to be less than 2^24 satoshi is a self-imposed limit while implementations are not yet considered stable, and temporary.
```
practicalswift: Nit: I suggest changing from "… is a self-imposed limit while implementations are not yet considered stable, and temporary" to "… is a temporary self-imposed limit while implementations are not yet considered stable".

dcousens: > is a temporary recommendation for implementations until the network is considered stable.

Issue #368, opened 2018-02-02
BOLT3 key can be ambiguous


dcousens:
The terminology phrase "key" doesn't typically refer to only a public key,  infact my lasting impression until I read the derivation/HTLC contracts was that it was the secret.

I think it should either be referred to as a "point" (consistent with `per_commitment_point`),  or "pubkey",  or "secretkey",  explicitly.
The distinction of when you are talking about the private key,  or the public key,  is IMHO often lost and therefore confusing from an implementation perspective.

Additionally,  you specify the derivation procedure for several key/points with:

> ### `localkey`, `remotekey`, `local_htlckey`, `remote_htlckey`, `local_delayedkey`, and `remote_delayedkey` Derivation

Maybe instead classify these key pairs as of the same "class"/"type" (a `commitmentkeypair`, 
 maybe),  and then specifically outline that the derivation procedure for that "type" is:

``` js
b = basepoint
h = SHA256(per_commitment_point || basepoint)
secretkey = b + h
pubkey = bG + hG
```

(etc)

PR #377, opened 2018-02-06
BOLT 2,3,5: always refer to shared/pubkey/private key.

rustyrussell:
Make it clear what kind of key we're talking about.  We use the abbreviation
pubkey for public key (as it's quite common to use in field names), but
generally spell out 'private'.

(I generally prefer 'secret' to 'private' but we use private far more often
already, and we use 'secret' for things which don't directly derive keys).

Fixes: #368

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -27,7 +27,7 @@ wpkh
 nextHmac
 basepoint
 streamKey
-localkey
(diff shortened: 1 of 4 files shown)


dcousens:
ACK,  if that wasn't clear :+1:

PR #384, opened 2018-02-09
BOLT-0 Glossary Reorg & Expansion

shannona:
This is a big change to the glossary in BOLT-0.

The overarcing changes are:

1.) Reorganized glossary to be alphabetical
2.) Add copious "See also:" lines to cross-reference containers, contents, and other related entries.

*Changed Definitions:* Peer and node were revised based in on discussions with Rusty; commitment revocation secret key was changed to be commitment revocation private key based on #377.

*New Definitions:* New words added based on their usage in the bolts were: announcement, closing transaction, final node, hop, origin node, processing,  node, receiving node, receiving peer, revoked commitment transaction, sending node, sending peer.

(I suspect the reorg will make the diffs almost useless, so this is the list of stuff to look at.)

### DIFF
--- a/00-introduction.md
+++ b/00-introduction.md
@@ -29,8 +29,8 @@ This is version 0.
    * A gossip message sent between *peers* intended to aid the discovery of a *channel* or a *node*.
 
 * `chain_hash`:
-   * The genesis hash of a
-     target blockchain. This allows *nodes* to create and reference *channels* on
+   * The uniquely identifying hash of the target blockchain (usually the genesis hash).
+     This allows *nodes* to create and reference *channels* on
      several blockchains. Nodes are to ignore any messages that reference a
      `chain_hash` that are unknown to them. Unlike `bitcoin-cli`, the hash is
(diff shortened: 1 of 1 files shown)


shannona:
Besides going over the updated and expanded terms, please also let me know if we want to use some slightly different formatting for the _see alsos_, presuming the cross-references are considered useful.

shannona:
From the PR: I am out of town next week, but after I get back I will have some definite time on my schedule, and can draft an intro for the spec, if we're comfortable with all the terms here.

shannona:
The update to the PR from today includes a long introduction to Lightning. My goal was to overview the system, to provide the context for all the BOLTs that followed, while simultaneously offering another way to link into the rest of the content. Hopefully it’ll meet your needs.

Of course, let me know for any changes that are required, particularly: (1) any inaccuracies in my explanations of the systems; and (2) any more general mistakes. But also let me know if (3) there are any questions you frequently get that could be addressed in this sort of intro.

PR #389, opened 2018-02-19
BOLT 2: ensure that we always have a non-dust output.

rustyrussell:
~~This requirement is subtle: firstly, we ensure that both reserve values are above
both dust limits.  But we also need to make sure that fees (which can eat into
reserves) always leave enough left: it makes sense to ensure that worst-case
fees are actually covered by the channel capacity, otherwise the channel is
likely to be useless anyway.~~

I re-read the fee payment section: we can simply assert that reserves are above dust.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -211,6 +211,8 @@ The receiving node MUST fail the channel if:
   - `funding_pubkey`, `revocation_basepoint`, `htlc_basepoint`, `payment_basepoint`, or `delayed_payment_basepoint`
 are not valid DER-encoded compressed secp256k1 pubkeys.
   - `dust_limit_satoshis` is greater than `channel_reserve_satoshis`.
+  - the funder's amount for the initial commitment transaction is sufficient for [fee payment](03-transactions.md#fee-payment).
+  - both `to_local` and `to_remote` amounts for the initial commitment transaction are less than or equal to `channel_reserve_satoshis`.
 
 The receiving node MUST NOT:
   - consider funds received, using `push_msat`, to be received until the funding transaction has reached sufficient depth.
@@ -223,7 +225,7 @@ Specifically, [the routing gossip protocol](07-routing-gossip.md) does not disca
 
 The *channel reserve* is specified by the peer's `channel_reserve_satoshis`: 1% of the channel total is suggested. Each side of a channel maintains this reserve so it always has something to lose if it were to try to broadcast an old, revoked commitment transaction. Initially, this reserve may not be met, as only one side has funds; but the protocol ensures that there is always progress toward meeting this reserve, and once met, it is maintained.
 
-The sender can unconditionally give initial funds to the receiver using a non-zero `push_msat` — this is one case where the normal reserve mechanism doesn't apply. However, like any other on-chain transaction, this payment is not certain until the funding transaction has been confirmed sufficiently (with a danger of double-spend until this occurs) and may require a separate method to prove payment via on-chain confirmation.
+The sender can unconditionally give initial funds to the receiver using a non-zero `push_msat`, but even in this case we ensure that the funder has sufficient remaining funds to pay fees and that one side has some amount it can spend (which also implies there is at least one non-dust output). Note that, like any other on-chain transaction, this payment is not certain until the funding transaction has been confirmed sufficiently (with a danger of double-spend until this occurs) and may require a separate method to prove payment via on-chain confirmation.
 
 The `feerate_per_kw` is generally only of concern to the sender (who pays the fees), but there is also the fee rate paid by HTLC transactions; thus, unreasonably large fee rates can also penalize the recipient.
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -231,7 +231,9 @@ Separating the `htlc_basepoint` from the `payment_basepoint` improves security: 
 
 The requirement that `channel_reserve_satoshis` is not considered dust
 according to `dust_limit_satoshis` eliminates cases where all outputs
-would be eliminated as dust.
+would be eliminated as dust.  The similar requirements in
+`accept_channel` ensure that both sides' `channel_reserve_satoshis`
+are above both `dust_limit_satoshis`.
 
 #### Future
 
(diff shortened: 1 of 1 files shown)


halseth:
Requiring channel reserve to be non-zero seems very restrictive to me, there are probably cases where a zero reserve is useful.

Maybe I'm forgetting something, but isn't it more natural to just require that the balance after subtracting fees is above the channel reserve (which can be zero)? This is what we do in `lnd`.

rustyrussell:
You can't control that.  i.e. I add a whole heap of txs at the same time you do.  Now we get the case where fees can't be paid...

02-peer-protocol.md
```diff
@@ -174,6 +174,7 @@ The sending node:
   - MUST set `push_msat` to equal or less than 1000 * `funding_satoshis`.
   - MUST set `funding_pubkey`, `revocation_basepoint`, `htlc_basepoint`, `payment_basepoint`, and `delayed_payment_basepoint` to valid DER-encoded, compressed, secp256k1 pubkeys.
   - MUST set `first_per_commitment_point` to the per-commitment point to be used for the initial commitment transaction, derived as specified in [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
+  - MUST `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis`.
```
halseth: `MUST set`

halseth:
Well, this prevents the case of non-dust output I guess (since the non-funder will have a balance always above the dust limit, and she doesn't pay any fees).

But the scenario you mention doesn't prevent the fees from eating *all* of the reserve, essentially creating a transaction with a insufficient fee, unable to get confirmed. This could again be handled by requiring the channel reserve to be big enough to always handle such a big chunk of HTLC fees, but then we wander even further into edge case land. Safest approach to me seems to be (as was also discussed briefly in regards to the `update_fee` edge case) to just fail the channel using the unrevoked commitment tx if such a situation occurs.

Shameless plug: I proposed a change to handle these cases in the ML thread "[Lightning-dev] Fee disentanglement for 1.1 spec?", that haven't gotten any response yet..

pm47:
> I re-read the fee payment section: we can simply assert that reserves are above dust.

The goal is to have at least a non-dust output post fee right?

> Well, this prevents the case of non-dust output I guess (since the non-funder will have a balance always above the dust limit, and she doesn't pay any fees).

I don't think it does: 
If the funder has all the money on his side (it's possible right after having created a channel with `push_msat`=0), fees are very high and eat almost all or exactly all of its output (meaning that resulting fee is correct), then we'll still have zero outputs?

> But the scenario you mention doesn't prevent the fees from eating all of the reserve, essentially creating a transaction with a insufficient fee, unable to get confirmed. [...] Safest approach to me seems to [...] just fail the channel using the unrevoked commitment tx if such a situation occurs.

This is what the fee payment section in BOLT 3 tells us to do in that case, so in that case I think it works.

> Shameless plug: I proposed a change to handle these cases in the ML thread "[Lightning-dev] Fee disentanglement for 1.1 spec?", that haven't gotten any response yet..

I've been meaning to read and answer to this in ~~weeks~~ months now, couldn't find the time... Fee management for dual-funded channels will be so much fun ;-)

halseth:
@pm47 You are right, totally wrong of me to assert that the non-funder will always have balance above the reserve! So I don't think this change actually helps ensuring we'll always have one output?

Still not convinced the best approach is not just to not let fees eat into reserve, and if any of the edge cases occur, fail the channel using the last commitment. That would also handle #393 I believe.

rustyrussell:
@pm47 is right; this problem needs a separate solution though.  See #393.

I think the answer to *that* is that:  The funder must be able to afford its own reserve plus fees for the initial commitment transaction plus least 1 HTLC (avoids creating useless channels).

This seems to prevent this case in general.  It undermines the original rationale of push_msat, but I'm coming to hate that thing anyway...

rustyrussell:
@pm47 is right; this problem needs a separate solution though.  See #393.

I think the answer to *that* is that:  The funder must be able to afford its own reserve plus fees for the initial commitment transaction plus least 1 HTLC (avoids creating useless channels).

This seems to prevent this case in general.  It undermines the original rationale of push_msat, but I'm coming to hate that thing anyway...

02-peer-protocol.md
```diff
@@ -174,6 +174,7 @@ The sending node:
   - MUST set `push_msat` to equal or less than 1000 * `funding_satoshis`.
   - MUST set `funding_pubkey`, `revocation_basepoint`, `htlc_basepoint`, `payment_basepoint`, and `delayed_payment_basepoint` to valid DER-encoded, compressed, secp256k1 pubkeys.
   - MUST set `first_per_commitment_point` to the per-commitment point to be used for the initial commitment transaction, derived as specified in [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
+  - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis`.
```
pm47: typo: "MUST set `channel_reserve_satoshis` [to]  greater than"?

02-peer-protocol.md
```diff
@@ -221,12 +225,18 @@ Specifically, [the routing gossip protocol](07-routing-gossip.md) does not disca
...
+The sender can unconditionally give initial funds to the receiver using a non-zero `push_msat`, but even in this case we ensure that the funder has sufficient remaining funds to pay fees and that one side has some amount it can spend (which also implies there is at least one non-dust output). Note that, like any other on-chain transaction, this payment is not certain until the funding transaction has been confirmed sufficiently (with a danger of double-spend until this occurs) and may require a separate method to prove payment via on-chain confirmation.
 
 The `feerate_per_kw` is generally only of concern to the sender (who pays the fees), but there is also the fee rate paid by HTLC transactions; thus, unreasonably large fee rates can also penalize the recipient.
 
 Separating the `htlc_basepoint` from the `payment_basepoint` improves security: a node needs the secret associated with the `htlc_basepoint` to produce HTLC signatures for the protocol, but the secret for the `payment_basepoint` can be in cold storage.
 
+The requirement that `channel_reserve_satoshis` is not considered dust
```
pm47: Is it enough though? Even if `channel_reserve_satoshis` > `dust_limit_satoshis` before fees, maybe it will become dust after fees are applied? That can happen with or without `push_msat`.

02-peer-protocol.md
rustyrussell: Then we get a repeated "to" (to greater than or equal to).  We can omit the final to but that reads worse to me...

02-peer-protocol.md
rustyrussell: This can't happen to both sides, however.  The 'effective capacity' of the channel is the amount post-reserve, and that must be greater than the greatest possible fees.  I think it works.

PR #392, opened 2018-03-05
Final: Channel range queries

rustyrussell:
Here's an early draft, please bikeshed @Roasbeef @cdecker @pm47 @sstone

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -24,6 +24,7 @@ multiple `node_announcement` messages, in order to update the node information.
   * [The `channel_announcement` Message](#the-channel_announcement-message)
   * [The `node_announcement` Message](#the-node_announcement-message)
   * [The `channel_update` Message](#the-channel_update-message)
+  * [Query Messages](#query-messages)
   * [Initial Sync](#initial-sync)
   * [Rebroadcasting](#rebroadcasting)
   * [HTLC Fees](#htlc-fees)
@@ -462,23 +463,30 @@ of two `channel_update`s within a single second.
 
 ## Initial Sync
 
+Note that the `initial_routing_sync` feature is overridden (and should
+be considered equal to 0) by the `gossip_queries` feature if the
+latter is negotiated.
+
+Note that `gossip_queries` won't work with older nodes, so the
+value of `initial_routing_sync` is still important to control
+interactions with them.
+
 ### Requirements
 
 An endpoint node:
-  - upon establishing a connection:
-    - SHOULD set the `init` message's `initial_routing_sync` flag to 1, to
-    negotiate an initial sync.
-  - if it requires a full copy of the other endpoint's routing state:
-    - SHOULD set the `initial_routing_sync` flag to 1.
-  - upon receiving an `init` message with the `initial_routing_sync` flag set to
-  1:
-    - SHOULD send `channel_announcement`s, `channel_update`s and
-    `node_announcement`s for all known channels and nodes, as if they were just
-    received.
-  - if the `initial_routing_sync` flag is set to 0, OR if the initial sync was
-  completed:
-    - SHOULD resume normal operation, as specified in the following
-    [Rebroadcasting](#rebroadcasting) section.
+  - if the `gossip_queries` feature is negotiated:
+	- MUST NOT relay any gossip messages unless explicitly requested.
+  - otherwise:
+    - if it requires a full copy of the other endpoint's routing state:
+      - SHOULD set the `initial_routing_sync` flag to 1.
+    - upon receiving an `init` message with the `initial_routing_sync` flag set to
+    1:
+      - SHOULD send gossip messages for all known channels and nodes, as if they were just
+      received.
+    - if the `initial_routing_sync` flag is set to 0, OR if the initial sync was
+    completed:
+      - SHOULD resume normal operation, as specified in the following
+      [Rebroadcasting](#rebroadcasting) section.
 
 ## Rebroadcasting
 
@@ -498,6 +506,8 @@ The final node:
         for its peers.
 
 An endpoint node:
+  - if the `gossip_queries` feature is negotiated:
+	- MUST not send gossip until it receives `gossip_timestamp_range`.
   - SHOULD flush outgoing gossip messages once every 60 seconds, independently of
   the arrival times of the messages.
     - Note: this results in staggered announcements that are unique (not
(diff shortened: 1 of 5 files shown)


07-routing-gossip.md
```diff
@@ -510,16 +524,197 @@ An endpoint node:
...
+which overlaps the ranges of the request.
+
+### The `gossip_timestamp_range` Message
+
+1. type: 265 (`gossip_timestamp_range`) (`gossip_queries`)
+2. data:
+    * [`32`:`chain_hash`]
+    * [`4`:`first_timestamp`]
+    * [`4`:`timestamp_range`]
+
```
sstone: `range` here has not the same meaning as in `query_channel_range`/`reply_channel_range`, `gossip_timestamp_filter` would be more appropriate ?

07-routing-gossip.md
```diff
@@ -510,16 +524,197 @@ An endpoint node:
...
+#### Rationale
+
+Since `channel_announcement` doesn't have a timestamp, we generate
+one.  As `channel_announcement` is usually followed immediately by a
+`channel_update`, that `channel_update` timestamp serves well: in the
+case where no `channel_update` has been received yet, any reasonable
+value will do: it could be the block time of the first possible
+announcement block (ie. six blocks following the funding transaction),
+or simply the this node received it.
```
sstone: typo: or simply the timestamp this node received it at ?

07-routing-gossip.md
```diff
@@ -510,16 +524,197 @@ An endpoint node:
...
+
+This does mean that a new node will consider a `channel_announcement`
+to be more recent than an old node would, but in the common case where
+`gossip_timestamp_range` is sent by the old node to restrict gossip to
+recent messages: in that case `channel_announcement` will be sent
+redundantly.  The new node will already need to ask the old node for
+old gossip, so it will also receive the `channel_announcement`.
+
```
sstone: There is still a problem with old inactive channels: if you prune and forget a channel, and later receive an announcement for this same channel, but no matching channel updates, you will broadcast again it (unless you set its timestamp to the timestamp of the block it appeared in ?). I think that broadcasting channel announcements only when we have a matching channel update should work. We could also group channel announcements and channel updates, it is less efficient and we'll send more data but it mitigated by the timestamp filter.

Roasbeef: Is there still an issue with old inactive channels? You'd start the gossip timestamp at the point beyond your "zombie horizon", so the stale channel update would never be sent to you right?

07-routing-gossip.md
rustyrussell: Agreed.

07-routing-gossip.md
rustyrussell: Ok, so as agreed in call, timestamp for announce filtering is that of any update. This means you don't send a useless announce which has no channel updates ever. It means you don't have to store the "timestamp" since you can just use the latest update, but you can, and you don't have to deal with the corner case "new update arrived, now announce timestamp is within the window so I have to send it too" since you can argue that you're using the older timestamp. You only need to consider that for the very first update.

sstone: Yes I think it works (we'll probably go with the "use latest timestamp" approach though). All that is missing now is the `gzip` option :)
We have updated endurance on testnet with a version of eclair that implements this PR

07-routing-gossip.md
```diff
@@ -462,23 +463,34 @@ of two `channel_update`s within a single second.
...
-    received.
-  - if the `initial_routing_sync` flag is set to 0, OR if the initial sync was
-  completed:
-    - SHOULD resume normal operation, as specified in the following
-    [Rebroadcasting](#rebroadcasting) section.
+  - if the `gossip_queries` feature is negotiated:
+	- MUST NOT relay any gossip messages unless explicitly requested.
+  - otherwise:
+    - upon establishing a connection:
+      - SHOULD set the `init` message's `initial_routing_sync` flag to 1, to
```
jimpo: I'm confused by this. On this line it says to always set the flag upon establishing a connection, and on the next line it says to only set the flag if you need a full copy of the routing state.

07-routing-gossip.md
```diff
@@ -510,16 +524,197 @@ An endpoint node:
...
+  - if does not maintain up-to-date channel information for `chain_hash`:
+	- MUST set `complete` to 0.
+  - otherwise:
+	- SHOULD set `complete` to 1.
+
+#### Rationale
+
+Future nodes may not have complete information; they certainly won't have
+complete information on unknown `chain_hash` chains.  While this `complete`
+field cannot be trusted, if 0 ir does indicate that the sender should search
```
jimpo: typo: s/ir/it/. I would actually just remove the "if" and "it": "While this `complete` field cannot be trusted, 0 does indicate...".

07-routing-gossip.md
jimpo: I also like the idea of withholding channel announcements until there is a corresponding update to timestamp it, especially compared to the ambiguity of "choose a timestamp that is reasonable." What is the downside to that?

07-routing-gossip.md
rustyrussell: Yes, that was actually in the prior text.  Will fix.

07-routing-gossip.md
rustyrussell: Yep, will fix.

07-routing-gossip.md
Roasbeef: Just to clarify, this the `timestamp_range` is in seconds, right? May be good to note that here to be precise.

sstone:
@Roasbeef spotted a potential issue: there is no easy way to add a new compression format, or to prevent you from sending data encoded in a format that your peer does not support.
I think we have 2 options:
- limit ourselves to `uncompressed` and `zlib`, and have everyone support both.  `zlib` already does a fairly good job, and the `tx index` part of short channel ids is basically random numbers so it may be difficult to do much better.
- specify that the first outgoing query message must be a `query_channel_range` message, which seems natural, and add an `expected_format` field to it. If the receiver does not support it they would fall back to `uncompress`.

rustyrussell:
We can add a new feature bit for new compression.  But frankly we should just do zlib, since everyone has to support it anyway.  While we can get factor-of-2 improvement with other options, it's not necessary AFAICT.

07-routing-gossip.md
```diff
@@ -510,16 +521,196 @@ An endpoint node:
...
+  - MUST NOT send this if it has sent a previous `query_channel_range` to this peer and not received `reply_channel_range`.
+  - MUST set `chain_hash` to the 32-byte hash that uniquely identifies the chain
+  that it wants the `reply_channel_range` to refer to
+  - MUST set `first_blocknum` to the first block it wants to know channels for
+  - MUST set `number_of_blocks` to 1 or greater.
+
+The receiver of `query_channel_range`:
+  - if it has not sent `reply_channel_range` to a previously received `query_channel_range` from this sender:
+    - MAY fail the connection.
+  - MUST respond with one or more `reply_channel_range` whose combined range
```
ZmnSCPxj: Multiple `reply_channel_range` may make it ambiguous whether we are violating "MUST NOT send this (`query_channel_range`) if it has sent a previous `query_channel_range` to this peer and not received `reply_channel_range`" requirement.

In particular if there are any data or message buffers between the operating code of the peers (e.g. c-lightning, where sending of messages is buffered in a message queue):

1.  Peer A sends a `query_channel_range`.
2.  Peer B sends two `reply_channel_range` in response.
3.  Peer A receives one `reply_channel_range` and thinks it has fulfilled the above requirement for the first `query_channel_range`.
4.  Peer A sends another `query_channel_range`.
5.  Peer B receives the above message, then gets stalled in processing due to OS scheduling or etc.
6.  Peer A receives the second `reply_channel_range` and thinks it has fulfilled the above requirement for the second `query_channel_range`.
7.  Peer A sends a third `query_channel_range`.
8.  Peer B is awakened by the OS again, and sees another `query_channel_range` even though it has not finished processing the second `query_channel_range`.  It then takes the above clause "MAY fail the connection" and disconnects.

sstone: We could add an explicit  `reply_channel_range_end` message here but I don't think that it is necessary: you can check the `first_blocknum` and `number_of_blocks` fields and tell when you've received all the replies to your query.

ZmnSCPxj: It would be good to clarify what rules exactly apply to determine if the last N `reply_channel_range` has fulfilled the most recent `query_channel_range` or not.  Later the text mentions using canned 1000-block `reply_channel_range`s, for instance; presumably a query is covered if all `reply_channel_range`s have fully covered the range, but for instance is it allowed to return the replies in any order, including random?

sstone: Short channel ids are ordered and replies must be sent in order (but it is not explicit in the current wording when replies span several messages).

07-routing-gossip.md
```diff
@@ -510,16 +521,196 @@ An endpoint node:
...
+gossip should be received.
+
+There are several messages which contain a long array of
+`short_channel_id`s (called `encoded_short_ids`) so we utilize a
+simple compression scheme: the first byte indicates the encoding, the
+rest contains the data.
+
+Encoding types:
+* `0`: uncompressed array of `short_channel_id` types, in ascending order.
+* `1`: array of `short_channel_id` types, in ascending order, compressed with zlib<sup>[1](#reference-1)</sup>
```
Roasbeef: Each new encoding type needs a distinct feature bit. Otherwise they're _all_ mandatory. Some clients may not want to package zlib, nor have to implement defensive decoding measures to ensure an enormous amount of memory isn't allocated due to a maliciously crafted payload.

sstone: Yes, `zlib` and `uncompressed` are mandatory and we'll probably never need to add another format (`zlib` already does a fairly good job).

Roasbeef: I mean why should zlib itself be mandatory? IMO it was a bit of a premature optimization (a few additional round trips really aren't that bad). It _forces_ implementations to implement defensive parsing strategies, or it may become a DoS vector if the spec doesn't explicitly pay out the risks and mitigation strategies.

07-routing-gossip.md
rustyrussell: I've changed the language appropriately:

```diff
 The sender of `query_channel_range`:
-  - MUST NOT send this if it has sent a previous `query_channel_range` to this peer and not received `reply_channel_range`.
+  - MUST NOT send this if it has sent a previous `query_channel_range` to this peer and not received all `reply_channel_range` replies.
```

and

```diff
 The receiver of `query_channel_range`:
-  - if it has not sent `reply_channel_range` to a previously received `query_channel_range` from this sender:
+  - if it has not sent all `reply_channel_range` to a previously received `query_channel_range` from this sender:
     - MAY fail the connection.
```

Since it's made clear that the receiver has to cover the entire range, the definition of `all` seems simple enough to leave implied?

07-routing-gossip.md
rustyrussell: zlib is pretty established and robust, and it's now deployed and thus too late :(

Any other compression methods would require a new feature bit.  For now, don't advertise gossip_queries if you don't support receiving zlib.

@Roasbeef pointed out that feature bits also allow us to do fine-grained disable if problems are found with an implementation.  **Let this be a lesson-learned**: every feature needs a feature bit, even if everyone intends to implement it!

rustyrussell: Meanwhile, I've added a paragraph on worst-case behavior:

```
Note that a 65535-byte zlib message can decompress into 67632120
bytes<sup>[2](#reference-2)</sup>, but since the only valid contents 
are unique 8-byte values, no more than 14 bytes can be duplicated
across the stream: as each duplicate takes at least 2 bits, no valid
contents could decompress to more then 3669960 bytes.
```

This number is very conservative: you still need some literals, and you still need the two-bytes between the 14 duplicated bytes, etc.

rustyrussell:
Appended two commits and rebased since it had a wordlist conflict...

rustyrussell:
Would appreciate final approval on this at next meeting!

sstone:
ACK

rustyrussell:
Rebased and squashed commits, with no textual changes.

cdecker:
Paging Dr. @Roasbeef for a final approval :wink:

PR #396, opened 2018-03-16
Clarify query semantics in BOLT10

### LINKED ISSUE #395: Clarify DNS seed query semantics in BOLT10
It's not immediately clear from reading BOLT10, how exactly the key/value pairs are passed to the DNS seed server.

Having received an explanation from @cdecker on the lightning-dev mailing list, I thought it would be useful to clarify a few things around this in the BOLT10 document.

ctp-tsteenholdt:
I've tried to clarify the query semantics in BOLT10 by briefly exemplifying how to pass the conditions to the DNS seed server. The pull-request aims to fix #395 that I created after a brief exchange with @cdecker on the lightning-dev mailing list.

### DIFF
--- a/10-dns-bootstrap.md
+++ b/10-dns-bootstrap.md
@@ -25,6 +25,12 @@ The following key-value pairs MUST be supported by a DNS seed:
  - `l`: `node_id`, the bech32-encoded `node_id` of a specific node, used to ask for a single node instead of a random selection. (default: null)
  - `n`: the number of desired reply records (default: 25)
 
+Conditions are passed in the DNS seed query as individual, dot-separated subdomain components.
+
+A query for `r0.a2.n10.lseed.bitcoinstats.com` would mean: Return 10 (`n10`) IPv4 (`a2`) records  for nodes supporting Bitcoin (`r0`).
+
+The DNS seed MUST evaluate the conditions from the _seed root domain_ and going up-the-tree, meaning right-to-left in a fully qualified domain name. In the example above, that would be: `n10`, then `a2`, then `r0`.
+If a condition (key) is specified more than once, the DNS seed MUST discard any earlier value for that condition and use the new value instead. For `n5.r0.a2.n10.lseed.bitcoinstats.com`, the result is then: ~~`n10`~~, `a2`, `r0`, `n5`.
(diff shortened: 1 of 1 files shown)


cdecker:
I think the example for address-types should be changed to be `SRV`. The reason being that having an `A` query implies that we are looking for IPv4 addresses, and `AAAA` implies that we are interested in IPv6, the only real ambiguity as to which address type comes when we are querying `SRV` records.

ctp-tsteenholdt:
I totally agree. I have updated the examples to use SRV queries instead.

ctp-tsteenholdt:
Since we're trying to clarify some of the things that may not be entirely clear, there are a few other things it may be relevant to address or define at the same time.

1. What's the _intended direction_ of interpreting the conditions?
2. What's the _result of the conflict_ if a condition is specified multiple times?

So using the thought-up case of `r0.a2.n5.a4.n10.seed.example.com`:

1. Should the DNS seed process the conditions from the seed root and "up the tree" or in the opposite direction?

For the sake of argument, I'm going to assume we want to take a (DNS-wise) logical approach, i.e. interpreting from the seed root and "up the tree", so in the order: `n10`, `a4`, `n5`, `a2`, `r0`.

2. What happens to `a2` when we meet `a4`? Should `a4` replace any current `a` condition or simply be ignored (or in the case of `a` perhaps even merged through a bitwise `or`)? Same question for `n5` when we meet `n10`.

Personally I think it would make sense to always replace currently set conditions, when we hit new conditions with the same key. This would allow us to easily add a condition to a query programmatically, without having to interpret any conditions that may be set already. Prepending a current query string with `a6.` would then guarantee that an earlier condition does not limit my results to either IPv4 or IPv6 nodes. As long as the length of the query string does not exceed a total of 253 chars (not counting the root `.` of course).

What do you think? Also, is this something that should really be discussed on the mailing list?

tmuncks:
@cdecker, I've had a couple of colleagues look through the changes and there seem to be a consensus that these updates improve understanding of the query semantics.

I'm eager to get your input on this, especially the parts that specifically needs agreement:
- Parsing direction
- Conflict handling

Any concerns or suggestions, just let me know and I'll be happy to try to address them all.

PR #397, opened 2018-03-20
Add clarification for Bolt11 prefix for regtest

DanielWeigl:
The document (and also BIP-0173) does not clarify which prefix bitcoin regtest should use, but bitcoin-core seems to have settled on `bcrt` (see https://github.com/bitcoin/bitcoin/issues/12314 ), thus the most concise way for bolt11 would be to use `lnbcrt...`

currently c-lightning and lnd are not compatible if you want to pay invoices in regtest mode, see:
- https://github.com/ElementsProject/lightning/issues/1241
- https://github.com/lightningnetwork/lnd/issues/882

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -324,3 +324,4 @@ scalarMult
 blindingFactor
 Mul
 unlinkable
+regtest


stevenroose:
See https://github.com/bitcoin/bips/pull/664.

And [this comment](https://github.com/bitcoin/bitcoin/issues/12314#issuecomment-375232961) from @sipa:
> regtest is a Bitcoin Core specific test network. It's not mentioned in the BIP because it's not something that affects Bitcoin.

DanielWeigl:
I understand that this is not necessarily something that needs to be strictly defined but oth I think it makes life much easier if all implementations use the same prefix (or make it configurable at least).

And I think only implicit mention it in the `(eg. ... )`  should be enough and wont do any harm - or?

rustyrussell:
I don't see a problem referencing it here; better to be clear.

dpc:
I'm looking around trying to find some arguments why would we even need a different prefix for a regtest network, [and I don't really see any, while there's a lot of reasons not to have it](https://github.com/bitcoin/bitcoin/issues/12314#issuecomment-580846197).

PR #404, opened 2018-04-06
Add MIME type recommendation to BOLT-11

icota:
Not using 'application/x-lightning' because x-prefixes are strongly discouraged as per IETF's RFC 6838: https://tools.ietf.org/html/rfc6838


jimpo:
Thoughts on `application/lightning-invoice` or `application/ln-invoice`? `application/lightning` doesn't feel specific enough. For example, maybe there will be a reason to transport raw LN messages like node_announcements or channel_announcements in a context that requires a MIME type.

icota:
@jimpo You are absolutely right. I prefer `application/lightning-invoice` as it's more consistent with the chosen URI scheme.

Roasbeef:
What's the precise use case for this?

icota:
@Roasbeef LN wallet apps may register themselves as handlers for our MIME type with the OS. When you tap your phone on an NFC payment terminal and data with a MIME type of `application/lightning-invoice` is read Android knows to wake up and forward that (BOLT-11) data to whatever app handles lightning. It's also useful for sharing via bluetooth, email, you name it

Roasbeef:
In that scenario, why would BOLT-11 even be used? As is, the format is meant to be human readable to a degree. When accepting information over NFC, it seems one would just use a custom file format? (which would be another BOLT all together)

icota:
Personally I wouldn't go reinventing the wheel for NFC just because a tiny bit is human readable. BOLT-11 works and we are already encoding it to QR codes which are definitely not human readable. NFC or bluetooth or whatever is just another medium. We already have a URI scheme why not a MIME type too?

rustyrussell:
You can't just self-assign MIME types.  There's actually a process for this; and you refer to the document which tells you that.

The only no-registered name I can see would be `prs.lightning-invoice` which would be a personal/experimental name.  Yech...

philipt18:
What happened with this proposal? Was a MIME type added?

icota:
Not that I know of.

PR #412, opened 2018-04-16
add file for CCv4

EvertonMelo:
Create file License in repository

PR #413, opened 2018-04-16
README adequation

EvertonMelo:
adequation of #412


rustyrussell:
Since #412 was dropped for being incorrect, I'm closing this one, too.

PR #411, opened 2018-04-16
BOLT-0 Glossary Reorg & Expansion

cdecker:
Supersedes PR #384 since I couldn't push to @shannona's branch. Quoting from the original PR:

>This is a big change to the glossary in BOLT-0.
>
> The overarcing changes are:
>
> 1.) Reorganized glossary to be alphabetical
> 2.) Add copious "See also:" lines to cross-reference containers, contents, and other related entries.
> 
> Changed Definitions: Peer and node were revised based in on discussions with Rusty; commitment revocation secret key was changed to be commitment revocation private key based on #377.
>
> New Definitions: New words added based on their usage in the bolts were: announcement, closing transaction, final node, hop, origin node, processing, node, receiving node, receiving peer, revoked commitment transaction, sending node, sending peer.
>
> (I suspect the reorg will make the diffs almost useless, so this is the list of stuff to look at.)

PR #415, opened 2018-04-18
BOLT0: Add introduction and background info to index

cdecker:
This was spun out of #411 (which itself was a second take on #384), as discussed during the 2018/04/16 spec meeting. We felt that the glossary changes are a really good addition on their own, but that the definition and explanation of LN would need separate sign-off.


cdecker:
I manually added @shannona as author, since this is his excellent work, I just happen to rebase and squash the commits :+1;

PR #417, opened 2018-04-25
BOLT4: `channel_update` is mandatory in `UPDATE` error messages

sstone:
We think that `mandatory` is more in line with the rest of the BOLT, but in any case the wording in the specs leaves room for interpretation (we believe it is mandatory and lnd for example believes it is optional) so it needs to be specified.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -655,6 +655,8 @@ The top byte of `failure_code` can be read as a set of flags:
 * 0x2000 (NODE): node failure (otherwise channel)
 * 0x1000 (UPDATE): new channel update enclosed
 
+Please note that the `channel_update` field is mandatory in `UPDATE` messages.
+
 The following `failure_code`s are defined:
 
 1. type: PERM|1 (`invalid_realm`)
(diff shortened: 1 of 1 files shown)


sstone:
There was a discussion about this some time ago (see https://github.com/lightningnetwork/lightning-rfc/pull/149) and it seems that the intent was indeed to make this field mandatory. 
However, if the rationale is that an update is not needed because if it was then the sender would have used a more specific error message then I'd rather we remove the UPDATE flag and the channel update from `temporary_channel_failure`, it would be cleaner (I still prefer the "mandatory" option though).

rustyrussell:
Yes, this seems correct.  If you can't give a channel_update, you need some generic error.

04-onion-routing.md
```diff
@@ -655,6 +655,8 @@ The top byte of `failure_code` can be read as a set of flags:
 * 0x2000 (NODE): node failure (otherwise channel)
 * 0x1000 (UPDATE): new channel update enclosed
 
+Please note that the `channel_update` field is mandatory in `UPDATE` messages.
```
cdecker: "in `UPDATE` messages" -> "in messages whose type includes the `UPDATE` bit"?

PR #419, opened 2018-04-26
BOLT2: Rephrase channel establishment introduction

C-Otto:
A channel may be established, but this is not mandatory. Furthermore, the connection initialization has to happen first.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -30,8 +30,8 @@ operation, and closing.
 
 ## Channel Establishment
 
-Channel establishment begins immediately after authentication and
-consists of the funding node (funder) sending an `open_channel` message,
+After authentication and initializing a connection, channel establishment may begin.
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -30,8 +30,8 @@ operation, and closing.
 
 ## Channel Establishment
 
-Channel establishment begins immediately after authentication and
-consists of the funding node (funder) sending an `open_channel` message,
+After authentication and initializing a connection, channel establishment may begin.
```
Roasbeef: I guess it SHOULD begin. Right? We want to promote liveness of channels.

cdecker: I think there are many good cases to have a connection in gossiping state. Random connection to learn the topology for example.

Roasbeef: Nvm, thought this was re-establish!

PR #435, opened 2018-05-28
BOLT 1: add testnet port

jonasnick:
Running current lightning and test network lightning nodes on the same host is currently needlessly complicated because both nodes try to listen at the same port. Standardizing a port for testnet would also simplify communicating node information. 19735 doesn't seem to have relevant usage elsewhere.


rustyrussell:
No, the protocol already handles different coins; there's no need to differentiate regtest, testnet, litecoin etc. 

Implementations could do a better job of realizing that a node has nothing interesting to say, however.

PR #436, opened 2018-05-29
BOLT 11: Break lines in route example

ysangkok:


cdecker:
Contents match, I propose applying this under the `typo-fix` rule

Issue #442, opened 2018-06-16
Clarify if amount multiplier in payment request is optional


btcontract:
https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#requirements-1 says that multiplier is obligatory:

> A reader SHOULD indicate if amount is unspecified, otherwise it MUST multiply amount by the multiplier value (if any) to derive the amount required for payment.

But later in https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#human-readable-part:

> amount: optional number in that currency, followed by an optional multiplier letter

There's a real world example of payment request without multiplier: https://github.com/btcontract/lnwallet/issues/58. If multiplier is optional, what denomination should be considered a default one?

PR #445, opened 2018-06-27
BOLT 11: reword Human-Readable Part Requirements in modern form.

rustyrussell:
a.k.a. landonmutch style, where requirements are explicit and bullet-pointed.

This also tightens requirements:
1. New requirement (previously implied) that writer use correct prefix.
2. Reader MUST rather than SHOULD fail malformed `amount`.

It also makes it clear that writer can omit the multiplier.

Fixes: #442

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -77,8 +77,6 @@ A reader:
 
 The `amount` is encoded into the human readable part, as it's fairly
 readable and a useful indicator of how much is being requested.
-The `multiplier` allows a more compressed representation for large
-amounts than presenting everything in millisatoshi.
 
 Donation addresses often don't have an associated amount, so `amount`
 is optional in that case. Usually a minimum payment is required for
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -54,21 +54,31 @@ The following `multiplier` letters are defined:
 
 ## Requirements
 
-A writer MUST include `amount` unless it will accept any payment amount.
-A writer MUST encode `amount` as a positive decimal integer with no leading zeroes and SHOULD use the shortest representation possible.
+A writer:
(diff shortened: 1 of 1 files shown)


11-payment-encoding.md
```diff
@@ -54,21 +54,31 @@ The following `multiplier` letters are defined:
...
+A reader:
+  - MUST fail if it does not understand the `prefix`
+  - If the `amount` is empty:
+	- SHOULD indicate if amount is unspecified
+  - Otherwise:
+	- MUST fail if `amount` contains a non-digit or is followed by
+      anything except a `multiplier` in the table above
+    - If the `multiplier` is present:
+	  - MUST multiply `amount` by the `multiplier` value to derive the
+        amount required for payment
```
toadlyBroodle: Looks good. Just for consistency sake: operators (if, otherwise, etc.) are lowercase and predicate phrases (MUST, SHOULD, etc.) are concluded with periods. I should be going over BOLT 11 soon at any rate, so I can make these changes then also.

11-payment-encoding.md
```diff
@@ -54,21 +54,31 @@ The following `multiplier` letters are defined:
...
 
 ## Rationale
 
 The `amount` is encoded into the human readable part, as it's fairly
 readable and a useful indicator of how much is being requested.
+The `multiplier` allows a more compressed representation for large
+amounts than presenting everything in millisatoshi.
```
cdecker: Actually the amounts are expressed in Bitcoin (hence the milli, nano, pico multipliers).

cdecker: But it should read bitcoins here, not millisatoshi. This is bound to create confusion otherwise.

rustyrussell: I will delete this sentence.

PR #458, opened 2018-07-26
added a list gossip protocol messanges

renepickhardt:
I have created a list of all gossip protocol messages which are being used for easy reference and lookup. 
Had to make it in mediawiki format instead of markdown since wikitables seem to be more expressive than markdown tables. I guess the appendix should be linked eather from BOLT0 or in BOLT1,2 and 7 respectively when the messanges are first introduced. I didn't put that in yet to make atomic PRs


ysangkok:
Why hardcode this when it could be generated with the tools provided in the repo?

rustyrussell:
These are not gossip messages, this looks like a complete protocol message index?

Anyway, yes, we should enhance the tools/ to generate it!

judondike38 (changes requested):
I need to add my bank account?

PR #462, opened 2018-08-03
Fix offered htlc to_local branch missing CLTV

junderw:
it says `# To local node via HTLC-timeout transaction (timelocked).` but no timelock OP anywhere in the script


junderw:
https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#htlc-timeout-and-htlc-success-transactions

hehe, RTFM

PR #465, opened 2018-08-07
Max capacity in channel update

### LINKED ISSUE #460: BOLT 7: Lack of capacity field in channel_announcement makes life difficult
What is the rationale for not including channel capacity into the channel_announcement message?

As things stand now, the only way to determine that is to check the blockchain, which is hard for mobile and/or light wallets. Even worse, the data is only there in the form of short_channel_id, which is incompatible with most block explorer APIs.

In other words, having to have access to the blockchain turns a route-finding tool from a simple program into a thing laden with many dependencies in order to fetch and parse the transactions. Not knowing the capacity of the channels drops routing success rates dramatically, especially for larger payments.

Is there a good reason to force us to do so that i'm missing?

rustyrussell:
Fixes #460

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -372,6 +372,7 @@ of *relaying* payments, not *sending* payments. When making a payment
     * [`8`:`htlc_minimum_msat`]
     * [`4`:`fee_base_msat`]
     * [`4`:`fee_proportional_millionths`]
+    * [`8`:`htlc_maximum_msat`] (option_channel_htlc_max)
 
 The `channel_flags` bitfield is used to indicate the direction of the channel: it
 identifies the node that this update originated from and signals various options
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -382,6 +384,19 @@ individual bits:
 | 0             | `direction` | Direction this update refers to. |
 | 1             | `disable`   | Disable the channel.             |
 
+The `message_flags` bitfield is used to indicate the presence of optional
+fields in the `channel_update` message:
+
+| Bit Position  | Name                      | Field                            |
+| ------------- | ------------------------- | -------------------------------- |
+| 0             | `option_channel_htlc_max` | `htlc_maximum_msat`              |
```
cdecker: I'd rather have this in the `channel_announcement`, maybe we can use the features bitfield to signal this? Anyway I expect everybody will just add this in perpetuity so a bit is not really needed.

cdecker:
ACK b77232eaa4ccc9c73318b434621469a5252468c0

halseth:
ACK

cdecker:
Needs one ack from either @sstone or @pm47 :-)

sstone:
ACK

PR #481, opened 2018-09-17
Make unit explicit encoded in BOLT11 invoice

Christewart:
I keep forgetting what the base currency unit is in BOLT11, so make it explicit.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -43,7 +43,7 @@ and MUST fail if the checksum is incorrect.
 The human-readable part of a Lightning invoice consists of two sections:
 1. `prefix`: `ln` + BIP-0173 currency prefix (e.g. `lnbc` for bitcoin mainnet, `lntb` for bitcoin testnet and `lnbcrt` for bitcoin regtest)
 1. `amount`: optional number in that currency, followed by an optional
-   `multiplier` letter
+   `multiplier` letter. The unit encoded here is the 'social' convention of a payment unit -- in the case of Bitcoin the unit is 'bitcoin' NOT satoshis.
 
 The following `multiplier` letters are defined:
 


rustyrussell:
"For the case of bitcoin, `amount` is in bitcoin, not satoshis"?

11-payment-encoding.md
```diff
@@ -43,7 +43,7 @@ and MUST fail if the checksum is incorrect.
 The human-readable part of a Lightning invoice consists of two sections:
 1. `prefix`: `ln` + BIP-0173 currency prefix (e.g. `lnbc` for bitcoin mainnet, `lntb` for bitcoin testnet and `lnbcrt` for bitcoin regtest)
 1. `amount`: optional number in that currency, followed by an optional
-   `multiplier` letter
+   `multiplier` letter. The unit encoded here is the 'social' convention of a payment unit -- in the case of Bitcoin it is 'bitcoin' NOT satoshis.
```
Roasbeef: Well you can't have a fraction of a satoshi, but I think you're saying here it should be the largest unit socially accepted?

Christewart: Isn't there an implicit assumption here that there is a 'base unit', and the multiplier is applied to that base unit? I guess I'm trying to explicitly say what the base unit is.

Christewart:
> "For the case of bitcoin, `amount` is in bitcoin, not satoshis"?

I was trying to be generic for multiple blockchains. In the case of the bitcoin blockchain, the "socially accepted" currency unit is bitcoin -- not satoshis.

11-payment-encoding.md
Christewart: Also you can't have a fraction of a satoshi, but in the current scheme it is possible to encode an invoice with a fraction of a satoshi no? `1p`?

halseth: You can pay invoices of `mSAT` value.

11-payment-encoding.md
```diff
@@ -43,7 +43,7 @@ and MUST fail if the checksum is incorrect.
 The human-readable part of a Lightning invoice consists of two sections:
 1. `prefix`: `ln` + BIP-0173 currency prefix (e.g. `lnbc` for bitcoin mainnet, `lntb` for bitcoin testnet and `lnbcrt` for bitcoin regtest)
 1. `amount`: optional number in that currency, followed by an optional
-   `multiplier` letter
+   `multiplier` letter. The unit encoded here is the 'social' convention of a payment unit -- in the case of Bitcoin the unit is 'bitcoin' NOT satoshis.
```
dcousens: `default unit` ?

PR #484, opened 2018-09-22
BOLT 2: readability edits

niftynei:
Reading through BOLT 2, I noticed a couple of things that could make it easier to understand. There's a few things I've added that I'm not certain are correct, I've annotated them with a comment.

### DIFF
--- a/00-introduction.md
+++ b/00-introduction.md
@@ -128,7 +128,7 @@ See [BOLT #11: Invoice Protocol for Lightning Payments](11-payment-encoding.md) 
    Each *peer* holds the other peer's signature for this transaction, so that each
    always has a commitment transaction that it can spend. After a new
    commitment transaction is negotiated, the old one is *revoked*.
-   * _See parts: commitment number, commitment revocation private key, HTLC, per-commitment secret_
+   * _See parts: commitment number, commitment revocation private key, HTLC, per-commitment secret, outpoint_
    * _See related: closing transaction, funding transaction, penalty transaction_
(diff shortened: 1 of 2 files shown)


00-introduction.md
```diff
@@ -185,6 +185,10 @@ See [BOLT #11: Invoice Protocol for Lightning Payments](11-payment-encoding.md)
    * _See category: node_
    * _See related: final node, processing node_
 
+* *Outpoint*:
```
niftynei: Not technically a BOLT 2 change, but it's used frequently enough that I realized it'd be nice to formally define. I recognize that it's pretty frequently used terminology within the bitcoin space, but it's nice to have it spelled out.

02-peer-protocol.md
```diff
@@ -1080,6 +1085,11 @@ messages are), they are independent of requirements here.
    * [`32`:`your_last_per_commitment_secret`] (option_data_loss_protect)
    * [`33`:`my_current_per_commitment_point`] (option_data_loss_protect)
 
+`next_local_commitment_number`: A commitment number is a 48-bit
+incrementing counter for each commitment transaction; counters
+are independent for each peer in the channel and start at 0.
+They're only relayed to the other node in the case of re-establishment.
```
niftynei: Per this BOLT, it appears to be the case that the `commitment_number` is only relayed in this particular message, but I didn't do an exhaustive check.

niftynei:
@rustyrussell updated!

PR #485, opened 2018-09-28
gossip: clarify what it means to 'discard a channel'

niftynei:
It's unclear what it means to `discard` a channel. Instead,
let's use the common phrase to `fail` a channel.

I'm assuming that `fail` is the same as `discard` here, but maybe it's meant to say discard the `message`? (But that seems unlikely, given the context).

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -480,7 +480,7 @@ The final node:
   - otherwise:
     - if `htlc_maximum_msat` is not present or greater than channel capacity:
 	  - MAY blacklist this `node_id`
-	  - SHOULD discard this channel.
+	  - SHOULD ignore this channel during route considerations.
 	- otherwise:
 	  - SHOULD consider the `htlc_maximum_msat` when routing.
 


niftynei:
Thanks for the clarification. I updated the commit to just add a bit of clarifying text. 

As you mention, the behavior for what to do if this is one of your immediate peers isn't immediately clear from the spec. The `htlc_maximum_msat` is provided mainly as a guideline for route-finding nodes -- should your channel peers care that you're advertising weird data?

07-routing-gossip.md
```diff
@@ -480,7 +480,7 @@ The final node:
   - otherwise:
     - if `htlc_maximum_msat` is not present or greater than channel capacity:
 	  - MAY blacklist this `node_id`
-	  - SHOULD discard this channel.
+	  - SHOULD discard this channel (from route considerations).
```
Roasbeef: How about ignore instead of discard?

niftynei:
Works for me. Updated.

PR #490, opened 2018-10-19
BOLT 4: clarify the maximum length of the route

araspitzu:
Clarify the maximum length of the route by removing any ambiguity with the term "hops".

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -73,7 +73,7 @@ A node:
 
 There are a number of conventions adhered to throughout this document:
 
- - Length: the maximum route length is limited to 19 _intermediate nodes_, the _origin node_ and _final node_ do not count toward the limit.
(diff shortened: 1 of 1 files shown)


rustyrussell:
This is wrong: it's limited to 19 intermediate nodes.  (20 channels)

04-onion-routing.md
```diff
@@ -73,7 +73,7 @@ A node:
 
 There are a number of conventions adhered to throughout this document:
 
- - Length: the maximum route length is limited to 20 hops.
+ - Length: the maximum route length is limited to 19 _intermediate nodes_, the _origin node_ and _final node_ do not count toward the limit.
```
halseth: I think we should keep the "20 hops" wording. The change can instead be an added clarification.

araspitzu: Sounds good to me, after all whatever removes the ambiguity will do.

Issue #492, opened 2018-10-22
Sighash type unspecified


OrfeasLitos:
The correct way to sign the input of a commitment transaction is not explicitly specified. In particular, the sighash type is not mentioned. In particular, [BOLT #2](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md) says:
> The sender MUST set: [...] `signature` to the valid signature using its `funding_pubkey` for the initial commitment transaction, as defined in [BOLT #3](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md).

And:

> The recipient: if `signature` is incorrect: MUST fail the channel.

No more details are added in [BOLT #3](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md).

After reading [BOLT #2](https://github.com/lightningnetwork/lightning-rfc/blob/master/02-peer-protocol.md) and [BOLT #3](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md) and searching the [lnd](https://github.com/lightningnetwork/lnd/) code, I deduced from context that all signatures are of `SIGHASH_ALL` type. Please correct me if I am wrong and consider explicitly stating how to correctly sign in the specification.

OrfeasLitos:
I created a relevant PR: #498

PR #498, opened 2018-10-31
Specify valid signature hash flags (SIGHASH_ALL)

OrfeasLitos:
Solves #492.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -806,7 +806,7 @@ HTLC-success) results in weights of:
 In the following:
  - It's assumed that *local* is the funder.
  - Private keys are displayed as 32 bytes plus a trailing 1 (Bitcoin's convention for "compressed" private keys, i.e. keys for which the public key is compressed).
- - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256).
+ - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256). A valid signature MUST sign all inputs and outputs of the relevant transaction (i.e MUST be created with a `SIGHASH_ALL` [signature hash](https://bitcoin.org/en/glossary/signature-hash)), unless explicitly stated otherwise. Note that clients MUST send the signature in compact encoding and not in Bitcoin-script format, thus the signature hash byte is not transmitted.
 
 The input for the funding transaction was created using a test chain
 with the following first two blocks; the second block contains a spendable


TheBlueMatt:
Cool! To nitpick wording, though, this seems to imply that you have to send the signature in Bitcoin-script format (ie including the sighash byte), when we currently only send the signature itself in compact encoding.

OrfeasLitos:
So, to clarify, currently each signature is created by signing all inputs and outputs (i.e. SIGHASH_ALL) but only the signature (compressed in some way) is sent over the wire, without any other data/flags?

OrfeasLitos:
Rephrased this part. It may have become a bit too verbose.

OrfeasLitos:
Let me know if you want any more changes before merging.

03-transactions.md
```diff
@@ -806,7 +806,7 @@ HTLC-success) results in weights of:
 In the following:
  - It's assumed that *local* is the funder.
  - Private keys are displayed as 32 bytes plus a trailing 1 (Bitcoin's convention for "compressed" private keys, i.e. keys for which the public key is compressed).
- - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256).
+ - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256). A valid signature MUST sign all inputs and outputs of the relevant transaction (i.e MUST be created with a `SIGHASH_ALL` [signature hash](https://bitcoin.org/en/glossary/signature-hash)). Note that clients MUST send the signature in compact encoding and not in Bitcoin-script format, thus the signature hash byte is not transmitted.
```
halseth: maybe add that it is `SIGHASH_ALL` unless explicitly specified, to allow future changes.

03-transactions.md
```diff
@@ -806,7 +806,7 @@ HTLC-success) results in weights of:
 In the following:
  - It's assumed that *local* is the funder.
  - Private keys are displayed as 32 bytes plus a trailing 1 (Bitcoin's convention for "compressed" private keys, i.e. keys for which the public key is compressed).
- - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256).
+ - Transaction signatures are all deterministic, using RFC6979 (using HMAC-SHA256). A valid signature MUST sign all inputs and outputs of the relevant transaction (i.e MUST be created with a `SIGHASH_ALL` [signature hash](https://bitcoin.org/en/glossary/signature-hash)), unless explicitly stated otherwise. Note that clients MUST send the signature in compact encoding and not in Bitcoin-script format, thus the signature hash byte is not transmitted.
```
renepickhardt: The problem with the url https://bitcoin.org/en/glossary/signature-hash is that it is just the masterbranch of some wiki (I understand it is rather unlikely to change) anyway wouldn't it make more sense to have wikipedia like permalink: https://github.com/bitcoin-dot-org/bitcoin.org/blob/c047e09f079acdb93fcbecffd5a4b891051519c3/_data/devdocs/en/guides/transactions.md#L549 (my suggestion is actually bad in the sense that it does not point directly to the information) maybe we can have a combination using your link but pointing out that it is commit c047e09f079acdb93fcbecffd5a4b891051519c3 or the bitcoin-dot-org repo?

OrfeasLitos: Is there a standard way of handling links that can possibly break in this project? I assume this is not the first time such an issue arises. In case this discussion has not taken place before, I see three options:
* Use the non-perma-link as it is now. Hope that it doesn't change, or, if it changes, someone will point it out and we will manually fix it.
* Add both links, the second in the lines of "[signature hash](https://bitcoin.org/en/glossary/signature-hash) (as per [this](https://github.com/bitcoin-dot-org/bitcoin.org/blob/c047e09f079acdb93fcbecffd5a4b891051519c3/_data/devdocs/en/guides/transactions.md#L549) definition)" or "[signature hash](https://bitcoin.org/en/glossary/signature-hash) ([permalink](https://github.com/bitcoin-dot-org/bitcoin.org/blob/c047e09f079acdb93fcbecffd5a4b891051519c3/_data/devdocs/en/guides/transactions.md#L549) to the intended site code)" or similar.
* Provide only the permalink.
* Do not provide links at all.
* The previous + provide the definition of the signature hash in the specification.

I would go with the first option. The provided link is not likely to change, so a little manual intervention in the future is not too much to ask for. Anyway, I believe that the most intensive use of this specification will take place when the development of related wallets is most active, i.e. in the near rather than far future.

The second option is rather unclear to the reader and is still inconsistent in case the first link breaks, so it only provides a temporary patch until we manually change it again. Furthermore, it will be highly confusing to the reader if the first link starts pointing to a different definition of the signature hash (the latter is highly unlikely).

The third and especially fourth option create the obvious and bigger problem that the reader is left to find the definition of the signature hash by herself. I believe that a link to the definition of the signature hash will be very helpful, especially since this is an obscure bitcoin construct.

Lastly, the fifth option does not follow the DRY principle. Defining core bitcoin constructs is not the purpose of the current project.

To sum up, the problem of broken links is not one that will be solved by this project.

PR #506, opened 2018-11-12
07-routing-gossip.md: Describe standard human-readable format for short channel ID.

ZmnSCPxj:
Uncertain if this is best place to describe human-readable format, but, this section is what describes the components of the scid.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -57,6 +57,15 @@ It is constructed as follows:
   2. the next 3 bytes: indicating the transaction index within the block
   3. the least significant 2 bytes: indicating the output index that pays to the channel.
 
+The standard human readable format for `short_channel_id` is created
+by printing the above components, in the order:
+block height, transaction index, and output index.
+Each component is printed as a decimal number,
+and separated from each other by the small letter `x`.
+For example, a `short_channel_id` might be written as `539268x845x1`,
+indicating a channel on the output 1 of the transaction at index 845
+of the block at height 539268.
+
 A node:
   - if the `open_channel` message has the `announce_channel` bit set AND a `shutdown` message has not been sent:
     - MUST send the `announcement_signatures` message.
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/506/commits/8a4e3fae4dc2646523cfa693d89d0d760023f71c

07-routing-gossip.md
```diff
@@ -57,6 +57,15 @@ It is constructed as follows:
...
   3. the least significant 2 bytes: indicating the output index that pays to the channel.
 
+The standard human readable format for `short_channel_id` is created
+by printing the above components, in the order:
+block height, transaction index, and output index.
+Each component is printed as a decimal number,
+and separated from each other by the small letter `x`.
+For example, a `short_channel_id` might be written as `539268x845x1`,
+indicating a channel on the output 1 of the transaction at index 845
+of the block at height 539268.
```
halseth: I'm not a native speaker, but should it be "index 845 *in* the block?

ZmnSCPxj: Construction "x of y" can mean "x is a part of y" or "x comes from y" or "x is owned by y", by my understanding of this language.  So we can consider "the transaction at index N of the block" as the transaction (which happens to be at index N) that is part of the block.

07-routing-gossip.md
```diff
@@ -76,6 +85,17 @@ A recipient node:
   - if it has sent AND received a valid `announcement_signatures` message:
     - SHOULD queue the `channel_announcement` message for its peers.
 
+#### Rationale
+
+The `short_channel_id` human readable format is designed
+so that double-clicking or double-tapping it will select the entire ID
+on most systems.
```
renepickhardt: ... on most operating systems ...

Jagkree:
ดึงสิ

PR #507, opened 2018-11-13
04-onion-routing.md: Prepare for extra onion blob encoding.

ZmnSCPxj:


ZmnSCPxj:
I used `per_hop_type` instead of agree upon `type`, but I feel `type` is imprecise as we also use the same term for message ID.

ZmnSCPxj:
I have also named `per_hop_type` 0 to `normal_per_hop`.

04-onion-routing.md
```diff
@@ -161,22 +161,22 @@ and has the following structure:
 
 1. type: `hops_data`
 2. data:
-   * [`1`:`realm`]
+   * [`1`:`per_hop_type`]
```
Roasbeef: Naming suggestion: packet type?

ZmnSCPxj: That does seem better.

t-bast:
Closing this old PR, superceded by #619

PR #508, opened 2018-11-14
BOLT 2: advise ping-before-commitment_signed on quiescent connections.

rustyrussell:
This seems to be a cause of stuck HTLCs on the network: c-lightning has
done this for a while now (since 0.6.1).

Decided-at: Adelaide Summit 2018

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -936,6 +936,8 @@ change the commitment transaction aside from the new revocation hash
 fee changes).
   - MUST include one `htlc_signature` for every HTLC transaction corresponding
   to BIP69 lexicographic ordering of the commitment transaction.
+  - if it has not recently received a message from the remote node:
+      - SHOULD use `ping` and await the reply `pong` before sending `commitment_signed`.
 
 A receiving node:
   - once all pending updates are applied:
@@ -953,6 +955,12 @@ A receiving node:
 There's little point offering spam updates: it implies a bug.
 
 The `num_htlcs` field is redundant, but makes the packet length check fully self-contained.
+
+The recommendation to require recent messages recognizes the reality
+that networks are unreliable: nodes might not realize their peers are
+offline until after sending `commitment_signed`.  Once
+`commitment_signed` is sent, the sender considers itself bound to
+those HTLCs, and cannot close incoming HTLCs.
 
 ### Completing the Transition to the Updated State: `revoke_and_ack`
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -954,6 +956,12 @@ There's little point offering spam updates: it implies a bug.
 
 The `num_htlcs` field is redundant, but makes the packet length check fully self-contained.
 
+The recommendation to require recent messages recognizes the reality
+that networks are unreliable: nodes might not realize their peers are
+offline until after sending `commitment_signed`.  Once
+`commitment_signed` is sent, the sender considers itself bound to
+those HTLCs, and cannot close incoming HTLCs.
```
```suggestion
those HTLCs, and cannot instantly fail related incoming HTLCs anymore.
```

02-peer-protocol.md
```diff
@@ -936,6 +936,8 @@ change the commitment transaction aside from the new revocation hash
 fee changes).
   - MUST include one `htlc_signature` for every HTLC transaction corresponding
   to BIP69 lexicographic ordering of the commitment transaction.
+  - if it has not recently received a message from the remote node:
+      - SHOULD use `ping` and await the reply `pong` before sending `commitment_signed`.
```
pm47: Can we also say alternatively that a node should regularly ping its peers? That's easier to implement and achieves almost the same goal.

TheBlueMatt:
Note that if you regularly ping your peers and disconnect if they take too long to pong you're guaranteed to always have "recently received a message from the remote node" so you'd be fine that way too, even with the current wording.

renepickhardt:
just to make sure I get this correctly. this does not mean we ping every node in the route that we are creating in the onion but just our direct channel partner. 
Since we assume every node on the route does so they will send an error message if they can't forward the HTLC since the pong did not arrive in time. 

What are the disadvantages with my suggestion that the payee pings all hopes on the path of the designated onion package? If any node does not pong in time we can reduce the traffic on all other nodes on the path up to that node.

02-peer-protocol.md
ZmnSCPxj: Agree with @pm47, "close" for HTLC seems wrong/too undetailed here.  Both @pm47  and @rustyrussell rewordings seem fine.

02-peer-protocol.md
```diff
@@ -954,6 +956,13 @@ There's little point offering spam updates: it implies a bug.
 
 The `num_htlcs` field is redundant, but makes the packet length check fully self-contained.
 
+The recommendation to require recent messages recognizes the reality
+that networks are unreliable: nodes might not realize their peers are
+offline until after sending `commitment_signed`.  Once
+`commitment_signed` is sent, the sender considers itself bound to
+those HTLCs, and cannot fail incoming the related incoming HTLCs until the
```
ZmnSCPxj: Wording problem? "fail incoming the related incoming"...?

PR #524, opened 2018-12-05
WIP: Dual Funding (v2 Channel Establishment protocol)

niftynei:
This is a first draft proposal for adding a second open channel version. This will empower either side of the channel to contribute to the funding transaction.

Broadly speaking, the largest change between this and version one of `open_channel` is that both sides now construct and broadcast the funding transaction. This adds a few more communication rounds, as the commitment transaction signatures can't be exchanged until both sides have the info necessary to construct the funding transaction.

Errata
- ~Employs strawman idea of using a sha of the revocation points for the channel_id as put forth here https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-December/001709.html~
- TODO: Needs size limits on witness data submissions
- Includes handling for RBF


02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+The receiving node:
+  - if is the `opener`:
+    - MUST fail the channel if:
+      - the sum of `input_info` and `output_info` is greater than
+        the `put_limit`.
+  - otherwise:
+    - MUST fail the channel if:
+      - `change_satoshis` would be negative.
+  - if has not yet sent a `funding_compose`:
+    - MUST send their `funding_compose` message.
```
rustyrussell: s/their/its/

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - otherwise:
+    - MUST use the sent and received `input_info` and `output_info`
+      to create the funding transaction, using `max_extra_witness_len`
```
rustyrussell: It's now called 'max_witness_len'.

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    - MAY send zero inputs and/or outputs.
+
+The receiving node:
+  - if is the `opener`:
+    - MUST fail the channel if:
+      - the sum of `input_info` and `output_info` is greater than
+        the `put_limit`.
+  - otherwise:
+    - MUST fail the channel if:
+      - `change_satoshis` would be negative.
```
rustyrussell: It may not be able to figure this out yet (fees).  So, perhaps this should be 'sum of input amounts < sum of output amounts'?  And add the change_satoshis test only once it has calculated the tx?

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - if has not yet sent a `funding_compose`:
+    - MUST send their `funding_compose` message.
+  - otherwise:
+    - MUST use the sent and received `input_info` and `output_info`
+      to create the funding transaction, using `max_extra_witness_len`
+      for each `input_info` and `feerate_per_kw_funding` as specified in
+      `open_channel2`.
+    - MUST send `commitment_signed2`.
```
rustyrussell: Can we just use `commitment_signed`.  It's identical, and performs the identical function, so I really think it deserves the same name...

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+
+#### Rationale
+Each node must have a complete set of the transaction inputs and outputs,
+to derive the funding transaction hash. This avoids information
+asymmetry between the nodes, as both sides share their input utxo set.
+
+`satoshis` is the value of the input or output. The opener must provide
+one change address for the transaction, which is flagged with a
+`satoshis` value of zero.
```
rustyrussell: You said this below as well; delete one?  We try to avoid spec-sounding language (i.e. "must") in Rationales to avoid confusion; I would say "The opener provides one output with `satoshis` equal zero, as a change address once the amount of change can be calculated".

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+`input_info`.`script` is the scriptPubkey data for the input.
+NB: for native SegWit inputs (P2WPKH and P2WSH) inputs, the `script` field will be empty. See [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#examples).
+
+The opening node must include at least one output with `satoshis` value
+of zero. This will be used for change, or discarded if its value is
+be below `dust_limit_satoshis`. Change is calculated as the sum of
+the opener's `input_info`.`satoshis` minus the estimated
+funding transaction size times the `feerate_per_kw_funding` minus the
+`funding_satoshis` minus all other `output_info`.`satoshis`.
```
rustyrussell: This Fee Calculation probably belongs in BOLT 3 with the others, FWIW.

rustyrussell: eg. it needs to / 1000 somewhere, as those do, and specify rounding....

02-peer-protocol.md
rustyrussell: Hmm, actually, I think this being negative should be allowed: in that case, funding_satoshis should be reduced.  In effect, this allows you to say "I don't want change, take the fee out of my initial balance".  Or perhaps we should make the change output optional for this case?

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+### The `funding_signed2` Message
+
+This message exchanges the witness data for the inputs that were
+originally sent in the `funding_compose` message.
+
+1. type: 60 (`funding_signed2`)
+2. data:
+  [`32`:`channel_id`]
+  [`2`:`num_witnesses`]
+  [`num_witnesses`\*`witness_stack`]
```
rustyrussell: Hmm, perhaps leave the * inside the quotes to avoid this \ ugliness?  That's what we do for [`num_htlcs*64`:`htlc_signature`] for example.

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+1. subsubtype: `witness_element`
+2. data:
+  * [`2`:`len`]
+  * [`len`:`witness`]
+
+#### Requirements
+The sending node:
+  - MUST set `witness` to the marshalled witness data for each of its
+    inputs, in funding transaction order. FIXME: add reference
```
rustyrussell: I realize this is quoting me, but the term I should have used is 'serialized' which IIRC is what the BIPs use.

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - if the recipient was the sender of `open_channel2`:
+    - SHOULD reply with its own `funding_signed2`.
+  - SHOULD apply `witness`es to the funding transaction and
+    broadcast the result.
+
+#### Rationale
+Exchanging witness data allows both sides to broadcast the funding
+transaction and maintains information symmetry.
+
+### The `funding_locked2` Message
```
rustyrussell: Again, prefer direct 'funding_locked' reuse.

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+indicated above.
+
+1. type: 62 (`init_rbf`)
+2. data:
+  * [`32`:`channel_id`]
+  * [`8`:`channel_reserve_satoshis`]
+  * [`4`:`feerate_per_kw`]
+  * [`4`:`feerate_per_kw_funding`]
+  * [`2`:`num_inputs`]
+  * [`num_inputs`\*`input_info`]
```
rustyrussell: num_additional_inputs might be clearer?

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - MUST include at least one output with the `satoshis` set to zero,
+    to be used as the change output.
+
+The receiving node:
+  - MUST return an error if:
+    - the `init_rbf` message does not include a change output
+      e.g. an `output_info` entry  with `satoshis` set to zero.
+    - the `feerate_per_kw_funding` is not greater than the previously
+      negotiated rate.
+    - `change_satoshis` is negative.
```
rustyrussell: I think we should allow "extract from channel" in this case, as well.

02-peer-protocol.md
```diff
@@ -390,6 +400,377 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - MUST have sent `open_channel2`.
+  - MUST send a `feerate_per_kw_funding` greater than the previously
+    negotiated rate.
+  - MUST update the `channel_reserve_satoshis` to reflect the new input balance.
+  - MAY update the `feerate_per_kw`, the commitment transaction feerate.
+  - MAY include new, additional inputs.
+  - MAY set the `num_inputs` to zero.
+  - MUST transmit all outputs.
+  - MUST include at least one output with the `satoshis` set to zero,
+    to be used as the change output.
```
rustyrussell: Maybe review this depending on decision about non-change channels above...

02-peer-protocol.md
niftynei: Right. The rationale was that it might be nice to decouple the two flows as that would allow future modification without needing to worry much about backwards compatibility.

02-peer-protocol.md
niftynei: This seems like a valid use case. Will remove the '`change_satoshis` must be positive' requirement.

Allowing negative change amounts means the resulting channel balance will be less than the indicated `funding_satoshis`, which ostensibly came from the user. Implementations would probably want to add a way for users to indicate where to take the fees out -- either the provided channel fund amount or in addition to that. (You could do this with the existing flow too, since it's just UX.)

My hesitation with making the `change_address` optional is that 1) it's an edge case (depends on available UTXO set) 2) it has a small chance of introducing bugs (i.e. broken implementation forgets to include a change address so you end up with a huge channel balance by accident). Leaving out the change address doesn't save you much in calculation overhead when picking out the input UTXO set either. The only thing it does save on is wire bytes, in the case where your funding amt works out exactly right.

02-peer-protocol.md
niftynei: inclusion of the `funding_tx_id` means we need a v2 for this

02-peer-protocol.md
```diff
@@ -406,6 +421,365 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+      as the change address.
+  - if is the `accepter`:
+    - consider the `*put_limit` the total of `num_inputs` plus
+      `num_outputs' from `funding_compose`, with minimum 2.
+    - MUST NOT send `input_info`s or `output_info` which
+      exceeds the `*put_limit`.
+    - MAY send zero inputs and/or outputs.
+
+The receiving node:
+  - if the total `input_info`.`satoshis` is less than the total `output_info`.`satoshis`
```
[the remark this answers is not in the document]
niftynei: Thinking about this more, I think we can simplify this requirement by exchanging the `scriptSig` data in addition to the txid/vout for inputs, and merely verifying that it is either `00` or one of the two acceptable v0 `redeemScript` formats for P2SH-P2W*. (i.e. `0014xxx` or `0020xxx`)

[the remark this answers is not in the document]
niftynei: Ok, updated to include the `scriptPubKey` of the desired output as well.

02-peer-protocol.md
niftynei: Updated to be 'optional' as well.

niftynei:
Working on the implementation of this for c-lightning, and I've hit a bit of a conundrum around what's the best way forward wrt the `channel_reserve_satoshi`.

In v1 of the channel establishment protocol, the entire funding balance is known as of the first `open_channel` message. In v2, the entire funding balance won't be known until after `accept_channel2` has been exchanged.

Here's a few options that I've come up with for how to deal with this. Leaning towards the 3rd option.

- leave it as is, and have both the opener and the accepter's values be based on the opener's funding_satoshis value (same as v1)
- leave the opener's reserve requirement to be the same, and have the accepter's value include their contribution as well. e.g. the opener sends `funding_satoshi: 10 000` and `channel_reserve_satoshi: 1 000`; accepter sends `funding_satoshis:10 000` and `channel_reserve_satoshi:2 000`.
- move the `channel_reserve_satoshi` field to the `funding_compose` message, so that the entire funding balance can be taken into account. e.g. both would send `channel_reserve_satoshi:2 000` (which is 1% of the entire funding balance).

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+The first commitment transaction has no HTLC's in it.
+
+Note that the `commitment_signed` message will include the `channel_id` derived from
+the `funding_txid`, instead of the `temporary_node_id`.
+
+
+### The `funding_signed2` Message
+
+This message contains witness data for for the inputs that were
```
darosior:
```suggestion
This message contains witness data for the inputs that were
```

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - MAY set the `num_inputs` to zero.
+  - MUST transmit all outputs, excluduing the channel funding output.
+  - MAY include at least one output with the `sats` set to zero,
+    to be used as the change output.
+
+The receiving node:
+  - MUST return an error if:
+    - the `feerate_per_kw_funding` is not greater than the previously
+      negotiated rate.
+  - MUST return an error, but not fail the channel if:
```
darosior: How does the remote node know that it should not fail the channel upon receiving the error ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    * [`channel_id`:`channel_id`]
+    * [`u64`:`funding_satoshis`]
+    * [`u32`:`feerate_per_kw`]
+    * [`u32`:`feerate_per_kw_funding`]
+    * [`u16`:`num_additional_inputs`]
+    * [`num_additional_inputs*input_info`:`input_info`]
+    * [`u16`:`num_outputs`]
+    * [`num_outputs*output_info`:`output_info`]
+
+#### Requirements
```
darosior: As it has been stated for other messages, maybe :
```
Either node:
    - MAY omit this message
```
?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+The sending node:
+  - MUST have sent `open_channel2`.
+  - MUST send a `feerate_per_kw_funding` greater than the most recently
+    negotiated rate.
+  - MAY update the `feerate_per_kw`, the commitment transaction feerate.
+  - MAY include additional inputs.
+  - MAY set the `num_inputs` to zero.
+  - MUST transmit all outputs, excluduing the channel funding output.
+  - MAY include at least one output with the `sats` set to zero,
```
darosior: Should we also specify which output to reduce ?

niftynei: Well, retransmitting all of the outputs would give you the opportunity to update the outputs as necessary.

The change output is calculated dynamically by each of the peers, and is identifiable in the output set as "the output of value zero", so you wouldn't need to update it. In fact, the change calculation should 1) drop the change output if it's below dust and 2) add or subtract any resultant amount to the opener's satoshis in the funding output. So, you wouldn't need to update the outputs on an rbf, unless you wanted to preserve  your funding output's value.

02-peer-protocol.md
niftynei: This is, at best, a case for the 'soft error' idea that @rustyrussell has been talking about. Clearly the node that sent you the `init_rbf` is confused, the polite thing to do in the case is to inform them that they've made a mistake. The other option, without a 'soft error' is probably to ignore the `init_rbf` request entirely.

You make a good point though, what we really want here is that a node MUST NOT send an `init_rbf` if it has already sent a `funding_locked` message.

02-peer-protocol.md
darosior: > So, you wouldn't need to update the outputs on an rbf, unless you wanted to preserve your funding output's value.

That's what I assumed: doesn't it make it more simple for both participants if the funding amount stayed the same but the fee bump was taken on the change output of the funder instead ?
Since both add inputs to the tx, they will most likely also include change outputs, at least that what I assumed when writing that.
My question was then whether to introduce an explicit MUST on the receiver side to check that the tx the funder sends does not greeds on the **fundee**'s change output (which implems would of course do, but ..)

But maybe I'm missing something here ?

niftynei: > That's what I assumed: doesn't it make it more simple for both participants if the funding amount stayed the same but the fee bump was taken on the change output of the funder instead ?

It's up to the funder to know what it's estimated change burden will be and add sufficient funds such that the total amount of funding is untouched by a fee-bump, if that's what they desire.

> does not greeds on the fundee's change output (which implems would of course do, but ..)

Fess fees never come out of the accepter's change; though it might (hypothetically) start eating out of their funding balance if the opener's balance is consumed by change. Reconsidering this and the point you made earlier, a `nack_rbf` message (or a 'soft-error') might be appropriate here. This would allow the second node to disallow a fee update that would eat into the entirety of the fundchannel amount.

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+`max_witness_len` is the total serialized length of the
+witness data that will be supplied (e.g. sizeof(varint) +
+sizeof(witness) for each) in `funding_signed2`.
+
+`input_info`.`script` is the scriptSig field for the input. Only applicable
+for P2SH-wrapped inputs.
+Native SegWit inputs (P2WPKH and P2WSH) inputs, will have an empty `script` field
+See [BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#examples).
+
+The `accepter` node may omit this message.
```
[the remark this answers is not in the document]
niftynei: ack, removed.

09-features.md
```diff
@@ -27,7 +27,8 @@ These flags may only be used in the `init` message:
 | 4/5   | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel                   | [BOLT #2][bolt02-open]       |
 | 6/7   | `gossip_queries`                 | More sophisticated gossip control                                         | [BOLT #7][bolt07-query]      |
 | 10/11 | `gossip_queries_ex`              | Gossip queries can include additional information                         | [BOLT #7][bolt07-query]      |
-| 12/13| `option_static_remotekey`     | Static key for remote output                                              | [BOLT #3](03-transactions.md)    |
+| 12/13 | `option_static_remotekey`        | Static key for remote output                                              | [BOLT #3](03-transactions.md)|
+| 28/29 | `option_dual_fund`               | Use v2 of channel open, enables dual funding                              | [BOLT #2](02-peer-protocol)  |
```
ariard: Should be rebased for flat features ofc but what context should we assign, IN- (you should let old nodes routes through you even if you enforce dual funding opening with your local peers) ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+#### Requirements:
+
+The accepting node:
+    - MAY respond with a `funding_satoshis` value of zero.
+
+#### Rationale
+
+Accepter sends their `funding_satoshi` value here instead of allowing the opener to derive
+it from their `funding_compose` response so that the opener can notionally decide whether
+to complete the opening without exposing their output set.
```
ariard, line 568: Should we require ownership proof of input to avoid opener freely probing utxo set of accepter ? Though it may be tricky if it's a non-templated P2WSH, accepter need to solve the script...

Also, maybe introduce concept of a quarantine utxo set ? If input as been announced as part of a cancelled dual-funding, it should be used in priority for next ones, and you should avoid to mix it with lambda onchain transactions.

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - MUST NOT send `funding_add_input` if it has already transmitted `funding_add_complete`
+  - MUST send this after `accept_channel2` has been exchanged, but before `funding_add_complete`
+  - MUST add all sent inputs to the funding transaction
+  - MUST NOT re-transmit inputs it has already received from the peer
+  - MUST ensure each `input_info` refers to a non-malleable (segwit) UTXO.
+  - if is the `opener`:
+    - MUST send at least one `funding_add_input` message
+    - MUST NOT send a total count of more than 64 inputs, across all `funding_add_input` messages.
+  - if is the `accepter`:
+    - MAY omit this message
```
ariard, line 617: Duplicated with https://github.com/lightningnetwork/lightning-rfc/pull/524/files#diff-3369c5aa1774fef2ff1e246979f223eaR620 ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    * [`u16`:`scriptlen`]
+    * [`scriptlen*byte`:`script`]
+
+##### Requirements
+
+The `accepter` node:
+  - MAY omit this message
+
+The sending node:
+  - MUST NOT send `funding_add_input` if it has already transmitted `funding_add_complete`
```
ariard, line 609: We don't write this common sense requirements in the current specs w.r.t funding_created which must come after open/accept sequence though obviously everyone encode this in their state machine ? IMO makes the spec unnecessarily verbose.

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - MUST send this after `accept_channel2` has been exchanged, but before `funding_add_complete`
+  - MUST add all sent inputs to the funding transaction
+  - MUST NOT re-transmit inputs it has already received from the peer
+  - MUST ensure each `input_info` refers to a non-malleable (segwit) UTXO.
```
ariard, line 612: You may check also that's a mature coinbase (but generally we may have issues with light-clients, unable to verify mempool rejection reasons and starting to RBF as crazy)

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+  - if is the `accepter`:
+    - MAY omit this message
+    - MUST NOT send a total count of more than 16 inputs, across all `funding_add_input` messages.
+
+The receiving node:
+  - MUST add all received inputs to the funding transaction
+  - MUST fail the channel if:
+    - it receives a duplicate input to one it sent previously
+    - if it receives this message after `funding_add_complete` is received.
+    - it receives an input that is malleable (P2SH/P2PKH)
```
ariard: Would be better to enforce than output is segwit, if implementor rejects input on a an enumeration of malleable inputs he may forget some (like P2PK, bare-multisig, ...).

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    * [`32*byte`:`temporary_channel_id`]
+    * [`u16`:`num_inputs`]
+    * [`num_inputs*input_info`:`input_info`]
+
+1. subtype: `input_info`
+2. data:
+    * [`u64`:`sats`]
+    * [`sha256`:`prevtx_txid`]
+    * [`u32`:`prevtx_vout`]
+    * [`u16`:`prevtx_scriptpubkey_len`]
```
ariard, line 600: I assume the reason for scriptpubkey inclusion is for bip 158 filters ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    - it receives a duplicate input to one it sent previously
+    - if it receives this message after `funding_add_complete` is received.
+    - it receives an input that is malleable (P2SH/P2PKH)
+  - if is the `opener`:
+    - MUST fail the channel if:
+      - the total count of `input_info`s is greater than 6
+
+
+##### Rationale
+Each node must have a complete set of the transaction inputs
```
ariard, line 633: Do we assume all nSequence to be set to SEQUENCE_FINAL ? We should at least have one of them signalling RBF.

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+      to create the funding transaction, using `max_witness_len`
+      for each `input_info` and `feerate_per_kw_funding` as specified in
+      `open_channel2`.
+    - MUST send `commitment_signed`.
+
+
+##### Rationale
+
+Each node must have a complete set of the transaction inputs and outputs,
+to derive the funding transaction and subsequent commitment signatures.
```
ariard, line 774: Do we assume nVersion=2 and nLocktime=0 ? If yes we may describe a template in BOLT3 (likely we need one to agree for input order)

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+Each node must have a complete set of the transaction outputs
+to derive the funding transaction and subsequent commitment signatures.
+
+`sats` is the satoshi value of the output.
+
+The opening node may include one output with `sats` value
+of zero. This will be used for change, or discarded if its value is
+be below `dust_limit_satoshis`.
+
+Change is calculated as the sum of the opener's `input_info`.`sats`
```
ariard, line 704: I'm not sure about the opener paying only the fees, it means if Alice funds with 1 input as opener but Bob puts max input 16, Bob utxo aggregation cost is externalized to Alice ? Sure you can hardcode or use a smart-fee-algo(tm) to reject but incentive-level it means accepter should seek this limit to free-ride ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+
+#### Rationale
+
+The first commitment transaction has no HTLC's in it.
+
+Note that the `commitment_signed` message will include the `channel_id` derived from
+the `funding_txid`, instead of the `temporary_node_id`.
+
+
+### The `funding_signed2` Message
```
ariard, line 814: Message name is a bit confusing, funding_signedv1 refers to accepter giving commitment tx signatures to fundee, here it refers to signatures exchange for the funding tx (even now wondering if a funding_signed distinct from commitment_signed wasn't a spec oversight..)

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    * [`num_outputs*output_info`:`output_info`]
+
+#### Requirements
+
+The sending node:
+  - MUST have sent `open_channel2`.
+  - MUST send a `feerate_per_kw_funding` greater than the most recently
+    negotiated rate.
+  - MAY update the `feerate_per_kw`, the commitment transaction feerate.
+  - MAY include additional inputs.
```
ariard, line 901: IIRC I think that hurts BIP 125 rule 2, you can't add new inputs to rbf'ed transactions

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+all of the previously relayed inputs plus the ones included here.
+
+All outputs must be transmitted in the `init_rbf` message, including
+ones communicated previously, except for the channel funding output.
+
+If a valid `funding_locked` message is received in the middle of an
+RBF workflow, the RBF attempt MUST be abandoned.
+
+
+### Acknowledging Replace-By-Fee `ack_rbf`
```
ariard, line 924: What problem the ack_rbf is handling ? Can't receiver of init_rbf directly replies with a commitment_signed ?

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+`witness` is the data for a witness element in a witness stack
+
+### Kicking Off Replace-By-Fee: `init_rbf`
+
+This message initiates the flow to create a replacement funding
+transaction.  It is sent by the channel opener, after the funding
+transaction has been broadcast but before a `funding_locked` message
+is exchanged.
+
+Once an `init_rbf` message has been successfully ack'd by the accepter
```
ariard, line 878: What if rbf is refused by party ? Maybe we should recommend to always have an output available to rbf and avoid input being stuck in the mempool for 2 weeks if funding tx feerate isn't enough to confirm

02-peer-protocol.md
```diff
@@ -430,6 +446,515 @@ would create a Denial of Service risk; therefore, forgetting it is recommended
...
+    corresponding to `prevtx_txid`:`prevtx_vout`
+  - MUST remember the details of this funding transaction.
+  - MUST NOT send a `witness_stack` whose length exceeds its
+    previously indicated `max_witness_len`.
+  - if it has NOT received a valid `commitment_signed` message:
+    - MUST NOT send a `funding_signed2` message.
+
+The receiving node:
+  - if the `witness_stack` length exceeds `max_witness_len`:
+    - MUST error.
```
t-bast: What does that mean exactly? Sending an error message to the other party and forgetting everything about this channel creation attempt?

I haven't thought about it deeply enough yet, but I feel that aborting during the funding flow may have some subtle edge cases. Are there some gotchas you ran into while implementing?

02-peer-protocol.md
niftynei: Vectors for revelation of your utxo set heavily depends on how much/how many inputs an accepter chooses to contribute to a channel, and will vary depending on the policy individual nodes set for their threshold and amount they've decided to contribute to a channel.

Currently, the c-lightning implementation mitigates this (slightly) by promoting any exposed utxos to be first eligible for spending. A repeated inquiry should expose the exact same utxos, until they're spent.

02-peer-protocol.md
niftynei: Oh good point. This isn't currently checked on funding transactions.

02-peer-protocol.md
niftynei: ACK

02-peer-protocol.md
niftynei: I believe you mean 'output available to cpfp' :)

02-peer-protocol.md
niftynei: Hmm. You've highlighted what I believe is the most subjective error case. 

How the opener chooses to handle this is hard to enforce. At this point, they have all the information necessary to successfully broadcast the transaction (the accepter is the only peer that transmits their signatures). If they choose to accept the lower effective feerate anyway and broadcast the tx, there's nothing stopping them. What we're looking to accomplish is incentivize the peer (accepter) to provide an accurate witness length, so that the agreed upon feerate is accurate. The way I've chosen to incentivize this is by threatening that the opener will cancel the open on them.

So yes, the opener would send the peer and error message and, ideally, spend an input to the tx such that the opening transaction was definitively invalidated.

02-peer-protocol.md
niftynei: good point, i think it is infact superfluous.

02-peer-protocol.md
niftynei: This is a misinterpretation of rule 2; an "unconfirmed input" is a utxo that has not yet been confirmed in a block (mined). 

>2. The replacement transaction may only include an unconfirmed input if that input was included in one of the original transactions. (An unconfirmed input spends an output from a currently-unconfirmed transaction.)

I think this is an oversight of the spec though; it should strictly exclude unconfirmed inputs (and thus disallow chaining)

02-peer-protocol.md
niftynei: Good point, this is underspecified.

02-peer-protocol.md
niftynei: This hadn't occurred to me, but yes. I believe the reason for their inclusion was to allow for a quick rejection if the script is not obviously a segwit script.

02-peer-protocol.md
niftynei: Yess, I've got a draft for how to re-distribute fees (and a tweak to this protocol) I'm about to send to the mailing list. :)

02-peer-protocol.md
niftynei: update: in order to avoid reversing the order of `commitment_signed` exchange, you need to get at least one message back from the peer before you send them the `commitment_signed` message.

09-features.md
niftynei: are we waiting on a spec change for 'N+/-', it's not currently listed in the features doc (but C+/- is)

02-peer-protocol.md
niftynei: i'm not sure this is common sense -- it needs to be specified somewhere

09-features.md
t-bast: Yes, we need to discuss this during the next meeting, but we still need some clarifying on feature bits (potentially the N+/N- modifiers).

02-peer-protocol.md
ariard: I'm fine with this for a first version. We may TLV ownership proof in a later one.

Maybe we should add privacy note somewhere on the risks, including some implementation advice like the one you're describing without this being mandatory.

02-peer-protocol.md
ariard: Okay IMO I find better to check in one place to avoid implementor assuming at the final check than scriptpubkey have been already validated (but not in a trusted way against blockchain view). Quick rejection will catch programming errors but not a malicious party.

02-peer-protocol.md
ariard: You're right, thanks the rule is more liberal than I believed.

Yes sure the LN spec should disallow unconfirmed input to avoid chaining. That would be a real attack vector for zero-conf channels if people keep doing them. Maybe even ask for few confirmations to avoid messy 1-depth reorg edges cases.

02-peer-protocol.md
ariard: Can't the `init_rbf` sender joins a `commitment_signed` and interpret the lack of getting back a `commitment_signed` as a refusal (though harder for the state machine I guess) ?

02-peer-protocol.md
ariard: Concerning the UTXO bloat it's likely they add a change output so they may use it for CPFP. 

I was thinking about the use-case of a zero-conf chans, for the party bearing the risk, adding a CPFP output and disabling RBF would be a lightweight mempool-enforced protection (though don't prevent malicious peer to directly send double-spend to miner..). 

Sure no need to make it mandatory, it's up to the party if they want to be safe against stuck-in-the-mempool cases.

ecdsa:
@rustyrussell suggested to add Lloyd Fournier's deterministic generation of funding output to this PR, see https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-December/002910.html

niftynei:
Superceded by #851

PR #539, opened 2019-01-07
BOLT 3: add test vectors for htlc-transactions in case where CLTV is used as tie-breaker for sorting

araspitzu:
Add a serialized transactions test vector for the edge case of sorting htlc-timeout-tx when there are multiple offered htlc with the same amount and preimage. The test vector reuses previous preimages and creates a case scenario with 1 received htlc and 2 offered, the two offered will have same scriptPubKey and redeemScript, but different CLTV value. It is asserted the order in which the htlc transactions should be kept internally and we assume the same order is used to construct the commitment_signed message. This complements #491 .


03-transactions.md
```diff
@@ -1263,6 +1263,26 @@ And, here are the test vectors themselves:
...
+    # HTLC 1 and 2 have CLTV 502 and 503, respectively, all htlc have preimage 0000000000000000000000000000000000000000000000000000000000000000
+    output commit_tx: 02000000000101bef67e4e2fb9ddeeb3461973cd4c62abb35050b1add772995b820b584a488489000000000038b02b8005e80300000000000022002052bfef0479d7b293c27e0f1eb294bea154c63a3294ef092c19af51409bce0e2ad007000000000000220020b9826699af5c86f289a6fe059b0d541048c6c9299f52ba223305f0920d953759d007000000000000220020b9826699af5c86f289a6fe059b0d541048c6c9299f52ba223305f0920d953759c0c62d0000000000160014ccf1af2f2aabee14bb40fa3851ab2301de843110a79f6a00000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e040047304402202254e1ab2b07e8c2c9a0108d7b547b6669ec4eccb2a19ebf68078eb54936ac4902203360cf0fa62315a1154428d34e8fe85133ea4d104ea9fccee57ffabe3aa337720147304402207fb2b0030a35d5ea93297fb87804b55fb43326c8e9f9f33cdd8417cb26a1d3ef02202de9f461f715ede216c03fdb8458dd4156e1d2cbf09e235055b9f72473abafcd01475221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae3e195220
```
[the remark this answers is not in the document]
araspitzu: Fixed the sat/msat, but i''m not sure I understand the HTLC1 and HTLC2 substituted? I just decided to stick with the sequential naming (HTLC1, HTLC2 as opposed to HLTC1a, HTLC1b) but I can change that if it's a problem.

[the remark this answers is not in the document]
araspitzu: Sorry for the delay, i misunderstood your comments at first. Now 8e8b4da defines at the top 2 new htlcs (#5 and #6) with a new preimage that is shared across them, the rest of the test remains unchanged.

ysangkok:
This seems like it would still be useful. @rustyrussell , the meeting logs say you [validated these vectors](http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-02-18-19.00.html). What was the result?

t-bast:
As discussed during the spec meeting, I added the corresponding anchor output test vector.
This is waiting for a second implementation to validate the test vectors.

03-transactions.md
```diff
@@ -1423,6 +1431,26 @@ And, here are the test vectors themselves:
...
+    # HTLC 5 and 6 have CLTV 505 and 506, respectively, and preimage 0505050505050505050505050505050505050505050505050505050505050505
+    output commit_tx: 02000000000101bef67e4e2fb9ddeeb3461973cd4c62abb35050b1add772995b820b584a488489000000000038b02b8005e80300000000000022002052bfef0479d7b293c27e0f1eb294bea154c63a3294ef092c19af51409bce0e2a8813000000000000220020305c12e1a0bc21e283c131cea1c66d68857d28b7b2fce0a6fbc40c164852121b8813000000000000220020305c12e1a0bc21e283c131cea1c66d68857d28b7b2fce0a6fbc40c164852121bc0c62d0000000000160014ccf1af2f2aabee14bb40fa3851ab2301de843110a79f6a00000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e04004730440220586f17b3f3f6eec96a0dc7040f1a33174c707e032010421fa1965497cd1c81e602204476e6a4cc17d433e8ad82edff2ffa47c954ecb52aad9cc34382befb66f6e00001473044022069797b84fad1e0e7b02e1edeec80cb6e0b6f01170700bca273788715f55e0c560220260833550768b1db3ed88f8c005cf2560821b1879ec69834cd9765de25ec350601475221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae3e195220
```
jkczyz, line 1442: Should the local and remote signatures be given for the `commit_tx` output like they are in the other test vectors?

t-bast: Sure, done in https://github.com/lightningnetwork/lightning-rfc/pull/539/commits/f8770b7dac0db2b7da6e75e9cfe870a82708f992

jkczyz, line 1444: `a79f6a0000000000` is the `to_local` output value, I believe, which is little-endian for 6_987_687. However, in RL I'm getting a value of 6_988_687. I'm seeing `c0c62d0000000000` (3_000_000) for the `to_remote` output value, as expected.

I think this means either our fee calculation is wrong or the test vector is wrong. Could someone verify this or correct my understanding?

t-bast: Interesting, and you diverge only on this test vector?
Doing the calculation by-hand, I find that the commit tx fee should be:

```
commit tx weight = 724
htlc output weight = 172
feerate = 253
fee = (724 + 3 * 172) * 253 / 1000 = 313 sat
```

So `to_local = 6_988_000 - 313 = 6_987_687` which seems to confirm the test vector.
You have exactly 1_000 sat more than that, which is the amount of the incoming HTLC, could you be somehow counting that in your `to_local`?

jkczyz: > Interesting, and you diverge only on this test vector?

Yeah, just this one.

> Doing the calculation by-hand, I find that the commit tx fee should be:
> 
> ```
> commit tx weight = 724
> htlc output weight = 172
> feerate = 253
> fee = (724 + 3 * 172) * 253 / 1000 = 313 sat
> ```
> 
> So `to_local = 6_988_000 - 313 = 6_987_687` which seems to confirm the test vector.
> You have exactly 1_000 sat more than that, which is the amount of the incoming HTLC, could you be somehow counting that in your `to_local`?

Ah, I think therein lies the problem! My subconscious churned away at this overnight:

Other than the first test which uses no HTLCs, the other tests use HTLCs 0-4 (according to lines 1013-1014 prior to this PR) which total 12_000_000 msats. I think that is where `to_local_msat: 6988000000` comes from. However, the new test uses HTLCs 0, 5, and 6. These total 11_000_000 msats, which should instead give `to_local_msat: 6989000000`.

Given the PR adds new HTLCs, I think the description in lines 1013-1014 needs to be updated. The language isn't very clear and is no longer accurate with the additional HTLCs.

t-bast: I see what you mean, but I'm not sure it makes sense to interpret the test case this way.

If we consider that we start with `to_local_msat = 7000000000` and `to_remote_msat = 3000000000` and then apply HTLCs 0-4, we should subtract `5000000 msat` from `to_local` and subtract `7000000 msat` from `to_remote`, resulting in `to_local_msat = 6995000000` and `to_remote_msat = 2993000000`.

Another way to read this would be to use the amounts if all HTLCs are fulfilled, but even in that case it doesn't work: we have `7000000 msat` incoming and `5000000 msat` outgoing, so the result would be `to_local_msat = 7002000000` (and `to_remote_msat = 2998000000`).

I can't speak for what the reason was when these test vectors were initially created, but I interpret the `to_local_msat` as simply saying "your main output has this value". I don't think the values of pending HTLCs should play any role here.

> Given the PR adds new HTLCs, I think the description in lines 1013-1014 needs to be updated.

That's right, the two new HTLCs should be described here, it would slightly help read these already hard-to-figure-out tests!

jkczyz: > I see what you mean, but I'm not sure it makes sense to interpret the test case this way.
> 
> If we consider that we start with `to_local_msat = 7000000000` and `to_remote_msat = 3000000000` and then apply HTLCs 0-4, we should subtract `5000000 msat` from `to_local` and subtract `7000000 msat` from `to_remote`, resulting in `to_local_msat = 6995000000` and `to_remote_msat = 2993000000`.
> 
> Another way to read this would be to use the amounts if all HTLCs are fulfilled, but even in that case it doesn't work: we have `7000000 msat` incoming and `5000000 msat` outgoing, so the result would be `to_local_msat = 7002000000` (and `to_remote_msat = 2998000000`).
> 
> I can't speak for what the reason was when these test vectors were initially created, but I interpret the `to_local_msat` as simply saying "your main output has this value". I don't think the values of pending HTLCs should play any role here.

While I don't profess to fully understand this myself, the other tests behave in this manner. Namely, for all test vectors, the "output commitment tx" has a `to_remote` output of 3_000_000 (`c0c62d0000000000` little-endian) regardless of which HTLCs are included.

t-bast: Indeed, I view this as an additional reason to decouple the `to_local_msat` value from the HTLCs used (which means the current test vector would be ok).

Can you explain why it's hard to replicate in rust-lightning? Aren't you simply initializing the commitment tx main outputs with the `to_local_msat` and `to_remote_msat` indicated in the test vector?

TheBlueMatt: The way the tests (currently) work is we build the full channel state machine and then get a commitment tx from it, not building a commitment tx from raw values. Thus we have to convert the numbers into state machine values (eg there is no amount-to-remote field, to-local tracking doesn't include HTLC values, etc).

t-bast: Got it, but then the HTLC amounts should not impact you, right? If you open a new channel, with the right initial amount (using `push_msat` I guess to have funds on the remote side) and set the `channel_reserve` accordingly, you should be able to have `to_local` and `to_remote` that match this test vector, regardless of the fact that it uses different HTLCs from the other vectors? So the issue probably lies somewhere else, unless I'm missing something...

jkczyz: Unfortunately, not. The HTLCs are used to calculate `to_local` and `to_remote` for the commitment transaction.

https://github.com/rust-bitcoin/rust-lightning/blob/9c9c88147e6a8fa637dc048f764ab8ecfea200d4/lightning/src/ln/channel.rs#L919-L925

So as the test vectors are written, `to_local`, `to_remote`, HTLCs, and fees must total the channel value (i.e., 10_000_000 sats). But for the new test vector they are 1_000 sats short.

t-bast: Got it, I'll change the HTLC amounts then to make it easier.

t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/539/commits/d178fb95907efabcec4e9c7a520e269a80eb15f8, let me know if that's easier for you.

jkczyz (commented):
@t-bast It occurred to me while testing this that adding the HTLCs in the order stated (i.e., the expected ordering) won't test the intended behavior since the HTLCs would be presorted. Case in point, shouldn't the outgoing HTLC ordering be switched in [Eclair's test](https://github.com/ACINQ/eclair/blob/dc364a1996548e405199fe38ad861a693baf4787/eclair-core/src/test/scala/fr/acinq/eclair/transactions/TestVectorsSpec.scala#L489-L490)?

Perhaps a comment in the BOLT on the new test vector would help convey this.

03-transactions.md
```diff
@@ -1436,22 +1436,22 @@ And, here are the test vectors themselves:
...
+    # local_htlc_signature = 30440220207ee7d74a0f223b5d05db49653c541d4a043f2ab3e88abbeeb5f1d1d761535902207de25db7f70a939eae6cec89d8487ad1e9c953cb0f30a5f84e5ade5736d6f9ce
+    # remote_htlc_signature = 3045022100de2757977d17699d092c1b05cebb52fa498de5d5fba6a60762f73aa7354ad8a50220199295de83336966e7387e8ebb2f105adc0db516260d6cd1e3c166f8b32d853e
+    output htlc_timeout_tx 2: 0200000000010175a53c9a465f56140f1c7b2c42b27f0e56b508e39cb6971f3c5b551cfda8e5e402000000000000000001e1120000000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e0500483045022100de2757977d17699d092c1b05cebb52fa498de5d5fba6a60762f73aa7354ad8a50220199295de83336966e7387e8ebb2f105adc0db516260d6cd1e3c166f8b32d853e014730440220207ee7d74a0f223b5d05db49653c541d4a043f2ab3e88abbeeb5f1d1d761535902207de25db7f70a939eae6cec89d8487ad1e9c953cb0f30a5f84e5ade5736d6f9ce01008576a91414011f7254d96b819c76986c277d115efce6f7b58763ac67210394854aa6eab5b2a8122cc726e9dded053a2184d88256816826d6231c068d4a5b7c820120876475527c21030d417a46946384f88d5f3337267c5e579765875dc4daca813e21734b140639e752ae67a9142002cc93ebefbb1b73f0af055dcc27a0b504ad7688ac6868fa010000
```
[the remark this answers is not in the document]
t-bast: Right, these test vectors weren't completely aligned with the format of other test vectors, I think they were partially hand-generated instead of relying on our tests, I fixed that in https://github.com/lightningnetwork/lightning-rfc/pull/539/commits/280e9603e9e4fba820aba2c3d27630c1477a7638

> Case in point, shouldn't the outgoing HTLC ordering be switched in Eclair's test?

I see what you mean, done in https://github.com/ACINQ/eclair/pull/1669
I think it's up to implementations to ensure this is properly tested.

03-transactions.md
```diff
@@ -1423,6 +1432,401 @@ And, here are the test vectors themselves:
...
+    output commit_tx: 02000000000101bef67e4e2fb9ddeeb3461973cd4c62abb35050b1add772995b820b584a488489000000000038b02b8007e80300000000000022002052bfef0479d7b293c27e0f1eb294bea154c63a3294ef092c19af51409bce0e2ad007000000000000220020403d394747cae42e98ff01734ad5c08f82ba123d3d9a620abda88989651e2ab5d007000000000000220020748eba944fedc8827f6b06bc44678f93c0f9e6078b35c6331ed31e75f8ce0c2db80b000000000000220020c20b5d1f8584fd90443e7b7b720136174fa4b9333c261d04dbbd012635c0f419a00f0000000000002200208c48d15160397c9731df9bc3b236656efb6665fbfe92b4a6878e88a499f741c4c0c62d0000000000160014cc1b07838e387deacd0e5232e1e8b49f4c29e484e0a06a00000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e040047304402206fc2d1f10ea59951eefac0b4b7c396a3c3d87b71ff0b019796ef4535beaf36f902201765b0181e514d04f4c8ad75659d7037be26cdb3f8bb6f78fe61decef484c3ea01473044022009b048187705a8cbc9ad73adbe5af148c3d012e1f067961486c822c7af08158c022006d66f3704cfab3eb2dc49dae24e4aa22a6910fc9b424007583204e3621af2e501475221023da092f6980e58d2c037173180e9a465476026ee50f96695963e8efe436f54eb21030e9f7b623d2ccc7c9bd44d66d5ce21ce504c0acf6385a132cec6d3c39fa711c152ae3e195220
```
rustyrussell, line 1490: This is correct; the txid is 2b887d4c1c59cd605144a1e2f971d168437db453f841f2fefb2c164f28ff84ab.

03-transactions.md
```diff
@@ -1423,6 +1432,401 @@ And, here are the test vectors themselves:
...
+    remote_htlc_signature = 304402207e0410e45454b0978a623f36a10626ef17b27d9ad44e2760f98cfa3efb37924f0220220bd8acd43ecaa916a80bd4f919c495a2c58982ce7c8625153f8596692a801d
+    # local_signature = 304402207cb324fa0de88f452ffa9389678127ebcf4cabe1dd848b8e076c1a1962bf34720220116ed922b12311bd602d67e60d2529917f21c5b82f25ff6506c0f87886b4dfd5
+    output htlc_success_tx 0: 020000000001018154ecccf11a5fb56c39654c4deb4d2296f83c69268280b94d021370c94e219700000000000000000001e8030000000000002200204adb4e2f00643db396dd120d4e7dc17625f5f2c11a40d857accc862d6b7dd80e050047304402206a6e59f18764a5bf8d4fa45eebc591566689441229c918b480fb2af8cc6a4aeb02205248f273be447684b33e3c8d1d85a8e0ca9fa0bae9ae33f0527ada9c162919a60147304402207cb324fa0de88f452ffa9389678127ebcf4cabe1dd848b8e076c1a1962bf34720220116ed922b12311bd602d67e60d2529917f21c5b82f25ff6506c0f87886b4dfd5012000000000000000000000000000000000000000000000000000000000000000008a76a91414011f7254d96b819c76986c277d115efce6f7b58763ac67210394854aa6eab5b2a8122cc726e9dded053a2184d88256816826d6231c068d4a5b7c8201208763a914b8bcb07f6344b42ab04250c86a6e8b75d3fdbbc688527c21030d417a46946384f88d5f3337267c5e579765875dc4daca813e21734b140639e752ae677502f401b175ac686800000000
```
rustyrussell, line 1503: This is wrong?  It should be spending the commit_tx above, but it isn't.  bitcoin-cli decodes this:

```
{
  "txid": "b8afcdfb3e0b49be541a0c24569f49d9ebd03d275f809780d0afde66bec01da2",
  "hash": "402ac1f6656f1f9adbff78d7191a4f30192390bdc2db187e3de1ca9de1c79c68",
  "version": 2,
  "size": 414,
  "vsize": 174,
  "weight": 696,
  "locktime": 0,
  "vin": [
    {
      "txid": "97214ec97013024db9808226693cf896224deb4d4c65396cb55f1af1ccec5481",
      "vout": 0,
```

rustyrussell (changes requested):
The option_static_remotekey HTLCs seem all wrong?

t-bast:
> The option_static_remotekey HTLCs seem all wrong?

That's worrying :sweat_smile: 
I'll investigate based on your comments.

03-transactions.md
t-bast: This is embarrassing, I have no idea how I managed to get that wrong...
These test vectors didn't even match what's in eclair, so I messed up either a copy-paste or a git rebase :thinking: 
That should be fixed in https://github.com/lightningnetwork/lightning-rfc/pull/539/commits/c6d13fb3f9196464d842dc579723cf6de35ec8f8

t-bast: @rustyrussell would you have some time to validate these? I think they should work this time :sweat_smile:

jkczyz:
> > The option_static_remotekey HTLCs seem all wrong?
> 
> That's worrying 😅
> I'll investigate based on your comments.

FYI, I confirmed that RL's test vectors do indeed match the corrected ones in c6d13fb3f9196464d842dc579723cf6de35ec8f8.

t-bast:
I have no idea how I messed that up, this one's completely on me!

shame

rustyrussell:
Vectors now check out!

Ack https://github.com/lightningnetwork/lightning-rfc/commit/c6d13fb3f9196464d842dc579723cf6de35ec8f8

There is some weirdness in the formatting, which should be cleaned up in a separate PR.
1. Some places use local_signature where it means local_htlc_signature.
2. `output htlc_success_tx 1` isn't clear whether this is output 1, or htlc id 1.  Both are used in different places, and we should probably say `output #1, htlc_success_tx id 1` etc.

t-bast:
Great, thanks for checking!

As discussed during a previous spec meeting (https://github.com/lightningnetwork/lightning-rfc/issues/838), merging this PR now that it has been verified by both rust-lightning and c-lightning.

> There is some weirdness in the formatting, which should be cleaned up in a separate PR.

I'll create a follow-up PR to apply these clarifications, that will indeed be useful for newcomers.

PR #544, opened 2019-01-14
BOLT 4: remove incorrect_payment_amount altogether.

rustyrussell:
914ebab9080ccccb0ff176cb16b7a6ba21e23bbe effectively deprecated this, but
left it for "reject if more than 2x expected amount" case.

Leaving it for gross overpayment still leaves an attack on the current
network in practice (all implementations I know of reject grossly excessive
payments), and removing it causes our code to nicely break when regenerating,
since that error code is now not defined anywhere.

[ @TheBlueMatt I invited you to the repo, so I can request reviews from you in future ]


cdecker:

PR #546, opened 2019-01-15
Clarify behavior of id field in update_add_htlc

OrfeasLitos:
Clarify that the id value is not reset after an update is complete


cdecker:

PR #548, opened 2019-01-16
Align offered and received HTLC scripts similarly

OrfeasLitos:


cdecker:

PR #549, opened 2019-01-17
BOLT3: RIPEMD160 -> HASH160 where appropriate

Christewart:
I believe we were misusing `RIPEMD160` where `HASH160` should have been used in BOLT3.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -680,8 +680,8 @@ The *expected weight* of an HTLC transaction is calculated as follows:
     accepted_htlc_script: 139 bytes
         - OP_DUP: 1 byte
         - OP_HASH160: 1 byte
-        - OP_DATA: 1 byte (RIPEMD160(SHA256(revocationpubkey)) length)
-        - RIPEMD160(SHA256(revocationpubkey)): 20 bytes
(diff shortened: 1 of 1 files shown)


03-transactions.md
```diff
@@ -126,7 +126,7 @@ This output sends funds to either an HTLC-timeout transaction after the HTLC-tim
             OP_DROP 2 OP_SWAP <local_htlcpubkey> 2 OP_CHECKMULTISIG
         OP_ELSE
             # To remote node with preimage.
-            OP_HASH160 <RIPEMD160(payment_hash)> OP_EQUALVERIFY
+            OP_HASH160 <HASH160(payment_hash)> OP_EQUALVERIFY
```
sstone: It's actually RIPEMD160. The trick is:
- R is a random 32 byte value chosen by the recipient
- H = sha256(R) is the payment hash used is the payment request and HTLC
 - so ripemd160(H) = ripemd160(sha256(R)) = hash160(R).

This script does not check that sha256(R) == H, instead it checks that hash160(R) == ripemd160(H), which saves 12 bytes.

PR #555, opened 2019-01-22
BOLT 4: Clarify final_incorrect_cltv_expiry data

nueno4:
use `update_add_htlc.cltv_expiry` like `final_incorrect_htlc_amount`.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -824,7 +824,7 @@ handling by the final node.
 
 1. type: 18 (`final_incorrect_cltv_expiry`)
 2. data:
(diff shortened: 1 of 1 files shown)


rustyrussell:
This is the final node, so it has to be incoming?  More precisely, it is the value in the update_add_htlc, not the value in the onion (the onion is presumably correct!).

nayuta-ueno:
`final_incorrect_htlc_amount` has `incoming_htlc_amt`.
Do you think `final_incorrect_htlc_amount`'s `incoming_htlc_amt` --> `amount_msat` ?

PR #519, opened 2018-12-02
BOLT 7: add extended channel queries

sstone:
Extended channel queries include extra timestamps (one for each channel_update), which
allows nodes that are often online to easily sync their routing tables.

PR #557, opened 2019-01-23
BOLT7: extend channel range queries with optional fields

sstone:
This is a new pull requests that supersedes #519 .

It addresses issues with the original proposal, mainly that it defined a new set of messages, adding complexity to a simple gossip protocol that we knew was limited in the first place.

This proposal does not add new messages, or feature bits, and is fully compatible with existing implementations. Instead of defining new messages it extends existing ones with additional data, that will be ignored by nodes which do not implement extended queries (see [BOLT #1](01-messaging.md)).

Nodes that support extended queries will append an additional extended query flag to their `query_channel_range` queries. If the receiver supports extended queries and understands this flag, it will append the requested additional data to its `reply_channel_range` message.

There is currently only one type of additional data: one timestamp and one checksum per `channel_update`.
The checksum is a simple Adler32 checksum computed over the `channel_update` with `timestamp` and `signature` omitted.
Together they can be used to avoid querying `channel_updates` that are older than the ones you already have, or that are newer but don't include new information.

Nodes can then append additional data to their `query_short_channel_ids` messages, which consists in one flag per short channel id and specifies what they would like to receive (`channel_announcement`, or/and one `channel_update` or both`).

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -587,6 +587,25 @@ contents could decompress to more then 3669960 bytes.
     * [`chain_hash`:`chain_hash`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
+    * [`query_short_channel_ids_tlv`]
+      * type: 1 (`query_flags`)
+      * data:
+        * [`1`:`encoding_type`]
+        * [`query_short_channel_ids_tlv_len-1`:`encoded_query_flags`]
+
+</a>
+
+`encoded_query_flags` is an array of bitfields, one varint per bitfield, one bitfield for each `short_channel_id`. Bits have the following meaning:
+
+| Bit Position  | Meaning                                  |
+| ------------- | ---------------------------------------- |
+| 0             | Sender wants `channel_announcement`      |
+| 1             | Sender wants `channel_update` for node 1 |
+| 2             | Sender wants `channel_update` for node 2 |
+| 3             | Sender wants `node_announcement` for node 1 |
+| 4             | Sender wants `node_announcement` for node 2 |
+
+Query flags must be minimally encoded, which means that one flag will be encoded with a single byte.
 
 1. type: 262 (`reply_short_channel_ids_end`) (`gossip_queries`)
 2. data:
@@ -611,6 +630,10 @@ The sender:
   - MAY send this if it receives a `channel_update` for a
    `short_channel_id` for which it has no `channel_announcement`.
   - SHOULD NOT send this if the channel referred to is not an unspent output.
+  - MAY include an optional `query_short_channel_ids_tlv`. If so:
+    - MUST set `encoding_type`, as for `encoded_short_ids`.
+    - Each query flag is a minimally-encoded varint.
+    - MUST encode one query flag per `short_channel_id`.
 
 The receiver:
   - if the first byte of `encoded_short_ids` is not a known encoding type:
@@ -619,8 +642,24 @@ The receiver:
     - MAY fail the connection.
   - if it has not sent `reply_short_channel_ids_end` to a previously received `query_short_channel_ids` from this sender:
     - MAY fail the connection.
-  - MUST respond to each known `short_channel_id` with a `channel_announcement`
-    and the latest `channel_update` for each end
+  - if the incoming message includes `query_short_channel_ids_tlv`:
+    - if `encoding_type` is not a known encoding type:
+      - MAY fail the connection
+    - if `encoded_query_flags` does not decode to exactly one flag per `short_channel_id`:
+      - MAY fail the connection.
+  - MUST respond to each known `short_channel_id`:
+    - if the incoming message does not include `encoded_query_flags`:
+      - with a `channel_announcement` and the latest `channel_update` for each end
+    - otherwise:
+      - We define `query_flag` for the Nth `short_channel_id` in
+        `encoded_short_ids` to be the Nth varint of the decoded
+        `encoded_query_flags`.
+      - if bit 0 of `query_flag` is set:
+        - MUST reply with a `channel_announcement`
+      - if bit 1 of `query_flag` is set and it has received a `channel_update` from `node_id_1`:
+        - MUST reply with the latest `channel_update` for `node_id_1`
+      - if bit 2 of `query_flag` is set and it has received a `channel_update` from `node_id_2`:
+        - MUST reply with the latest `channel_update` for `node_id_2`
 	- SHOULD NOT wait for the next outgoing gossip flush to send these.
   - MUST follow with any `node_announcement`s for each `channel_announcement`
 	- SHOULD avoid sending duplicate `node_announcements` in response to a single `query_short_channel_ids`.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -666,6 +666,10 @@ The receiver:
         - MUST reply with the latest `channel_update` for `node_id_1`
       - if bit 2 of `query_flag` is set and it has received a `channel_update` from `node_id_2`:
         - MUST reply with the latest `channel_update` for `node_id_2`
+      - if bit 3 of `query_flag` is set and it has received a `node_announcement` from `node_id_1`:
+        - MUST reply with the latest `node_announcement` for `node_id_1`
+      - if bit 4 of `query_flag` is set and it has received a `node_announcement` from `node_id_2`:
+        - MUST reply with the latest `node_announcement` for `node_id_2`
 	- SHOULD NOT wait for the next outgoing gossip flush to send these.
   - MUST follow with any `node_announcement`s for each `channel_announcement`
 	- SHOULD avoid sending duplicate `node_announcements` in response to a single `query_short_channel_ids`.
@@ -699,7 +703,7 @@ timeouts.  It also causes a natural ratelimiting of queries.
 2. types:
     1. type: 1 (`query_option`)
     2. data:
-        * [`1`:`query_option_flags`]
+        * [`varint`:`query_option_flags`]
 
 `query_option_flags` is a bitfield represented as a minimally-encoded varint. Bits have the following meaning:
 
@@ -784,8 +788,8 @@ The receiver of `query_channel_range`:
       - SHOULD set `complete` to 1.
 
 If the incoming message includes `query_option`, the receiver MAY append additional information to its reply:
-- if bit 0 in `query_option_flags` is set, the receive MAY append a `timestamps_tlv` that contains `channel_update` timestamps for all `short_chanel_id`s in `encoded_short_ids`
-- if bit 1 in `query_option_flags` is set, the receive MAY append a `checksums_tlv` that contains `channel_update` checksums for all `short_chanel_id`s in `encoded_short_ids`
+- if bit 0 in `query_option_flags` is set, the receiver MAY append a `timestamps_tlv` that contains `channel_update` timestamps for all `short_chanel_id`s in `encoded_short_ids`
+- if bit 1 in `query_option_flags` is set, the receiver MAY append a `checksums_tlv` that contains `channel_update` checksums for all `short_chanel_id`s in `encoded_short_ids`
 
 
 #### Rationale
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -587,6 +587,8 @@ Query messages can be extended with optional fields that can help reduce the num
 - timestamp-based filtering of `channel_update` messages: only ask for `channel_update` messages that are newer than the ones you already have.
 - checksum-based filtering of `channel_update` messages: only ask for `channel_update` messages that carry different information from the ones you already have.
 
+Nodes can signal that they support extended gossip queries with the `gossip_queries_ex` feature bit.
+
 ### The `query_short_channel_ids`/`reply_short_channel_ids_end` Messages
 
 1. type: 261 (`query_short_channel_ids`) (`gossip_queries`)
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -580,6 +580,11 @@ are unique 8-byte values, no more than 14 bytes can be duplicated
 across the stream: as each duplicate takes at least 2 bits, no valid
 contents could decompress to more then 3669960 bytes.
 
+Query messages can be extended with optional fields that can help reduce the number of messages needed to synchronize routing tables by enabling:
+
+- timestamp-based filtering of `channel_update` messages: only ask for `channel_update` messages that are newer than the ones you already have.
+- checksum-based filtering of `channel_update` messages: only ask for `channel_update` messages that carry different information from the ones you already have.
+
 ### The `query_short_channel_ids`/`reply_short_channel_ids_end` Messages
 
 1. type: 261 (`query_short_channel_ids`) (`gossip_queries`)
@@ -687,6 +692,19 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`chain_hash`:`chain_hash`]
     * [`u32`:`first_blocknum`]
     * [`u32`:`number_of_blocks`]
+    * [`query_channel_range_tlv`]
+      * type: 1 (`query_option`)
+      * data:
+        * [`1`:`query_option_flags`]
+
+`query_option_flags` is a bitfield represented as a minimally-encoded varint. Bits have the following meaning:
+
+| Bit Position  | Meaning                 |
+| ------------- | ----------------------- |
+| 0             | Sender wants timestamps |
+| 1             | Sender wants checksums  |
+
+Though it is possible, it would not be very useful to ask for checksums without asking for timestamps too: the receiving node may have an older `channel_update` with a different checksum, asking for it would be useless. And if a `channel_update` checksum is actually 0 (which is quite unlikely) it will not be queried.
 
 1. type: 264 (`reply_channel_range`) (`gossip_queries`)
 2. data:
@@ -696,8 +714,36 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`byte`:`complete`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
+    * [`reply_channel_range_tlv`]
+      * type: 1 (`timestamps_tlv`)
+      * data:
+        * [`1`:`encoding_type`]
+        * [`timestamps_tlv_len-1`:`encoded_timestamps`]
+      * type: 3 (`checksums_tlv`)
+      * data:
+        * [`checksums_tlv_len`:`encoded_checksums`]
 
-This allows a query for channels within specific blocks.
+For a single `channel_update`, timestamps are encoded as:
+
+* [`4`:`timestamp_node_id_1`]
+* [`4`:`timestamp_node_id_2`]
+
+Where:
+* `timestamp_node_id_1` is the timestamp of the `channel_update` for `node_id_1`, or 0 if there was no `channel_update` from that node.
+* `timestamp_node_id_2` is the timestamp of the `channel_update` for `node_id_2`, or 0 if there was no `channel_update` from that node.
+
+For a single `channel_update`, checksums are encoded as:
+
+* [`4`:`checksum_node_id_1`]
+* [`4`:`checksum_node_id_2`]
+
+Where:
+* `checksum_node_id_1` is the checksum of the `channel_update` for `node_id_1`, or 0 if there was no `channel_update` from that node.
+* `checksum_node_id_2` is the checksum of the `channel_update` for `node_id_2`, or 0 if there was no `channel_update` from that node.
+
+The checksum of a `channel_update` is the CRC32 checksum of this `channel_update` without its `signature` and `timestamp` fields.
+
+This allows to query for channels within specific blocks.
 
 #### Requirements
 
@@ -707,6 +753,7 @@ The sender of `query_channel_range`:
   that it wants the `reply_channel_range` to refer to
   - MUST set `first_blocknum` to the first block it wants to know channels for
   - MUST set `number_of_blocks` to 1 or greater.
+  - MAY append an additional `query_channel_range_tlv`, which specifies the type of extended information it would like to receive.  
 
 The receiver of `query_channel_range`:
   - if it has not sent all `reply_channel_range` to a previously received `query_channel_range` from this sender:
@@ -724,11 +771,18 @@ The receiver of `query_channel_range`:
     - otherwise:
       - SHOULD set `complete` to 1.
 
+If the incoming message includes a `query_channel_range_tlv` that it understands, the receiver MAY append additional information to its reply.
+- if bit 0 in `query_option_flags` is set, the receive MAY append a `timestamps_tlv` that contains `channel_update` timestamps for all `short_chanel_id`s in `encoded_short_ids`
+- if bit 1 in `query_option_flags` is set, the receive MAY append a `checksums_tlv` that contains `channel_update` checksums for all `short_chanel_id`s in `encoded_short_ids`
+
+
 #### Rationale
 
 A single response might be too large for a single packet, and also a peer can
 store canned results for (say) 1000-block ranges, and simply offer each reply
 which overlaps the ranges of the request.
+
+The addition of timestamp and checksum fields allow a peer to omit querying for redundant updates.
 
 ### The `gossip_timestamp_filter` Message
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -574,6 +574,8 @@ Encoding types:
 * `0`: uncompressed array of `short_channel_id` types, in ascending order.
 * `1`: array of `short_channel_id` types, in ascending order, compressed with zlib deflate<sup>[1](#reference-1)</sup>
 
+This encoding is also used for arrays of other types (timestamps, flags, ...), and specified with an `encoded_` prefix. For example, `encoded_timestamps` is an array of timestamps than can be either compressed (with a `1` prefix) or uncompressed (with a `0` prefix).
+
 Note that a 65535-byte zlib message can decompress into 67632120
 bytes<sup>[2](#reference-2)</sup>, but since the only valid contents
 are unique 8-byte values, no more than 14 bytes can be duplicated
@@ -592,14 +594,13 @@ Query messages can be extended with optional fields that can help reduce the num
     * [`chain_hash`:`chain_hash`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
-    * [`tlvs`:`query_short_channel_ids_tlvs`]
+    * [`query_short_channel_ids_tlvs`:`tlvs`]
 
-1. tlvs: `query_short_channel_ids_tlv`
+1. tlvs: `query_short_channel_ids_tlvs`
 2. types:
     1. type: 1 (`query_flags`)
     2. data:
-        * [`byte`:`encoding_type`]
-        * [`len-1`:`encoded_query_flags`]
+        * [`...*byte`:`encoded_query_flags`]
 
 `encoded_query_flags` is an array of bitfields, one varint per bitfield, one bitfield for each `short_channel_id`. Bits have the following meaning:
 
@@ -697,7 +698,7 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`chain_hash`:`chain_hash`]
     * [`u32`:`first_blocknum`]
     * [`u32`:`number_of_blocks`]
-    * [`tlvs`:`query_channel_range_tlvs`]
+    * [`query_channel_range_tlvs`:`tlvs`]
 
 1. tlvs: `query_channel_range_tlvs`
 2. types:
@@ -722,18 +723,16 @@ Though it is possible, it would not be very useful to ask for checksums without 
     * [`byte`:`complete`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
-    * [`tlvs`:`reply_channel_range_tlvs`]
+    * [`reply_channel_range_tlvs`:`tlvs`]
 
 1. tlvs: `query_channel_range_tlvs`
 2. types:
     1. type: 1 (`timestamps_tlv`)
     2. data:
-        * [`byte`:`encoding_type`]
-        * [`len-1`:`encoded_timestamps`]
+        * [`...*byte`:`encoded_timestamps`]
     1. type: 3 (`checksums_tlv`)
     2. data:
-        * [`byte`:`encoding_type`]
-        * [`len-1`:`encoded_checksums`]
+        * [`...*byte`:`checksums`]
 
 For a single `channel_update`, timestamps are encoded as:
 
@@ -759,7 +758,7 @@ Where:
 
 The checksum of a `channel_update` is the CRC32C checksum as specified in [RFC3720](https://tools.ietf.org/html/rfc3720#appendix-B.4) of this `channel_update` without its `signature` and `timestamp` fields.
 
-This allows to query for channels within specific blocks.
+This allows to query for channels within specific blocks. 
 
 #### Requirements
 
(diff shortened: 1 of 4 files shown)


sstone:
For completeness, there is even a simpler solution: leave `query_channel_range` as is (no extra flag to signify that you want extended data), and if the receiver supports extended queries they always include extended data in their `reply_channel_range`. 

I chose instead to explicitly signal support for extended queries in `query_channel_range`, which is more bandwidth efficient (replies won't include extended data unless you ask for it).

07-routing-gossip.md
```diff
@@ -609,6 +619,15 @@ The receiver:
   - otherwise:
 	- SHOULD set `complete` to 1.
 
+If the incoming message include `option_encoded_query_flags`, instead of systematically replying with
+a `channel_announcement` and the latest `channel_update`s it knows, the receiver:
+  - MUST check that `option_encoded_query_flags` decodes to exactly one flag per `short_channel_id`, and
+  MAY fail the connection if it is not the case
+  - for each `short_channel_id`:
+    - if bit 1 of the corresponding query_flag is set, MUST reply with a `channel_announcement`
```
pm47: You probably meant bits 0-1-2?

```
    - if bit 0 of the corresponding query_flag is set, MUST reply with a `channel_announcement`
    - if bit 1 of the corresponding query_flag is set, MUST reply with the `channel_update` for `node_id_1` 
    - if bit 2 of the corresponding query_flag is set, MUST reply with the `channel_update` for `node_id_2`
```

pm47:
@rustyrussell does this correspond to feature `option_fec_gossip` in #571?

pm47:
For those interested, this feature is now implemented on endurance: 03933884aaf1d6b108397e5efe5c86bcf2d8ca8d2f700eda99db9214fc2712b134@34.250.234.192:9735

07-routing-gossip.md
sstone: Yes I meant bits 0-1-2, fixing the PR, thanks!

sstone:
There is no need for a specific feature bit for this: additional "extended" data appended to query messages will simply be ignored by nodes which do not implement extended queries (see [BOLT 1](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#lightning-message-format)). It's one of the reasons why this PR is much better than my first proposal: extended queries are completely optional and compatible with current implementations and deployed nodes.

`option_fec_gossip` refers to new messages to efficiently sync routing tables based on set reconciliation techniques (IBLT, minisketch, ...) that have not been defined yet.

pm47:
That's how I've implemented it, but yeah it's under specified.

Le mer. 20 mars 2019 à 20:59, Rusty Russell <notifications@github.com> a
écrit :

rustyrussell:
OK, so I've implemented this.  I've kept the existing logic to send node_announcements for any channel_announcement we send (uniqified).  Am testing the vectors now...

rustyrussell:
OK, I've got a fixup which I applied to your INV gossip PR, which should have gone here instead.  Will push.  Also, rebased onto v1.0, otherwise I get unrelated regressions...

rustyrussell:
... and another proposed cleanup (no semantic changes, just integrating the requirements better, I think).

sstone:
This PR has been discussed in our dev meetings a few times and there are still open points, I'll try and summarise them here. They were raised mostly by @Roasbeef  and @cfromknecht , If I've missed anything please add a question/comment which I'll try and address here.
 
**Why do we need this ? We could use heuristics, be more optimistic on the sender's side and work with a "good enough" routing table without trying to get all updates asap. Better heuristics could also be use to broadcast enable/disable updates and reduce "flapping" gossip.** 

It's very true. But this PR really is a fix for current channel queries which are very good for learning about channels you don't have, but almost unusable for updating the ones you already have. Basically they're just a way of reconciling your list of channel ids with your peer's. Then what ? You've learned nothing about what has changed. If you do nothing, you will get updates  as your payments fails with an "update" error, but it's bad UX, especially for mobile nodes when users turn them on, make a single or just a few payments, then turn them off again. And gossip filters (which say “send me everything that is more recent than timestamp X”) don’t really help when you’re offline most of the time since you don’t know what you’re missing to begin with.

I think adding timestamps, and a flag to query specific items and not everything, should not be controversial, especially since now they're completely optional. Do we all agree on that ?

**Why do we also need a checksum ? We should use better heuristics to fix the flapping channels issue.**

Because then you have the option of ignoring updates that don't change routing policies (or simply flagging them as low priority). Again, better heuristics for broadcasting these updates, as well as stricter peer selection/banning heuristics, would help a lot, but having the option to know that an update does not really change anything * before * you've downloaded it is very useful (and this proposal very cheap from a bandwitdh/cpu point of view). 

**Should we use TLV for the new optional fields ?**

I'm now having second thoughts about this one. Current proposal is simple and consistent, I think that using TLV would just slow things down and would not brings us anything ? But I'll give it a try.

sstone:
I've updated this PR:
- optional fields now use TLV format
- timestamps and checksums are now independent

These changes heavily borrow on Rusty's tlv branch (mistakes are mine of course :))

Here's an updated test vector:

``` json
{
  "msg" : {
    "type" : "QueryChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 100000,
    "numberOfBlocks" : 1500,
    "extensions" : [ ]
  },
  "hex" : "01070f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206000186a0000005dc"
}
{
  "msg" : {
    "type" : "QueryChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 35000,
    "numberOfBlocks" : 100,
    "extensions" : [ "WANT_TIMESTAMPS | WANT_CHECKSUMS" ]
  },
  "hex" : "01070f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206000088b800000064010103"
}
{
  "msg" : {
    "type" : "ReplyChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 756230,
    "numberOfBlocks" : 1500,
    "complete" : 1,
    "shortChannelIds" : {
      "encoding" : "UNCOMPRESSED",
      "array" : [ "0x0x142", "0x0x15465", "0x69x42692" ]
    }
  },
  "hex" : "01080f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206000b8a06000005dc01001900000000000000008e0000000000003c69000000000045a6c4"
}
{
  "msg" : {
    "type" : "ReplyChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 1600,
    "numberOfBlocks" : 110,
    "complete" : 1,
    "shortChannelIds" : {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ "0x0x142", "0x0x15465", "0x4x3318" ]
    }
  },
  "hex" : "01080f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206000006400000006e01001601789c636000833e08659309a65878be010010a9023a"
}
{
  "msg" : {
    "type" : "ReplyChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 122334,
    "numberOfBlocks" : 1500,
    "complete" : 1,
    "shortChannelIds" : {
      "encoding" : "UNCOMPRESSED",
      "array" : [ "0x0x12355", "0x7x30934", "0x70x57793" ]
    },
    "timestamps" : {
      "encoding" : "UNCOMPRESSED",
      "timestamps" : [ {
        "timestamp1" : 164545,
        "timestamp2" : 948165
      }, {
        "timestamp1" : 489645,
        "timestamp2" : 4786864
      }, {
        "timestamp1" : 46456,
        "timestamp2" : 9788415
      } ]
    },
    "checksums" : {
      "checksums" : [ {
        "checksum1" : 1111,
        "checksum2" : 2222
      }, {
        "checksum1" : 3333,
        "checksum2" : 4444
      }, {
        "checksum1" : 5555,
        "checksum2" : 6666
      } ]
    }
  },
  "hex" : "01080f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e22060001ddde000005dc01001900000000000000304300000000000778d6000000000046e1c1011900000282c1000e77c5000778ad00490ab00000b57800955bff031800000457000008ae00000d050000115c000015b300001a0a"
}
{
  "msg" : {
    "type" : "ReplyChannelRange",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "firstBlockNum" : 122334,
    "numberOfBlocks" : 1500,
    "complete" : 1,
    "shortChannelIds" : {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ "0x0x12355", "0x7x30934", "0x70x57793" ]
    },
    "timestamps" : {
      "encoding" : "COMPRESSED_ZLIB",
      "timestamps" : [ {
        "timestamp1" : 164545,
        "timestamp2" : 948165
      }, {
        "timestamp1" : 489645,
        "timestamp2" : 4786864
      }, {
        "timestamp1" : 46456,
        "timestamp2" : 9788415
      } ]
    },
    "checksums" : {
      "checksums" : [ {
        "checksum1" : 1111,
        "checksum2" : 2222
      }, {
        "checksum1" : 3333,
        "checksum2" : 4444
      }, {
        "checksum1" : 5555,
        "checksum2" : 6666
      } ]
    }
  },
  "hex" : "01080f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e22060001ddde000005dc01001801789c63600001036730c55e710d4cbb3d3c080017c303b1012201789c63606a3ac8c0577e9481bd622d8327d7060686ad150c53a3ff0300554707db031800000457000008ae00000d050000115c000015b300001a0a"
}
{
  "msg" : {
    "type" : "QueryShortChannelIds",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "shortChannelIds" : {
      "encoding" : "UNCOMPRESSED",
      "array" : [ "0x0x142", "0x0x15465", "0x69x42692" ]
    },
    "extensions" : [ ]
  },
  "hex" : "01050f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206001900000000000000008e0000000000003c69000000000045a6c4"
}
{
  "msg" : {
    "type" : "QueryShortChannelIds",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "shortChannelIds" : {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ "0x0x4564", "0x2x47550", "0x69x42692" ]
    },
    "extensions" : [ ]
  },
  "hex" : "01050f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206001801789c63600001c12b608a69e73e30edbaec0800203b040e"
}
{
  "msg" : {
    "type" : "QueryShortChannelIds",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "shortChannelIds" : {
      "encoding" : "UNCOMPRESSED",
      "array" : [ "0x0x12232", "0x0x15556", "0x69x42692" ]
    },
    "extensions" : [ {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ 1, 2, 4 ]
    } ]
  },
  "hex" : "01050f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e22060019000000000000002fc80000000000003cc4000000000045a6c4010c01789c6364620100000e0008"
}
{
  "msg" : {
    "type" : "QueryShortChannelIds",
    "chainHash" : "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206",
    "shortChannelIds" : {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ "0x0x14200", "0x0x46645", "0x69x42692" ]
    },
    "extensions" : [ {
      "encoding" : "COMPRESSED_ZLIB",
      "array" : [ 1, 2, 4 ]
    } ]
  },
  "hex" : "01050f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206001801789c63600001f30a30c5b0cd144cb92e3b020017c6034a010c01789c6364620100000e0008"
}
```

07-routing-gossip.md
```diff
@@ -636,6 +690,34 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`1`:`complete`]
     * [`2`:`len`]
     * [`len`:`encoded_short_ids`]
+    * [`query_channel_range_tlv`]
```
cfromknecht: the name `query_channel_range_tlv` confused me a bit. might it make more sense to call it `reply_channel_range_tlv`? there is a reference further down that would also need to be updated. correct if i'm misunderstanding the rationale :)

07-routing-gossip.md
```diff
@@ -636,6 +690,34 @@ timeouts.  It also causes a natural ratelimiting of queries.
...
+For a single `channel_update`, checksums are encoded as:
+
+* [`4`:`checksum_node_id_1`]
+* [`4`:`checksum_node_id_2`]
+
+Where:
+* `checksum_node_id_1` is the checksum of the `channel_update` for `node_id_1`, or 0 if there was no `channel_update` from that node.
+* `checksum_node_id_2` is the checksum of the `channel_update` for `node_id_2`, or 0 if there was no `channel_update` from that node.
+
+The checksum of a `channel_update` is the Adler32 checksum of this `channel_update` without its `signature` and `timestamp` fields.
```
Roasbeef: Any rationale as to why this particular algorithm was chosen?

07-routing-gossip.md
```diff
@@ -559,13 +559,34 @@ are unique 8-byte values, no more than 14 bytes can be duplicated
...
     * [`len`:`encoded_short_ids`]
+    * [`query_short_channel_ids_tlv`]
+      * type: 1 (`query_flags`)
+      * data:
+        * [`1`:`encoding_type`]
+        * [`len-1`:`encoded_query_flags`]
+
+</a>
+
+`encoded_query_flags` is an array of bitfields, one for each channel id. Bits have the following meaning:
```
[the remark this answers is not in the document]
sstone: Default endianness is big endian as per BOLT 1, I don't think it needs to be specified ? 
Yes using `short_channel_id` is more consistent with what we already have I'll rename

07-routing-gossip.md
```diff
@@ -559,13 +559,34 @@ are unique 8-byte values, no more than 14 bytes can be duplicated
...
 1. type: 261 (`query_short_channel_ids`) (`gossip_queries`)
 2. data:
     * [`32`:`chain_hash`]
     * [`2`:`len`]
     * [`len`:`encoded_short_ids`]
+    * [`query_short_channel_ids_tlv`]
+      * type: 1 (`query_flags`)
+      * data:
+        * [`1`:`encoding_type`]
+        * [`len-1`:`encoded_query_flags`]
```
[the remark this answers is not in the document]
sstone: `encoded_query_flags` is a binary blob of size `len - 1`. If `encoding_type` is zero then yes its size should be `num_sids`, but if its one then you need to decompress it first, then the length of the decompressed buffer should be `num_sids`. I'll rephrase and make this more explicit

07-routing-gossip.md
sstone: You're right, I will rename this field

07-routing-gossip.md
sstone: Because it's even simpler and faster than CRC32 and should be as good for our use case (afaik it's less collision-resistant than CRC32 for very small or very large messages).

07-routing-gossip.md
```diff
@@ -636,6 +690,34 @@ timeouts.  It also causes a natural ratelimiting of queries.
...
+* `timestamp_node_id_1` is the timestamp of the `channel_update` for `node_id_1`, or 0 if there was no `channel_update` from that node.
+* `timestamp_node_id_2` is the timestamp of the `channel_update` for `node_id_2`, or 0 if there was no `channel_update` from that node.
+
+For a single `channel_update`, checksums are encoded as:
+
+* [`4`:`checksum_node_id_1`]
+* [`4`:`checksum_node_id_2`]
+
+Where:
+* `checksum_node_id_1` is the checksum of the `channel_update` for `node_id_1`, or 0 if there was no `channel_update` from that node.
```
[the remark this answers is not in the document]
sstone: Will do, it's probably not very useful to request checksums without timestamps.

07-routing-gossip.md
sstone: Feedback was clearly in favour of CRC32, so I've changed the checksum algorithm

07-routing-gossip.md
```diff
@@ -559,13 +559,34 @@ are unique 8-byte values, no more than 14 bytes can be duplicated
 across the stream: as each duplicate takes at least 2 bits, no valid
 contents could decompress to more then 3669960 bytes.
 
+Query messages can be extended with optional fields that can help reduce the number of messages needed to synchronize routing tables by enabling:
+
+- timestamp-based filtering of `channel_update` messages: don't ask for `channel_update` messages that
+ are older than the ones you already have.
```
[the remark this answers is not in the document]
sstone: I've rephrased this section to make it clear that you want to ask for data that is newer than what you have

t-bast:
ACK `1978561`

sstone:
Since #607 has been approved (yes !), I've implemented a suggestion by @t-bast to use minimally-encoded varints instead of single bytes for query flags. Given the number of options atm they will all still be encoded on a single byte, and test vectors remain valid. I think we're good to go now and hope this will be merged with the TLV PR.

rustyrussell:
OK, I took https://github.com/lightningnetwork/lightning-rfc/pull/557/commits/19785614d6be5a6fbbc79ccc77dc8efa13b4e1df and rebased it on master.

I also split it into separate commits: logically this adds two (useful) features, so they should be separate commits IMHO.  Certainly easier to implement that way.  I also reformatted so the tooling can extract them properly.

Two minor fixups, as well.  Consider simply replacing it with this if you're happy?  It's a bit of a mess now!

https://github.com/rustyrussell/lightning-rfc/tree/guilt/pr-557

sstone:
Thanks! 

I've rebased against master, and specified that checksums use CRC32C, but left out the tooling-related reformatting (I have not made up my mind yet:)) and the use of a byte prefix to specify an encoding type for checksums because I don't think that compression will help here (unless we get many 0 timestamps) but I don't mind adding it if you feel strongly about it (and it's just one byte).

I see your point about cleaning up the commit history but I'm not sure about changing it yet. Some of them are not relevant now that we've switched to TLV but this PR has been open for some time and I thought that for now we should keep its history unchanged. And when we merge we'll squash and cleanup the commit message. But I'll take anything that makes this more "mergeable" so if you (or anyone else) strongly feels that we should have a cleaner commit history now I'm more than happy to reset my branch and point it to yours.

07-routing-gossip.md
```diff
@@ -647,6 +692,21 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`chain_hash`:`chain_hash`]
     * [`u32`:`first_blocknum`]
     * [`u32`:`number_of_blocks`]
+    * [`byte`:`option_extended_query_flag`]
+    * [`byte`:`query_flag`] (`option_extended_query_flags`)
+    * [`query_channel_range_tlv`]
+      * type: 1 (`query_option`)
+      * data:
+        * [`byte`:`query_option_flags`]
```
cfromknecht: this definition says that the type is `byte`, but elsewhere it is described as a minimally encoded varint. which varint encoding is being proposed?

07-routing-gossip.md
```diff
@@ -618,8 +646,24 @@ The receiver:
...
+      - with a `channel_announcement` and the latest `channel_update` for each end
+    - otherwise:
+      - We define `query_flag` for the Nth `short_channel_id` in
+        `encoded_short_ids` to be the Nth varint of the decoded
+        `encoded_query_flags`.
+      - if bit 0 of `query_flag` is set:
+        - MUST reply with a `channel_announcement`
+      - if bit 1 of `query_flag` is set and it has received a `channel_update` from `node_id_1`:
+        - MUST reply with the latest `channel_update` for `node_id_1`
+      - if bit 2 of `query_flag` is set and it has received a `channel_update` from `node_id_2`:
```
cfromknecht: are we missing the requirements for bits 3 and 4?

07-routing-gossip.md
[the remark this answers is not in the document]
cfromknecht: this length might also be inaccurate if we are encoding them as varints rather than bytes, but tbh i don't have a better suggestion on how to describe the expected length :P

rustyrussell:
> I've rebased against master, and specified that checksums use CRC32C, but left out the tooling-related reformatting (I have not made up my mind yet:)) and the use of a byte prefix to specify an encoding type for checksums because I don't think that compression will help here (unless we get many 0 timestamps) but I don't mind adding it if you feel strongly about it (and it's just one byte).

Agreed it's probably a noop.  But everything else called `encoded_` is of that format, so it makes sense.

rustyrussell:
Happy for you to keep appending :) 

The reformat is needed for tools/extract-formats.py, so I can actually code this up. But I can carry it on top.

I definitely want the two parts as separate commits, though.  They're independent, and simple to implement that way; I got quite confused initially as I was implementing.  I think a feature bit might be nice, too: this lets you selectively query peers who support this, rather than intuiting support from their replies.  From previous experience, this should be two separate feature bits.

sstone:
@rustyrussell So you would have 2 feature bits for 2 different functional changes:
- extended queries, where you can specify what you want (channel announcement, channel updates, node announcement)
- extended information, where nodes can append additional data such as timestamps and checksums

For me these changes work together and I don't really see how you could use one without the other: extended info are not useful if you cannot specify what you want, and without extended info you won't know what to query ?

I agree that being able to tell in advance if extended queries are supported could be useful, but then I would just add one new feature bit and not two.

07-routing-gossip.md
sstone: Right, it should be a  `varint`  (using the LN definition of `varint`)

sstone:
@rustyrussell I've included your formatting changes. I think this PR can be merged without defining specific feature bits (which could be added later if needed) but I'll see how the spec meeting goes.

rustyrussell:
Minor cleanups.  I can now parse this into our source code (with appropriate changes on our side!):

```diff
diff --git a/07-routing-gossip.md b/07-routing-gossip.md
index 722a159..497d5c4 100644
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -591,14 +591,14 @@ Query messages can be extended with optional fields that can help reduce the num
     * [`chain_hash`:`chain_hash`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
-    * [`tlvs`:`query_short_channel_ids_tlvs`]
+    * [`query_short_channel_ids_tlvs`:`tlvs`]
 
-1. tlvs: `query_short_channel_ids_tlv`
+1. tlvs: `query_short_channel_ids_tlvs`
 2. types:
     1. type: 1 (`query_flags`)
     2. data:
         * [`byte`:`encoding_type`]
-        * [`len-1*byte`:`encoded_query_flags`]
+        * [`...*byte`:`encoded_query_flags`]
 
 `encoded_query_flags` is an array of bitfields, one varint per bitfield, one bitfield for each `short_channel_id`. Bits have the following meaning:
 
@@ -696,7 +696,7 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`chain_hash`:`chain_hash`]
     * [`u32`:`first_blocknum`]
     * [`u32`:`number_of_blocks`]
-    * [`tlvs`:`query_channel_range_tlvs`]
+    * [`query_channel_range_tlvs`:`tlvs`]
 
 1. tlvs: `query_channel_range_tlvs`
 2. types:
@@ -721,18 +721,18 @@ Though it is possible, it would not be very useful to ask for checksums without
     * [`byte`:`complete`]
     * [`u16`:`len`]
     * [`len*byte`:`encoded_short_ids`]
-    * [`tlvs`:`reply_channel_range_tlvs`]
+    * [`reply_channel_range_tlvs`:`tlvs`]
 
-1. tlvs: `query_channel_range_tlvs`
+1. tlvs: `reply_channel_range_tlvs`
 2. types:
     1. type: 1 (`timestamps_tlv`)
     2. data:
         * [`byte`:`encoding_type`]
-        * [`len-1*byte`:`encoded_timestamps`]
+        * [`...*byte`:`encoded_timestamps`]
     1. type: 3 (`checksums_tlv`)
     2. data:
         * [`byte`:`encoding_type`]
-        * [`len-1*byte`:`encoded_checksums`]
+        * [`...*byte`:`encoded_checksums`]
 
 For a single `channel_update`, timestamps are encoded as:
 
diff --git a/tools/extract-formats.py b/tools/extract-formats.py
index 1b6a754..2ae87cc 100755
--- a/tools/extract-formats.py
+++ b/tools/extract-formats.py
@@ -29,7 +29,7 @@ tlvline = re.compile(
 subtypeline = re.compile(
     '1\. subtype: `(?P<name>[A-Za-z2_]+)`( \(`?(?P<option>[^)`]*)`\))?')
 dataline = re.compile(
-    '\s+\* \[`(?P<typefield>[-_a-zA-Z0-9*+]+)`:`(?P<name>[_a-z0-9]+)`\]( \(`?(?P<option>[^)`]*)`?\))?')
+    '\s+\* \[`(?P<typefield>[.-_a-zA-Z0-9*+]+)`:`(?P<name>[_a-z0-9]+)`\]( \(`?(?P<option>[^)`]*)`?\))?')
 
 # Generator to give us one line at a time.
 def next_line(args, lines):
@@ -121,12 +121,12 @@ def parse_type(genline, output, name, value, option, in_tlv=None):
 #    1. type: 1 (`query_flags`)
 #    2. data:
 #      * [`byte`:`encoding_type`]
-#      * [`tlv_len-1*byte`:`encoded_query_flags`]
+#      * [`...*byte`:`encoded_query_flags`]
 #
 # output:
 #  tlvtype,query_short_channel_ids_tlvs,query_flags,1
 #  tlvdata,query_short_channel_ids_tlvs,query_flags,encoding_type,byte,
-#  tlvdata,query_short_channel_ids_tlvs,query_flags,encoded_query_flags,byte,tlv_len-1
+#  tlvdata,query_short_channel_ids_tlvs,query_flags,encoded_query_flags,byte,...
 def parse_tlv(genline, output, name, option):
     i, line = next(genline)
 
```

rustyrussell:
OK, I've just reproduced your test vectors, with two caveats:

1. I had to change my zlib compression level from Z_BEST_COMPRESSION to Z_DEFAULT_COMPRESSION (turns out this is actually written in the second byte of the zlib output!).

2. I had to remove the [`byte`:`encoding_type`] from encoded_checksums, and change it to:

```
     1. type: 3 (`checksums_tlv`)
     2. data:
        * [`...*channel_update_checksums`:`encoded_checksums`]
```

Which is nice an explicit.

sstone:
Nice! Let me see if I can generate test vectors with Z_BEST_COMPRESSION settings which makes more sense here. 
If changing the description of TLV fields makes them easier to parse automatically then I'm all for it. What about:
```
1. tlvs: `query_channel_range_tlvs`
2. types:
    1. type: 1 (`timestamps_tlv`)
    2. data:
        * [`encoded(array(channel_update_timestamps))`:`encoded_timestamps`]
    1. type: 3 (`checksums_tlv`)
    2. data:
        * [`encoded(array(channel_update_checksums))`:`encoded_checksums`]
``` 
Here `array()` and `encoded()` could be defined in BOLT01 ?
Of course this syntax could be changed to `encoded*array*type` or similar if it's more usable. Wdyt ?

sstone:
@rustyrussell using Z_BEST_COMPRESSION I can generate the same values as CL except for chekcksum TLV since you don't yet implement uncompressed/zlib encoding for them yet.

rustyrussell:
> Here `array()` and `encoded()` could be defined in BOLT01 ?

... and in `tools/extract-format.py`.  Let's do that as a separate change.

rustyrussell:
OK, so meeting agreed to remove compression flag and use a straight array for checksums:

http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-08-05-20.03.html

So I'm going to merge this for c-lightning (only with --enable-experimental-features) for next release: https://github.com/ElementsProject/lightning/pull/2900

I'm not completely opposed to removing the EXPERIMENTAL_FEATURES conditional in the next few days (we tag -rc1 on the 10th of alternate months) if spec is sorted, feature flag is added, and Eclair confirms interoperation in practice.  But that's a tight timeline!

sstone:
@rustyrussell I've changed the formatting and added a JSON test vector. Its format is just a bit different from the one you used but I could get your test to pass with a few changes

rustyrussell:
OK, I found a bug in my code while implementing the protocol tests; my checksums are wrong.

I assume the crc32 is not supposed to cover the 2 `type` bytes at the start of the channel_update?  ie. our code (now!):

```
	assert(tal_count(channel_update) > 2 + 64 + 32 + 8 + 4);
	sum = crc32c(0, channel_update + 2 + 64, 32 + 8);
	sum = crc32c(sum, channel_update + 2 + 64 + 32 + 8 + 4,
		     tal_count(channel_update) - (64 + 2 + 32 + 8 + 4));
```

And for a specific test cases:

1.  signature=76df7e70c63cc2b63ef1c062b99c6d934a80ef2fd4dae9e1d86d277f47674af3255a97fa52ade7f129263f591ed784996eba6383135896cc117a438c80293282 chain_hash=06226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f short_channel_id=103x1x0 timestamp=1565587763 message_flags=0 channel_flags=0 cltv_expiry_delta=144 htlc_minimum_msat=0 fee_base_msat=1000 fee_proportional_millionths=10
`010276df7e70c63cc2b63ef1c062b99c6d934a80ef2fd4dae9e1d86d277f47674af3255a97fa52ade7f129263f591ed784996eba6383135896cc117a438c8029328206226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f00006700000100005d50f933000000900000000000000000000003e80000000a`
crc32 = 0x1112fa30

2. signature=06737e9e18d3e4d0ab4066ccaecdcc10e648c5f1c5413f1610747e0d463fa7fa39c1b02ea2fd694275ecfefe4fe9631f24afd182ab75b805e16cd550941f858c chain_hash=06226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f short_channel_id=109x1x0 timestamp=1565587765 message_flags=1 channel_flags=0 cltv_expiry_delta=48 htlc_minimum_msat=0 fee_base_msat=100 fee_proportional_millionths=11 htlc_maximum_msat=100000
`010206737e9e18d3e4d0ab4066ccaecdcc10e648c5f1c5413f1610747e0d463fa7fa39c1b02ea2fd694275ecfefe4fe9631f24afd182ab75b805e16cd550941f858c06226e46111a0b59caaf126043eb5bbf28c34f3a5e332a1fc7b2b73cf188910f00006d00000100005d50f935010000300000000000000000000000640000000b00000000000186a0`
crc32 = f32ce968

sstone:
@rustyrussell thanks! we do skip the first 2 bytes of the encoded `channel_update` and find the same checksums, I'll make new test vectors and include your data.

t-bast:
CL and Eclair have implemented this and correctly inter-operate.
LL said implementation will come later, but concept acked.
@cfromknecht can you tell us if you're ok with the current state of the PR and merging it as-is?

PR #571, opened 2019-02-12
Feature bit unification and renaming

rustyrussell:
This is 5 commits, each standalone:

1. Rename "local" to "node" features, "global" to "channel" features.  The original idea was that
   "local" features only affect direct peers.  That still stands, but the new names are clearer.
   Note: There are still no "channel" features defined.
2. Note that features share the same number space, needed for #5.
3. Remove init message requirement to disconnect on unknown `channelfeatures`.  This just
    means you can't open a channel.  Unknown node features means you can't even talk.  
4. Explicitly mirror channelfeatures in channel_announcement.  These are things you might need to 
    know to use that specific channel, or might be required, depending on the feature itself.   Removes
    the "don't relay these if they have an unknown feature" rule.
5. Explicitly mirror what channelfeatures we support in `node_announcement` (usually as optional
    features, of course).  This lets you find nodes with specific features.

## Concrete Examples

Consider a node which supports a d-log-HTLC channel, and doesn't support legacy HTLCs on that channel:

1. This is clearly a channelfeature.
2. It may have negotiated this with the peer as an optional or required feature.
3. It must advertize this in channel_announcement as a required feature, since you need to understand it to use the channel at all.
4. It gets advertized in node_announcement, as required or optional, depending on 2 above.

Now, consider a node which can create wumbo (> 2^24 satoshi) channels:

1. This is a nodefeature.
2. It's (presumably) optional.
3. It doesn't appear in channel_announcement.
4. It gets advertized in node_announcement, as (presumably) optional.

Finally, consider a node which can handle wumbo HTLCs:

1. This is a channelfeature.
2. It's (presumably) optional.
3. It gets advertized in channel_announcement as optional, since you don't need to understand it.
4. It gets advertized in node_announcmenet, as (presumably) optional.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -727,7 +727,7 @@ In the case where the `channel_announcement` is nonetheless missed,
 ## Initial Sync
 
 If a node requires an initial sync of gossip messages, it will be flagged
-in the `init` message, via a feature flag ([BOLT #9](09-features.md#assigned-localfeatures-flags)).
+in the `init` message, via a feature flag ([BOLT #9](09-features.md#assigned-nodefeatures-flags)).
 
 Note that the `initial_routing_sync` feature is overridden (and should
 be considered equal to 0) by the `gossip_queries` feature if the
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -248,8 +248,8 @@ nodes not associated with an already known channel are ignored.
 1. type: 257 (`node_announcement`)
 2. data:
    * [`64`:`signature`]
-   * [`2`:`flen`]
-   * [`flen`:`features`]
+   * [`2`:`cflen`]
+   * [`cflen`:`combinedfeatures`]
    * [`4`:`timestamp`]
    * [`33`:`node_id`]
    * [`3`:`rgb_color`]
@@ -304,8 +304,9 @@ The origin node:
   to 0.
   - SHOULD ensure `ipv4_addr` AND `ipv6_addr` are routable addresses.
   - MUST NOT include more than one `address descriptor` of the same type.
-  - SHOULD set `flen` to the minimum length required to hold the `features`
+  - SHOULD set `cflen` to the minimum length required to hold the `combinedfeatures`
   bits it sets.
+  - MUST set `combinedfeatures` to the logical OR of `nodefeatures` and `channelfeatures`.
 
 The receiving node:
   - if `node_id` is NOT a valid compressed public key:
@@ -316,12 +317,6 @@ The receiving node:
 any future fields appended to the end):
     - SHOULD fail the connection.
     - MUST NOT process the message further.
-  - if `features` field contains _unknown even bits_:
-    - MUST NOT parse the remainder of the message.
-    - MAY discard the message altogether.
-    - SHOULD NOT connect to the node.
-  - MAY forward `node_announcement`s that contain an _unknown_ `features` _bit_,
-  regardless of if it has parsed the announcement or not.
   - SHOULD ignore the first `address descriptor` that does NOT match the types
   defined above.
   - if `addrlen` is insufficient to hold the address descriptors of the
@@ -342,11 +337,6 @@ any future fields appended to the end):
     - SHOULD insinuate their self-signed origins.
 
 ### Rationale
-
-New node features are possible in the future: backwards compatible (or
-optional) ones will have _odd_ `feature` _bits_, incompatible ones will have
-_even_ `feature` _bits_. These may be propagated by nodes even if they
-cannot process the announcements themselves.
 
 New address types may be added in the future; as address descriptors have
 to be ordered in ascending order, unknown ones can be safely ignored.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -140,7 +140,7 @@ announcement message: this is accomplished by having a signature from each
     * [`64`:`bitcoin_signature_1`]
     * [`64`:`bitcoin_signature_2`]
     * [`2`:`len`]
-    * [`len`:`features`]
+    * [`len`:`channelfeatures`]
     * [`32`:`chain_hash`]
     * [`8`:`short_channel_id`]
     * [`33`:`node_id_1`]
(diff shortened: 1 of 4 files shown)


t-bast:
I think this is a good clarification.
ACK from me for next meeting.

t-bast:
As discussed in the IRC meeting, people still find the names confusing so here are my suggestions, I'll let the consensus judge if they're better or worse ;)

- Suggestions for renaming `local` features
  - `connection` or `conn` features (since it only impacts people I'm connected to)
  - `peer` features (it only impacts my direct peers)
  - `neighbour` or `neighbor` features (meh)
  - keep `local` (I honestly think `local` is a good name, it's more the `global` part that I found a bit misleading)
- Suggestions for renaming `global` features
  - `network` features (they are interesting to the whole network)
  - `routing` features (if we assume that routing is the only thing we're using remote nodes for - not sure if that will always be true)

I'll edit this comment if I have other ideas.

rustyrussell:
I like `local` and `routing` features, TBH.

Roasbeef:
Thanks for those examples! I understand the underlying motivation now after reading through them. As for names, my vote goes for `node` features (former global) and `link` features.

Roasbeef:
It seems the only "pure" link feature we have is `initial-routing-sync`, the rest can lie in either domain.

t-bast:
> As for names, my vote goes for node features (former global) and link features.

I think `node` is confusing. Rusty proposed it to replace the current `local` and you're now proposing it to replace the current `global`. That clearly indicates that it's misleading, doesn't it? :)

I also think that `link` is confusing, I can see arguments for using it to replace either `local` or `global`...
Because it could mean that it defines properties of your local links (channels?) so it's a `local` feature, or it's a `global` feature because it defines properties that remote non-neighbor nodes should know about your links (channel?).

cfromknecht:
As discussed on IRC, the naming that makes the most sense to me personally is:
 - `global -> node`: functionally equivalent to service bits in bitcoin, apart from the fact that they are gossiped. Emphasize that the features are applicable to the node as a whole (depending on the feature, that it applies to _all_ of a node's channels), e.g. can this node: support variable frame onions, extended gossip queries, DLP, link-level payment sharding, etc.
 - `local -> connection`: emphasizes that these feature bits are only applied for the duration of the connection, and may change upon a subsequent reconnect, e.g. will this connection: allow funding wumbo channels, extended gossip queries, DLP, etc.

To me `routing` is too specific, as there are already several features that could be advertised that have nothing to do with routing (and we are likely to add more in the future). `local` is okay, but it's a little ambiguous what that means and could be clearer.

07-routing-gossip.md
```diff
@@ -173,16 +173,16 @@ The origin node:
...
   it sets.
+  - MUST set a `channelfeatures` bit for each `channelfeature` negotiated for
+    the channel, according to [BOLT #9](09-features.md).
 
 The receiving node:
   - MUST verify the integrity AND authenticity of the message by verifying the
   signatures.
-  - if there is an unknown even bit in the `features` field:
-    - MUST NOT parse the remainder of the message.
+  - if there is an unknown even bit in the `channelfeatures` field:
```
cfromknecht: question about commit message: why would `key-exchange-instead-of-hash-preimage` be advertised as even? why is not okay to advertise it as optional and infer whatever the sender signals/uses?

07-routing-gossip.md
```diff
@@ -316,12 +317,6 @@ The receiving node:
 any future fields appended to the end):
     - SHOULD fail the connection.
     - MUST NOT process the message further.
-  - if `features` field contains _unknown even bits_:
-    - MUST NOT parse the remainder of the message.
-    - MAY discard the message altogether.
-    - SHOULD NOT connect to the node.
-  - MAY forward `node_announcement`s that contain an _unknown_ `features` _bit_,
```
cfromknecht: this seems like a good change, we probably shouldn't be forwarding messages that we were unable to parse

07-routing-gossip.md
```diff
@@ -316,12 +317,6 @@ The receiving node:
 any future fields appended to the end):
     - SHOULD fail the connection.
     - MUST NOT process the message further.
-  - if `features` field contains _unknown even bits_:
```
cfromknecht: if i understand correctly, this will removes the ability to partition the gossip network as a result of nodes not understanding an even feature bit? do we still want to retain the ability to do so via unknown TLV types?

07-routing-gossip.md
rustyrussell: If you don't support the old style any more, you would advertize it as even, right?

t-bast:
I quite like renaming `local` to `connection` (seems like we all have a rough agreement on that one).
I'm not completely convinced by `node` to replace `global` but my best proposal (`network`) isn't perfect either, so I could live with it.

07-routing-gossip.md
rustyrussell: Hmm, that's an excellent question!  I... don't know.  Allowing "invalid" TLVs to propagate seems like a burden on implementations, but maybe it's a feature somehow?

07-routing-gossip.md
rustyrussell: Will modify commit msg.  it was:

> Each feature can itself specify how it's advertized here: an
key-exchange-instead-of-hash-preimage feature would need to advertize
as even (you need to understand it to use it), for example, but a
wumbo feature would advertize as odd.

Now:

> Each feature can itself specify how it's advertized here: an
key-exchange-instead-of-hash-preimage feature which didn't support hash
preimages any more would need to advertize
as even (you need to understand it to use it), for example, but a
wumbo feature would advertize as odd.

Roasbeef:
node+connection SGTM!

07-routing-gossip.md
cfromknecht: thanks for explaining, clear to me now :)

07-routing-gossip.md
cfromknecht: In due time we may discover, I see arguments for both sides 🤷‍♂️

PR #572, opened 2019-02-12
Specify tlv format, initial MPP

rustyrussell:


rustyrussell:
Too ambitious, stepping back for a moment.

PR #578, opened 2019-02-20
BOLT 11: change default expiry to 86400 seconds (1 day)

prusnak:
context: https://twitter.com/francispouliot_/status/1098193142170439680

> We open channel if no route found, wait 6 confs (1h), pay. Default Bolt11 expiry in C-Lightning/LND is 3600 secs (1h). Payment failures are guaranteed by default UX configs.
> 
> Should be 86400 (1 day).
> 
> The average time it takes to open a channel is exactly the time it takes for an invoice to expire. If an invoice expires, I need to send an email to the user, he must come back to the app, create a new Bolt11 w/same node, and we try again. Changing 3600 to 86400 solves it
> 

fix in c-lightning: https://github.com/FrancisPouliot/lightning/commit/3f0d93bfd919213531c07d26b84ec9cb653237f5

fix in LND: https://github.com/prusnak/lnd/commit/0b6212c3d9a88a2f66fd736647f9bc88f56c18a0


rustyrussell:
This leaves it uncertain for older clients, who would see an empty expiry as meaning 1 hour, not 1 day.  However, the failure mode in this case is fairly harmless, and nodes can explicitly use `x=86400` during the transition.

PR #584, opened 2019-02-28
[WIP] BOLT 7: Inventory-based gossip

sstone:
This is a proposal for adding inventory messages to the gossip protocol. With the current design, nodes which receive gossip messages (node announcements, channel announcements and updates) will relay them to their peers, which will result in duplicates for nodes that are connected to many different peers.

Nodes which support the `option_inv_gossip` feature will instead broadcast inventory messages which contain identifiers for gossip messages they've received. Receiving node will compare these identifiers to their local view of the routing table, and ask for missing or outdated messages using channel queries. This implies that `option_inv_gossip` cannot be used without `gossip_queries`.

It builds upon PR #571 (unification of feature bits, which includes a definition for `option_inv_gossip`) and #557 (extended channel queries).
More specifically, it is a "companion" PR to #557:
- extended channel queries are used to efficiently synchronise routing tables between a node that is often offline and a very limited number of peers
- inventory-based gossip is used to minimize duplicate gossip traffic for nodes that are connected to many different peers
- they share similar concern: how to efficiently advertise information about channel announcement and updates

In fact, the inventory message that I propose to use here is almost identical to an extended `reply_channel_range` message and includes short channel ids, checksums and timestamps, to allow receiving to efficiently query messages based on content (checksum) and timestamps.

Node announcements have been left out on purpose because they have a much more limited impact on bandwidth than channel updates, and because gains would be smaller (you would still need to advertise node ids which are 33 bytes long).


07-routing-gossip.md
```diff
@@ -559,13 +559,21 @@ are unique 8-byte values, no more than 14 bytes can be duplicated
...
+
 ### The `query_short_channel_ids`/`reply_short_channel_ids_end` Messages
 
 1. type: 261 (`query_short_channel_ids`) (`gossip_queries`)
 2. data:
     * [`32`:`chain_hash`]
     * [`2`:`len`]
     * [`len`:`encoded_short_ids`]
+    * [`2`:`option_len`]
+    * [`len`:`option_encoded_query_flags`]
```
rustyrussell: `len` needs to be `option_len` here.

07-routing-gossip.md
```diff
@@ -636,6 +656,9 @@ timeouts.  It also causes a natural ratelimiting of queries.
     * [`1`:`complete`]
     * [`2`:`len`]
     * [`len`:`encoded_short_ids`]
+    * [`1`:`option_extended_query_flag`]
+    * [`2`:`option_extended_info_len`]
+    * [`len`:`option_extended_info`]
```
rustyrussell: Here too, `len` -> `option_extended_info_len`.

n1bor:
Other thought was to get nodes to opt-out of updates for most channels from most peers. Since node has a full map of the network it could tell most peers not to send updates for a list of channels that it knows it will get quicker from multiple other peers.
Some info here: https://github.com/ACINQ/eclair/issues/864

pm47:
Fwiw we can already opt out entirely of updates on a peer by peer basis by
not sending `gossip_timestamp_filter`.

07-routing-gossip.md
```diff
@@ -798,6 +838,33 @@ and allows bootstrapping for new nodes as well as updating for nodes that
 have been offline for some time.  The `gossip_queries` option
 allows for more refined synchronization.
 
+### Inventory-based Rebroadcasting
+
+With the simple rebroadcasting scheme described above, nodes that are connected to many different peers will receive duplicates for the same gossip messages.
+
+A more efficient scheme based on inventory messages, similar to how transactions and blocks are propagated on the bitcoin p2p network, can optionally be used: if the `option_inv_gossip` feature bit is set, instead of broadcasting `channel_annoucement` and `channel_update` messages, nodes will broadcast a `gossip_inventory` message which is similar to a `reply_channel_range` message and can be used to filter announcements by timestamp or by content (checksum):
+
+1. type: 266 (`gossip_inventory`)
```
Roasbeef: Rather than a single message for all inventory types I think we should instead have distinct messages as this allows us to open encode certain fields that are relevant if a node wants to filter out a class of advertisements. For example a `channel_inv` message would have things like the size of the channel, or the nodes that it's connecting included. This would let nodes performing filtering such as _ignoring_ all channels that are below a certain size. With the checksum approach the node has no idea what it's about to fetch, and will waste bandwidth by fetching something that it will instantly abandon. 

While we're at it, we may also need to extend the prior queries in order to allow fetching a node announcement.

n1bor:
Just some stats on scale of issues:
https://lightningconductor.net/grafana/d/CZQF3HzWz/lightning-stats?orgId=1&from=now-7d&to=now&fullscreen&panelId=22
Is for node:
03c436af41160a355fc1ed230a64f6a64bcbd2ae50f12171d1318f9782602be601@mainnet.lightningconductor.net:9735

Roasbeef:
What implementation and version is that node running? I see about an order
of magnitude less traffic on my node personally.

n1bor:
A 2 week old version of Eclair.
https://github.com/ACINQ/eclair/commit/9032da5326a46a295c6581ef0ce6e3d84087da60
They did commit an optimisation a week ago. Will move onto that in a few day.
I think my stats are good as I am just scraping the log file with logstat. And I did double check a few of the peeks and they were real.

pm47:
Those are orthogonal things: this PR reduces the bandwidth used by gossip, all things remaining equal (even if the number of `channel_update`s remained the same). 

@n1bor seems to have a pathological case of flapping channels, which caused eclair to generate a constant stream of `channel_update`s. That's an implementation problem, not a spec one, and our recent fix (https://github.com/ACINQ/eclair/commit/fb84dfb855ddc1ee0a42897351ac5506a28198da) alleviates this issue.

n1bor:
Updated to HEAD de5a782 - lets see if it settles down in the next few hours.

07-routing-gossip.md
sstone: I've rebased to get the latest changes from #557, and I'm using TLV encoding now. I've adding a node inventory messages (just a list of node ids). 
I don't think that we need to include channel capacity in the inv message: you would have to trust the sender anyway, and tiny channels will progressively go away.
I don't think we should include the node ids it's connecting: how would you use it ? You don't know what a new channel enables from a routing pov until you have the rest of the graph

rustyrussell:
OK, I'm working on TLV parsing, so some cleanups (and introduce some conventions).  Nothing significant changed in the format, except removal of some redundant lengths inside TLV records.

07-routing-gossip.md
```diff
@@ -798,6 +897,42 @@ and allows bootstrapping for new nodes as well as updating for nodes that
...
+   1. type: 5 (`checksums`)
+   2. data:
+        * [`tlv_len`:`encoded_checksums`]
+   1. type: 7 (`node_inv`)
+   2. data:
+      * [`tlv_len`:`node_ids`]
+ 
+`channel_inv` contains information about channel announcements and updates. It uses the exact same encoding as `reply_channel_range` and includes:
+ - a list of `short_channel_id`s
+ - an optional `timestamps`, which includes `channel_update` timestamps
```
cfromknecht: is it possible to avoid the redundancy of sending timestamps for both edges every time?

07-routing-gossip.md
```diff
@@ -798,6 +897,42 @@ and allows bootstrapping for new nodes as well as updating for nodes that
...
+   1. type: 3 (`timestamps`)
+   2. data:
+	  * [`1`:`encoding_type`]
+	  * [`tlv_len-1`:`encoded_timestamps`]
+   1. type: 5 (`checksums`)
+   2. data:
+        * [`tlv_len`:`encoded_checksums`]
+   1. type: 7 (`node_inv`)
+   2. data:
+      * [`tlv_len`:`node_ids`]
```
cfromknecht: from my understanding, #557 doesn't have a way of querying for `node_announcement` without the corresponding `short_channel_id`. Can you explain how one would retrieve the `short_channel_id` from this message to use in the query? If `node_ids` is a different size from `encoded_short_ids`, which one should the receiver use?

07-routing-gossip.md
```diff
@@ -798,6 +897,42 @@ and allows bootstrapping for new nodes as well as updating for nodes that
...
+ 
+`channel_inv` contains information about channel announcements and updates. It uses the exact same encoding as `reply_channel_range` and includes:
+ - a list of `short_channel_id`s
+ - an optional `timestamps`, which includes `channel_update` timestamps
+ - an optional `checksums`, which includes `channel_update` checksums
+
+ `node_inv` contains a list of node ids, 33 bytes for each node id.
```
cfromknecht: i imagined this would also include a timestamp, otherwise we may request redundant node announcements

07-routing-gossip.md
```diff
@@ -798,6 +897,42 @@ and allows bootstrapping for new nodes as well as updating for nodes that
...
+A more efficient scheme based on inventory messages, similar to how transactions and blocks are propagated on the bitcoin p2p network, can optionally be used: if the `option_inv_gossip` feature bit is set, instead of broadcasting `channel_annoucement` and `channel_update` messages, nodes will broadcast a `gossip_inventory` message which is similar to a `reply_channel_range` message and can be used to filter announcements by timestamp or by content (checksum):
+
+1. type: 266 (`gossip_inventory`)
+2. data:
+    * [`32`:`chain_hash`]
+	* [`tlvs`:`gossip_inventory_tlvs`]
+
+1. tlvs: `gossip_inventory_tlvs`
+2. types:
+   1. type: 1 (`channel_inv`)
```
cfromknecht: is the idea here that this will only include the `short_channel_ids` of new `channel_updates`? and then the receiver can request the `channel_announcement` if it doesn't know of it?

07-routing-gossip.md
sstone: Not atm. You can choose to not send timestamps at all, but if you do you must include timestamps for all updates. I'm not sure how sending some but not all of them could be used ?

07-routing-gossip.md
sstone: That's a good point and now I see why @Roasbeef wanted a different INV message for node announcements. A simple answer would be to include `node_id`, `timestamp` and `short_channel_id` in the INV message, receiver can then use the provided `short_channel_id` to query the node announcement.

07-routing-gossip.md
sstone: Yes, see comment above.

07-routing-gossip.md
sstone: Yes that's correct.

cfromknecht: Gotcha, makes sense to me. No sense announcing a channel_announcement if we don't have any updates for it!

TheBlueMatt:
This needs a rebase. If we're gonna add more stuff to the route-sync state machine, can we improve it? The current design is super complicated, and its just kinda grown organically without a lot of thought to the end state. It has grown based on whats available without a lot of care to whats allowed (eg it relies heavily on sequence id fields being timestamps, which the spec doesn't even require, not to mention stale updates my propagate late and uses zlib as a cheap way to communicate data, instead of a purpose-designed thing like minisketch). At various points, I've suggested we move towards where Bitcoin Core is going and:

 * Start with an initial sync based on last timestamp fields with a two hour buffer (we need to constrain the timestamp fields to be at least after block header timestamps, but those are only +/- 2 hours, so we can't assume more than that).
 * Use minisketch (or just an inv list, if we don't want to take the dependency) to figure out what you need that was missed maybe because it wasn't relayed in a timely fashion or because its somewhere in the two hour buffer,
 * Use invs thereafter.

This would remove a *ton* of existing complexity.

PR #585, opened 2019-03-02
BOLT 2 : Remove dust option in sending no-change commitment tx

ariard:
Assuming amount of to_remote or to_local output is decreased at htlc commitment, even committing a dust one should always modify commitment tx. And bolt2 also prohibits commitment tx without outputs.


ariard:
Closing, forget the case where HTLC may be in msat range, and so not tampering to_remote/to_local output.

PR #589, opened 2019-03-26
 BOLT 7: fix message name(gossip_timestamp_range -> gossip_timestamp_filter)

Yamaguchi:
It is prefered to be written as "gossip_timestamp_filter" instead of "gossip_stamp_range" ?


t-bast:
Looks like this hard been fixed by another change.

PR #591, opened 2019-03-29
Must not filter local gossip

pm47:
We should never filter out `channel_update`s for direct channels to the peer (even if no `gossip_timestamp_filter` has been set), otherwise we wouldn't be able to use the channels for sending or relaying outgoing payments.

This would make it safe to only enable gossip of a subset of our peers.


t-bast:
Looks like this is a subet of #684, should we close in favor of it?

rustyrussell:
OOps, agreed, #684 was a dup of this.  My bad :(

PR #594, opened 2019-03-31
Change opcode reference format

alexbosworth:
`OP_CHECKLOCKTIMEVERIFY` and `OP_CSV` use an inconsistent naming convention in the document, since `OP_CLTV`/`OP_CLV` is not used but `OP_CSV` is used. Update OP_CSV to match the `OP_CHECKLOCKTIMEVERIFY` convention by using `OP_CHECKSEQUENCEVERIFY` instead

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -94,14 +94,14 @@ The amounts for each output MUST be rounded down to whole satoshis. If this amou
 
 #### `to_local` Output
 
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/594/commits/be9a9229cfd9e62b9c964200a848972bfdfb331f

PR #595, opened 2019-04-02
BOLT2: rephrase cltv_expiry_delta text about accepting HTLCs

harding:
Saying "the risk is only to the node *accepting* the HTLC" is confusing
because merely accepting an HTLC is risk-free.  The risk comes from
accepting *responsibility to route the payment*, i.e. offering an HTLC
of your own in the next channel on the path, where a too-small
difference in the HTLC values could end up with you cheated out of a
payment.

This revised paragraph hopefully makes that clearer.

(My thanks to @jnewbery and @cdecker for helping me understand what was going on here.)

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -670,10 +670,10 @@ the forwarding case (B). `min_final_cltv_expiry` is the minimum difference
 between HTLC CLTV timeout and the current block height, for the
 terminal case (C).
 
-Note that if this value is too low for a channel, the risk is only to
-the node *accepting* the HTLC, not the node offering it. For this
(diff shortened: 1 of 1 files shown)


jnewbery:
This seems like a good clarification to me. ACK.

cdecker:
I like it :smiley: 

ACK https://github.com/lightningnetwork/lightning-rfc/pull/595/commits/390882933c04a770e25e8cef0d4ebc49f6569bde

cfromknecht:
Nice rewording! 

ACK 390882933c04a770e25e8cef0d4ebc49f6569bde

PR #590, opened 2019-03-27
BOLT 9: add features wumbo and wumborama

araspitzu:
This proposal formalizes the discussion happened on the mailing list about lifting the channel size limit. Two features are defined here:

-  `option_i_wumbo_you_wumbo` means I'm willing to accept channel larger than 2^24  (local feature)
-  `option_wumborama` means I offer and I'm willing to accept `option_i_wumbo_you_wumbo` from anyone (global feature)

Peers MUST set `option_i_wumbo_you_wumbo` in the local features of the `init` message if they set the `option_wumborama` in the global features, the proposal reuses the bits defined in #571

PR #596, opened 2019-04-02
Single-option large channel proposal

araspitzu:
This PR allows peers to lift the channel size limit with a single option `option_support_large_channel`, it subsumes the previous PR #590. The option means that the offering peer supports _opening_ large channels, a peer willing to send a large htlc should refer to `channel_update.htlc_maximum_msat` to check if a certain channel can route that.


09-features.md
```diff
@@ -29,7 +29,12 @@ These flags may only be used in the `init` message:
 
 ## Assigned `globalfeatures` flags
 
-There are currently no `globalfeatures` flags.
+These flags may only be used in the `init` and `node_announcement` message:
+
+| Bits | Name             |Description                                     | Link                                                                |
+|------|------------------|------------------------------------------------|---------------------------------------------------------------------|
+| 16/17 | `option_support_large_channel` | Can create large channels		   | [Rationale](#Rationale) |
```
Roasbeef: Why start at 16/17 instead of zero? The existing local feature bits followed the pattern of always starting new features from the lowest unassigned value, which in this case is zero.

araspitzu: Because it took in account the feature bits unification. I agree that if we aren't going to merge #571 we should start the `globalfeatures` from zero but **if** we later merge it then we need to update this feature bits (localfeature 0/1 is already defined), in that scenario is better to start from 16/17 to avoid collisions.

rustyrussell:
Technically, this is a local/node feature since it only effects direct peers.

And #571 supports advertizing those features via node_announcement, see https://github.com/lightningnetwork/lightning-rfc/pull/571/commits/14ced3b6c33961f1adc3699994c21a4abc4958f8

araspitzu:
@rustyrussell right! I took the wrong bits from the temporary assignment in #571, this was meant to use the nodefeature bits as in what previously was called `option_i_wumbo_you_wumbo` with bits 14/15, I will update the PR.

t-bast:
After re-reading through #571, I agree with @rustyrussell: this is a `node` feature.

In #605 there are both `node` and `channel` feature flags for this. I think the `option_wumborama` `channel` flags aren't useful, should we remove them? Or can you explain why you think they are useful? In my opinion just looking at the channel capacity and `htlc_maximum_msat` give you all the information you need for routing.

araspitzu:
@t-bast indeed this PR does not propose to use `option_wumborama` on the channel flags but uses a single-option and it should go in the node features (once we unify feature bits). As per description the option means that the offering peer _supports_ opening large channels, a peer willing to send a large htlc should refer to `channel_update.htlc_maximum_msat` to check if a certain channel can route that.

07-routing-gossip.md
```diff
@@ -510,7 +510,7 @@ The receiving node:
   - if the `option_channel_htlc_max` bit of `message_flags` is 0:
     - MUST consider `htlc_maximum_msat` not to be present.
   - otherwise:
-    - if `htlc_maximum_msat` is not present or greater than channel capacity:
+    - if `htlc_maximum_msat` is not present, greater than channel capacity or greater than 2^32:
```
rustyrussell: I don't think this change is right?

09-features.md
```diff
@@ -29,7 +29,12 @@ These flags may only be used in the `init` message:
 
 ## Assigned `globalfeatures` flags
 
-There are currently no `globalfeatures` flags.
+These flags may only be used in the `init` and `node_announcement` message:
+
+| Bits | Name             |Description                                     | Link                                                                |
+|------|------------------|------------------------------------------------|---------------------------------------------------------------------|
+| 14/15 | `option_support_large_channel` | Can create large channels		   | [Rationale](#Rationale) |
```
rustyrussell: I think that link is broken, since Rationale is in a separate BOLT?

07-routing-gossip.md
araspitzu: It does make sense to reject channel_updates with `htlc_maximum_msat` greater than 2^32 as this (currently) is the largest htlc possible, in fact the section above about the sending node says:
``` 
     if the htlc_maximum_msat field is present:
      ...
     MUST set htlc_maximum_msat to the maximum value it will send through this channel for a single HTLC. 
     for channels with chain_hash identifying the Bitcoin blockchain:
           MUST set this to less than 2^32.
```
Maybe is far fetched to include this change here but it's sensible for the receiving node to check the `htlc_maximum_msat` is consistent with the limits.

09-features.md
araspitzu: Isn't it below at line 45? [Rationale](https://github.com/araspitzu/lightning-rfc/blob/large_channel_support/09-features.md#rationale)

araspitzu:
Summary of the latest changes: 
- Rebased on #571 
- Updated feature bits to 8/9 
- Moved rationale to BOLT-2 replacing old section about the current limit

07-routing-gossip.md
araspitzu: I reverted the controversial change in 0c78b3d

araspitzu:
Summary of the changes up to 8af0535:
- Uses flat features from #666 
- Uses feature bits 14/15 (they're mostly placeholders until actual bits are assigned)

09-features.md
```diff
@@ -1,43 +1,55 @@
...
+| 0/1   | `option_data_loss_protect`       | Requires or supports extra `channel_reestablish` fields   | IN       | [BOLT #2][bolt02-retransmit]          |
+| 3     | `initial_routing_sync`           | Sending node needs a complete routing information dump    | I        | [BOLT #7][bolt07-sync]                |
+| 4/5   | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel   | IN       | [BOLT #2][bolt02-open]                |
+| 6/7   | `gossip_queries`                 | More sophisticated gossip control                         | IN       | [BOLT #7][bolt07-query]               |
+| 8/9   | `var_onion_optin`                | Requires/supports variable-length routing onion payloads  | IN       | [Routing Onion Specification][bolt04] |
+| 10/11 | `gossip_queries_ex`              | Gossip queries can include additional information         | IN       | [BOLT #7][bolt07-query]               |
+| 14/15 | `option_support_large_channel`   | Can create large channels                                 | INC      | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
```
t-bast: Should it be INC+? I think that in `channel_announcement`, we would either not set any of those bits (if the channel isn't a large channel) or set it as `even` if the channel is a large channel.

But maybe that's short-sighted, if we allow changing the capacity of a channel in the future setting the odd bit would make sense for a channel that's not large yet, but can become large in the future?

araspitzu: Right, restricting it to `C+` (required bits) makes more sense because it's ambiguous to have optionality here, either it's a large channel or it's not. I'm not sure about changing the capacity of the channel but i suppose this would come with a new `channel_announcement` and if that's the case the new instance can contain the updated (even) bits if it is a large channel.

araspitzu:
Commit d5a8978 resolves conflicts with master and updates the feature bits to 48/49, those requested in the waiting room #605

09-features.md
```diff
@@ -37,6 +37,7 @@ The Context column decodes as follows:
 | 12/13 | `option_static_remotekey`        | Static key for remote output                              | IN       |                   | [BOLT #3](03-transactions.md)         |
 | 14/15 | `payment_secret`                 | Node supports `payment_secret` field                      | IN9      | `var_onion_optin` | [Routing Onion Specification][bolt04] |
 | 16/17 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | `payment_secret`  | [BOLT #4][bolt04-mpp]                 |
+| 48/49 | `option_support_large_channel`   | Can create large channels                                 | INC+     | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
```
t-bast: nit: this is now missing a column: 

```suggestion
| 48/49 | `option_support_large_channel`   | Can create large channels                                 | INC+     | | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
```

araspitzu:
As discussed on the last spec meeting I've updated the feature bits to 18/19 in commit 660dd2e

t-bast:
Merging as agreed during the meeting: http://www.erisian.com.au/meetbot/lightning-dev/2020/lightning-dev.2020-02-17-19.06.html

A follow-up PR with an advisory for confirmation scaling depending on funding amount will be submitted shortly.

PR #601, opened 2019-04-22
BOLT 3: Explicit description of implicitly enforced timelocks on HTLC outputs

nkohen:
Added descriptions of how a 2-of-2 multisignature verification is used for enforcing timelocks when timing out on-chain offered HTLCs as well as spending on-chain received HTLCs in the success case.

I say implicitly enforced to mean that these timelocks do not show up explicitly in any of the HTLC output scripts.

It took me a lot of time and effort to reason these facts out of the existing documentation so I figured I'd propose to add it.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -981,6 +981,8 @@ offline until after sending `commitment_signed`.  Once
 those HTLCs, and cannot fail the related incoming HTLCs until the
 output HTLCs are fully resolved.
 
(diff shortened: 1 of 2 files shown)


cdecker:
Thanks for the contribution @nkohen. While reading the PR we noticed that you mention committing to the time out but never mention how that timeout is specified. It would be nice to have a reference to the `locktime` specifying the time out in your description.

nkohen:
Thanks for the comment @cdecker! I changed my comment to reference the `locktime` field of the HTLC-timeout transaction :)

PR #604, opened 2019-05-06
Onion format variation a-la @roasbeef (vs #593)

rustyrussell:
This splits the difference between #593 from @cdecker and comments from @Roasbeef 

I based it on #593 so the differences are clear.  It leaves realm 0 alone (naming it legacy), just defines realms 1-15. 

These are TLVs, which in is a win for the final node since `short_channel_id` is unnecessary there.

(See #603 which explicitly describes that encoding smaller numbers with fewer than 8 bytes in a TLV is allowed).


t-bast:
I'm fine with this small tweak. If I'm not missing anything it achieves the same tradeoffs as @cdecker's proposal, in fact the only change is that the least significant bits of the `realm` are used to encode the number of frames instead of the most significant bits (and we commit to using TLV for all those 16 `realm` values which is what we decided anyway).

We have room for new messages in two different places:

- in `realms` `0x01` to `0x0f` by leveraging the TLV types that aren't yet used
- in all the other `realm` values with complete liberty over the format of the hop payload

I think that this is clearly enough for the foreseeable future (and it achieves both backwards-compatibility and simplicity, which is very important to get it deployed smoothly).

In the longer term, we know we want to change a few things in a version 1 of the onion message (per-hop versioning for example). Version 1 can be an opportunity to provide even more room for new messages if developers can't achieve what they want with version 0.

04-onion-routing.md
```diff
@@ -181,17 +181,14 @@ Each `hop_payload` has the following structure:
...
    * [`padding_len`: `padding`]
    * [`32`: `HMAC`]
 
 Notice that since the `hop_payload` is instantiated once per hop, the subscript `_i` may be used in the remainder of this document to refer to the `hop_payload` and its fields destined for hop `i`.
 
-The `hop_payload` consists of at least one `frame` followed by up to 15 additional `frame`s.
-The number of additional frames allocated to the current hop is determined by the 4 most significant bits of `num_frames_and_realm`, while the 4 least significant bits determine the payload format.
-Therefore the number of frames allocated to the current hop is given by `num_frames = (num_frames_and_realm >> 4) + 1`.
-For simplification we will use `hop_payload_len` to refer to `num_frames * FRAME_SIZE`.
+The `hop_payload` consists of at least one `frame` followed by up to 15 additional `frame`s.  For simplification we will use `hop_payload_len` to refer to `num_frames * FRAME_SIZE`.
```
Roasbeef: If the entire byte is available, why restrict it to 15 instead of ~20 which lets nodes utilize all the available space?

Roasbeef:
Thinking about this more, I really like the signaling unification and also
clean slate for the payload for each hop. In comparison, the other proposal
made the payloads into a hybrid of the old and new, while with this,
payloads are just opaque blobs to be passed up (which lets us do things
like shave the short id for the final hop) . Also to my reading, the prior
proposal left the current 13 padding bytes unutilized, while with this
there's no longer any distinction.

cdecker:
It has one small downside of having to keep a mapping of `realm -> num_frames` in the code processing the onion, whereas #593 was making that simply the first few bits, but like mentioned it's a minor thing.

rustyrussell:
Closed in favor of revisions to #593

04-onion-routing.md
```diff
@@ -237,13 +270,48 @@ Field descriptions:
...
 may lead to extraneous routing failure.
 
+## `tlv_hop_data` Payload Format
+
+This is a more flexible format, which avoids the redundant
+`short_channel_id` field for the final node.  The range of `realm`
+values also allows use of more than one frame.
+
+1. tlv: `tlv_hop_data`
+2. types:
```
Roasbeef: Rather than unroll these fields and take up 4 types, we can instead have them all packed under a _single_ type. All the fields are fixed length, so there's no additional signalling overhead and we also save a few bytes as well. When allocating space in the TLV namespace we should prefer to pack several types into one, if all the types are related to a discrete use case. This model promotes more efficient utilization of the payload space.

cdecker: Well like we discussed in the IRC meeting some of the types would be omitted, for example in the last `hop_payload`. So while I agree that we may avoid defining types that must always be used in combination, we should also keep in mind that swapping out or omitting individual types is also an important feature here.

cdecker:
Closed in favor of PR #619 as discussed during the IRC meeting on 2019/06/10.

PR #606, opened 2019-05-06
BOLT2: past fulfillment deadline fails channel

jtimon:
This seems a typo. It seems it should say channel rather than connection.
If not, why fail the connection (presumably to reconnect later? or to ban the node?) instead of failing the channel?
Or are they just equivalent?


cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/606/commits/95fd655d98fa8dc0b6bd737e72f27a0f431a867c

cdecker:
Merging as discussed during the [last specification meeting](http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-05-13-20.03.html)

PR #609, opened 2019-05-13
BOLT5: Unify two practically redundant paragraphs

jtimon:
The two paragraph seem quite redundant and repetitive. Perhaps the resulting one can even be simplified even further.

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -435,14 +435,10 @@ own timeout still applies as an upper bound.
 ## HTLC Output Handling: Remote Commitment, Remote Offers
 
 Each HTLC output can only be spent by the recipient if it uses the payment
-preimage. If a node does not possess the preimage (and doesn't discover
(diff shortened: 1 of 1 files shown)


cdecker:
ACK 4f35fd391b275af45b01f3c2567e003f1931bdce

jtimon:
As discussed on IRC, the second paragraph seems just a more specific version of the first one. So I could simply remove the first paragraph and leave the second one as is. Or I could just leave it at what I have now, or follow some other suggestion too.

jtimon:
Please, chose between commit 1 alone or both squashed. I think we all agree both options are improvements, I don't care either way.

cdecker:
Merging like discussed during the [last specification meeting](http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-05-13-20.03.html)

jtimon:
I could have squashed, but I guess it doesn't really matter.

PR #611, opened 2019-05-23
Trampoline routing proposal

t-bast:
This PR highlights proposed changes to enable trampoline routing (once #593 and #603 are accepted).

Everything is currently grouped in a single file to make the initial review process easier. Once we start converging on the proposal, the final changes should probably be split between bolts 2, 4 and 7 (unless you feel this should be a new bolt) and the wording should be in formal spec format (with more details on some sections).

### DIFF
--- a/wip-trampoline.md
+++ b/wip-trampoline.md
@@ -1,0 +1,840 @@
+# Trampoline Onion Routing
+
+This document describes the additions we make to the existing onion routing to
+enable the use of trampoline hops.
+
+As the network grows, more bandwidth and storage will be required to keep an
+up-to-date view of the whole network. Finding a payment path will also require
+more computing power, making it unsustainable for constrained devices.
+
+Constrained devices should only keep a view of a small part of the network and
+leverage trampoline nodes to route payments. This proposal still uses an onion
+created by the payer so it doesn't sacrifice anonymity (in most cases it will
+even provide a bigger anonymity set).
+
+Nodes that keep track of the whole network should advertise support for
+`option_trampoline_routing`. This is an opportunity for them to earn more fees
+than with the default onion routing.
+
+A payer selects a few trampoline nodes and builds a corresponding trampoline
+onion TLV. It then embeds that trampoline onion in the last `per_hop` payload
+of an `onion_packet` destined to the first trampoline node. Computing routes
+between trampoline nodes is deferred to the trampoline nodes themselves.
+
+Merchants may include a few trampoline nodes that are close to them in their
+invoice to help payers select a good last trampoline node. This is optional and
+payers aren't required to use this routing hint.
+
+An interesting side-effect is that trampoline routing combines nicely with
+rendezvous routing. With a simple new TLV type and more data in the invoice,
+rendezvous routing can easily work on top of trampoline routing, with a
+construction similar to what is described [here][rendezvous].
+
+## Table of Contents
+
+* [Packet Structure](#packet-structure)
+  * [Trampoline Onion](#trampoline-onion)
+  * [Usage](#usage)
+  * [Paying to non-trampoline nodes](#paying-to-non-trampoline-nodes)
+* [Fees and CLTV requirements](#fees-and-cltv-requirements)
+* [Network Pruning](#network-pruning)
+  * [The `node_update` Message](#the-node_update-message)
+  * [Fees and CLTV estimation](#fees-and-cltv-estimation)
+  * [Filtering gossip messages](#filtering-gossip-messages)
+* [Examples](#examples)
+  * [Merchant supporting trampoline payments](#merchant-supporting-trampoline-payments)
+  * [Merchant without trampoline support](#merchant-without-trampoline-support)
+* [References](#references)
+
+## Packet Structure
+
+### Trampoline Onion
+
+The trampoline onion is a fixed-size TLV packet with the following structure:
+
+1. type: `trampoline_onion_packet`
+2. data:
+   * [`1`:`0x08`] (`type`)
+   * [`3`:`0xfde202`] (`length`)
+   * [`1`:`version`]
+   * [`33`:`public_key`]
+   * [`672`:`trampoline_hops_data`]
+   * [`32`:`hmac`]
+
+This results in a `742`-bytes fixed-size packet where the value has exactly the
+same format as the `onion_packet` with a smaller `hops_data`.
+
+The `trampoline_hops_data` field is a list of `trampoline_hop_payload`s. A
+`trampoline_hop_payload` is a structure that holds obfuscations of the next
+trampoline hop's address, transfer information, and its associated HMAC. It
+uses the same format as the `onion_packet`'s variable-size `per_hop` payload,
+meaning that it can contain an arbitrary TLV stream:
+
+1. type: `trampoline_hops_data`
+2. data:
+   * [`n1`:`trampoline_hop_payload`]
+   * [`n2`:`trampoline_hop_payload`]
+   * ...
+   * `filler`
+
+Each `trampoline_hop_payload` has the following structure:
+
+1. type: `trampoline_hop_payload`
+2. data:
+   * [`var_int`:`length`]
+   * [`length`:`tlv_stream`]
+   * [`32`:`hmac`]
+
+The `tlv_stream` contains data necessary to route to the next trampoline hop.
+The following types MAY be used to encode that data:
+
+* type: `node_id`
+* data:
+  * [`1`:`0x02`] (`type`)
+  * [`1`:`0x21`] (`length`)
+  * [`33`:`node_id`]
+* type: `amount_msat`
+* data:
+  * [`1`:`0x04`] (`type`)
+  * [`1`:`length`]
+  * [`length`:`amount_msat_value`]
+* type: `cltv`
+* data:
+  * [`1`:`0x06`] (`type`)
+  * [`1`:`length`]
+  * [`length`:`cltv_value`]
+
+### Usage
+
+A node that wants to rely on trampoline nodes to relay payments may use a
+`trampoline_onion_packet` in the TLV stream of the _last_ hop of an
+`onion_packet`:
+
+1. type: `onion_packet`
+2. data:
+   * [`1`:`version`]
+   * [`33`:`public_key`]
+   * [`n1`:`hop_payload`] (normal hop payload)
+   * [`n2`:`hop_payload`] (normal hop payload)
+   * ...
+   * [`n`:`hop_payload`] (hop payload containing a `trampoline_onion_packet`)
+   * `filler`
+   * [`32`:`hmac`]
+
+The structure of the _last_ `hop_payload` is:
+
+1. type: `hop_payload`
+2. data:
+   * [`var_int`:`length`]
+   * [`742`:`trampoline_onion_packet`]
+   * `optional TLVs can be added here`
+   * [`32`:`hmac`]
+
+### Paying to non-trampoline nodes
+
+If the recipient of the payment doesn't support trampoline routing, we rely on
+the last trampoline node to convert the last hop to a standard onion payment.
+
+The last `trampoline_hop_payload` must contain a `recipient_info` TLV packet
+with the following structure:
+
+1. type: `recipient_info`
+2. data:
+   * [`1`:`0x0a`] (`type`)
+   * [`1`:`0x01`] (`length`)
+   * [`1`:`0x00`] (`option_trampoline_routing`)
+
+Note that it reveals the identity of the recipient and the amount paid to the
+last trampoline node (but it doesn't reveal the identity of the payer).
+This can be avoided if the recipient uses rendezvous routing to hide its
+identity. Otherwise recipients should support `option_trampoline_routing` to
+preserve their anonymity (and can advertise a high `fee` if they don't want to
+route too many payments).
+
+### Requirements
+
+A sending node:
+
+* MUST verify that each hop in the `trampoline_onion_packet` supports `option_trampoline_routing`
+* MUST include a `recipient_info` with `option_trampoline_routing` set to `0x00` TLV if the recipient doesn't support `option_trampoline_routing`
+* MUST encrypt the `trampoline_onion_packet` with the same construction as `onion_packet`
+* MUST use a different `session_key` for the `trampoline_onion_packet` and the `onion_packet`
+* MAY use other TLV types than `node_id`, `amount_msat` and `cltv` in the `trampoline_hop_payload`
+* MAY include additional TLV types in `trampoline_hop_payload`s
+* MUST put the `trampoline_onion_packet` in the _last_ hop's payload of the `onion_packet`
+
+When processing a `trampoline_onion_packet`, a receiving node:
+
+* If it doesn't support `option_trampoline_routing`:
+  * MUST report a route failure to the origin node
+* MUST process the `trampoline_onion_packet` as an `onion_packet`
+* If it encounters an unknown _even_ TLV type inside the `trampoline_onion_packet`:
+  * MUST report a route failure to the origin node
+* If it doesn't have enough data to locate the next hop:
+  * MUST report a route failure to the origin node
+* MUST compute a route to the next trampoline hop:
+  * If it cannot find a route that satisfies `cltv` requirements:
+    * MUST report a route failure to the origin node
+  * If it cannot find a route that satisfies `fees` requirements:
+    * MUST report a route failure to the origin node
+  * If it has a channel open with the next trampoline hop:
+    * MAY use that channel to build a single-hop route
+  * If a `recipient_info` TLV is received with `option_trampoline_routing` set to `0x00`:
+    * MUST convert the peeled `trampoline_onion_packet` to an `onion_packet`
+  * Otherwise:
+    * MUST include the peeled `trampoline_onion_packet` in the last `hop_payload`
+* SHOULD return errors as specified in the [error handling section](#returning-errors)
+
+### Rationale
+
+This construction allows nodes with an incomplete view of the network to
+delegate the construction of parts of the route to trampoline nodes.
+
+The origin node only needs to select a set of trampoline nodes and to know a
+route to the first trampoline node. Each trampoline node is responsible for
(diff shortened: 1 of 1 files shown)


cdecker:
FWIW I think it is a great idea to have an overall design doc like this that collects all the details of the incremental change in one place. We agreed on having a "living" spec, but I still like having a well-written and easy to follow bundle that explains the rationale more in depth :+1:

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Packet Structure
+
+### Trampoline Onion
+
+We define a fixed-size trampoline onion packet that can be nested inside onion
+packets. We define `N` to be the maximum number of trampoline nodes (excluding
+the sender node and the destination node).
+
+The trampoline packet always contains `2*N*FRAME_SIZE + 66` bytes.
```
cdecker: I'm not clear where the `2*FRAME_SIZE` per hop comes from.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+
+Note: we use the same onion format as the standard onion which wastes frames,
+but that theoretically allows another level of nesting (mind blown).
+However not really usable given the space constraints...
+
+TODO: maybe this should be version 1 of the overall onion version namespace?
+
+TODO: maybe we can do something more clever with TLV? At the cost of a second
+onion parser (trampoline onion parser)? Use the top-level version in the parser
+to discriminate?
```
cdecker: If we stick with 65 byte frames we can reuse the onion processing from the outer onion.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+   * [`32`:`HMAC`]
+
+For this specification, some restrictions apply:
+
+* `realm` MUST be `0x01` to indicate that the payload contains a TLV stream
+and `num_frames` MUST correctly reflect the number of frames used.
+* the TLV stream MUST contain the `type` `0x01` (`next_trampoline_hop`)
+* `next_trampoline_hop`'s length MUST be `33+8+4=45` bytes
+* the TLV stream MAY contain other optional types
+* `padding` MUST use only `0x00` bytes
```
cdecker: I'd prefer sticking with a raw TLV on the top-level that is easier to change later, rather than rebuilding the legacy `hop_data_v0` format in this new layer.

wip-trampoline.md
cdecker: I see, it comes from the requirement of 45 bytes of payload, however if we just describe this as a TLV payload the rest resolves itself. That is if we just say we have `N` frames available, and allow any trampoline hop to use as many as they want we're not imposing this limitation right away.

Who knows, we might come up with a more concise way of addressing a node, and the amount to forward and we'd suddenly fit into a single frame.

Long story short: I think it's a mistake to hard-code the `2*FRAME_SIZE` here.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+1. type: `onion_packet`
+2. data:
+   * [`1`:`version`]
+   * [`33`:`public_key`]
+   * [`n1*FRAME_SIZE`:`hop_payload`]
+   * [`n2*FRAME_SIZE`:`hop_payload`]
+   * ...
+   * [`2*(N+1)*FRAME_SIZE`:`trampoline_onion_packet_payload`]
+   * `filler`
+   * [`32`:`hmac`]
```
cdecker: So in this example `hop_payload`s of `n1`, `n2` and so forth are the path to get to the trampoline node, and then the `trampoline_onion_packet_payload` is used to find a continuation at the trampoline, right?

I don't understand the rationale of `2*(N+1)*FRAME_SIZE` which is bigger than the payload we have at our disposal.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+The structure of the last payload is:
+
+1. type: `trampoline_onion_packet_payload`
+2. data:
+   * [`1`:`realm`]
+   * [`1`:`type`]
+   * [`3`:`length`]
+   * [`2*N*FRAME_SIZE + 66`:`trampoline_onion_packet`]
+   * `optional TLV can be added here`
+   * `padding`
```
cdecker: Why not just define a TLV type and make it a completely normal TLV type-length-value pair? That way we could even give additional instructions to the trampoline, instead of having an optional TLV at the end.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+improved.
+
+Upon receiving a packet with type `trampoline_onion_packet_payload`:
+
+* The receiving node is the destination for this v0 onion packet
+* It extracts the `trampoline_onion_packet` and is able to unwrap one layer to
+discover the next trampoline node
+* It creates a new v0 onion packet to the next trampoline node
+* It puts the unwrapped `trampoline_onion_packet` in the payload for the last
+hop
```
cdecker: Notice that we also need a downgrade option: while we may want to rely on trampolines to forward a payment through the network and each of the trampolines just receives an inner onion it knows how to process, the recipient may not support trampolines at all. Hence the last trampoline may be asked to downgrade to a regular `hop_payload` so that the onion arriving at the recipient is just a classical onion.

cdecker: While this seems to be addressed in the example below I can't find a description of how this should really work.

wip-trampoline.md
```diff
@@ -0,0 +1,138 @@
...
+P -> R1 -> R2 -> R3 -> M
+
+R1, R2 and R3 need to have the `option_trampoline` feature flag set.
+Setting this flag implies that you fully comply with the trampoline
+specification.
+
+M doesn't need to understand trampoline nor multi-frame: the last trampoline
+hop will use a v0 onion to route the payment to M.
+
+M will usually include a few "close" trampoline nodes in its invoices to help
```
cdecker: Well, if `M` doesn't know about trampoline, how is he supposed to find close-by trampoline nodes?

wip-trampoline.md
```diff
@@ -25,6 +25,15 @@ but that theoretically allows another level of nesting (mind blown).
...
 TODO: maybe this should be version 1 of the overall onion version namespace?
+It can then also be used as a top-level onion (smaller than the v0 onion).
+It's like the v0 onion but lets each hop choose how they route to the next hop.
+If they have a direct channel they directly send the unwrapped v1 onion to the
+next hop.
+If they don't they wrap the v1 onion inside a v0 to get to the next hop via
+intermediate hops.
+Even though the onion packet is smaller, you don't know if previous hops used
+the direct route or went through v0 wrappings, so you have even greater doubts
+about your position in the route than with only v0.
```
cdecker: I disagree with the fact that v1 should be a trampoline onion. The onion construction we have is versatile enough to be reused as is for the trampoline onion.

The chances of that happening is virtually 0 and we shouldn't be optimizing for this case, and you are basically giving away to an observer that this was a trampoline onion and not a routing onion, by just observing a smaller packet go over the encrypted connection.

wip-trampoline.md
```diff
@@ -1,60 +1,40 @@
...
-
-TODO: maybe we can do something more clever with TLV? At the cost of a second
-onion parser (trampoline onion parser)? Use the top-level version in the parser
-to discriminate?
-
-TODO: version maybe redundant with the TLV type that will include this?
-But keeping it allows using this packet to be used as a top-level onion packet
-where we basically let each hop choose which channel to use to route to the
-next hop. Is this useful?
+This results in a `706`-bytes fixed-size packet.
```
cdecker: Just as a side-note: this leaves us with 594 bytes for the remaining hops, or a maximum routing distance of 9 between two consecutive trampolines.

cdecker: It also means that the maximum route length that is actually enforced by the protocol is `9 * 8 = 72`, which gives us a 72x multiplier for griefing attacks...

wip-trampoline.md
```diff
@@ -63,36 +43,67 @@ Each `trampoline_hop_payload` has the following structure:
 1. type: `trampoline_hop_payload`
 2. data:
    * [`1`:`num_frames_and_realm`]
-   * [`1`:`type`]
-   * [`1`:`length`]
+   * [`47`:`next_trampoline_node`]
```
cdecker: Like mentioned before `next_trampoline_node` should itself be expressed as TLV-pairs in a single TLV, in case we want to switch out the way we address the nodes, allowing us to reduce the footprint.

wip-trampoline.md
```diff
@@ -63,36 +43,67 @@ Each `trampoline_hop_payload` has the following structure:
...
-   * `padding`
-   * [`32`:`HMAC`]
 
-For this specification, some restrictions apply:
+#### Requirements
+
+A sending node:
+
+* MUST set the `realm` part of `num_frames_and_realm` to `0x01` to indicate
+that the payload contains a TLV stream
```
cdecker: This itself implies that the entire payload is just one TLV.

wip-trampoline.md
```diff
@@ -187,12 +187,97 @@ themselves.
...
+2. data:
+   * [`64`:`signature`]
+   * [`32`:`chain_hash`]
+   * [`33`:`node_id`]
+   * [`4`:`timestamp`]
+   * [`2`:`cltv_expiry_delta`]
+   * [`8`:`htlc_minimum_msat`]
+   * [`8`:`htlc_maximum_msat`]
+   * [`4`:`fee_base_msat`]
+   * [`4`:`fee_proportional_millionths`]
```
cdecker: Unless this message is also gated by the existence of a channel with the node as an endpoint it becomes a really cheap DoS attack vector. It's also hard for a node with a partial network view to even verify that there is a channel matching this node.

wip-trampoline.md
```diff
@@ -187,12 +187,97 @@ themselves.
...
+
+TODO: use a TLV stream for `node_update`.
+TODO: should use an odd type (261?) because it's safe to ignore if unsupported?
```
cdecker: Definitely :+1:

wip-trampoline.md
```diff
@@ -187,6 +187,9 @@ themselves.
 
 This grows the maximum route length to `9 * 8 = 72` total hops.
 
+The same kind of mechanism can be used to easily unlock rendezvous routing by
+simply defining a new TLV type (that will extend the `trampoline_onion` type).
```
cdecker: Uhm, actually the rendez-vous routing as proposed [here](https://github.com/lightningnetwork/lightning-rfc/wiki/Rendez-vous-mechanism-on-top-of-Sphinx) works fine without the need to extend the `trampoline_onion`. It just needs a `switch_ephemeral_key` TLV entry and we're good to go :wink:

wip-trampoline.md
```diff
@@ -229,33 +229,48 @@ TODO: use a TLV stream for `node_update`.
...
+If a node is not willing to relay trampoline payments, it will simply never
+send a `node_update`. This lets the network discriminate nodes that are able to
+relay trampoline payments from nodes that use it but simply as sender or
+destination.
 
 ### Fees and CLTV estimation
 
-TODO: detail heuristics that can be used by trampoline nodes to estimate cltv
-and fees.
+Trampoline nodes need to estimate a `cltv_expiry_delta` and `fee` that would
```
cdecker: Trampoline nodes should be encouraged to not be constrained and keep a full view of the topology.

wip-trampoline.md
```diff
@@ -229,33 +229,48 @@ TODO: use a TLV stream for `node_update`.
...
+allow them to route to any other trampoline node while being competitive with
+other nodes. This is a great opportunity to incentivize nodes to open channels
+between each other to minimize the cost of trampoline hops. This is also a
+great opportunity for nodes to implement smart fee estimation algorithms as
+a competitive advantage.
```
cdecker: Is it? I would expect the trampoline to just do trial-and-error with the maximum fee allocated to themselves just like the original sender would. That's not really estimating fees at all. They even get told how much funds MUST arrive at the next trampoline.

wip-trampoline.md
t-bast: This was an old commit and was removed afterwards.
Good news is that with yesterday's proposal during the IRC meeting, we're getting rid of `FRAME_SIZE` entirely ;)

wip-trampoline.md
t-bast: Same, this was an old commit. I plan on removing `FRAME_SIZE` entirely, so it's going to be a lot nicer. I'll also squash all the commits together to hide the old commits.

wip-trampoline.md
t-bast: I agree, I'll clarify this while removing the fixed frames entirely.

wip-trampoline.md
t-bast: > So in this example hop_payloads of n1, n2 and so forth are the path to get to the trampoline node, and then the trampoline_onion_packet_payload is used to find a continuation at the trampoline, right?

Exactly.

> I don't understand the rationale of 2*(N+1)*FRAME_SIZE which is bigger than the payload we have at our disposal.

I think the confusion comes from the distinction between `FRAME_SIZE` and `FRAME_SIZE_V1`.
Removing fixed sizes entirely will make the proposal easier to understand.

wip-trampoline.md
t-bast: I'll investigate that, it's a bit of a trade-off between flexibility and space.
Thanks for the suggestion!

wip-trampoline.md
t-bast: Very good point, for now I assumed that the destination would need to handle trampoline (and ideally provide an invoice with hints about close trampoline nodes).
However you're right that we should make this work for legacy payee nodes. I'll figure something out.

wip-trampoline.md
t-bast: Magic ¯\_(ツ)_/¯
This was an old commit, I fixed this by explicitly stating that M needed to understand trampoline, but in fact I must make this work for payees that don't support it. I'll work on that.

wip-trampoline.md
t-bast: This is an important point, this is a fundamental design choice. Let's discuss it in the global conversation for visibility (your comment was made on an old commit).

wip-trampoline.md
t-bast: I'll explore that option.

wip-trampoline.md
t-bast: Nodes will apply the same filters for this message as what they do for `channel_update`s.
A node producing the `node_update` can only send it to peers with whom it has a channel and can't send `node_update`s more than once every X seconds.
Most nodes that support the `node_update` message will be routing/trampoline nodes and will have the full network view, so they can drop messages from nodes that don't have any channel.

You have a very good point though for the constrained nodes that have only a view of their neighbourhood (let's call one of those Alice). If one of Alice's neighbours is malicious and forwards `node_update`s for nodes that don't have any channel, Alice can't verify that and might include that node in her trampoline routes which is an issue. However, Alice will regularly change her list of trampoline nodes based on randomness and reliability: payment failure will trigger the eviction of the malicious node.

That needs to be detailed more thoroughly though.

wip-trampoline.md
t-bast: That `switch_ephemeral_key` is a cute trick, good job ;)
With trampoline we're able to do rendezvous without switching the ephemeral key. But your proposal with `switch_ephemeral_key` works great too and might require less space.

wip-trampoline.md
t-bast: Yes definitely. I think this incentivizes them to do it because it's the best way for them to be more economically viable.

wip-trampoline.md
t-bast: > They even get told how much funds MUST arrive at the next trampoline.

But that will often be based on the fee they advertised (via their estimation algorithm) ;)

> I would expect the trampoline to just do trial-and-error

They could, but there's a risk they will end up locking up funds if one of the nodes in the route is unresponsive and they have to wait until the timeout. It's hard to predict because I think the heuristics will depend on the reliability of the overall network, but I think both options could make sense.

t-bast:
Thanks for all the feedback @cdecker, really appreciate it (especially since this is a somewhat dense early draft).

> We agreed on having a "living" spec, but I still like having a well-written and easy to follow bundle that explains the rationale more in depth

I think we could use the wiki for that (like what you did for rendezvous). Once the trampoline proposal is integrated to the spec (in spec format, probably split between multiple bolt pages), a wiki page to explain the rationale and give an overall view of how all pieces come together is going to be useful. What do you think?

To sum up your comments for other readers, here are the important things I need to fix (let me know if there is something I missed or misunderstood):

* The fixed-size framing is messy and makes the proposal hard to follow: thanks to yesterday's variable-length frames proposal, I'll remove fixed frames entirely and it should be a lot simpler
* I should use finer-grained TLVs
* The payee doesn't need to understand trampoline (last hop needs to switch to v0 onion if needed)
* I need to clarify the restrictions on `node_update`s to make sure this doesn't introduce an attack vector

> you are basically giving away to an observer that this was a trampoline onion and not a routing onion, by just observing a smaller packet

If I understand this comment correctly, you'd like me to remove the proposal to define onion v1 as a 706-bytes packet. Instead I could only define the trampoline onion as a 706-bytes TLV that lives inside a v0 onion (and we would keep only 1366-bytes sized onion packets). Would that work for you?

Side-note: it looks like old commits made it harder to review - do you think I should squash everything to start from a clean slate and not confuse reviewers?

wip-trampoline.md
```diff
@@ -0,0 +1,840 @@
...
+Each `trampoline_hop_payload` has the following structure:
+
+1. type: `trampoline_hop_payload`
+2. data:
+   * [`var_int`:`length`]
+   * [`length`:`tlv_stream`]
+   * [`32`:`hmac`]
+
+The `tlv_stream` contains data necessary to route to the next trampoline hop.
+The following types MAY be used to encode that data:
```
t-bast: It feels like many TLV types will be useful accross many message types.
I'd like to have a centralized place where we could register known TLVs and their assigned `type` (like what we have in bolt #9). I think that all the TLVs I'm using here (except the `trampoline_onion`) will be useful to other messages and should be specified outside of this PR.

t-bast:
Closing in favor of #654

PR #612, opened 2019-05-24
BOLT9: Format table

jtimon:
I know this is kind of stupid, but it makes it easier to read in plain text.
I totally understand if people don't think it's worth it.
It doesn't affect how github displays the file, the markdown is still correct.


t-bast:
@rustyrussell could we get a second approval on this and merge it?

PR #615, opened 2019-05-27
Rephrase last node payload requirements

OrfeasLitos:
Mention that `outgoing_cltv_value` has to be equal to `min_final_cltv_expiry` and `amt_to_forward` has to be equal to `amount`, as seen in the invoice

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -291,8 +291,10 @@ using an alternate channel.
 
 When building the route, the origin node MUST use a payload for
 the final node with the following values:
-* `outgoing_cltv_value`: set to the final expiry specified by the recipient
-* `amt_to_forward`: set to the final amount specified by the recipient
+* `outgoing_cltv_value`: set to the final expiry specified by the recipient (e.g.
+  `min_final_cltv_expiry` from a [BOLT #11](11-payment-encoding.md) payment invoice)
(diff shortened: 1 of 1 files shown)


OrfeasLitos:
The old wording however leaves it unclear where these numbers come from. It definitely took me a while to understand what it meant, but a spec should try to help the reader. What do you think about something like this:
"""
* `outgoing_cltv_value`: set to the final expiry specified by the recipient (e.g. `min_final_cltv_expiry` in invoice)
* `amt_to_forward`: set to the final amount specified by the recipient (e.g. `amount` in invoice)
"""
or any other version that mentions `min_final_cltv_expiry` and `amount` from the invoice, but also allows for other ways of defining `outgoing_cltv_value` and `amt_to_forward`?

t-bast:
That sounds good!
I would replace `in invoice` by `from a [BOLT #11](11-payment-encoding.md) payment invoice` but that's nit.

OrfeasLitos:
Great, I'll change it now

t-bast:
@cdecker could we get your opinion/ack on that small change?

cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/615/commits/ae0146f1bf7cb6f265a510935290559a422dcf1f

PR #617, opened 2019-05-30
BOLT4: Correct hop count clarification

jtimon:
The 20 hop count excludes the origin but includes the final node, right? thus it's 19 intermediate nodes but 20 channels.
This feels like a typo, please, ignore me if I'm wrong.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -101,7 +101,7 @@ There are a number of conventions adhered to throughout this document:
 
 # Clarifications
 
-The longest route supported has 20 hops without counting the _origin node_ and _final node_, thus 19 _intermediate nodes_ and a maximum of 20 channels to be traversed.
+The longest route supported has 20 hops without counting the _origin node_, but counting _final node_, thus 19 _intermediate nodes_ and a maximum of 20 channels to be traversed.
 
 # Key Generation
 


t-bast:
I think you're right that the sentence is incorrect but your proposal isn't correct either (unless I'm not interpreting `hops` and `nodes` correctly).
I think this should be right:

> The longest route supported has 20 hops (including the _origin node_ and _final node_), thus 19 _intermediate nodes_ and a maximum of 20 channels to be traversed.

t-bast:
You're right that the sentence is confusing, because `hops` refer to intermediate nodes:

> The packet is routed through a number of intermediate nodes, called hops.

I think that it would be less confusing to say:

> The longest route supported has 19 _intermediate nodes_ and a maximum of 20 channels to be traversed.

jtimon:
> I think that it would be less confusing to say:
> 
> > The longest route supported has 19 _intermediate nodes_ and a maximum of 20 channels to be traversed.

Yeah, I like your sentence better too, it's simpler.
Happy to change it to that if more people agree.

Roasbeef:
Worth noting that after #619, the notion of the maximum hop count may become a bit more fuzzy.

cfromknecht:
I think with #619 the max path length is 26 hops, can anyone else verify that?

t-bast:
The hop count has been completely removed from the bolt with the variable-length payload update, so I'm closing this.
@jtimon feel free to re-open if I missed something.

PR #618, opened 2019-06-05
BOLT 2: Must not send identical fee changes

pm47:
See https://github.com/ElementsProject/lightning/issues/2701 for context.

I went with *MUST NOT* to be consistent with:

> A sending node:
>  - MUST NOT send a `commitment_signed` message that does not include any
updates.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1049,6 +1049,7 @@ given in [BOLT #3](03-transactions.md#fee-calculation).
 The node _responsible_ for paying the Bitcoin fee:
   - SHOULD send `update_fee` to ensure the current fee rate is sufficient (by a
       significant margin) for timely processing of the commitment transaction.
+  - MUST NOT send identical consecutive `update_fee`s.
 
 The node _not responsible_ for paying the Bitcoin fee:
   - MUST NOT send `update_fee`.


ZmnSCPxj:
Is it allowed to do `update_fee A` `update_fee B` `update_fee A`, and what would our software do in that case?

t-bast:
I think this is not a problem @ZmnSCPxj.
If after `update_fee A` you sent a corresponding `commitment_signed`, and do the same after `update_fee B` and the next `update_fee A` everything works - you haven't sent duplicates in a row.
If you wait to batch `commitment_signed` (which you should) you might have sent `update_fee A`, `update_fee B` then `update_fee A` (which is ok since there aren't any subsequent duplicates) and when you update your commit tx you should detect that it hasn't changed since your last `commitment_signed` and shouldn't send a duplicate `commitment_signed`.
Does that make sense?

pm47:
I realize now that I have made a confusion between no-changes (as in: no `update_*` messages) and identical (as in: the resulting new commitment is the same as the previous one) commitments.

Depending on the interpretation BOLT 2 doesn't allow no-changes commitments, but it may or may not allow identical commitments.

I would go with the interpretation that BOLT 2 allows forbids no-changes but allows identical commitments, because it is simpler and there is no point in considering this a spam protection since you can always bump the fee by tiny amounts.

ZmnSCPxj:
@t-bast it seems there would be some flex in how to interpret the spec in this area, as per @pm47 reply above.  If I do `update_fee A` `update_fee B` `update_fee A` `commitment_signed` is this considered "no change" and thus invalid under the current BOLT wording or is this considered "identical, but change still occurred" and thus valid?

t-bast:
Good point. I think that `update_fee A` `update_fee B` `update_fee A` `commitment_signed` would be ok.
Some changes happened, resulting in an identical `commitment_signed` so I think it should fall under the `allow identical commitments` rule.

PR #621, opened 2019-06-16
BOLT #7: receiving node requirements related to timestamp for channel_update message

ugamkamat:
While going through BOLT #7 I was confused by this receiving node requirement for `channel_update` message: "if the `timestamp` is equal to the last-received `channel_update` AND the fields (other than `signature`) differ: MAY blacklist this `node_id`; MAY forget all channels associated with it." On the outset it might read to "`timestamp` should be same, the `signature` should be same as previous but if the other data is different; then blacklist the `node_id`". However, if `signature` is same as previous update but data is different, then the `signature` would be rendered invalid. And since `signature` check is done prior to the `timestamp` check, this situation of checking may not arise. So this check seems to be redundant. However, Mark H corrected me on stackexchange (https://bitcoin.stackexchange.com/questions/88350/bolt-7-can-receiving-node-requirements-in-channel-update-message-give-rise-to) by saying that a valid `signature` must be provided for this `channel_update` and `timestamp` to go through. This pull request slightly corrects the language behind this check and adds a paragraph in the rationale to explain that point. 

In correcting the point, I have said fields below `timestamp` differs for the following reason: (1) `short_channel_id` and `chain_hash` requirements are processed prior to `timestamp` and (2) valid `signature` check is also done prior to the `timestamp`.

While creating this pull request, I also noticed a slight typo in the same section. The sentence prior to the one I mentioned above says that if the `timestamp` is not greater than the previous message, the node should ignore the message. And the second sentence is about comparing if `timestamp` is equal. But again, the equal to check might not be processed as the message will be discarded if it is not greater. I have changed that to "greater than or equal to". Would love your comments and suggestions.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -498,8 +498,8 @@ The receiving node:
   `channel_update` for this `short_channel_id` AND for `node_id`:
     - SHOULD ignore the message.
   - otherwise:
-    - if the `timestamp` is equal to the last-received `channel_update`,
-    AND the fields differ (with a valid `signature` for the message):
+    - if the `timestamp` is equal to the last-received `channel_update` for this
+    `short_channel_id` AND `node_id`, AND the fields below `timestamp` differ:
       - MAY blacklist this `node_id`.
       - MAY forget all channels associated with it.
   - if the `timestamp` is unreasonably far in the future:
@@ -523,6 +523,13 @@ either too far in the future or have not been updated in two weeks; so it
 makes sense to have it be a UNIX timestamp (i.e. seconds since UTC
 1970-01-01). This cannot be a hard requirement, however, given the possible case
 of two `channel_update`s within a single second.
+
+It is assumed that more than one `channel_update` message changing the channel 
+parameters in the same second is a DoS attempt, and therefore, the node responsible 
+for signing such messages ought to be blacklisted. However, a node can send a same 
+`channel_update` message with a different signature (changing the nonce in signature 
+signing), and hence fields apart from signature are checked to see if the channel 
+parameters have changed for the same timestamp.
 
 The explicit `option_channel_htlc_max` flag to indicate the presence
 of `htlc_maximum_msat` (rather than having `htlc_maximum_msat` implied
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -534,7 +534,7 @@ for signing such messages may be blacklisted. However, a node may send a same
 `channel_update` message with a different signature (changing the nonce in signature 
 signing), and hence fields apart from signature are checked to see if the channel 
 parameters have changed for the same timestamp. It is also important to note that 
-ECDSA signatures are malleable. So, an intermediatary node who received the `channel_update` 
+ECDSA signatures are malleable. So, an intermediate node who received the `channel_update` 
 message can rebroadcast it just by changing the `s` component of signature with `-s`. 
 This should however not result in the blacklist of the `node_id` from where
 the message originated.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -526,10 +526,14 @@ of two `channel_update`s within a single second.
 
 It is assumed that more than one `channel_update` message changing the channel 
 parameters in the same second may be a DoS attempt, and therefore, the node responsible 
-for signing such messages may be blacklisted. However, a node can send a same 
+for signing such messages may be blacklisted. However, a node may send a same 
 `channel_update` message with a different signature (changing the nonce in signature 
 signing), and hence fields apart from signature are checked to see if the channel 
-parameters have changed for the same timestamp.
+parameters have changed for the same timestamp. It is also important to note that 
+ECDSA signatures are malleable. So, an intermediatary node who received the `channel_update` 
+message can rebroadcast it just by changing the `s` component of signature with `-s`. 
+This should however not result in the blacklist of the `node_id` from where
+the message originated.
 
 The explicit `option_channel_htlc_max` flag to indicate the presence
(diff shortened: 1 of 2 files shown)


07-routing-gossip.md
```diff
@@ -494,12 +494,12 @@ The receiving node:
   - if the specified `chain_hash` value is unknown (meaning it isn't active on
   the specified chain):
     - MUST ignore the channel update.
-  - if `timestamp` is NOT greater than that of the last-received
+  - if `timestamp` is NOT greater than OR equal to that of the last-received
```
t-bast: Maybe replacing `NOT greater than OR equal to` by `lower than` makes it clearer?

07-routing-gossip.md
```diff
@@ -524,6 +524,13 @@ makes sense to have it be a UNIX timestamp (i.e. seconds since UTC
 1970-01-01). This cannot be a hard requirement, however, given the possible case
 of two `channel_update`s within a single second.
 
+It is assumed that more than one `channel_update` message changing the channel 
+parameters in the same second is a DoS attempt, and therefore, the node responsible
```
t-bast: I'd be less strict on categorizing something as a DoS attempt, you'd need more historical data of a peer misbehaving before putting it in a blacklist.

```suggestion
parameters in the same second may be a DoS attempt, and therefore, the node responsible 
```

07-routing-gossip.md
```diff
@@ -524,6 +524,13 @@ makes sense to have it be a UNIX timestamp (i.e. seconds since UTC
 1970-01-01). This cannot be a hard requirement, however, given the possible case
 of two `channel_update`s within a single second.
 
+It is assumed that more than one `channel_update` message changing the channel 
+parameters in the same second is a DoS attempt, and therefore, the node responsible 
+for signing such messages ought to be blacklisted. However, a node can send a same
```
t-bast:
```suggestion
for signing such messages may be blacklisted. However, a node can send a same 
```

07-routing-gossip.md
ugamkamat: Thanks for the review! I have applied both your changes to your commitment. I'll also change the timestamp config to 'lower than' rather than 'greater than equal to'

07-routing-gossip.md
```diff
@@ -524,6 +524,13 @@ makes sense to have it be a UNIX timestamp (i.e. seconds since UTC
 1970-01-01). This cannot be a hard requirement, however, given the possible case
 of two `channel_update`s within a single second.
 
+It is assumed that more than one `channel_update` message changing the channel 
+parameters in the same second may be a DoS attempt, and therefore, the node responsible 
+for signing such messages may be blacklisted. However, a node can send a same 
+`channel_update` message with a different signature (changing the nonce in signature
```
cfromknecht: important to note that since ECDSA signatures are malleable, the signatures can be different even if the node only signed once.

ugamkamat: Very fair point. I'm incorporating this in the rationale.

07-routing-gossip.md
```diff
@@ -494,12 +494,12 @@ The receiving node:
   - if the specified `chain_hash` value is unknown (meaning it isn't active on
   the specified chain):
     - MUST ignore the channel update.
-  - if `timestamp` is NOT greater than that of the last-received
+  - if `timestamp` is lower than that of the last-received
```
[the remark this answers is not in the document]
ugamkamat: @cfromknecht Thanks! I included your suggestion for the condition where the `timestamp` is equal. But in order to not leave other holes, I slightly changed it in terms of order than what you suggested. Let me know if its good.

t-bast:
@ugamkamat could you update to fix the typo (which would make the build succeed)?
@cfromknecht could you provide feedback/ack?

ugamkamat:
@t-bast Done, typo is fixed

t-bast:
The build is still failing because you introduce words that aren't in the spellcheck list yet.
Could you add "DoS" and "ECDSA" to the `.aspell.en.pws` file?

ugamkamat:
@t-bast Done. Since this was my first time adding a commit, I wasn't entirely sure how to update that file. I have created a PR with those words added.

ugamkamat:
@t-bast So, I have pushed the file to this PR. Let me know if that is correct.

t-bast:
ACK `839d5b3`
We need two approvals to commit, so I'll wait for @cfromknecht or someone else to chime in.

07-routing-gossip.md
```diff
@@ -497,14 +497,18 @@ The receiving node:
...
+    `short_channel_id` AND `node_id`:
+  	- if the fields below `timestamp` differ:
+		- MAY blacklist this `node_id`.
+		- MAY forget all channels associated with it.
+	- if the fields below `timestamp` are equal:
+		- SHOULD ignore this message	
+  - if `timestamp` is lower than that of the last-received
   `channel_update` for this `short_channel_id` AND for `node_id`:
     - SHOULD ignore the message.
   - otherwise:
```
jtimon: I guess this otherwise could have been removed too?

ugamkamat: @jtimon No that otherwise is needed. Just that the `if` and `otherwise` should be indented on the previous otherwise. I think there was a formatting error when creating the pull request.

ugamkamat: Is reverting the commit and making changes the right way to go?

t-bast: No we don't rewrite history. You can submit a tiny indentation fix PR and it should be approved very quickly.

ugamkamat: @t-bast cool. Will submit one soon.

t-bast: Thanks!

PR #623, opened 2019-06-17
I can has extension?

ZmnSCPxj:
Proposal for permissionless development of experimental features.

I believe this can be added to the spec almost entirely as-is, although it does remove 25% of the possible space of available messages for itself.


13-i-can-has-extension.md
```diff
@@ -0,0 +1,191 @@
...
+Two message types, `i_can_has_extension` (type 49152) and
+`i_has_a_extensions` (type 49153) are used to inform about extensions
+and enable an extension to be used over the peer protocol.
+
+Message types 49154 (`0xC002`) to 65535 (`0xFFFF`) are to be used
+for extension messages.
+These message types are _multiplexed_,
+meaning that all extensions share the same message type numbers.
+Messages for a specific extension are called _extension-specific_
+messages.
```
renepickhardt: How is the specific extension enabled?  (further below I see via `I_can_has_extension`) what if two peers enable extensions concurrently? Is only the recipient matching incoming messages to the previously received `I_can_has_extension` message? Shouldn't there be some form of handshake in which two peers negotiate that they will now use this particular extension? Or shouldn't there at least in message types 49154 to 65535 always be the requirement to send the current extension vector?

13-i-can-has-extension.md
```diff
@@ -0,0 +1,191 @@
...
+which it uses for received messages.
+The `i_can_has_extension` message is used by the sender
+to change the "currently enabled extension" of the receiver.
+
+## Requirements
+
+Before sending an extension-specific message to the peer,
+the sending node MUST ensure that
+the "currently enabled extension" of the receiver is the extension
+the sending node intends to use.
```
renepickhardt: as mentioned above how can that go without handshake and a "connection"?

13-i-can-has-extension.md
```diff
@@ -0,0 +1,191 @@
...
+peer connection is established.
+
+## The `i_has_a_extensions` Message
+
+This message contains a set of extensions supported by a node.
+
+1. type: 49153 (`i_has_a_extensions`)
+2. data:
+   * [`2` : `num_extensions`]
+   * [`num_extensions * 20` : `extensions`]
```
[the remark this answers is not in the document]
ZmnSCPxj: Possibly support sending this message multiple times.

13-i-can-has-extension.md
```diff
@@ -0,0 +1,191 @@
...
+
+### Requirements
+
+The sender:
+  - MUST send this message as soon as the connection is
+    established.
```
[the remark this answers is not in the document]
ZmnSCPxj: The intent is that the sender must send this as soon as possible, so that the recipient can know immediately what extensions the sender has.

However, it is also the intent, that this message is optional, so possibly better wording can be had here.

13-i-can-has-extension.md
ZmnSCPxj: Only the recipient matches incoming messages to the previously requested extension of the sender.  So one side can be talking in one extension, while the other side is talking in a different extension, and it is up to the listener to remember what "language" it is supposed to receive.  This reflects the fact that we can expect the messages to arrive at the same order (thanks to TCP) but we cannot expect both peers to execute in tight synchrony.

13-i-can-has-extension.md
ZmnSCPxj: BOLT 8 already specifies an ongoing connection; all Lightning messages are sent via BOLT 8.

dr-orlovsky:
What is the current status of these proposal? It would be great to have it accepted

ZmnSCPxj:
I do not know proper procedure for how to get these merged, though am interested in somehow getting it in.

ZmnSCPxj:
Subsumed by #705

PR #632, opened 2019-07-09
Update .aspell.en.pws

ugamkamat:


t-bast:
You should do this directly in the PR concerned, ie #621 
Can you please do these changes in a new commit in #621 ?

PR #634, opened 2019-07-10
BOLT 2: remove local/remote from reestablish field names.

rustyrussell:
(No spec change, just wording)

The "local" and "remote" here are just *confusing*.  Each side says
where it's at, and the other side retransmits based on that.

We could call it 'number_of_next_commitment_i_expect_to_receive' and
'number_of_next_revocation_i_expect_to_receive' but that's getting
silly.

These names were a major source of confusion while writing tests!

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1115,12 +1115,12 @@ messages are), they are independent of requirements here.
 1. type: 136 (`channel_reestablish`)
 2. data:
    * [`channel_id`:`channel_id`]
-   * [`u64`:`next_local_commitment_number`]
-   * [`u64`:`next_remote_revocation_number`]
+   * [`u64`:`next_commitment_number`]
+   * [`u64`:`next_revocation_number`]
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -1264,15 +1264,15 @@ A re-transmittal of `revoke_and_ack` should never be asked for after a
...
 by the remote node.
 
-Note that the `next_local_commitment_number` starts at 1, since
+Note that the `next_commitment_number` starts at 1, since
 commitment number 0 is created during opening.
-`next_remote_revocation_number` will be 0 until the
-`commitment_signed` for commitment number 1 is received, at which
-point the revocation for commitment number 0 is sent.
+`next_revocation_number` will be 0 until the
+`commitment_signed` for commitment number 1 is send and then
```
niftynei:
```suggestion
`commitment_signed` for commitment number 1 is sent and
```

PR #639, opened 2019-07-11
BOLT 5 : data_loss let you salvage received HTLC on remote commitment

ariard:
Thanks to data_loss_protect, your remote peer is going to give you his current per_commitment_point if channel_reestablish are exchanged. If he owns a fully-signed commitment_transaction and broadcast it, using your basepoints you should be able to derive keys even for remote_htlcpubkey and get back, non only your to_remote output, but also HTLC outputs.


Roasbeef:
The assumption here is that if you're behind in commitment state, then you likely don't also have the required information to reconstruct the HTLC script itself.

ariard:
Oh right, was thinking only in a key derivation side, agree it's highly likely you won't have preimage/timelock to rebuild scripts

PR #640, opened 2019-07-12
BOLT01: swap CompactSize for BigSize in TLV format

cfromknecht:
This commit modifies the varint encoding used for TLV types and lengths
to use a custom format called BigSize. The format is identical to
bitcoin's CompactSize, except it replaces the use of little-endian
encodings for multi-byte values with big-endian. This is done to prevent
mixing endianness within the protocol, since otherwise CompactSize would
be the first introduction of little-endian encodings.

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -341,3 +341,5 @@ optimizations
 structs
 CompactSize
 encodings
+bigsize
+BigSize
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -20,6 +20,7 @@ All data fields are unsigned big-endian unless otherwise specified.
     * [The `error` Message](#the-error-message)
   * [Control Messages](#control-messages)
     * [The `ping` and `pong` Messages](#the-ping-and-pong-messages)
+  * [Appendix B: BigSize Test Vectors](#appendix-b-bigsize-test-vectors)
   * [Acknowledgments](#acknowledgments)
   * [References](#references)
   * [Authors](#authors)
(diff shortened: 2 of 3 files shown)


01-messaging.md
```diff
@@ -95,16 +96,21 @@ A `tlv_record` represents a single field, encoded in the form:
 * [`varint`: `length`]
 * [`length`: `value`]
 
+A `varint` is a variable-length, unsigned integer encoding using the
+[BigSize](#appendix-b-bigsize-test-vectors) format, which resembles the bitcoin
+CompactSize encoding but uses big-endian for multi-bye values rather than
```
t-bast:
```suggestion
CompactSize encoding but uses big-endian for multi-byte values rather than
```

.aspell.en.pws
```diff
@@ -341,3 +341,28 @@ optimizations
...
+json
+testReadVarInt
+testWriteVarInt
+ReadVarInt
+WriteVarInt
+varIntTest
+tlv
+ExpErr
+Fatalf
+fc
```
t-bast: It's a bit weird to have to add all of those in the spell-checker...
I think that if we always wrap them in back-ticks (e.g. `0xfc`) the spell-checker doesn't look at them, which might be a better solution.

cfromknecht: good point, i gave a shot at modifying the spell checker to ignore code blocks, but not sure that's the best approach

rustyrussell:
Ack.

We ended up not reusing the Bitcoin routines and reimplementing them for this, so using Bitcoin endianness here was a lose. Wasn't sure if other implementations were the same.

JSON for testcases is a bit useless without a programmatic extractor IMHO, but I'm just happy to have test vectors at all. We can have a Grand Testcase Unification later...

rustyrussell:
PS. I've lost count of how many beers I owe @cfromknecht now, but I suspect it may be more than we can drink in Berlin...

t-bast:
ACK. Let's decide on this once and for all, big-endian everywhere it is!

rustyrussell:
Tested-Ack: 8038471806a45d0fa15ea0d646050d0df25b53ba

.aspell.en.pws
niftynei: i'm +1 for ignoring code blocks fwiw

PR #642, opened 2019-07-17
option_static_remotekey

rustyrussell:
This is now *just* removing key rotation (no CSV symmetry).  I've used the placeholder optnum 48/49, we can assign a proper one at the meeting if accepted.

I have implemented this in c-lightning as an EXPERIMENTAL_FEATURES option, but not yet merged.  It seems to work, and I also have protocol tests for it.


02-peer-protocol.md
```diff
@@ -1166,12 +1178,14 @@ The sending node:
...
   next `revoke_and_ack` message it expects to receive.
-  - if it supports `option_data_loss_protect`:
+  - if it supports `option_data_loss_protect` or `option_static_remotekey`:
     - if `next_remote_revocation_number` equals 0:
       - MUST set `your_last_per_commitment_secret` to all zeroes
     - otherwise:
       - MUST set `your_last_per_commitment_secret` to the last `per_commitment_secret`
     it received
+  - if `option_static_remotekey` applies to the commitment transaction:
+	- MUST NOT include `my_current_per_commitment_point`.
```
cfromknecht: does this mean that it should be blank?

Obviously haven't tried to implement it yet, but I'm wondering, do we have to modify this at all? Is it possible to keep all of the DLP handling/derivation the same, but just use the raw `payment_basepoint` in `to_remote` outputs?

rustyrussell: No, I want to kill it.  It's under-defined, it's been a source of problems, and even now I suspect we're seeing disagreements between c-lightning and lndL I'm seeing sync error far too often.

So, not blank, literally not present.  Of course, recipient will ignore anyway, but next extension may rely on this by appending a TLV...

rustyrussell:
> Thanks @rustyrussell, definitely like the size of diff :)
> 
> Do we also want to modify the derivations described [in BOLT 3](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#localpubkey-remotepubkey-local_htlcpubkey-remote_htlcpubkey-local_delayedpubkey-and-remote_delayedpubkey-derivation) to say that that the `payment_basepoint` is not tweaked?

Huh, somehow my rebase lost that part!  I def did edit that, will push...

02-peer-protocol.md
cfromknecht: It still seems orthogonal to not tweaking the `to_remote` outputs. I can't imagine adding more logic in this area is going to improve the issues we're seeing... :)

02-peer-protocol.md
```diff
@@ -1298,6 +1321,13 @@ is valid. However, this also means the fallen-behind node has revealed this
 fact (though not provably: it could be lying), and the other node could use this to
 broadcast a previous state.
 
+`option_static_remotekey` removes the changing `to_remote` key,
+so the `my_current_per_commitment_point` is unnecessary and thus
+removed, but the disclosure of previous secret still allows
+fall-behind detection.  An implementation can offer both, however, and
+fall back to the `option_data_loss_protect` behavior if
+`option_simplified_commitment` is not negotiated.
```
araspitzu:
```suggestion
`option_static_remotekey` is not negotiated.
```

02-peer-protocol.md
Roasbeef: If we want to add additional fields to this in the future, then we can't have it be absent, since w/o TLV, we don't actually have "optional fields".

02-peer-protocol.md
```diff
@@ -377,6 +383,12 @@ The recipient:
   - on receipt of a valid `funding_signed`:
     - SHOULD broadcast the funding transaction.
 
+#### Rationale
+
+We decide on `option_static_remotekey` at this point when we first have to generate the commitment
+transaction.  Even if a later reconnection does not negotiate this parameter, this channel will honor it.
```
Roasbeef: What does the second sentence mean? That this is actually a _connection_ level value, rather than local to this specific channel negotiation?

02-peer-protocol.md
```diff
@@ -1205,8 +1219,17 @@ A node:
       - SHOULD fail the channel.
 
  A receiving node:
-  - if it supports `option_data_loss_protect`, AND the `option_data_loss_protect`
-  fields are present:
+  - if `option_static_remotekey` applies to the commitment transaction:
+    - if `next_remote_revocation_number` is greater than expected above, AND
+    `your_last_per_commitment_secret` is correct for that
```
Roasbeef: Not a big deal, but there's a bit of duplication here and below.

09-features.md
```diff
@@ -26,6 +26,7 @@ These flags may only be used in the `init` message:
 | 3  | `initial_routing_sync` | Indicates that the sending node needs a complete routing information dump | [BOLT #7](07-routing-gossip.md#initial-sync) |
 | 4/5  | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
 | 6/7  | `gossip_queries`           | More sophisticated gossip control | [BOLT #7](07-routing-gossip.md#query-messages) |
+| 48/49| `option_static_remotekey`   | Static key for remote output | [BOLT #3](03-transactions.md) |
```
Roasbeef: IMO this should also be a global feature bit, so nodes can opt to _only_ open channels to other nodes that support this new "safu commitment".

Roasbeef: Also why not in order? So 10/11?

02-peer-protocol.md
```diff
@@ -366,6 +366,12 @@ This message introduces the `channel_id` to identify the channel. It's derived f
 
 #### Requirements
 
+Both peers:
```
Roasbeef: I think we may also want to define a new bit in the funding flags within the `OpenChannel` message to allow nodes to modulate this value, since there'll be optional _and_ required feature bits for it. It'll take some time for channels to switch over (since I don't foresee the entire networking shutting down and transitioning to this new channel type). For some period of time, implementations will need to understand both the old and new commitment types, and a node can't feasible _only_ use this new channel type. 

If we add a new funding flag, then we'll be able to easily flip it via a config option or the like to ensure we all still understand the old and new formats. This will also be useful for new implementations as they'll still need to interface with the rest of the legacy network. 

In short I propose two signalling venues:
   * the global feature bit
   * a new defined in in the `open_channel` message's funding flags

02-peer-protocol.md
rustyrussell: You can't turn it off by reconnecting without the feature later.  If someone downgrades, it's undefined.

02-peer-protocol.md
rustyrussell: Of course we can.  If we negotiate static_remotekey (which will eventually become compulsory), then the TLV goes immediately after it.  We won't support TLVs without static_remotekey I expect.

03-transactions.md
```diff
@@ -459,6 +459,19 @@ secrets are known (i.e. the private keys corresponding to `localpubkey`, `local_
...
 
+### `remotepubkey` Derivation
+
+If `option_static_remotekey` is negotiated the `remotepubkey` is simply the
+remote node's `payment_basepoint`, otherwise it is calculated as above using
+the remote node's `payment_basepoint`.
+
+The simplified derivation means that a node can spend a commitment
+transaction even if it has lost data and doesn't know the
+corresponding `payment_basepoint`.  A watchtower could correlate
```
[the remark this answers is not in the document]
rustyrussell: Fixed

02-peer-protocol.md
rustyrussell: Indeed.

I decided to do that so eventually when we remove the `option_data_loss_protect` once we're in a world of rainbows and unicorns, the spec will read better and we can just remove the latter paragraph.

09-features.md
rustyrussell: Because local features will also be advertized when we stop bikeshedding #571...

The real bit will be assigned when we finally ack this and are doing interop...

02-peer-protocol.md
rustyrussell: I think current feature negotiation is sufficient.  The open_channel flag seems completely redundant?

rustyrussell:
Rebased to fix minor conflicts, and added minor fixes in separate commit.

rustyrussell:
OK, I am now running this on my two main test nodes:

mainnet: 024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605
testnet: 031a3478d481b92e3c28810228252898c5f0d82fc4d07f5210c4f34d4aba56b769

02-peer-protocol.md
Roasbeef: > We won't support TLVs without static_remotekey I expect.

This _may_ be the case, but it seems short sighted to block of that path all together, when there's little cost to leaving the field as is. Newer implementations will simply write zeroes for that field, and then being the actual TLV parsing after that boundary. It's a fundamental trait that pre-TLV there're no optional fields, therefore we can't just "delete" them without adding additional parsing complexity. 

The other downside of "removing" this field, is that it makes parsing more contextual. This path forces parsing to be aware of the higher level context of which feature bits are online vs just reading the data, then interpreting it later.

cfromknecht: deleting it also a privacy leak, in that reveals what type of channel you have to anyone inspecting the size of your encrypted packets

rustyrussell: OK, as per meeting agreement, we'll leave it and ignore it.  http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-09-02-20.06.html

sstone:
It just hit me that switching to static remote keys without adding a CSV delay, as for the to-local keys, is making incentives for using mutual vs force close worse than they are today.

With the current specs, not having a delay on the to-remote is already a problem: if you can get your peer to publish their commit tx, you get a UTXO that you can spend right away, while they have to wait to spend theirs. However, most implementations will use BIP32 wallets and will want spend this output (and pay onchain fees) and send it to a wallet key, as it makes backup/restore of onchain funds much easier (you just need your seed). So it's still better to negotiate a mutual close whenever possible, as it will return your funds directly to your BIP32 wallet.

With this proposal, without adding a CSV delay to the to-remote output, implementation will most likely set their static remote key to one of their BIP32 wallet key, which means that in practice getting your peers to publish their commit tx will always be the fastest solution, and could also be the cheapest if fees are going up and you don't want to negotiate.

This is why I would strongly prefer that we add a CSV delay to the to-remote output: users would then have an incentive to mutual close channels whenever possible, and not play the "get your peer to close first" game.

Roasbeef:
Adding a CSV value makes the recovery case less self contained. The most
important part about this change is making recovery of funds in the case of
disaster much easier. I'd say the current nonce usage is the #1 reason
users lose money on the network: they delete their app or state for w/e
reason, then have lost hope even if they have a seed of some sort. Even
just yesterday I met a user that lost funds in this very fashion. Remedying
this situation should be amongst our highest priorities.

The CSV change is a larger change that is more subtle and has affects on
the incentives of the system that should be considered in isolation of
making Lightning safer for the end user. In terms of fees/chain space, the
straight shot output also reduces total on chain traffic as there're less
sweeps required per force close.

IMO, the bigger incentive issue at play is the fact that the initiator is
the one that pays all fees on the commitment transaction atm.

sstone:
> Adding a CSV value makes the recovery case less self contained.
What does that mean ?

Having you peer include in its commit tx an output that sends directly to your wallet feels just plain wrong. What is the point of even negotiating a mutual close then ?

Roasbeef:
It means that if I lose data, then when the other party force closes, the
funds go directly to my wallet. Surely you've encountered users that have
deleted an app for w/e reason or their application state and have lost
funds. This makes the recovery case even easier, resulting in less users
funds being lost, or put at risk needing to be rescued via manual action.
This is the #1 thing that confuses users: what do you mean if they force
close the funds don't go directly into my wallet?? I need to receive some
information from the other party to receive my funds once again? What if
they don't always send me that information after they close out the channel
(which only lnd does), does that mean my coins are lost?

As is today, you'd want to do a cooperative close as you'll negotiate lower
fees since the fee on the commitment transaction serves as an upper limit
on the fees paid. A cooperative close also allows you to more freely send
the funds elsewhere from the channel vs a force close which means the funds
must be sent to a predetermined key.

SomberNight:
> With this proposal, without adding a CSV delay to the to-remote output, implementation will most likely set their static remote key to one of their BIP32 wallet key
> Having you peer include in its commit tx an output that sends directly to your wallet 

Note that the to-remote-output pays to a p2wpkh script; and longer term there will be wallets who would not by default derive p2wpkh addresses for their main on-chain addresses. E.g. when witness v1 gets soft forked in, some wallets might want to only have witness v1 addresses, and those will then need to sweep the to-remote outputs, at which point a cooperative close paying directly to one of their main witness v1 addresses (as soon as will be allowed for a cooperative close to do so, as it is not atm) would be cheaper.

My point is that it's not necessarily a true assumption that the commitment tx having a static p2wpkh output can always directly pay to a main wallet address.

Roasbeef:
Witness v1 is some ways off. Even then, wallets won't cease to use BIP 49,
as there'll need to be a transition period. Even now not all outputs use
witness v0, and I don't expect there ever to be 100% utilization.

SomberNight:
There are already wallets that only derive witness v0 outputs, e.g. Wasabi and Electrum.

sstone:
@Roasbeef I don't see why adding a delay here prevents you from recovering your funds ? My point is that not having such a delay was bad before and we're making it worse. 

The fact that you currently need you peer's commit point is a different problem. the to-remote key could have been derived from the remote commit point (I don't think it makes security/privacy worse) and it would have made recovery much simpler.

@SomberNight well we'll have eltoo by then won't we :) but it may not be soon...

Roasbeef:
>  I don't see why adding a delay here prevents you from recovering your funds ? 

It doesn't _prevent_ it, it makes it less self contained: recovery logic now needs additional data (to reconstruct the script). 

> My point is that not having such a delay was bad before and we're making it worse.

With my response being that the _real_ incentive issue here is the initiator always paying the fees whenever a channel closes. 

> The fact that you currently need you peer's commit point is a different problem. the to-remote key could have been derived from the remote commit point (I don't think it makes security/privacy worse) and it would have made recovery much simpler.


This is the very problem that's being fixed with this change.

02-peer-protocol.md
Roasbeef: I realized we may actually still need to _check_ it, even if we're using this new format. 

`lnd` has a recovery mode where in order to trigger the other party to force close if we have zero state transitions (no commit secret to check as nothing has been revoked), we'd purposefully set an invalid per commitment point. However, if the other party ignores it, then everything else looks fine: "state zero local, state zero remote, aight let's boot this thing up". So now Carol can't signal to Bob that they've lost data and Bob should force close.

Roasbeef: Put in less implementation specific terms: we need a way to signal that a party has lost data while they're on state zero, so the other person can force close. Previously `lnd` relied on the fact that we'd always check the commitment point. As a result, we'll (on our end implementation wise) will continue to always check that point. This isn't _mandatory_ persay, but would be nice if eclair/CL still checked that point to support this recovery case.

rustyrussell:
@Roasbeef Meeting nodes say this is pending your ack to merge?  Ping!

Roasbeef:
@rustyrussell needs a rebase! I'll make another PR to add that [note re compatibility with detecting proper data loss from state zero](https://github.com/lightningnetwork/lightning-rfc/pull/642#discussion_r325445545).

Roasbeef:
Actually, the current text looks fine, scratch my last comment, doing one final pass over everything.

cfromknecht:
needs rebase!!

rustyrussell:
Trivial rebase, and collapsed into a single commit.

Xekyo:
Clarifying question: does this allow only to set an address from the Lightning Client's wallet, or would it be allowed to set an arbitrary address?

Roasbeef:
That's an implementation detail.

blacktemplar:
I have a question for this change: With this change is it possible to use low fees on the commitment transaction and in the case of a force close use CPFP to get the transaction through? I really hope this is possible some day since if the fee market gets serious the high fees for commitment transactions serving as an upper bound will be a big problem in my opinion. Using CPFP would enable a much more dynamic fee system and every participant could decide for themselves how fast they need the assets / how much they want to pay for fees.

PR #643, opened 2019-07-17
Base AMP

rustyrussell:
This is a commit-separated and TLV-enhanced version of #511 by @ZmnSCPxj 

We need the total amount, due to amountless invoices.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -248,18 +248,111 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 8 (`payment_data`)
+    2. data:
+        * [`32*byte`:`payment_secret`]
+        * [`tu64`:`total_msat`]
 
 ### Requirements
 
 The writer:
-  - MUST include `amt_to_forward` and `outgoing_cltv_value` for every node.
-  - MUST include `short_channel_id` for every non-final node.
-  - MUST NOT include `short_channel_id` for the final node.
+  - Unless `node_announcement`, `init` message or the [BOLT #11](11-payment-encoding.md#tagged-fields) offers feature `var_onion_optin`:
+    - MUST use the legacy payload format instead.
+  - For every node:
+    - MUST include `amt_to_forward` and `outgoing_cltv_value`.
+  - For every non-final node:
+    - MUST include `short_channel_id`
+	- MUST NOT include `payment_data`
+  - For the final node:
+    - MUST NOT include `short_channel_id`
+    - if the recipient provided `payment_secret`:
+      - MUST include `payment_data`
+	  - MUST set `payment_secret` to the one provided
+	  - MUST set `total_msat` to the total amount it will send
 
 The reader:
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
+  - if it is the final node:
+    - MUST treat `total_msat` as if it were equal to `amt_to_forward` if it
+	  is not present.
 
-The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format).
+The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format)
+and [below](#basic-multi-part-payments).
+
+### Basic Multi-Part Payments
+
+An HTLC may be part of a larger "multi-part" payment: such
+"base" atomic multipath payments will use the same `payment_hash` for
+all paths.
+
+Note that `amt_to_forward` is the amount for this HTLC only: a
+`total_msat` field containing a greater value is a promise by the
+ultimate sender that the rest of the payment will follow in succeeding
+HTLCs; we call these outstanding HTLCs which have the same preimage,
+an "HTLC set".
+
+#### Requirements
+
+The writer:
+  - if the invoice offers the `basic_mpp` feature:
+    - MAY send more than one HTLC to pay the invoice.
+    - MUST use the same `payment_hash` on all HTLCs in the set.
+    - SHOULD send all payments at approximately the same time.
+    - SHOULD try to use diverse paths to the recipient for each HTLC.
+    - SHOULD retry and/or re-divide HTLCs which fail.
+    - if the invoice specifies an `amount`:
+       - MUST set `total_msat` to at least that `amount`, and less
+         than or equal to twice `amount`.
+    - otherwise:
+      - MUST set `total_msat` to the amount it wishes to pay.
+    - MUST ensure that the total `amount_msat` of the HTLC set which arrives at the payee
+      is equal to `total_msat`.
+    - MUST NOT send another HTLC if the total `amount_msat` of the HTLC set is already greater or equal to `total_msat`.
+  - otherwise:
+    - MUST set `total_msat` equal to `amt_to_forward`.
+
+The final node:
+  - MUST fail the HTLC if dictated by Requirements under [Failure Messages](#failure-messages)
+    - Note: "amount paid" specified there is the `total_msat` field.
+  - if it does not support `basic_mpp`:
+    - MUST fail the HTLC if `total_msat` is not exactly equal to `amt_to_forward`.
+  - otherwise, if it supports `basic_mpp`:
+    - MUST add it to the HTLC set corresponding to that `payment_hash`.
+    - SHOULD fail the entire HTLC set if `total_msat` is not the same for
+      all HTLCs in the set.
+    - if the total `amount_msat` of this HTLC set equals `total_msat`:
+      - SHOULD fulfill all HTLCs in the HTLC set
+    - otherwise, if the total `amount_msat` of this HTLC set is less than
+	  `total_msat`:
+      - MUST NOT fulfill any HTLCs in the HTLC set
+      - MUST fail all HTLCs in the HTLC set after some reasonable timeout.
+        - SHOULD wait for at least 60 seconds after the initial HTLC.
+        - SHOULD use `mpp_timeout` for the failure message.
+    - if it fulfills any HTLCs in the HTLC set:
+       - MUST fulfill the entire HTLC set.
+
+#### Rationale
+
+If `basic_mpp` is present it causes a delay to allow other partial
+payments to combine.  The total amount must be sufficient for the
+desired payment, just as it must be for single payments.  But this must
+be reasonably bounded to avoid a denial-of-service.
+
+Because invoices do not necessarily specify an amount, and because
+payers can add noise to the final amount, the total amount must be
+sent explicitly.  The requirements allow exceeding this slightly, as
+it simplifies adding noise to the amount when splitting, as well as
+scenarios in which the senders are genuinely independent (friends
+splitting a bill, for example).
+
+The restriction on sending an HTLC once the set is over the agreed total prevents the preimage being released before all
+the partial payments have arrived: that would allow any intermediate
+node to immediately claim any outstanding partial payments.
+
+An implementation may choose not to fulfill an HTLC set which
+otherwise meets the amount criterion (eg. some other failure, or
+invoice timeout), however if it were to fulfill only some of them,
+intermediary nodes could simply claim the remaining ones.
 
 # Accepting and Forwarding a Payment
 
@@ -313,6 +406,8 @@ using an alternate channel.
 When building the route, the origin node MUST use a payload for
 the final node with the following values:
 
+* `payment_secret`: set to the payment secret specified by the recipient (e.g.
+  `payment_secret` from a [BOLT #11](11-payment-encoding.md) payment invoice)
 * `outgoing_cltv_value`: set to the final expiry specified by the recipient (e.g.
   `min_final_cltv_expiry` from a [BOLT #11](11-payment-encoding.md) payment invoice)
 * `amt_to_forward`: set to the final amount specified by the recipient (e.g. `amount`
@@ -827,9 +922,10 @@ handling by the processing node.
    * [`u64`:`htlc_msat`]
    * [`u32`:`height`]
 
-The `payment_hash` is unknown to the final node, the amount for that
-`payment_hash` is incorrect or the CLTV expiry of the htlc is too close to the
-current block height for safe handling.
+The `payment_hash` is unknown to the final node, the `payment_secret` doesn't
+match the `payment_hash`, the amount for that `payment_hash` is incorrect or
+the CLTV expiry of the htlc is too close to the current block height for safe
+handling.
 
 The `htlc_msat` parameter is superfluous, but left in for backwards
 compatibility. The value of `htlc_msat` always matches the amount specified in
@@ -884,6 +980,11 @@ The decrypted onion per-hop payload was not understood by the processing node
 or is incomplete. If the failure can be narrowed down to a specific tlv type in
 the payload, the erring node may include that `type` and its byte `offset` in
 the decrypted byte stream.
+
+1. type: 23 (`mpp_timeout`)
+
+The complete amount of the multi-part payment was not received within a
+reasonable time.
 
 ### Requirements
 
@@ -952,6 +1053,10 @@ An _intermediate hop_ MUST NOT, but the _final node_:
   - if the payment hash has already been paid:
     - MAY treat the payment hash as unknown.
     - MAY succeed in accepting the HTLC.
+  - if the `payment_secret` doesn't match the expected value for that `payment_hash`,
+    or the `payment_secret` is required and is not present:
+    - MUST fail the HTLC.
+    - MUST return an `incorrect_or_unknown_payment_details` error.
   - if the amount paid is less than the amount expected:
     - MUST fail the HTLC.
     - MUST return an `incorrect_or_unknown_payment_details` error.
--- a/09-features.md
+++ b/09-features.md
@@ -35,6 +35,8 @@ The Context column decodes as follows:
 | 8/9   | `var_onion_optin`                | Requires/supports variable-length routing onion payloads  | IN       | [Routing Onion Specification][bolt04] |
 | 10/11 | `gossip_queries_ex`              | Gossip queries can include additional information         | IN       | [BOLT #7][bolt07-query]               |
 | 12/13| `option_static_remotekey`         | Static key for remote output                              | IN       | [BOLT #3](03-transactions.md)         |
+| 14/15 | `payment_secret`                 | Node supports `payment_secret` field                      | IN9      | [Routing Onion Specification][bolt04]   |
+| 16/17 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | [BOLT #3][bolt04-mpp]                 |
 
 ## Requirements
 
@@ -67,3 +69,4 @@ This work is licensed under a [Creative Commons Attribution 4.0 International Li
 [bolt04]: 04-onion-routing.md
 [bolt07-sync]: 07-routing-gossip.md#initial-sync
 [bolt07-query]: 07-routing-gossip.md#query-messages
+[bolt04-mpp]: 04-onion-routing.md#basic-multi-part-payments
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -282,7 +282,7 @@ The field is big-endian.  The least-significant bit is numbered 0,
 which is _even_, and the next most significant bit is numbered 1,
 which is _odd_.
 
-Note that the `payment_secret` feature probing attacks from nodes
+Note that the `payment_secret` feature prevents probing attacks from nodes
 along the path, but only if made compulsory: yet doing so will break
 older clients which do not understand the feature.
 
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -129,6 +129,7 @@ Each Tagged Field is of the form:
 Currently defined tagged fields are:
 
 * `p` (1): `data_length` 52. 256-bit SHA256 payment_hash. Preimage of this provides proof of payment.
+* `s` (16): `data_length` 52. This 256-bit secret prevents forwarding nodes from probing the payment recipient.
 * `d` (13): `data_length` variable. Short description of purpose of payment (UTF-8), e.g. '1 cup of coffee' or 'ナンセンス 1杯'
 * `n` (19): `data_length` 53. 33-byte public key of the payee node
 * `h` (23): `data_length` 52. 256-bit description of purpose of payment (SHA256). This is used to commit to an associated description that is over 639 bytes, but the transport mechanism for the description in that case is transport specific and not defined here.
@@ -148,7 +149,7 @@ Currently defined tagged fields are:
 ### Requirements
 
 A writer:
-  - MUST include exactly one `p` field.
+  - MUST include exactly one `p` and `s` fields.
   - MUST set `payment_hash` to the SHA2 256-bit hash of the `payment_preimage`
   that will be given in return for payment.
   - MUST include either exactly one `d` or exactly one `h` field.
@@ -191,8 +192,8 @@ A writer:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
 
 A reader:
-  - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h`, or
-  `n` fields that do NOT have `data_length`s of 52, 52, or 53, respectively.
+  - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h`, `s` or
+  `n` fields that do NOT have `data_length`s of 52, 52, 52 or 53, respectively.
   - if the `9` field contains unknown _odd_ bits that are non-zero:
     - MUST ignore the bit.
   - if the `9` field contains unknown _even_ bits that are non-zero:
@@ -202,6 +203,8 @@ A reader:
   description.
   - if a valid `n` field is provided:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
+  - if there is a valid `s` field:
+    - MUST use that as [`payment_secret`](04-onion-routing.md#tlv_payload-payload-format)
 
 ### Rationale
 
@@ -279,6 +282,34 @@ The field is big-endian.  The least-significant bit is numbered 0,
 which is _even_, and the next most significant bit is numbered 1,
 which is _odd_.
 
+Note that the `payment_secret` feature probing attacks from nodes
+along the path, but only if made compulsory: yet doing so will break
+older clients which do not understand the feature.
+
+### Requirements
+
+A writer:
+  - if `payment_secret` feature is set:
+    - MUST include an `s` field.
+  - otherwise:
+    - MUST NOT include an `s` field.
+  - if the `payment_secret` field is required in the onion:
+    - MUST set the even feature `payment_secret`.
+  - If the final node supports [Basic multi-part payments](04-onion-routing.md#basic-multi-part-payments):
+    - MUST set the `basic_mpp` feature.
+  - Otherwise:
+    - MUST NOT set the `basic_mpp` feature.
+  - if it sets either `payment_secret` or `basic_mpp` features:
+    - MUST set the `var_onion_optin` feature.
+  - MUST set `var_onion_optin` if and only if it supports that feature.
+
+A reader:
+  - if the `basic_mpp` feature is offered in the invoice:
+    - MAY pay using [Basic multi-part payments](04-onion-routing.md#basic-multi-part-payments).
+  - otherwise:
+    - MUST NOT use [Basic multi-part payments](04-onion-routing.md#basic-multi-part-payments).
+
+
 # Payer / Payee Interactions
 
 These are generally defined by the rest of the Lightning BOLT series,
@@ -318,7 +349,7 @@ https://github.com/rustyrussell/lightning-payencode
 
 # Examples
 
-NB: all the following examples are signed with `priv_key`=`e126f68f7eafcc8b74f54d269fe206be715000f94dac067d1c04a8ca3b2db734`.
+NB: all the following examples are signed with `priv_key`=`e126f68f7eafcc8b74f54d269fe206be715000f94dac067d1c04a8ca3b2db734`.  Also, the first 9 examples are legacy: modern invoices have an `s` field. 
 
 > ### Please make a donation of any amount using payment_hash 0001020304050607080900010203040506070809000102030405060708090102 to me @03e7156ae33b0a208d0744199163177e909e80176e55d97a2f221ede0f934dd9ad
 > lnbc1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdpl2pkx2ctnv5sxxmmwwd5kgetjypeh2ursdae8g6twvus8g6rfwvs8qun0dfjkxaq8rkx3yf5tcsyz3d73gafnh3cax9rn449d9p5uxz9ezhhypd0elx87sjle52x86fux2ypatgddc6k63n7erqz25le42c4u4ecky03ylcqca784w
@@ -535,8 +566,8 @@ Breakdown:
   * `6c6e626332306d0b25fe64570d0e496dbd9f8b0d000dbb44824f751380da37c6dba89b14f6f92047d63f576e304021a00008101820283038404800081018202830384048000810182028303840480810243500c318a1e0a628b34025e8c9019ab6d09b64c2b3c66a693d0dc63194b02481931000` hex of data for signing (prefix + data after separator up to the start of the signature)
   * `399a8b167029fda8564fd2e99912236b0b8017e7d17e416ae17307812c92cf42` hex of SHA256 of the preimage
 
-> ### Please send $30 for coffee beans to the same peer, which supports features 1 and 9
-> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9qzsze992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq73t7cl
+> ### Please send $30 for coffee beans to the same peer, which supports features 15 and 99, using secret 0x1111111111111111111111111111111111111111111111111111111111111111
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdeessp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q5sqqqqqqqqqqqqqqqpqqq4u9s93jtgysm3mrwll70zr697y3mf902hvxwej0v7c62rsltw83ng0pu8w3j230sluc5gxkdmm9dvpy9y6ggtjd2w544mzdrcs42t7sqdkcy8h
 
 Breakdown:
 
@@ -548,14 +579,17 @@ Breakdown:
 * `d`: short description
   * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20)
   * `vdhkven9v5sxyetpdees`: 'coffee beans'
+* `s`: payment secret
+  * `p5`: `data_length` (`p` = 1, `5` = 20; 1 * 32 + 20 == 52)
+  * `zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs`: 0x1111111111111111111111111111111111111111111111111111111111111111
 * `9`: features
-  * `qz`: `data_length` (`q` = 0, `z` = 2; 0 * 32 + 2 == 2)
-  * `sz`: b1000000010
-* `e992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq`: signature
-* `73t7cl`: Bech32 checksum
+  * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20) 4
+  * `sqqqqqqqqqqqqqqqpqqq`: b1000....00001000000000000000
+* `pqqq4u9s93jtgysm3mrwll70zr697y3mf902hvxwej0v7c62rsltw83ng0pu8w3j230sluc5gxkdmm9dvpy9y6ggtjd2w544mzdrcs42t7sq`: signature
+* `dkcy8h`: Bech32 checksum
 
-> # Same, but using invalid unknown feature 100
-> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9q4pqqqqqqqqqqqqqqqqqqszk3ed62snp73037h4py4gry05eltlp0uezm2w9ajnerhmxzhzhsu40g9mgyx5v3ad4aqwkmvyftzk4k9zenz90mhjcy9hcevc7r3lx2sphzfxz7
+> # Same, but adding invalid unknown feature 100
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdeessp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q4psqqqqqqqqqqqqqqqpqqqqu7fz6pjqczdm3jp3qps7xntj2w2mm70e0ckhw3c5xk9p36pvk3sewn7ncaex6uzfq0vtqzy28se6pcwn790vxex7xystzumhg55p6qq9wq7td
 
 Breakdown:
 
@@ -567,11 +601,14 @@ Breakdown:
 * `d`: short description
   * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20)
   * `vdhkven9v5sxyetpdees`: 'coffee beans'
+* `s`: payment secret
+  * `p5`: `data_length` (`p` = 1, `5` = 20; 1 * 32 + 20 == 52)
+  * `zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs`: 0x1111111111111111111111111111111111111111111111111111111111111111
 * `9`: features
   * `q4`: `data_length` (`q` = 0, `4` = 21; 0 * 32 + 21 == 21)
-  * `pqqqqqqqqqqqqqqqqqqsz`: b00001...(90 zeroes)...1000000010
-* `k3ed62snp73037h4py4gry05eltlp0uezm2w9ajnerhmxzhzhsu40g9mgyx5v3ad4aqwkmvyftzk4k9zenz90mhjcy9hcevc7r3lx2sp`: signature
-* `hzfxz7`: Bech32 checksum
+  * `psqqqqqqqqqqqqqqqpqqqq`: b000011000....00001000000000000000
+* `u7fz6pjqczdm3jp3qps7xntj2w2mm70e0ckhw3c5xk9p36pvk3sewn7ncaex6uzfq0vtqzy28se6pcwn790vxex7xystzumhg55p6qq`: signature
+* `9wq7td`: Bech32 checksum
 
 # Authors
 
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -367,6 +367,8 @@ fips
 rfc
 varint
 CompactSize
+multipath
+mpp
 tlvs
 snprintf
 GitHub


04-onion-routing.md
```diff
@@ -257,6 +257,9 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 8 (`basic_mpp`)
```
ZmnSCPxj: Indentation and numbering seem strange to me.

11-payment-encoding.md
```diff
@@ -277,6 +277,10 @@ The field is big-endian.  The least-significant bit is numbered 0,
 which is _even_, and the next most significant bit is numbered 1,
 which is _odd_.
 
+| Bits | Name        |Description            | Link                                           |
+|------|-------------|-----------------------|------------------------------------------------|
+| 0/1  | `basic_mpp` | Payee understands mpp.| [BOLT #4](04-onion-routing.md#basic-multi-part-payments) |
```
ZmnSCPxj: What would be the point of using the even bit?

For odd bits, it seems to me to mean, "payer *may* use `basic_mpp`".

For even bits, what is it mean? "Payer *must* use `basic_mpp`"?

04-onion-routing.md
pm47: Even types are used because those are mandatory fields, does that help?

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
 The reader:
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
 
-The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format).
+The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format)
+and [below](#basic-multi-part-payments).
+
+### Basic Multi-Part Payments
```
t-bast: I'm not sure this should be in Bolt 4. Apart from the TLV that signals support, all the AMP stuff isn't really related to onion encryption or routing. Maybe AMP should have its own Bolt? Not sure but if it becomes bigger (when we add more advanced AMP features) it will feel messy to have it here, won't it?

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+      - MUST use the same `payment_preimage` on all HTLCs in the set.
+      - MUST set `total_msat` to the amount it wishes to pay.
+      - SHOULD send all payments at approximately the same time.
+      - SHOULD try to use diverse paths to the recipient for each HTLC.
+      - SHOULD retry and/or re-divide HTLCs which fail.
+      - if the invoice specifies an `amount`:
+          - MUST set `total_msat` to at least that `amount`, and less
+            than or equal to twice `amount`.
+      - MUST ensure that the total `amount_msat` of the HTLC set which arrive at the payee
+        is at greater or equal to `total_msat`.
```
t-bast: nit: "is at greater or equal" should be "is greater or equal"

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+      - MUST ensure that no (incomplete) non-failed subset of the HTLC set adds
+        to a total `amount_msat` greater or equal to `total_msat`.
+
+The reader:
+  - if `basic_mpp` is present:
+    - MUST fail the HTLC if it would otherwise fail a single HTLC of value
+      `total_msat`, `payment_hash` or `ctlv_expiry`.
+    - otherwise:
+	  - MUST add it to the HTLC set corresponding to that `payment_hash`.
+	  - SHOULD fail the entire HTLC set if `total_msat` is not the same for
```
t-bast: Shouldn't we upgrade that "SHOULD" to a "MUST"? Is there a use-case where we want to allow senders to specify different `total_msat`s?

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+#### Rationale
+
+If `basic_mpp` is present it causes a delay to allow other partial
+payments to combine.  The total amount must be sufficient for the
+desired payment, just as it must be for single payments.
+
+Because invoices do not necessarily specify an amount, and because
+payers can add noise to the final amount, the total amount must be
+sent explicitly.  The requirements allow exceeding this slightly, as
+it simplfies adding noise to the amount when splitting, as well as
```
t-bast: nit: "simplfies" -> "simplifies"

11-payment-encoding.md
```diff
@@ -140,6 +141,9 @@ Currently defined tagged fields are:
    * `fee_base_msat` (32 bits, big-endian)
    * `fee_proportional_millionths` (32 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `9` (5): `data_length` variable. One or more bytes containing features
```
t-bast: Not related to that change, but it seems like Bolt 11 rolled out a custom TLV before TLV was cool.
Should we update Bolt 11 to use a TLV stream?

04-onion-routing.md
ZmnSCPxj: Numbering I am referring to is before `.` . I.e. `1. type` then `2. data` then `1. type`.

Also next line is more indent but starts at `2.`

rustyrussell: Yes, these are a list of separate TLV record types.

11-payment-encoding.md
rustyrussell: We always assign in pairs.  In future, it's theoretically possible you could set this if no single channel had sufficient capacity, I guess, but it's a bit weird!

04-onion-routing.md
ZmnSCPxj: Seems we should change the formatting somehow?  If you format like that, this is what Markdown will produce:

> 1. type
> 2. data
> 1. type
>    2. data

04-onion-routing.md
rustyrussell: Yeah, BOLT4 is where all the "node" semantics ended up, this just continues the trend.

Separating the handling into a new bolt is probably a good idea, but a fairly large and invasive job which I'd prefer to happen while we're not actively churning the spec!

04-onion-routing.md
rustyrussell: Well, it's not your problem if they do this: you can just ignore it if you want to.

I tend to reserve MUST for things which will either stop something from working, or cause a security issue.

11-payment-encoding.md
rustyrussell: Yeah, if we were doing it again we probably would.  But bolt11 is natively 5-bit units, so it's a little boutique.

rustyrussell:
Typos fixed, spellcheck made happy, rebased to solve conflict.

No non-formatting changes.

04-onion-routing.md
rustyrussell: Ah, yuck.  I missed that it misrenders.  Let me see if I can fix that somehow :(

rustyrussell: OK, I think the neatest change is to change them to '*' and fix the tooling to allow that:

```markdown
1. tlvs: `tlv_payload`
2. types:
    * type: 0 (`destination_signal`)
    * data:
	    * Empty
    * type: 2 (`amt_to_forward`)
    * data:
        * [`tu64`:`amt_to_forward`]
    * type: 4 (`outgoing_cltv_value`)
    * data:
        * [`tu32`:`outgoing_cltv_value`]
    * type: 6 (`short_channel_id`)
    * data:
    * type: 8 (`basic_mpp`)
	* data:
	    * [`tu64`:`total_msat`]
```

This needs to be tweaked to parse anyway, let's wait until #631 which fixes the tooling, then I'll add a fixup for this...

11-payment-encoding.md
t-bast: Ok we save more space than TLV by using only 5 bits, understood.
We can discuss that later to see if it's worth changing.

04-onion-routing.md
```diff
@@ -238,11 +233,124 @@ Field descriptions:
...
+
+The `amt_to_forward` value will be the amount for this partial payment
+only.  The `basic_mpp` flag is a promise by the ultimate sender that
+the rest of the payment will follow in succeeding HTLCs; we call these
+HTLCs, which have the same preimage, an "HTLC set".
+
+#### Requirements
+
+The writer:
+  - MUST not include `basic_mpp` for any non-final node.
```
flack: `not` should be uppercase

ysangkok: why not? It is RFC 2119.

rustyrussell:
More typo fixes...

Roasbeef:
Can be rebased now that the onion stuff has been merged in.

11-payment-encoding.md
Roasbeef: This is already covered for regular AMP, but for regular MPP, I think we should also include a random identifier here and require that the sender include this for end-to-end security (in the onion). Otherwise, if it's a 0-value invoice, then intermediate nodes can probe, and then use that information to possibly take most/all of the HTLC as "fees".

11-payment-encoding.md
Roasbeef: We may also want to add an additional section here w.r.t path finding advisory using routing hints. If the amount of the invoice is greater than available inbound bandwidth of any of their channels, then they can "guide" the sender with regular or "special" routing hints to increase the likelihood of success.

04-onion-routing.md
```diff
@@ -257,6 +257,9 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 8 (`basic_mpp`)
+	2. data:
+	    * [`tu64`:`total_msat`]
```
Roasbeef: I suggest a minimal two-tuple of `(amt, randID)` as mentioned in my earlier comment.

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+  - MUST not include `basic_mpp` for any non-final node.
+  - if `basic_mpp` feature was not set in the invoice:
+    - MUST NOT include `basic_mpp` for the final node.
+  - otherwise:
+    - MAY include `basic_mpp` for the final node.
+    - if it does include `basic_mpp`:
+      - MAY send more than one HTLC to pay the invoice.
+      - MUST use the same `payment_preimage` on all HTLCs in the set.
+      - MUST set `total_msat` to the amount it wishes to pay.
+      - SHOULD send all payments at approximately the same time.
```
Roasbeef: This conflicts w/ two bullets below (try to re-derive if they fail). 

We should also have a section below here detailing failure modes due to _partial_ settles, as it's possible that some portion of the funds just never get to the receiver, and become windfall for intermediate nodes.

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+    - MUST fail the HTLC if it would otherwise fail a single HTLC of value
+      `total_msat`, `payment_hash` or `ctlv_expiry`.
+    - otherwise:
+	  - MUST add it to the HTLC set corresponding to that `payment_hash`.
+	  - SHOULD fail the entire HTLC set if `total_msat` is not the same for
+	    all HTLCs in the set.
+    - if the total `amount_msat` of this HTLC set equals or exceeds `total_msat`:
+      - SHOULD fulfill all HTLCs in the HTLC set
+    - otherwise:
+      - SHOULD wait for at least 60 seconds for remaining members of the HTLC set.
```
Roasbeef: A static value here isn't very encompassing. It may be the case that a higher level application has the sender purposefully delay the remaining shards.

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+
+#### Rationale
+
+If `basic_mpp` is present it causes a delay to allow other partial
+payments to combine.  The total amount must be sufficient for the
+desired payment, just as it must be for single payments.
+
+Because invoices do not necessarily specify an amount, and because
+payers can add noise to the final amount, the total amount must be
+sent explicitly.  The requirements allow exceeding this slightly, as
```
Roasbeef: If we don't add a mandatory random identifier along the way, then we should _discourage_ sending more than the invoice amount.

11-payment-encoding.md
t-bast: Great point, there is indeed an attack vector here for the next-to-last node!
Your suggested fix assumes that the payment request will stay hidden from the intermediate nodes though (otherwise it doesn't fix anything if the intermediate nodes also know this random identifier).
It could be important to highlight this point to applications that want to leverage MPP: using MPP for donations should be a two-steps process, where a user who wants to donate needs to go to your website to get an invoice generated for him (and only him). That invoice must not be shared publicly.

04-onion-routing.md
t-bast: > We should also have a section below here detailing failure modes due to partial settles, as it's possible that some portion of the funds just never get to the receiver, and become windfall for intermediate nodes.

Can you explain what you mean? I didn't get that.

04-onion-routing.md
t-bast: True but I think the receiver should enforce a maximum duration for a receive session, otherwise a malicious sender can lock funds from a lot of people almost indefinitely (if what you were thinking of was some kind of sliding window on the receiver end where you reset the timeout each time you receive a valid share).

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+        to a total `amount_msat` greater or equal to `total_msat`.
+
+The reader:
+  - if `basic_mpp` is present:
+    - MUST fail the HTLC if it would otherwise fail a single HTLC of value
+      `total_msat`, `payment_hash` or `ctlv_expiry`.
+    - otherwise:
+      - MUST add it to the HTLC set corresponding to that `payment_hash`.
+      - SHOULD fail the entire HTLC set if `total_msat` is not the same for
+        all HTLCs in the set.
```
joostjager: Does this mean that an adversary can prevent an amp payment from completion by periodically sending a very small shard with a random `total_msat`, thereby cancelling any shards that have been acquired?

joostjager: I guess those htlcs are cancelled early in the process because their `total_msat` doesn't match what is in the invoice database. How would it work in that case for zero-valued invoices? First htlc that arrives sets the amount?

04-onion-routing.md
joostjager: Could it be useful to specify the hold timeout in the onion payload, so that the sender picks this value themselves?

04-onion-routing.md
joostjager: Shouldn't `randID` be defined as a separate type, so that it can be used to pay non-amp to zero-valued invoices?

11-payment-encoding.md
```diff
@@ -253,6 +268,19 @@ engines that support SQL or other dynamically interpreted querying languages.
...
+Feature bits allow forward and backward compatibility, and follow the
+_it's ok to be odd_ rule.
+
+The field is big-endian.  The least-significant bit is numbered 0,
+which is _even_, and the next most significant bit is numbered 1,
+which is _odd_.
+
+| Bits | Name        |Description            | Link                                           |
+|------|-------------|-----------------------|------------------------------------------------|
+| 0/1  | `basic_mpp` | Payee understands M.P.P.| [BOLT #4](04-onion-routing.md#basic-multi-part-payments) |
```
joostjager: Will there also be a global feature bit that advertises mpp in the graph, so that spontaneous AMP payments can be made too?

I think that in that case the invoice flag is still required, because the payer may not have the (up to date) node features of the payee? Does this mean that `basic_mpp` implies `var_onion_optin`?

04-onion-routing.md
rustyrussell: Yeah, we need to pick something, and it can't really be up to the sender or recipient since it effects the entire network.  Implementations *will* pick something, so might as well specify it.

04-onion-routing.md
rustyrussell: You're right, will change to ignore & fail if total_msat is different to avoid griefing.

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+      - MUST use the same `payment_preimage` on all HTLCs in the set.
+      - MUST set `total_msat` to the amount it wishes to pay.
+      - SHOULD send all payments at approximately the same time.
+      - SHOULD try to use diverse paths to the recipient for each HTLC.
+      - SHOULD retry and/or re-divide HTLCs which fail.
+      - if the invoice specifies an `amount`:
+          - MUST set `total_msat` to at least that `amount`, and less
+            than or equal to twice `amount`.
+      - MUST ensure that the total `amount_msat` of the HTLC set which arrive at the payee
+        is greater or equal to `total_msat`.
```
joostjager: What is the rationale for allowing the total htlc set `amount_msat` to be greater than `total_msat`? And should this be capped too, to be consistent with single path payments? Otherwise the following would be possible (if I understood this correctly): invoice amt: 1000 sat, `total_msat`: 1000 sat, first htlc: 500 sat, second htlc: 1 btc?

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+      - MUST ensure that no (incomplete) non-failed subset of the HTLC set adds
+        to a total `amount_msat` greater or equal to `total_msat`.
```
joostjager: What does this mean exactly? If the sender has enough htlcs in flight to pay the invoice but the htlcs are not settled, they shouldn't add another htlc?

04-onion-routing.md
joostjager: We must make sure that an attacker can't cancel other shards, but also that they can't block the genuine shards from being accepted. If the attacker is the first to set the tracked `total_msat` to a random value, they could possibly block the payment for the timeout duration (and repeat this thereafter).

joostjager: In that light, the proposed `randID` may actually be required not only to prevent fee stealing with 0-valued invoices, but also to keep an attacker from interfering with the multi path settle process.

04-onion-routing.md
joostjager: In addition to the seconds timeout, does something need to be specified in terms of block height? It could be that the first htlc comes in, meets the invoice cltv delta requirement, but its value isn't high enough to settle the invoice. The htlc is held. Then a block comes in. At that point, the cltv delta requirement isn't met anymore. Should the htlc be canceled?

04-onion-routing.md
t-bast: Exactly that's it. If you've sent an HTLC set that pays the total amount, you should wait for (some of) them to fail or settle. If you send another HTLC you will end up paying more than what you intended to pay.

04-onion-routing.md
joostjager: Or borrow the `stream_id` from #658 to keep the htlc sets apart. The first set that gets all its htlcs in is settled, other sets are cancelled.

joostjager: Or borrow the stream_id from #658 to keep the htlc sets apart. The first set that gets all its htlcs in is settled, other sets are cancelled.

joostjager: One thing that the receiver-generated `randID` prevents and the sender-generated not, is the attack that @t-bast described: the penultimate node can find out whether they are the penultimate node by crafting a new single hop route instead of forwarding as they should. If the newly crafted route succeeds, they know that the next node is the final destination. They still get the preimage and can pull their incoming htlc.

The attacker can generate a new sender-generated `randID` for this. But if the receiver would match on a receiver-generated `randID`, the attack wouldn't work. 

This only applies to regular, non-mpp payments I think.

joostjager: >The first set that gets all its htlcs in is settled, other sets are cancelled.

This is contradictory to being able to pay to settled invoices (to prevent leaking whether an invoice is paid or not). We could cancel all other sets, mark the invoice as settled, but if a new set comes in after that, we'd accept it again. Then we could just as well settle all sets that meet the requirements always.

joostjager: In the penultimate node attack-scenario above, I think they can also still do fee sniping on zero valued mmp invoices. The attacker crafts a new route with a new, lower `totalAmt` (single shard), and buys the preimage cheaply.

joostjager: >IIUC the attack isn't limited to the penultimate node, just easier because they have less guess on the amount, right?

Yes, sounds like it. @t-bast was there anything specific in the second-to-last case?

t-bast: Correct, anyone in the route could probe like that, but it's a lot easier for the penultimate node: it only has to test the next node in the route (which it knows because it is in the onion).

If you are the second-to-last node, you need to have the intuition of which node to probe, which is a lot harder in practice (unless you already suspect who the recipient is).

joostjager: Ok, thanks for that clarification! What about calling it "invoice secret"? (is this a better term than 'receiver-generated randid`?)

joostjager: Actually, we started calling this `payment_address` which is more descriptive in the context of #658.

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+all paths.
+
+The `amt_to_forward` value will be the amount for this partial payment
+only.  The `basic_mpp` flag is a promise by the ultimate sender that
+the rest of the payment will follow in succeeding HTLCs; we call these
+HTLCs, which have the same preimage, an "HTLC set".
+
+#### Requirements
+
+The writer:
```
ariard, line 296: nit: Not introduced here but I found the writer/reader wording a bit weird outside bolt-11 shouldn't we stick with sending node/receiving node?

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+  - MUST not include `basic_mpp` for any non-final node.
+  - if `basic_mpp` feature was not set in the invoice:
+    - MUST NOT include `basic_mpp` for the final node.
+  - otherwise:
+    - MAY include `basic_mpp` for the final node.
+    - if it does include `basic_mpp`:
+      - MAY send more than one HTLC to pay the invoice.
+      - MUST use the same `payment_preimage` on all HTLCs in the set.
```
ariard: Maybe we should detail all set HTLCs have same final_ctlv  ? I mean the `expiry` field from invoice should scope all HTLC buddies but they will have different first_cltv due to different cltv_delta on their mutual paths.

04-onion-routing.md
ariard: Agree better to specify in terms of block height. If CLTV height is shared between all HTLCs, you should be able to wait until it minus some block height delta to land an unilateral closing.

Also with current wording it's not clear what receiving node should do after delay expiration, back to step upper ("SHOULD fulfill ...") or go to the one below ("MUST fail ...")

04-onion-routing.md
```diff
@@ -269,7 +272,81 @@ The writer:
...
+scenarios in which the senders are genuinely independent (friends
+splitting a bill, for example).
+
+The subset restriction prevents the preimage being released before all
+the partial payments have arrived: that would allow any intermediate
+node to immediately claim any outstanding partial payments.
+
+An implementation may choose not to fulfill an HTLC set which
+otherwise meets the amount criterion (eg. some other failure, or
+invoice timeout), however if it were to fulfill only some of them,
```
ariard, line 354: In the same way, if you try to pull some HTLCs of the set before all of them are locked on final_hop channels it may be a risk? Let's say if you pick up different paths and some nodes share the same operator, if you provide preimage for HTLC#1 but HTLC#3 isn't locked yet, preimage could be replay on HTLC#3 first links without payment being forwarded until payee. This point should be better underscored.

04-onion-routing.md
rustyrussell: No strict subset.  This has been a source of confusion, though, so we need to reword it.

04-onion-routing.md
rustyrussell: These are routed through multiple nodes though, so it's best to be clear.

04-onion-routing.md
rustyrussell: This fails in the case of independent payers (eg. bill splitting), and it also restricts fuzzing on re-attempts.  You now have to decide the total fuzz up-front, and thus leak more information if you have to split a payment since it will always add to the previous amount exactly.

And note that overpaying is consistent with existing payments, requiring exact payment would not be.

04-onion-routing.md
rustyrussell: > Agree better to specify in terms of block height. If CLTV height is shared between all HTLCs, you should be able to wait until it minus some block height delta to land an unilateral closing.

No, you're tying up channels across the network.  We're not going to hold things for many minutes before this becomes a spam attack.  Certainly not waiting for blocks!

> Also with current wording it's not clear what receiving node should do after delay expiration, back to step upper ("SHOULD fulfill ...") or go to the one below ("MUST fail ...")

Yes, it should be clear that you must fail them.

04-onion-routing.md
rustyrussell: There's no reason I can think of to specify this though?

11-payment-encoding.md
rustyrussell: Yes, and yes.  I'll add that specifically.

04-onion-routing.md
```diff
@@ -248,18 +248,104 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format)
+and [below](#basic-multi-part-payments).
+
+### Basic Multi-Part Payments
+
+An HTLC may be part of a larger "multi-part" payment: such
+"base" atomic multipath payments will use the same `payment_hash` for
+all paths.
+
+Note that `amt_to_forward` is the amount for for this HTLC only: a
```
t-bast:
```suggestion
Note that `amt_to_forward` is the amount for this HTLC only: a
```

04-onion-routing.md
```diff
@@ -248,18 +248,104 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+HTLCs; we call these outstanding HTLCs which have the same preimage,
+an "HTLC set".
 
-The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format).
+#### Requirements
+
+The writer:
+  - if the invoice offers the `basic_mpp` feature:
+    - MAY send more than one HTLC to pay the invoice.
+    - MUST use the same `payment_preimage` on all HTLCs in the set.
```
t-bast:
```suggestion
    - MUST use the same `payment_hash` on all HTLCs in the set.
```

04-onion-routing.md
```diff
@@ -248,18 +248,104 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+    - if the invoice specifies an `amount`:
+       - MUST set `total_msat` to at least that `amount`, and less
+         than or equal to twice `amount`.
+    - otherwise:
+      - MUST set `total_msat` to the amount it wishes to pay.
+    - MUST ensure that the total `amount_msat` of the HTLC set which arrives at the payee
+      is greater or equal to `total_msat`.
+    - MUST NOT send another HTLC if the total `amount_msat` of the HTLC set is already greater or equal to `total_msat`.
+  - otherwise:
+    - MUST set `total_msat` equal to `amt_to_forward`.
```
t-bast: It seems to me that if the invoice doesn't support the `basic_mpp` feature, we shouldn't send this _even_-numbered tlv. IMO the invoice feature bit is the signal that this tlv field is understood.

04-onion-routing.md
ariard: IIRC, we used "processing node" in other parts of the spec to name intermediary nodes

04-onion-routing.md
ariard: Having same final_ctlv encumbering every HTLC of the set should avoid one being expired and timeout while others being fulfilled due to misbehavior and so a partial payment.. But would require doing the same on whole multiple paths which is not realistic..

04-onion-routing.md
```diff
@@ -248,18 +248,104 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 8 (`total_msat`)
+    2. data:
+        * [`tu64`:`total_msat`]
```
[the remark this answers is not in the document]
joostjager: Today t-bast and I discussed the tlv payload a bit more with the idea in mind to consider the receiver id as a more generic `payment_address` (taken from #658). This is a property that has meaning for any type of payment. It is like a box on the destination node in which the payment is dropped. So the destination of a Lightning payment would consist of a node pubkey and a payment address. 

Depending on the type of payment, there may be additional fields required. For an mpp payment there is a specific `payment_hash` that needs to be used on the htlc to get a proof of payment for the address that is being paid to. For an amp payment, there is a `set_id` and `shares`.

For mpp, the `payment_address` is single use and needs to be generated randomly. But for amp or a different form of spontaneous payment, it could actually be something that is derived from other data. For example, the hash of a user name. Users could have their own deposit boxes on a node and people donating don't need an invoice to donate to a specific user. They just include the user name hash as the payment address in the payload.

As we don't know yet exactly how this `payment_address` will be used, it is probably safer to use our standard 32 byte length for it. If it will be used with a hash, it will provide sufficient collision resistance. And there may also be other forms of composition that benefit from having this standard size.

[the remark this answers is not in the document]
t-bast: tl;dr would be to change the tlv to:

1. type: 8 (`basic_mpp`)
2. data:
    * [`32*byte`:`payment_addr`]
    * [`u64`:`total_msat`]

And add this `payment_addr` to the invoice (needs some changes in Bolt 11 similar to the ones I made in #671)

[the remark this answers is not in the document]
rustyrussell: I assumed we'd add some nonce on top of this, yes.

But for my "offers" proposal (still, mainly in my head!), the nonce would be a merkle tree, so 32-bytes is a nice number.  In simplest incarnation (base AMP), it's provided in the invoice.

I'll add it.

04-onion-routing.md
rustyrussell: Yes, since people have already implemented TLV handling pre-mpp, we should make it odd.

04-onion-routing.md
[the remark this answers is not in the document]
t-bast: Good point, 32 bytes is interesting because it fits a merkle root which opens the door for many applications.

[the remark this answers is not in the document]
rustyrussell: I made it variable, because why not?

Rebased your PR at the end with minor changes to make it fit.

04-onion-routing.md
```diff
@@ -245,6 +245,9 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 2 (`amt_to_forward`)
     2. data:
         * [`tu64`:`amt_to_forward`]
+    1. type: 3 (`payment_secret`)
```
rustyrussell: This should be even, as we agreed to it in advance.

rustyrussell:
OK, I applied [proposed changes](http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-09-30-20.03.html) as a series of separate fixups for easy review.

I'll squash them together at the end once we're all happy.

t-bast:
After playing quite a bit with my implementation, I'm happy with the current state of the proposal. Are there pending comments to address?

The only thing I'd like to discuss is the invoice feature bits. I'm leaning more and more towards making it compatible with local/global/flat/unified/??? feature bits (so we'd put those feature bits in Bolt 9 and make sure they don't overlap with other feature bits). WDYT?

Roasbeef:
Direction seems to be: a feature bit to signal that you support the feature, then make the TLV fields themselves required. As a result, since they're now required, we can now just combine the fields into a single type. The suggested ordering is: `secret || amount`, since the the `secret` is fixed sized, we can truncate the size of the amount saving more space.

t-bast:
SGTM. The new TLV field should be something like:

```
1. type: 8 (`multipart`)
2. data:
    * [`tu64`:`total_msat`]
```

I do think the two invoice feature bits pairs should stay separate though. You may want the invoice to say that you only accept whole payments (`amount_to_forward` == `total_amount`) but you require a `payment_secret` to guard against probing. I don't see how we could do that if we have a single feature bits pair.

rustyrussell:
OK, as agreed at meeting, I used standard feature bits for BOLT11 features.  That means an explicit "basic_mpp" feature bit is required, though I overloaded the "var_onion_optin" feature a little: if it's even you *have* to use a secret.

09-features.md
```diff
@@ -27,6 +27,7 @@ These flags may only be used in the `init` message:
 | 4/5   | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel                   | [BOLT #2][bolt02-open]       |
 | 6/7   | `gossip_queries`                 | More sophisticated gossip control                                         | [BOLT #7][bolt07-query]      |
 | 10/11 | `gossip_queries_ex`              | Gossip queries can include additional information                         | [BOLT #7][bolt07-query]      |
+| 12/13 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | [BOLT #3][bolt04-mpp]               |
```
cfromknecht: i've been looking at this and wondering, what does it meant to _require_ a multi-path payment. if the sender can get there in one shard, why would the receiver require them to use multiple paths? even then, the granularity would be 1 htlc or >1 htlc, which doesn't seem very useful.

i'm starting to think we don't _really_ need a bit to indicate whether we can or cannot assemble multiple htlcs, and that this could just indicate optional/required for sending the `payment_data` record. most of complexity in mpp is on the sender side anyway, so idk how much we would be saving people by allowing them to choose between 1 and >1 on the receiver-side.

12/13 is already `option_static_remotekey`

11-payment-encoding.md
```diff
@@ -202,6 +203,8 @@ A reader:
   description.
   - if a valid `n` field is provided:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
+  - if there is a valid `s` field:
```
cfromknecht: gate with `basic_mpp` feature bit?

11-payment-encoding.md
```diff
@@ -278,6 +281,35 @@ The field is big-endian.  The least-significant bit is numbered 0,
 which is _even_, and the next most significant bit is numbered 1,
 which is _odd_.
 
+Features are listed in [09-features.md](BOLT #9) [FIXME: Or will be after
+flattened features is merged!].
+
+Note that the `var_onion_optin` feature is slightly overloaded to indicate
+support for the `s` field: making this compulsory prevents probing attacks
```
cfromknecht: would be in favor of not overloading `var_onion_optin` with the `s` field. there's a huge gap between adding TLV onion and actually being able to implement this proposal, perhaps they should be separate?

09-features.md
t-bast: I have some additional thoughts too. The reason we initially split the `payment_secret` and `multi_part` feature bits was to allow light payment receivers to require a `payment_secret` while not supporting multi-part. I still think this is a valid scenario and I think we can't achieve it now that there's no feature bit for `payment_secret` (unless we use `var_onion_optin` to implicitly do that as suggested, but I'm not a huge fan).

Receiving multi-part payments means you have to keep some state (partial HTLC set received) until either the payment succeeds or times out. It's hard to guess what other people will want to do in the future, but I can imagine a tiny payment processor in a vending machine that wouldn't want this overhead and would only support single-part payment, while enforcing the `payment_secret`.

I think our feature bit set should allow that kind of tuning, so I think it would be better to have a pair of feature bits for multi-part (14/15, even though the even one wouldn't make much sense) and a pair of feature bits for payment secret (16/17). This way a very constrained payment processor chip can set 16 without setting 14 nor 15.

Do you think this is overkill?

11-payment-encoding.md
t-bast: There's a huge gap between implementing `var_onion_optin` and implementing MPP. But implementing support for payment secrets is really simple, right?
So making `var_onion_optin` compulsory to imply that you need to send the payment secret (potentially for existing single-part payments) seems acceptable, even though I'd prefer a dedicated feature bit to make it more explicit (but at the cost of more feature bits in the system...).

09-features.md
joostjager: I think if you look at the total processing power that is required for a Lightning node, the tracking of a multi-path set is not significant. I'd rather add such specialized feature bit only when that "light receiver" actually presents itself.

04-onion-routing.md
```diff
@@ -248,18 +251,110 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+  - For the final node:
+    - MUST NOT include `short_channel_id`
+    - MUST include `total_msat`
+    - if the recipient provided `payment_secret`:
+	  - MUST include that `payment_secret`
 
 The reader:
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
+  - if it is the final node:
+    - MUST treat `total_msat` as if it were equal to `amt_to_forward` if it
```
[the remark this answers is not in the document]
rustyrussell: I thought LND and Eclair already implemented TLV without this?  In which case, this applies.

04-onion-routing.md
```diff
@@ -248,18 +248,111 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+  - otherwise:
+    - MUST set `total_msat` equal to `amt_to_forward`.
+
+The final node:
+  - MUST fail the HTLC if dictated by Requirements under [Failure Messages](#failure-messages)
+    - Note: "amount paid" specified there is the `total_msat` field.
+  - if it does not support `basic_mpp`:
+    - MUST fail the HTLC if `total_msat` is not exactly equal to `amt_to_forward`.
+  - otherwise, if it supports `basic_mpp`:
+    - MUST add it to the HTLC set corresponding to that `payment_hash`.
```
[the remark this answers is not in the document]
rustyrussell: The first requirement here is "MUST fail the HTLC if dictated by Requirements under [Failure Messages](#failure-messages)" which covers that and similar?

It's a bit awkward, but I think wholesale reformatting of the requirements should be done separately.

09-features.md
rustyrussell: If you don't have incoming capacity in a single channel, you might indicate you NEED MPP, maybe?

I think keeping the features separate is cleaner.

11-payment-encoding.md
rustyrussell: It is gated on the writer side.  But what's a reader to do if the `s` is set and mpp isn't, or vice versa?  I chose to keep it simple: if `s` is there, use it.

04-onion-routing.md
```diff
@@ -248,18 +248,111 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+        * [`32*byte`:`payment_secret`]
+        * [`tu64`:`total_msat`]
 
 ### Requirements
 
 The writer:
-  - MUST include `amt_to_forward` and `outgoing_cltv_value` for every node.
-  - MUST include `short_channel_id` for every non-final node.
-  - MUST NOT include `short_channel_id` for the final node.
+  - Unless `node_announcement` contains `var_onion_optin` or the [BOLT #11](11-payment-encoding.md#tagged-fields) invoice contains an `s` field:
```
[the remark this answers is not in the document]
rustyrussell: Sure, but you can't add payment_data unless you have the payment_atgnwt!

t-bast:
Sorry for missing yesterday's meeting, here are my updated comments on the points raised yesterday:

- naming: `payment_nonce` was my first choice, but since `n` was already assigned in Bolt 11 I changed to `payment_secret` (which is somewhat misleading because it's not entirely secret), but that can be named `payment_nonce` and use something else than `n` in the invoice. Other options: `payment_id`, `invoice_id`?
- splitting in two features instead of overloading `var_onion_optin` (`payment_thing` and `mpp`): thanks, I really wanted that separation to happen ;)

I'll get these latest changes implemented in Eclair today, will update here if anything looks weird.

t-bast:
One question about the encoding of `payment_data` (specifically the `tu64` part).
If encoding a payment secret `0x1111111111111111111111111111111111111111111111111111111111111111` and a `total_msat` of 1105 (0x0451), is the result:

```
0x082211111111111111111111111111111111111111111111111111111111111111110451
```
or
```
0x08221111111111111111111111111111111111111111111111111111111111111111020451
```

(ie is `tu64` explicitly prefixed by its length or implicitly by the length of the whole TLV record?)

I prefer the second option because it's more composable (you can have multiple `tu64` in the same TLV record) whereas the first one feels hackish (especially to only save 1 byte).

cfromknecht:
@t-bast currently as specified it’s the former, so the length is implied by the size of the record. indeed we could have a separate prefix-truncated uint (`ptu64`?) encoding as you suggest when multiple variable length values are being encoded. Note that for the current application `ptu64` would only save one byte over just making a whole new record, so it’s not necessary in this case but could be useful in the future

rustyrussell:
> Sorry for missing yesterday's meeting, here are my updated comments on the points raised yesterday:
> 
>     * naming: `payment_nonce` was my first choice, but since `n` was already assigned in Bolt 11 I changed to `payment_secret` (which is somewhat misleading because it's not entirely secret), but that can be named `payment_nonce` and use something else than `n` in the invoice. Other options: `payment_id`, `invoice_id`?

payment_handle?

>     * splitting in two features instead of overloading `var_onion_optin` (`payment_thing` and `mpp`): thanks, I really wanted that separation to happen ;)

Yep, done.

rustyrussell:
> One question about the encoding of `payment_data` (specifically the `tu64` part).
> If encoding a payment secret `0x1111111111111111111111111111111111111111111111111111111111111111` and a `total_msat` of 1105 (`0x0451`), is the result:
> 
> ```
> 0x08 22 1111111111111111111111111111111111111111111111111111111111111111 0451
> ```
> 
> or
> 
> ```
> 0x08 23 1111111111111111111111111111111111111111111111111111111111111111 020451
> ```
> 
> (ie is `tu64` explicitly prefixed by its length or implicitly by the length of the whole TLV record?)
> 
> I prefer the second option because it's more composable (you can have multiple `tu64` in the same TLV record) whereas the first one feels hackish (especially to only save 1 byte).

tu64 is defined as the former.  Your argument extends to "just add a separate tlv", since this only saves one byte :)  Turns out a single value is a common case, worth optimizing IMHO.

t-bast:
> tu64 is defined as the former. Your argument extends to "just add a separate tlv", since this only saves one byte :) Turns out a single value is a common case, worth optimizing IMHO.

Got it, I feel it's a bit overkill that we optimize this for at most a one byte gain per TLV record against more edge cases to handle in the encoder, but if you both feel it's worth it I'll adapt ;)

11-payment-encoding.md
```diff
@@ -278,6 +281,37 @@ The field is big-endian.  The least-significant bit is numbered 0,
 which is _even_, and the next most significant bit is numbered 1,
 which is _odd_.
 
+Features are listed in [09-features.md](BOLT #9) [FIXME: Or will be after
+flattened features is merged!].
+
+Note that the `payment_secret` feature probing attacks from nodes
```
t-bast:
```suggestion
Note that the `payment_secret` feature prevents probing attacks from nodes
```

rustyrussell:
> > tu64 is defined as the former. Your argument extends to "just add a separate tlv", since this only saves one byte :) Turns out a single value is a common case, worth optimizing IMHO.
> 
> Got it, I feel it's a bit overkill that we optimize this for at most a one byte gain per TLV record against more edge cases to handle in the encoder, but if you both feel it's worth it I'll adapt ;)

In particular, it gives us back the bytes in the onion which TLV costs us, making the trade-off more palatable.

rustyrussell:
Trivial squash and rebase on master.

04-onion-routing.md
```diff
@@ -248,18 +248,111 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+    - MUST set `total_msat` equal to `amt_to_forward`.
+
+The final node:
+  - MUST fail the HTLC if dictated by Requirements under [Failure Messages](#failure-messages)
+    - Note: "amount paid" specified there is the `total_msat` field.
+  - if it does not support `basic_mpp`:
+    - MUST fail the HTLC if `total_msat` is not exactly equal to `amt_to_forward`.
+  - otherwise, if it supports `basic_mpp`:
+    - MUST add it to the HTLC set corresponding to that `payment_hash`.
+    - SHOULD fail the entire HTLC set if `total_msat` is not the same for
```
cfromknecht: another thing that came up for us when implementing the receiver logic: let's say one mpp HTLC has arrived, but is lower than `total_msat`. then a legacy payment arrives settling the full amount, should the receiver settle the legacy payment and cancel the mpp HTLC? or keep waiting for mpp?

it could be that the payment_secret is optional and the sender doesn't understand it, but rejecting the legacy means it's actually compulsory. tbh this situation only seems totally resolvable once it is fully required, but i wonder, is one of the approaches better than the other?

04-onion-routing.md
```diff
@@ -248,18 +248,111 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
...
+be reasonably bounded to avoid a denial-of-service.
+
+Because invoices do not necessarily specify an amount, and because
+payers can add noise to the final amount, the total amount must be
+sent explicitly.  The requirements allow exceeding this slightly, as
+it simplifies adding noise to the amount when splitting, as well as
+scenarios in which the senders are genuinely independent (friends
+splitting a bill, for example).
+
+The restriction on sending an HTLC once the set is over the agreed total prevents the preimage being released before all
```
cfromknecht: nit: long line

04-onion-routing.md
t-bast: Honestly this looks like something that will never happen: IIUC you're assuming that two different people are trying to pay the same invoice, right?
Because if Alice sent the first mpp HTLC, there's no reason for her to send a legacy payment next. So it has to come from someone else?

If that's the scenario you're exploring, I believe the legacy payment "wins". It's the first payment that pays the full amount, so it should receive the preimage.

I think this raises an interesting point about proof-of-payment. If multiple people are collaborating to pay an invoice with MPP, they currently all receive the preimage, but they didn't pay the whole thing. The proof-of-payment is now weaker than before: it only says that you paid part of that invoice. Is that a problem? Not sure.

cfromknecht: > Because if Alice sent the first mpp HTLC, there's no reason for her to send a legacy payment next. So it has to come from someone else?

An intermediary can see that the first MPP HTLC is stuck being held, then start probing with regular payments. In the case of a zero value invoice, they could settle the invoice for a value much lower than `total_msat`, then settle back and take the difference.

Until we have mandatory payment secrets, I think it's actually safer to reject legacy payments if you have already started constructing an MPP HTLC set, that's how we have it implemented in our draft. Allowing legacy to override would reintroduce the probing vector afaict.

cfromknecht: > I think this raises an interesting point about proof-of-payment. If multiple people are collaborating to pay an invoice with MPP, they currently all receive the preimage, but they didn't pay the whole thing. The proof-of-payment is now weaker than before: it only says that you paid part of that invoice. Is that a problem? Not sure.

More evidence that we should stop calling preimages proofs of payment, because they aren't :)

t-bast: > An intermediary can see that the first MPP HTLC is stuck being held, then start probing with regular payments.

Got it. However, I think there's another way to resolve that. In my opinion if the client is doing MPP (sending multiple shards), then it must include the payment secret. This prevents the probing you're mentioning.
And in that case the receiver simply needs to enforce that other HTLCs for that payment_hash include the payment_secret.

It's really hard to express correctly the intent with feature bits though. In my mind, when the invoice contains `basic_mpp_optional || payment_secret_optional`, this means:

- I allow legacy payments (single HTLC without `payment_secret`)
- Or MPP payments but in that case the `payment_secret` is required

WDYT?

09-features.md
```diff
@@ -35,6 +35,8 @@ The Context column decodes as follows:
 | 8/9   | `var_onion_optin`                | Requires/supports variable-length routing onion payloads  | IN       | [Routing Onion Specification][bolt04] |
 | 10/11 | `gossip_queries_ex`              | Gossip queries can include additional information         | IN       | [BOLT #7][bolt07-query]               |
 | 12/13| `option_static_remotekey`         | Static key for remote output                              | IN       | [BOLT #3](03-transactions.md)         |
+| 14/15 | `payment_secret`                 | Node supports `payment_secret` field                      | IN9      | [Routing Onion Specification][bolt04]   |
+| 16/17 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | [BOLT #3][bolt04-mpp]                 |
```
cfromknecht: what is the rationale for setting this in these in the init message? so that they might be used for informing channel funding?

rustyrussell: Yeah; since node_announcements are not 100% reliable, this lets you tell about the direct neighbors at least.

04-onion-routing.md
rustyrussell: This is fixed by requiring the payment secret (see #712 which is a separate PR since this was already accepted at the meeting).

In this case, if the legacy payment arrives first, it's accepted (assuming that you haven't made payment_secret compulsory yet!).  If mpp arrives first, the secret requirement is now locked in, which is a security win.

> More evidence that we should stop calling preimages proofs of payment, because they aren't :)

They absolutely *are*.  They are proof that payment was made.  They *don't* prove who paid it, which indeed sucketh!

rustyrussell:
OK, @t-bast has completed interoperability testing between Eclair and c-lightning (successfully, though he discovered some UX suckage as we only allow making such payment manually). As a result, the PR in c-lightning is now merged into master (https://github.coAm/ElementsProject/lightning/pull/3335).

Normally, this would mean I would now merge this PR into the spec, *however* I think we should delay pending @cfromknecht since there are outstanding questions?

04-onion-routing.md
cfromknecht: > This is fixed by requiring the payment secret (see #712 which is a separate PR since this was already accepted at the meeting).

Chatted on IRC w/ @rustyrussell, indeed #712 prevents this along with the receiver rejecting legacy payments if they advertised required payment secrets (which is already in this pr).

11-payment-encoding.md
cfromknecht: sgtm!

11-payment-encoding.md
cfromknecht: still unresolved?

09-features.md
cfromknecht: sounds good, it can't hurt to do so

11-payment-encoding.md
rustyrussell: Ah, this was fixed in #712 which reworks this paragraph anyway.  Let me roll it in here, then I can rebase #712

rustyrussell:
Merged after successful @t-bast  Eclair <-> clightning interop test.  :cake: :fireworks:

PR #645, opened 2019-07-17
BOLT 7: update interpretation of `complete` field within query messages

wpaulino:
The current interpretation seems to have been copied over from reply_short_channel_ids_end. That message should only be sent once at the end, so the interpretation there makes sense. This is not the case for reply_channel_range however, since multiple of these messages can be sent. When responding to a query_channel_range, if the response spans multiple messages and the message being sent is not the last one, the field should be `0`, otherwise `1`.


rustyrussell:
Yeah, it's a bit useless in practice.  The idea is that if complete == 0, you should definitely query other peer(s), but in practice it's not that useful.

But this message only occurs at the end, so no need to repurpose the field.

wpaulino:
> But this message only occurs at the end, so no need to repurpose the field.

Ah yes, that was my mistake. The current interpretation for `reply_short_channel_ids_end` seems correct indeed, but I don't think this should also apply to `reply_channel_range` given that receivers of `query_channel_range` can span their response over multiple messages.

PR #647, opened 2019-07-18
BOLT07: Inventory Gossip

cfromknecht:
Took a stab at drafting some requirements for an Inventory-based Gossip proposal. This builds off of #557 which allows for more fine-grained querying of a remote node for gossip messages. The primary motivation is to reduce redundant bandwidth due to gossiping duplicate messages.

As it stands, the proposal adds a new `inv` message carrying:
   * The `short_channel_id`s of newly accepted channels
   * The `short_channel_id`, `direction`, and `timestamp` of the latest `channel_update` in the last window.

Once the receiving node filters the message for any new data, it can request anything that is unknown or newer than info it already has using the `extended_query_flags` features.

The feature is gated by a new feature bit `option_inv_gossip`, temporarily assigned to 50/51.

Notes:
 - The proposal is most probably incomplete, just hoping to establish a high-level direction before proceeding further.
 - This does not affect the propagation of `node_announcement`s at the moment, since #557 doesn't offer a way to do so without knowing the `short_channel_id`. However, the bulk of the messages being sent on the network are `channel_update`s so this is where our time is best spent atm.

Thanks in advance for any feedback!


sstone:
Hi @cfromknecht ! Did you have a look at #584 ? It also builds on #557 and has been opened for some time (I though ). The major difference with your PR is that:
- I reuse extended data defined for `reply_channel_range`, since they carry information that I think are relevant for an INV message (i.e you want data you can query against before you ask for gossip messages). It stems from a discussion on IRC where I realised that extended queries are functionally equivalent to what I think is needed for inventory messages
- Node announcements can also be included in the INV message (@Roasbeef proposed to use a separate INV message for node announcements but I don't think it's necessary)

cfromknecht:
Hey @sstone, somehow I missed #584 😅I'll leave some questions for you there!

PR #650, opened 2019-07-19
Simple tooling fixes

rustyrussell:
The first two are in #631 already, but the third one is new: @ZmnSCPxj points out in #643 that describing TLVs using ordered lists causes them to be merged on formatting, so we should use unordered lists (which is also logically consistent).


rustyrussell:
Merged as agreed at http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-08-05-20.03.html

PR #654, opened 2019-08-02
Trampoline Routing

t-bast:
This proposal allows nodes running on constrained devices to sync only a small portion of the network
and leverage trampoline nodes to calculate the missing parts of the payment route.

The main idea is to use _layered onions_: a normal onion contains a smaller onion for the last hop of the route, and that smaller onion contains routing information to reach the next trampoline hop.

This PR is split in two files. Reviewers should start with `proposals/trampoline.md`, where concepts and designs are presented at a higher level. This document lets reviewers see the big picture and how all the pieces work together.

Then reviewers can move on to `wip-trampoline.md` which contains an aggregate of the spec sections that will be incorporated into existing bolts. This is following the usual spec format, this is where we'll work on the nitty-gritty details once we agree on the concept.

We don't need to implement all features from the start. Trampoline could be a succession of independent features, each bringing an incremental addition to the protocol:

- The onion construction and payment relaying: it will provide a storage and computation improvement for nodes running on phones (but won't provide any improvement to bandwidth usage)
- The network pruning / gossip filters part can be added afterwards and will provide bandwidth improvements (faster sync time and faster time-to-first-payment for mobile apps)
- Then we can add trampoline routing hints to invoices and `node_update` messages in a third PR

### DIFF
--- a/proposals/trampoline.md
+++ b/proposals/trampoline.md
@@ -1,0 +1,346 @@
+# Trampoline Onion Routing
+
+## Table of Contents
+
+* [Proposal](#proposal)
+  * [Introduction](#introduction)
+  * [Overview](#overview)
+  * [Trampoline onion](#trampoline-onion)
+  * [Gossip](#gossip)
+  * [Filtering gossip messages](#filtering-gossip-messages)
+    * [The `channel_update_filter`](#the-channel_update_filter)
+    * [The `node_update_filter`](#the-node_update_filter)
+  * [Privacy](#privacy)
+  * [Multi-Part trampoline](#multi-part-trampoline)
+* [Open questions](#open-questions)
+  * [Building a trampoline route](#building-a-trampoline-route)
+
+## Proposal
+
+### Introduction
+
+As the network grows, more bandwidth and storage will be required to keep an
+up-to-date view of the whole network. Finding a payment path will also require
+more computing power, making it unsustainable for constrained devices.
+
+Constrained devices should only keep a view of a small part of the network and
+leverage trampoline nodes to route payments.
+
+### Overview
+
+Nodes that are able to calculate routes on behalf of other nodes will advertise
+support for `trampoline_routing`. This is an opportunity for them to earn more
+fees than with the default onion routing.
+
+A payer selects a few trampoline nodes and builds a corresponding trampoline onion.
+It then embeds that trampoline onion in the last `hop_payload` of an `onion_packet`
+destined to the first trampoline node. Computing routes between trampoline nodes
+is deferred to the trampoline nodes themselves.
+
+Note that this construction still uses an onion created by the payer so it doesn't
+sacrifice privacy (in most cases it may even provide a bigger anonymity set).
+
+### Trampoline onion
+
+A trampoline onion uses the same construction as normal payment onions but with
+a smaller size, to allow embedding it inside a normal payment onion.
+
+For example, Alice may want to pay Carol using trampoline. Alice inserts
+trampoline nodes between her and Carol. For this example we use a single
+trampoline (Bob) but Alice may use more than one trampoline.
+
+Alice then finds a route to the first trampoline Bob. The complete route is:
+
+```text
+        +--> Irvin --> Iris --+          +--> ... --+
+        |                     |          |          |
+        |                     |          |          |
+Alice --+                     +--> Bob --+          +--> Carol
+```
+
+The corresponding `update_add_htlc` message would look like:
+
+```text
+                        update_add_htlc
++---------------------------------------------------------------------------+
+| channel_id                                                                |
++---------------------------------------------------------------------------+
+| htlc_id                                                                   |
++---------------------------------------------------------------------------+
+| amount_msat                                                               |
++---------------------------------------------------------------------------+
+| payment_hash                                                              |
++---------------------------------------------------------------------------+
+| cltv_expiry                                                               |
++---------------------------------------------------------------------------+
+|               onion_routing_packet                                        |
+| +-----------------------------------------------------------------------+ |
+| | version                                                               | |
+| +-----------------------------------------------------------------------+ |
+| | public_key                                                            | |
+| +-----------------------------------------------------------------------+ |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward                             |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | outgoing_cltv_value                        | normal onion payload | | |
+| | +--------------------------------------------+ for Irvin            | | |
+| | | short_channel_id                           |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward                             |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | outgoing_cltv_value                        | normal onion payload | | |
+| | +--------------------------------------------+ for Iris             | | |
+| | | short_channel_id                           |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward                             |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | outgoing_cltv_value                        |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | payment_secret                             |                      | | |
+| | +--------------------------------------------+ normal onion payload | | |
+| | | total_amount                               | for Bob              | | |
+| | +--------------------------------------------+                      | | |
+| | |        trampoline_onion                    |                      | | |
+| | | +---------------------+--------------------|                      | | |
+| | | | amt_to_forward      |                    |                      | | |
+| | | +---------------------+ trampoline payload |                      | | |
+| | | | outgoing_cltv_value | for Bob            |                      | | |
+| | | +---------------------+                    |                      | | |
+| | | | outgoing_node_id    |                    |                      | | |
+| | | +---------------------+--------------------|                      | | |
+| | | +---------------------+--------------------|                      | | |
+| | | | amt_to_forward      |                    |                      | | |
+| | | +---------------------+ trampoline payload |                      | | |
+| | | | outgoing_cltv_value | for Carol (final   |                      | | |
+| | | +---------------------+ recipient)         |                      | | |
+| | | | payment_secret      |                    |                      | | |
+| | | +---------------------+                    |                      | | |
+| | | | total_amount        |                    |                      | | |
+| | | +---------------------+--------------------|                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| +-----------------------------------------------------------------------+ |
+| | hmac                                                                  | |
+| +-----------------------------------------------------------------------+ |
++---------------------------------------------------------------------------+
+```
+
+When Bob receives the trampoline onion, he discovers that the next node is Carol.
+Bob finds a route to Carol, builds a new payment onion and includes the peeled
+trampoline onion in the final payload. This process may repeat with multiple
+trampoline hops until we reach the final recipient.
+
+### Gossip
+
+The main goal of trampoline routing is to reduce the amount of gossip that
+constrained nodes need to sync.
+
+These nodes should only keep track of nearby channels (with an `N`-radius
+heuristic for example). These nearby channels will be used to build an onion
+route to a first trampoline node.
+
+These nodes could even choose to ignore channels with a capacity lower than
+`min_chan_capacity` to reduce the amount of gossip they need to sync at start-up.
+
+`N` and `min_chan_capacity` are configured by the node and not advertised to
+the network.
+
+But these nodes also need to store information about nodes that are outside of
+their neighborhood to use as trampoline hops. A solution is to introduce a
+`node_update` gossip message containing the information needed to use nodes as
+trampoline hops (similar to the `channel_update` message). Constrained nodes would
+synchronize a subset of the network's trampoline nodes, based on whatever heuristic
+matches their use-case.
+
+Trampoline nodes need to estimate a `cltv_expiry_delta` and `fee` that allows them
+to route to any other trampoline node while being competitive with other nodes,
+and broadcast this information via the `node_update` message.
+
+This is a great opportunity to incentivize nodes to open channels between each
+other to minimize the cost of trampoline hops. This is also a great opportunity
+for nodes to implement smart fee estimation algorithms as a competitive advantage.
+
+Nodes may be very conservative and advertise their worst case `fee` and
+`cltv_expiry_delta`, corresponding to the furthest node they can reach.
+
+On the contrary, nodes may apply statistical analysis of the network to find a
+lower `fee` and `cltv_expiry_delta` that would not allow them to reach all other
+trampoline nodes but would work for most cases. Such nodes may choose to route
+some payments at a loss to keep reliability high, attract more payments by building
+a good reliability reputation and benefit from an overall gain.
+
+Trampoline nodes may accept payments with a fee lower than what they advertised
+if they're still able to route the payment in an economically viable way (because
+they have a direct channel or a low-cost route to the next trampoline hop for example).
+
+That means that payers may choose to ignore advertised fees entirely if they think
+the fee/cltv they're using will still be able to route properly and retry with
+a higher fee on route failures. Similarly to the existing `UPDATE` onion errors,
+failures could include a `node_update` message containing the fee/ctlv needed to
+correctly route the payment.
+
+### Filtering gossip messages
+
+Constrained nodes should listen to `node_update` messages and store some of them
+to be used as trampoline hops. Nodes are free to choose their own heuristics for
+trampoline node selection (some randomness is desired for anonymity).
+
+While this reduces storage requirements on constrained nodes, it doesn't reduce
+their bandwidth requirements: constrained nodes still need to listen to
+`node_update` and `channel_update` messages (even though they will ignore most of them).
+
+Nodes can reduce bandwidth usage by applying gossip filters before forwarding
+gossip messages. Constrained nodes may require their peers to support a
+`gossip_filters` feature. Per-connection filters are negotiated with each peer
+after `init`. When receiving gossip messages, the remote node must apply the
+negotiated filters and only forward to the local node the messages that match the filter.
+
+Note that the remote node can decide to ignore the filters entirely and forward
+every gossip to the local node: in that case the local node may close all channels
+with that remote node, fail the connection and open channels to more cooperative nodes.
+
+#### The `channel_update_filter`
+
+The local node may send a `channel_update_filter` to the remote node. The remote
+node must apply this filter before forwarding `channel_update`s. This allows the
+local node to keep an up-to-date view of only a small portion of the network.
+
+The filter could simply be a `distance` integer. The remote node should only
+forward `channel_update`s of nodes that are at most `distance` hops away from
+the remote node itself.
+
+Computing this filter is inexpensive for small `distance`s. The remote node should
+reject `distance`s that would be too costly to evaluate.
+
+#### The `node_update_filter`
+
+The local node may send a `node_update_filter` to the remote node. The remote
+node should apply this filter before forwarding `node_update`s. This allows the
+local node to only receive `node_update`s it cares about to periodically refresh
+its list of trampoline nodes and stay up-to-date with their fee rate and cltv
+requirements.
+
+The local node may choose to keep the same set of trampoline nodes if they are
+reliable enough. Or the local node may choose to rotate its set of trampoline
+nodes regularly to improve anonymity and test new payment paths. This decision
+should be configurable by the user/implementation.
+
+A simple heuristic to frequently rotate the set of trampoline nodes would be to
+compare their `node_id` to `sha256(local_node_id || latest_block_hash)` and
+update this every `N` blocks. The `node_update_filter` could be based on the
+distance between these two values. Then the local node chooses which trampoline
+nodes to keep depending on their advertised fees/cltv and reliability reputation
+(based on historical data if it's possible to collect such data).
+
+### Privacy
+
+Trampoline routing allows constrained devices to send and receive payments
+without sacrificing privacy.
+
+Senders are advised to sync a small portion of the graph (their local neighborhood).
+This allows them to insert "normal" hops before the first trampoline node, thus
+protecting their privacy with the same guarantees than normal payments.
+
+Receivers may use route-blinding or rendezvous to hide their exact location.
+Even without such measures, since the trampoline onion may contain multiple hops,
+intermediate trampoline nodes can't know their position in the route and thus
+can't tell whether the next node is the final recipient or not.
+
+### Multi-Part Trampoline
+
+Trampoline routing combines nicely with multi-part payments. When multi-part
+payment is used, we can let trampoline nodes combine all the incoming partial
+payments before forwarding. Once the totality of the payment is received, the
+trampoline node can choose the most efficient way to re-split it to reach the
+next trampoline node.
+
+For example, Alice could split a payment in 3 parts to reach a trampoline node,
+which would then split it in only 2 parts to reach the destination (we use a
+ingle hop between trampoline nodes for simplicity, but any number of intermediate
+nodes could be used):
+
+```text
+                                                     HTLC(1500 msat, 600112 cltv)
+                                                    +---------------------------+
+                                                    | amount_fwd: 1500 msat     |
+                                                    | expiry: 600112            |
+            HTLC(1560 msat, 600124 cltv)            | payment_secret: aaaaa     |
+             +-----------------------+              | total_amount: 2800 msat   |
+             | amount_fwd: 1500 msat |              | trampoline_onion:         |                                                                  HTLC(1100 msat, 600000 cltv)
+             | expiry: 600112        |              | +-----------------------+ |                                                                 +-----------------------------+
+        +--> | channel_id: 3         | ---> I1 ---> | | amount_fwd: 2500 msat | | --+                                                             | amount_fwd: 1100 msat       |
+        |    |-----------------------|              | | expiry: 600000        | |   |                                                             | expiry: 600000              |
+        |    |     (encrypted)       |              | | node_id: Bob          | |   |                                                             | payment_secret: xxxxx       |
+        |    +-----------------------+              | +-----------------------+ |   |                     HTLC(1150 msat, 600080 cltv)            | total_amount: 2500 msat     |
+        |                                           | |      (encrypted)      | |   |                      +-----------------------+              | trampoline_onion:           |
+        |                                           | +-----------------------+ |   |                      | amount_fwd: 1100 msat |              | +-------------------------+ |
+        |                                           +---------------------------+   |                      | expiry: 600000        |              | | amount_fwd: 2500 msat   | |
+        |                                           |             EOF           |   |                 +--> | channel_id: 561       | ---> I4 ---> | | expiry: 600000          | | --+
+        |                                           +---------------------------+   |                 |    |-----------------------|              | | total_amount: 2500 msat | |   |
+        |                                            HTLC(800 msat, 600112 cltv)    |                 |    |     (encrypted)       |              | | payment_secret: yyyyy   | |   |
+        |                                           +---------------------------+   |                 |    +-----------------------+              | +-------------------------+ |   |
+        |                                           | amount_fwd: 800 msat      |   |                 |                                           | |         EOF             | |   |
+        |                                           | expiry: 600112            |   |                 |                                           | +-------------------------+ |   |
+        |   HTLC(820 msat, 600130 cltv)             | payment_secret: aaaaa     |   |                 |                                           +-----------------------------+   |
+        |    +-----------------------+              | total_amount: 2800 msat   |   |                 |                                           |             EOF             |   |
+        |    | amount_fwd: 800 msat  |              | trampoline_onion:         |   |                 |                                           +-----------------------------+   |
+        |    | expiry: 600112        |              | +-----------------------+ |   |                 |                                                                             |
+Alice --+--> | channel_id: 5         | ---> I2 ---> | | amount_fwd: 2500 msat | | --+--> Trampoline --+                                                                             +--> Bob
+        |    |-----------------------|              | | expiry: 600000        | |   |  (fee 170 msat) |                                            HTLC(1400 msat, 600000 cltv)     |
+        |    |     (encrypted)       |              | | node_id: Bob          | |   |    (delta 32)   |                                           +-----------------------------+   |
+        |    +-----------------------+              | +-----------------------+ |   |                 |                                           | amount_fwd: 1400 msat       |   |
+        |                                           | |      (encrypted)      | |   |                 |                                           | expiry: 600000              |   |
+        |                                           | +-----------------------+ |   |                 |                                           | payment_secret: xxxxx       |   |
+        |                                           +---------------------------+   |                 |   HTLC(1480 msat, 600065 cltv)            | total_amount: 2500 msat     |   |
+        |                                           |             EOF           |   |                 |    +-----------------------+              | trampoline_onion:           |   |
+        |                                           +---------------------------+   |                 |    | amount_fwd: 1400 msat |              | +-------------------------+ |   |
+        |                                            HTLC(500 msat, 600112 cltv)    |                 |    | expiry: 600000        |              | | amount_fwd: 2500 msat   | |   |
+        |                                           +---------------------------+   |                 +--> | channel_id: 1105      | ---> I5 ---> | | expiry: 600000          | | --+
+        |                                           | amount_fwd: 500 msat      |   |                      |-----------------------|              | | total_amount: 2500 msat | |
+        |                                           | expiry: 600112            |   |                      |     (encrypted)       |              | | payment_secret: yyyyy   | |
+        |   HTLC(510 msat, 600120 cltv)             | payment_secret: aaaaa     |   |                      +-----------------------+              | +-------------------------+ |
+        |    +-----------------------+              | total_amount: 2800 msat   |   |                                                             | |         EOF             | |
+        |    | amount_fwd: 500 msat  |              | trampoline_onion:         |   |                                                             | +-------------------------+ |
+        |    | expiry: 600112        |              | +-----------------------+ |   |                                                             +-----------------------------+
+        +--> | channel_id: 7         | ---> I3 ---> | | amount_fwd: 2500 msat | | --+                                                             |             EOF             |
+             |-----------------------|              | | expiry: 600000        | |                                                                 +-----------------------------+
+             |     (encrypted)       |              | | node_id: Bob          | |
+             +-----------------------+              | +-----------------------+ |
+                                                    | |      (encrypted)      | |
+                                                    | +-----------------------+ |
+                                                    +---------------------------+
+                                                    |             EOF           |
+                                                    +---------------------------+
+```
+
+## Open questions
+
+### Building a trampoline route
+
+Once a node has a set of trampoline nodes, how does it choose which ones to use
+to efficiently reach the final recipient?
+
+The best solution for privacy is to choose them at random as long as the total
+fee fits what the payer is ready to accept. This ensures that we are doing a
+random walk between the trampoline nodes instead of taking a path that could be
+guessed by an attacker. The downside is that the path may be wasteful, but that
+is a trade-off that needs to be made between efficiency and privacy.
+
+Another solution would be to have the recipient explicitly provide close trampoline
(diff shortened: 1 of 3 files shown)


wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+This document describes the additions we make to the existing onion routing to enable the use of trampoline nodes.
+
+As the network grows, more bandwidth and storage will be required to keep an up-to-date view of the whole network.
+Finding a payment path will also require more computing power, making it unsustainable for constrained devices.
+
+Constrained devices should only keep a view of a small part of the network and leverage trampoline nodes to route
+payments. This proposal still uses an onion created by the payer so it doesn't sacrifice privacy (in most cases it
+will even provide a bigger anonymity set).
+
+Nodes that keep track of the whole network should advertise support for `trampoline_routing`. This is an opportunity
```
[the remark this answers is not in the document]
t-bast: I was pretty sure you would comment something like that, especially after reading your last email on routing ;)

I totally agree that in the long run, we don't want/need to ask trampoline nodes to sync the whole network and accurately estimate fees (we have retries to make up for inaccuracy and failures). But we must take this problem one step at a time, and I think this goes in the right direction. This proposal doesn't force trampoline nodes to sync the whole network and I think it slowly goes towards a packet-switched, internet-like network where trampoline nodes would act as AS routers.

I'm not afraid of the cryptographic/mac part: we will find a way to make it work. So bottom line is: I agree with you on the long-term goals and I think this is a first step towards that.

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+### Paying to non-trampoline nodes
+
+If the recipient of the payment doesn't support trampoline routing, the last trampoline node must convert the last hop
+to a standard onion payment. The `recipient_features` tlv is used to detect that the next node doesn't support
+trampoline.
+
+Note that it reveals the identity of the recipient and the amount paid to the last trampoline node (but it doesn't
+reveal the identity of the payer). This can be avoided if the recipient uses rendez-vous routing to hide its identity.
+Otherwise recipients should support `trampoline_routing` to preserve their anonymity (and can advertise a high `fee`
+if they don't want to be chosen to route too many payments, or never send a `node_update` if they don't want to route
```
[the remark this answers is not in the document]
t-bast: I agree. As always in lightning payments, the trade-offs you make are privacy vs cost-efficiency. Privacy-concerned nodes should support trampoline. In my ideal world everyone is privacy-concerned and does that.

I included this statement mainly to show that payers can use trampoline payments even if payees don't want to fully support it (same for the section on sending to non-trampoline-aware nodes). But I expect this statement to be removed before we merge to the spec.

wip-trampoline.md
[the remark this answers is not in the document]
ZmnSCPxj: Okay.  I am afraid I am skimming the details for now.  It seems fine for a first cut.

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+
+As the network grows, more bandwidth and storage will be required to keep an up-to-date view of the whole network.
+Finding a payment path will also require more computing power, making it unsustainable for constrained devices.
+
+Constrained devices should only keep a view of a small part of the network and leverage trampoline nodes to route
+payments. This proposal still uses an onion created by the payer so it doesn't sacrifice privacy (in most cases it
+will even provide a bigger anonymity set).
+
+Nodes that keep track of the whole network should advertise support for `trampoline_routing`. This is an opportunity
+for them to earn more fees than with the default onion routing.
```
ariard: What's about network partition of constrained nodes by channel squatting of concurrent trampoline node servicing same network area ? Intuition would say channel squatting would be more efficient than now as trampoline node as a better view of the network than constrained one

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+        * [`tu64`:`amt_to_forward`]
+    * type: 4 (`outgoing_cltv_value`)
+    * data:
+        * [`tu32`:`outgoing_cltv_value`]
+    * type: 10 (`next_node_id`)
+    * data:
+        * [`point`:`next_node_id`]
+    * type: 12 (`recipient_features`)
+    * data:
+        * [`length*byte`:`recipient_features`]
```
ariard: Is recipient features only global ones or also local or invoice ones ? E.g  where payee need to understand multi-payment part

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+   * `filler`
+   * [`32*byte`:`hmac`]
+
+### Paying to non-trampoline nodes
+
+If the recipient of the payment doesn't support trampoline routing, the last trampoline node must convert the last hop
+to a standard onion payment. The `recipient_features` tlv is used to detect that the next node doesn't support
+trampoline.
+
+Note that it reveals the identity of the recipient and the amount paid to the last trampoline node (but it doesn't
```
ariard: Hmmm should it be mandatory for payee to flag itself as trampoline_routing ? Otherwise payer would be only decision maker to leak payee privacy to last trampoline node. With current onion packet construction, I think payer can't force leakage of last hop as payee. That would be a regression for payee while maybe having full-resources.

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+
+## Packet Structure
+
+### Trampoline Onion
+
+The trampoline onion is a fixed-size tlv packet with the following structure:
+
+1. type: 14 (`trampoline_onion_packet`)
+2. data:
+   * [`byte`:`version`]
```
ariard, line 41: Version number should be `0x00`, do we need a magic number before to distinguish between multiple types of  embedded payload ?

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+reveal the identity of the payer). This can be avoided if the recipient uses rendez-vous routing to hide its identity.
+Otherwise recipients should support `trampoline_routing` to preserve their anonymity (and can advertise a high `fee`
+if they don't want to be chosen to route too many payments, or never send a `node_update` if they don't want to route
+any trampoline payment).
+
+### Requirements
+
+A sending node:
+
+* MUST verify that each hop in the `trampoline_onion_packet` supports `trampoline_routing`
```
ariard, line 110: Is final hop (the recipient) must support trampoline_routing ? Not clear with rational above.

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+* MUST include a `recipient_features` with the `trampoline_routing` flags set to `0` if the recipient doesn't support `trampoline_routing`
+* MUST encrypt the `trampoline_onion_packet` with the same construction as `onion_packet`
+* MUST use a different `session_key` for the `trampoline_onion_packet` and the `onion_packet`
```
ariard, line 112: How to derive second session_key ? BIP32 HD derivation style based on ephemeral pubkey/node_id ?

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+  * MUST report a route failure to the origin node
+* If it doesn't have enough data to locate the next hop:
+  * MUST report a route failure to the origin node
+* MUST compute a route to the next trampoline hop:
+  * If it cannot find a route that satisfies `cltv` requirements:
+    * MUST report a route failure to the origin node
+  * If it cannot find a route that satisfies `fees` requirements:
+    * MUST report a route failure to the origin node
+  * If it has a channel available with the next trampoline hop:
+    * MAY use that channel in a single-hop route
```
ariard: Does it mean `trampoline_onion_packet` is passed as it is in an `update_add_htlc` ? (need to update `update_add_htlc` ?)

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+The trampoline onion is a fixed-size tlv packet with the following structure:
+
+1. type: 14 (`trampoline_onion_packet`)
+2. data:
+   * [`byte`:`version`]
+   * [`point`:`public_key`]
+   * [`400*byte`:`hop_payloads`]
+   * [`32*byte`:`hmac`]
+
+This results in a `466`-bytes fixed-size tlv where the value has exactly the same format as the `onion_packet` with
```
ariard, line 46: With a fixed `466`-bytes fixed-sizes is maximum size implicitly 120 hops (20 * 6) ? I think not given you have variable size payload, but even variable there is a minimum size to describe meaningfully HTLC routed..

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+        * [`u32`:`fee_base_msat`]
+        * [`u32`:`fee_proportional_millionths`]
+
+It has a structure similar to the `channel_update` message but is channel agnostic.
+`Node_update`s should be relayed the same way `channel_update`s are relayed (staggered broadcast).
+
+Trampoline nodes should only accept and relay `node_update`s from nodes that have open channels with inbound and
+outbound capacity.
+
+If a node is not willing to relay trampoline payments, it will simply never send a `node_update`. This lets the network
```
ariard: How a node opt-out of being a trampoline one after some time ? (like while opening/closing/splicing channels)

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+    * data:
+        * [`u32`:`timestamp`]
+        * [`u16`:`cltv_expiry_delta`]
+        * [`u32`:`fee_base_msat`]
+        * [`u32`:`fee_proportional_millionths`]
+
+It has a structure similar to the `channel_update` message but is channel agnostic.
+`Node_update`s should be relayed the same way `channel_update`s are relayed (staggered broadcast).
+
+Trampoline nodes should only accept and relay `node_update`s from nodes that have open channels with inbound and
```
ariard: Why the mandatory both inbound and outbound ? Seems acceptable to only trampoline route upstream or downstream packets

wip-trampoline.md
```diff
@@ -0,0 +1,780 @@
...
+It uses a new mechanism that allows switching the `session_key` in the middle of the route.
+
+Trampoline routing provides another option to implement rendez-vous routing (with some similarities).
+The recipient creates a small onion from his rendez-vous point to himself and embeds it inside a payment request.
+The sender includes that small onion in his normal onion payload and it gets relayed to the last trampoline node.
+That last trampoline node then uses the provided rendez-vous onion to reach the recipient.
+
+## Appendix C: Trampoline and MPP
+
+Trampoline routing combines nicely with multi-part payments. When multi-part payment is used, we can allow trampoline
```
ariard: Hmmm if a trampoline node has a good view of a network, knowing every last partial payment node, it should be able to do path intersections from them to discover who is the original sender. IMO, a randomly-generated one-shot tunnel with a minimal length would be better than splitting payment if we want to preserve payer-privacy towards first trampoline node. But this whole point need real analysis.

wip-trampoline.md
t-bast: Not sure what you mean there, could you expand?
Are you implying that trampoline nodes in the same area would have an incentive to attack each other's inbound capacity?
Do not forget that most likely all current routing nodes (nodes running on servers) will support trampoline. That means that a vast majority of the public network graph will support trampoline. Users of trampoline (mobile / constrained nodes) usually don't advertise their node or their channels.

wip-trampoline.md
t-bast: Global and local are being merged into a single set of feature bits, which aligns well with that field.
I don't think invoice features are useful there; but maybe I'm missing a use-case (if that's the case, another tlv for invoice features can be added later).

wip-trampoline.md
t-bast: Payer can always leak payee privacy, even today (since payer has this information). It can put information in ignored bits at the end of any message, or in an odd tlv field, etc.
I strongly encourage payees to support trampoline because it protects their privacy and automatically re-balances channel capacity (by routing payments).
But in order to have a reasonable upgrade path, we need to support paying to legacy nodes that don't know about trampoline (which is why this feature is necessary). I'm not happy about it, but I don't see a better solution at the moment...

wip-trampoline.md
t-bast: I kept the version to be able to completely re-use the existing onion code, but it's true that it's fixed at `0x00`.
I think different embedded payloads should be signaled by a different tlv type (which acts as a magic number). Does that make sense or am I missing your point?

wip-trampoline.md
t-bast: No, I meant `hop` as intermediate `hop`. I should clarify that to correctly align with the rationale.

wip-trampoline.md
t-bast: For normal payments in Bolt 4, we don't specify how the session key should be derived (just that it needs to be random). It's the same for that second session key: it's up to the implementation to choose how they want to derive randomness. I'd say just pure random bytes32.

wip-trampoline.md
t-bast: Never directly, it's embedded in a normal payment onion: no changes are required to the existing htlc messages. It's just that the normal payment onion will contain only one payload, and that payload will contain the trampoline onion.

wip-trampoline.md
t-bast: I would need to re-do the maths, but I think it's less than that.
With the trampoline onion inside a normal onion, the normal onion route can be at most ~15 hops.
And the number of trampoline hops will be at most 4 or 5 (would need to double-check).
So we would have maximum 60-hops route. And an attacker wouldn't control the route taken between trampoline hops so in reality the route will likely be less than that (this shouldn't increase too much the effectiveness of a capacity attack).

wip-trampoline.md
t-bast: I think this will work fine. Trampoline nodes simply forward "normal" htlcs.
If you suddenly switch off the trampoline feature bit and refuse to route new trampoline payments, this wouldn't break anything on your side. It may be an issue for constrained nodes that still think you are a trampoline node, but they should route around you on failures and refresh their trampoline set at some point, evicting you from their DB.

wip-trampoline.md
t-bast: You're totally right, this sentence doesn't make any sense...a node cannot know another node's channels balance so it's impossible to figure out that a node doesn't have inbound or outbound capacity. Don't know why I wrote that :thinking:

wip-trampoline.md
t-bast: Interesting point, thanks for raising this! This is an issue with normal AMP, isn't it? The recipient could de-anonymize the payer by doing path intersection (even though in reality, the payer's channels are likely unannounced channels so the recipient can't de-anonymize the first hop).

If I understand your attack correctly, it seems that on the contrary trampoline helps anonymity in that case. Because in that case a node is only able to de-anonymize the previous node, thus discovering who the previous trampoline hop was. But thanks to onion encryption, this doesn't tell the attacker whether that node was the payer or an intermediate trampoline node! Unless the payer doesn't advertise trampoline feature bits, in which case we are in the same spot as normal AMP privacy-wise.

t-bast:
> I really can see trampoline nodes being partial blinded themselves and just announce their connections to other trampolines ones

I completely agree. I'm not mentioning it yet because it is a larger change and I want to roll-out such things incrementally, but the next step after that first version of trampoline is very likely something like that.

wip-trampoline.md
ariard: Yes as you said, was thinking about trampoline node attacking other trampoline node inbound capacity. It's a fuzzy concern if we have a lot of trampoline nodes and cost partition attack is high, but not if we have few some of them. Saying most likely all current routing nodes will support trampoline is IMO hard without empirical data from deployment, so maybe will should beware to not ask to high-requirement for being one (like also supporting wumbo channels)

wip-trampoline.md
ariard: Cool for merging both local/global features. Yeah would be easy to upgrade but doesn't the last version of base AMP use invoice feature bits to signal support of feature?

wip-trampoline.md
ariard: You're right TLV should make the work to dissociate between embedded payloads.

wip-trampoline.md
ariard: I think there is a difference between intentionally having to tweak protocol messages to leak info and just leaking info by following faithful execution of protocol. Not happy too, I'm wandering if we can solve it with another routing node of indirection but naive solutions I can think of are gross..

wip-trampoline.md
ariard: Gotcha, keep this explication somewhere when you split/refactor proposal!

wip-trampoline.md
ariard: Okay so a bit confusing to explicit something which is not a special case in fact? IMO would just drop it, but your call :)

wip-trampoline.md
ariard: Ah you right for the math didn't subtract the trampoline onion, maybe in future versions, sending node should be able to pick trampoline onion size as a parameter, and so increase number of trampolines (privacy would be a good reason)

wip-trampoline.md
ariard: IMO would be nice to disable support cleanly to avoid useless failures and not having mission control/autopilot downgrading your score for bad reason. A bitfield `node_flags` a la `channel_update` would be cool, mostly if we have other routing features

wip-trampoline.md
ariard: Let's say you have following topology:

```    
        ______  B __________
     /                      \
   /                         \
 A  ___________ C  _________ T
   \                        /
     \ _______ D  _________/

```

If all channels are publicly knew and that's the whole graph, and A is using T as trampoline node by sending a AMP payment through B, C and D, that's super easy to deduct than A is payer from T viewpoint. Now this topology is super dumb, and private channels would obfuscate real graph and make whole analysis probabilistic-based. So payer should use only private channels as first links to broadcast AMP payment. Like you said it should keep you safe but that's easy to mess up in practice ("Oh my two first channels are private, that should be enough, don't care than my third one is public !")

wip-trampoline.md
t-bast: I don't think this will be an issue, it's an implementation detail. If the invoice specified that AMP could be used, the sender will include the `total_amount` TLV. That lets the trampoline node detect that AMP may be used to route to the destination (without the need for an explicit feature bit). But I agree that some future invoice feature bits might require us to add an `invoice_features` TLV here.

wip-trampoline.md
t-bast: > I think there is a difference between intentionally having to tweak protocol messages to leak info and just leaking info by following faithful execution of protocol

That's true. There is definitely room for improvement here, I'll think about this further too. If you find a work-around let me know ;)

wip-trampoline.md
t-bast: Yeah that's true, this requirement can probably be removed. It's an implementation detail and there's no reason implementers wouldn't use a single-hop route if they have one.

wip-trampoline.md
t-bast: Yeah I'm trying to balance privacy (where long routes are better) with capacity attacks (where long routes are bad). This is an inherent trade-off with today's lightning and it's a real pain ;)

wip-trampoline.md
t-bast: Interesting idea, that definitely could be done. Since you supported trampoline at some point that means your node contains the code for the `node_update` message, so you could send a last `node_update` setting a `disable` bit. That would allow a clean downgrade. I'll include that.

wip-trampoline.md
t-bast: But the example you're sharing showcases an issue with AMP _without_ trampoline ;)
If this is a non-trampoline AMP where T is the recipient, T is able to deduce who the payer is.

In the trampoline case, T is only able to deduce that A was the previous trampoline node (which may be the payer of just another intermediate trampoline node). So it's just like knowing the previous hop in a normal payment route.

ariard: I'm dumb, you're right I forgot a vertice from T the trampoline to some E node the payee.

Yes I think you're right that T is only able to deduce that A was the previous trampoline node (not necessarily the payer), but to keep analysis blurred we should recommend to use multiple trampoline/or private channels. If it's knew that a particular implementation uses only one-hop trampoline privacy would be broken.

t-bast: I completely agree. Implementations should never hard-code or bias too much towards a one-hop trampoline route. I'll make stress this more in the rationale.

wip-trampoline.md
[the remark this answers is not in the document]
jtimon: I don't know, I would still rephrase it to "nodes that want to act as trampoline nodes" or something like that. Since it is not a requirement for trampoline nodes to keep track of the whole network, "Nodes that keep track of the whole network" seems unnecessarily confusing.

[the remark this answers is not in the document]
t-bast: Good suggestion. I'll do another big pass on this PR once I'm done with the implementation in eclair, I learnt a lot while implementing so there are a few points that have changed slightly.

wip-trampoline.md
t-bast: This isn't a concern anymore as invoice features now use the same bit namespace as other features. Yay!

wip-trampoline.md
t-bast: I'm leaning more and more towards removing the ability to pay legacy nodes via this trampoline-to-legacy mechanism (unless we find a better way to do it than what I currently have).

wip-trampoline.md
```diff
@@ -0,0 +1,842 @@
...
+to a standard onion payment. The `outgoing_node_features` field is used to detect that the next node doesn't support
+trampoline.
+
+Note that this reveals the identity of the recipient and the amount paid to the last trampoline node (but it doesn't
+reveal the identity of the payer). It also means that we'll need to add the invoice routing hints to the payload for
+the last trampoline node, otherwise it's likely that it won't be able to find a route to the recipient.
+
+Recipients should support `trampoline_routing` to properly preserve their anonymity.
+
+Question: should we even allow sending to non-trampoline-aware nodes? Maybe it's best to disallow it entirely.
```
jtimon: I really dislike that you can be paid via trampoline and without knowing it, your identity and the amount are revealed to some hop. Not only you don't know, but also you cannot opt out of this. This doesn't feel right to me.
Additionally, requiring the recipient to be trampoline aware should simplify the protocol.
If we hope for all nodes to eventually be trampoline-aware, this shouldn't be a big deal in the long term anyway.

So my preference would be to simply disallow it entirely. Or in other words, requiring recipients of trampoline payments to be trampoline-aware.

ZmnSCPxj: I agree with @jtimon here.  This tends to put more load on each node, requiring each node to actually have a local routemap.  I have a number of ideas to compress routemaps as well.

t-bast: I agree too, I initially added that to make the upgrade smoother but I'd like to remove it before spec acceptance.
I already have another comment on this PR from December 5th where I stated that ;)

> I'm leaning more and more towards removing the ability to pay legacy nodes via this trampoline-to-legacy mechanism (unless we find a better way to do it than what I currently have).

It feels weird and it's bad for privacy. It will be removed from the proposal in the next iteration.

TheBlueMatt:
Personally I think moving in this direction is a bad idea for two reasons:

 * One of the key pitches for lightning is its privacy properties - properties which today are somewhat weak, but which can be enabled only with sufficient route graph data to select a sufficiently long, and diverse path. Just look at the motivation section here for why this is going to even further erode practical lightning privacy - folks can and will use this to hold a routedb that contains just a group of very large nodes and use all of their routing dbs, probably losing a lot of privacy. This is a cost for everyone, not just the users of unfortunate implementations that do this - not only is their privacy set reduced, but lightning gains a reputation of not being private, and moves Lightning towards utter crap like Interledger instead of a well-designed user-first system.

 * At a high level, this seems like premature optimization. I'm highly dubious that the current routedb can't simply be (and shouldn't simply be) pruned by learning which nodes are regularly offline/fail to route regularly and have this be a viable routing table. If churn in the routedb is an issue for you, stop fetching it all the time and use AMP to send a payment simultaneously along multiple paths, retrying the missing paths down a successful path to ensure quick payment success. If initial fetch of the routedb is an issue for you, a) fetch well-connected subsets from redundant central servers on startup and b) work on schnorr signatures in the channel announcements to reduce the number of sigchecks and amount of bandwith used. Don't jump to three steps back in the privacy guarantees lightning provides until you've exhausted low-hanging fruit to get towards a similar fetch time.

ZmnSCPxj:
@TheBlueMatt a sufficiently well-designed trampoline routing allows for payee secrecy from payer.  Currently the payer has to know the payee or a node that directly has a channel to the payee.  With a trampoline routing scheme that allows independent pre-concatenation, the payee can provide an encrypted trampoline route to itself in an invoice (i.e. rendezvous routing, i.e. very much like Tor hidden services) as well as payee-pays-fees (the payer is obligated to pay only up to the trampoline node, with any fees from the trampoline node to the payee being shouldered by the payee --- useful for c\*stodial exchanges that are willing to publish their own node as a trampoline node, their ~~victims~~ clients would make a short trampoline from the exchange node to one or two intermediaries then to themselves, with fees along the way shouldered by the client).  Basically trampoline is simply the extension of the previous rendezvous routing proposal, extended to more than one hop.

Of course, that assumes we want the payee to have secrecy of which node on the network it actually is.

And I am not at all certain the actual current proposal supports independent concatenation (i.e. having different sources encrypt different hops, which would be needed for rendezvous).


> folks can and will use this to hold a routedb that contains just a group of very large nodes and use all of their routing dbs, probably losing a lot of privacy
>
> ...
>
> I'm highly dubious that the current routedb can't simply be (and shouldn't simply be) pruned by learning which nodes are regularly offline/fail to route regularly and have this be a viable routing table

Given that large nodes tend to be the most online/successful, I think the two sets converge.  Admittedly the latter set will probably start out larger than the former set, which improves risk-sharing, and probably dynamic growth will ensure the latter set will always be larger than the former set.

ZmnSCPxj:
@TheBlueMatt part of your privacy concerns is also mitigated if reduced-routemap nodes can also be trampolines.  This ***requires*** that we allow independent pre-concatenation.

Suppose there exists a node with a f\*cking tiny routemap: it only knows its direct peers, and one large, high-uptime server that it uses for all paying anyone other than its direct peers (it does not even know a direct path to that server: it randomly selects a direct peer and puts a trampoline containing that peer -> server -> payee).

The privacy of that node would be greatly improved if that node were to publicly publish itself as a rendezvous node.  That would be viable if and only if trampolines supported pre-concatenation, i.e. if that node could take a pre-existing trampoline onion and prepend a few hops onto it.

Suppose the large server receives a trampoline request. It could be:

* This node with a f\*cking tiny routemap wants to pay directly to the next hop in the trampoline onion.  In which case, bam, no privacy.
* This node with a f\*cking tiny routemap was actually selected by *another* node as a trampoline hop.
  *  What actually happened was that it received an encrypted trampoline onion to a node that is not a direct peer.
     * So what this node did was take this onion and prepend the large server, then prepend one of its direct peers.
     * The rest of the trampoline onion might have more than one hops, or the next hop could simply be the direct payee, but that depends on the behavior of the other node.

This is basically a form of delegation.

Admittedly, this is still not perfect when considering risk-sharing principle.  This simply ensures more routes get to the large server: not only do non-myopic nodes have a chance to select the large server as a trampoline hop, but every myopic node that signals itself as a trampoline hop, and remembers this large server, will ensure that use of those myopic nodes as trampoline hops is nothing more than an alias for using the large server as a trampoline hop.  So much for trampoline delegation.

TheBlueMatt:
I'm aware of the benefits of trampoline routing for payee-secrecy, and would probably be in favor of trampoline routing with that as the explicit and only goal, but this proposal is not that, and you need go no further than the opening motivation in this PR to see that. Hence my vocal objection. A proposal for trampoline routing with an explicit goal of helping clients reduce their routemap size by relying on third parties to do their routing for them seems like about five steps backwards for lightning (and if you want to do this to your poor users, you can do it yourself without polluting the spec with it).

ZmnSCPxj:
> (and if you want to do this to your poor users, you can do it yourself without polluting the spec with it).

Seems a good point.

t-bast:
The reason I'm pushing this proposal is to foster discussion and brainstorm new ideas.
I know how much time it takes to properly review any non-trivial spec addition (remember `var_onion_optin`) which is why I'm sharing the state of my current research and experimentations, but I'm hoping to improve it thanks to the community's feedback.
This has already proven useful: for example ZmnSCPxj's suggestion to find more flexible route concatenation may prove useful if we find a way to pull it off, and I'll be exploring that (and hopefully others will be researching that too because they're following the discussion).

Now onto the technical feedback, I disagree with your claims on the privacy impact.
I believe that your concerns apply to our current source-routing solution equally, and allowing some kind of partial delegation is actually a net improvement for both privacy and usability.
I'd also like to point out that full source-routing is actually a special case of trampoline (with 0-hops routes between trampoline nodes).

Privacy is only possible with a big enough network and some randomness in the route selection.
You simply can't expect a mobile wallet to sync that whole graph (and run Dijkstra on it), so there has to be some sort of pruning.
If implementations prune the network graph that is indeed effectively reducing the privacy set.
That's completely independent of trampoline and is true today: and what's worse is that all nodes are likely to end up with almost the same pruned graph because they still need to be able to compute a full route to any other node. With trampoline that pruning will be more diverse because you don't need to be able to compute a full route to any other node, only to a reasonable number of trampoline nodes. Selecting a diverse set of trampoline nodes in your route (ideally sampled randomly) will give you a more diverse final route than doing it with full source-routing because it won't provide an optimal route. That costs you more fees, but is more private because it breaks heuristics by doing a partial random walk.

Once we have a satisfactory solution to concatenate routes, it's highly likely that a majority of routing nodes will activate that feature (more opportunities to relay payments). That means you have a large set of potential nodes to sample, avoiding the issues you mention. I believe that, if done correctly, random sampling can provide more privacy and more opportunities for smaller nodes to relay payments.

The current proposal should certainly be improved and that's why I'm asking for feedback.
However I disagree with your statement that moving in this direction is a bad idea because in my opinion all the issues you're listing already exist today, so we need to change something.

>  (and if you want to do this to your poor users, you can do it yourself without polluting the spec with it)

Not really: the privacy benefits of trampoline can only be met if it is largely deployed on the network, hence the need for standardization. Of course a single-node deployment would be bad for privacy, but, again, the same is true for our current routing system.

sstone:
Trampoline routing also creates an incentive for relaying nodes to properly maintain a "good" view of the network. Such an incentive does not exist today since relaying nodes don't need a routing table to forward payments (you need routing info when you want to send payment, not when you're forwarding or receiving them).

ZmnSCPxj:
> Trampoline routing also creates an incentive for relaying nodes to properly maintain a "good" view of the network

Why should such an incentive ***need*** to exist, for some reason of improved privacy or security or censorship-freedom or scaling?  Why require the spending of resources if improvements are otherwise not achieved anyway?  This should be considered a cost of trampoline, not a benefit, unless you can show that "if some other hypothetical feature XYZZY, not related to Trampoline, were existent, and had the property that it required relaying nodes to maintain a good view of the network, then we should enable that feature ***RIGHT NOW*** because it forces relaying nodes to maintain a good view of the network because ${REASON}".

t-bast:
Well because the routing nodes are critical to enabling payment routing. Nothing to do with with privacy or security, but simply to make the system work in the first place. If the network is very private and very secure but the graph isn't connected, then Lightning is useless. People _could_ make anonymous and secure payments, but in reality they can't because there's no routes to their destinations.

You have a much better chance of having a healthy, well-connected graph when the nodes actually care about adding the right edges to improve connectivity, don't you think? That's a healthy incentive in my opinion to have routing nodes care about the graph instead of only wallets caring about it. But that's a side comment, we should probably focus on the main arguments about privacy/efficiency instead of forking the discussion too much.

ZmnSCPxj:
> You have a much better chance of having a healthy, well-connected graph when the nodes actually care about adding the right edges to improve connectivity, don't you think? That's a healthy incentive in my opinion to have routing nodes care about the graph instead of only wallets caring about it.

That is why we have fees.  Fee-maximizing routing nodes will already strive to improve connectivity locally for themselves because by doing so, they can undercut competition and gain more fees, and multiple fee-maximizing routing nodes will then greatly improve the overall connectivity of the entire graph. Given that, an additional incentive is unnecessary and thus not an argument *for* this proposal regardless.

>  But that's a side comment, we should probably focus on the main arguments about privacy/efficiency instead of forking the discussion too much.

Yes.

TheBlueMatt:
> Trampoline routing also creates an incentive for relaying nodes to properly maintain a "good" view of the network.

Aside from agreeing with @ZmnSCPxj here, I don't understand this concept. In today's lightning network, if every routing node in the network had literally zero view of other nodes in the network outside of their immediate neighbors, that would cause absolutely zero problems. They have no need to know about anything other than their neighbors except to open new channels.

To reiterate, in general, I think the points about payee-privacy are really important, and if there were a proposal for trampoline routing which *only* supported payee-privacy in the form of invoice trampolines (with associated "hiding the fact that the end recipient is X from the intermediate node by having multiple hops be trampoines"), I probably would have less of an issue with it, but I find it somewhat astounding that we have to argue about whether its ok to export routing decisions to your next-hop, and given that somehow we are, would likely object to trampoline routing in general just to avoid people doing stupid things.

t-bast:
> I think the points about payee-privacy are really important, and if there were a proposal for trampoline routing which only supported payee-privacy in the form of invoice trampolines (with associated "hiding the fact that the end recipient is X from the intermediate node by having multiple hops be trampoines"), I probably would have less of an issue with it

If I understand your point correctly, I think we're in alignment there (so I really don't understand why you reject the whole proposal that strongly).

It's true that I didn't remove yet the "pay to legacy recipient" section so it looks like you're loosing payee privacy every time you're using trampoline; however there are two open comments on this PR where I said that it will be removed in the next commits.

That means a payee signals support for trampoline (in the invoice). And a payer uses multiple trampoline hops, so it seems to me that we're not losing any privacy compared to today's routing.

I understand that this PR is big and takes time to review, so it's easy to dismiss it without reading the whole thing: that's because I wanted to give more context to help readers, fuel discussions, and make sure everyone sees the full picture. Maybe it was a mistake; it makes everyone focus only on the points they don't like and dismiss the rest. I'll push an update soon to at least remove the pay-to-legacy part: that will make the PR smaller and remove something we all agree is a bad idea for payee privacy.

proposals/trampoline.md
```diff
@@ -0,0 +1,346 @@
...
+| | | outgoing_cltv_value                        | normal onion payload | | |
+| | +--------------------------------------------+ for Iris             | | |
+| | | short_channel_id                           |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward                             |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | outgoing_cltv_value                        |                      | | |
+| | +--------------------------------------------+                      | | |
+| | | payment_secret                             |                      | | |
```
cdecker, line 101: I'm not sure I understand why trampoline nodes should require access to the `payment_secret`, shouldn't that be in the trampoline onion, destined for the recipient, or at most the last trampoline before the destination that doesn't support trampoline only? If the latter was the intended semantics it'd need a note to this regard.

cdecker: Or are they intended to be decoy secrets?

t-bast:
> I'd rather make the new message more expressive: each trampoline node can select up to n "adjacent" trampoline nodes, and provide "virtual channel" stats on how to get to those nodes

I agree, we could get rid of the `node_update` and leverage the `node_announcement` instead with something like what you suggest. I had investigated this solution before, coupled with a better way of actively querying for `node_announcement`s it could work. It's definitely worth investigating, I'll think about it more.

proposals/trampoline.md
t-bast: I should make this clearer, there are two distinct `payment_secret`s here to allow MPP to be aggregated at trampoline nodes before being re-split by the trampoline node when forwarding.

Let me know if these ASCII arts are better: https://github.com/ACINQ/eclair/blob/ce3629c98b07cdc914bd397fe45dd4bfd8338392/eclair-core/src/main/scala/fr/acinq/eclair/wire/Onion.scala#L59

The idea is that we need to support MPP payments split between multiple different trampoline routes (those need a `payment_secret` in the inner, trampoline onion) and using MPP between trampoline nodes to forward trampoline payments efficiently (those needs a different, unrelated `payment_secret` in the outer, normal onion).

cdecker: Ah ok, that makes sense, but wouldn't that also be possible just by aggregating across the `payment_hash`? Barring accidental reuse of the hash for different destinations (which is inherently unsafe anyway) it should work just as well.

FWIW even the concurrent reuse case is just a special case of the general reuse case, so I don't think it needs special treatment.

t-bast: That would work but you'd lose the probing protection that the `payment_secret` brings, so a node Mallory that's just before a trampoline node Ted could guess that Ted is either the payment final recipient or a trampoline node. Not sure they could steal fees, but they might want to try which is bad for the payment success rate, so it felt better to keep the `payment_secret` mechanism.

t-bast:
Closing this PR in favor of:

- https://github.com/lightningnetwork/lightning-rfc/pull/829 for the high-level view
- https://github.com/lightningnetwork/lightning-rfc/pull/836 for the onion format

PR #656, opened 2019-08-06
Bolt11 feature bits

rustyrussell:
We will want these for the payment-nonce proposal, since we'd eventually like to make that compulsory.  So I'm pushing this now so we can all deploy it ASAP.

(Note: this PR is simply extract from #643 )

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -533,6 +533,44 @@ Breakdown:
   * `6c6e626332306d0b25fe64570d0e496dbd9f8b0d000dbb44824f751380da37c6dba89b14f6f92047d63f576e304021a00008101820283038404800081018202830384048000810182028303840480810243500c318a1e0a628b34025e8c9019ab6d09b64c2b3c66a693d0dc63194b02481931000` hex of data for signing (prefix + data after separator up to the start of the signature)
   * `399a8b167029fda8564fd2e99912236b0b8017e7d17e416ae17307812c92cf42` hex of SHA256 of the preimage
 
+> ### Please send $30 for coffee beans to the same peer, which supports features 1 and 9
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9qzsze992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq73t7cl
+
+Breakdown:
+
+* `lnbc`: prefix, Lightning on Bitcoin mainnet
+* `25m`: amount (25 milli-bitcoin)
+* `1`: Bech32 separator
+* `pvjluez`: timestamp (1496314658)
+* `p`: payment hash...
+* `d`: short description
+  * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20)
+  * `vdhkven9v5sxyetpdees`: 'coffee beans'
+* `9`: features
+  * `qz`: `data_length` (`q` = 0, `z` = 2; 0 * 32 + 2 == 2)
+  * `sz`: b1000000010
+* `e992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq`: signature
+* `73t7cl`: Bech32 checksum
+
+> # Same, but using invalid unknown feature 100
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9q4pqqqqqqqqqqqqqqqqqqszk3ed62snp73037h4py4gry05eltlp0uezm2w9ajnerhmxzhzhsu40g9mgyx5v3ad4aqwkmvyftzk4k9zenz90mhjcy9hcevc7r3lx2sphzfxz7
+
+Breakdown:
+
+* `lnbc`: prefix, Lightning on Bitcoin mainnet
+* `25m`: amount (25 milli-bitcoin)
+* `1`: Bech32 separator
+* `pvjluez`: timestamp (1496314658)
+* `p`: payment hash...
+* `d`: short description
+  * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20)
+  * `vdhkven9v5sxyetpdees`: 'coffee beans'
+* `9`: features
+  * `q4`: `data_length` (`q` = 0, `4` = 21; 0 * 32 + 21 == 21)
+  * `pqqqqqqqqqqqqqqqqqqsz`: b00001...(90 zeroes)...1000000010
+* `k3ed62snp73037h4py4gry05eltlp0uezm2w9ajnerhmxzhzhsu40g9mgyx5v3ad4aqwkmvyftzk4k9zenz90mhjcy9hcevc7r3lx2sp`: signature
+* `hzfxz7`: Bech32 checksum
+
 # Authors
 
 [ FIXME: ]
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -9,6 +9,7 @@ over Lightning.
   * [Human-Readable Part](#human-readable-part)
   * [Data Part](#data-part)
     * [Tagged Fields](#tagged-fields)
+    * [Feature Bits](#feature-bits)
   * [Payer / Payee Interactions](#payer--payee-interactions)
     * [Payer / Payee Requirements](#payer--payee-requirements)
   * [Implementation](#implementation)
@@ -140,6 +141,9 @@ Currently defined tagged fields are:
    * `fee_base_msat` (32 bits, big-endian)
    * `fee_proportional_millionths` (32 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `9` (5): `data_length` variable. One or more bytes containing features
+  supported or required for receiving this payment.
+  See [Feature Bits](#feature-bits).
 
 ### Requirements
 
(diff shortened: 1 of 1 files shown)


rustyrussell:
As agreed at http://www.erisian.com.au/meetbot/lightning-dev/2019/lightning-dev.2019-08-05-20.03.html

11-payment-encoding.md
```diff
@@ -140,6 +141,9 @@ Currently defined tagged fields are:
    * `fee_base_msat` (32 bits, big-endian)
    * `fee_proportional_millionths` (32 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `9` (5): `data_length` variable. One or more bytes containing features
```
Roasbeef: Any particular reason why `9` is chosen here when the existing defined tags are letters?

11-payment-encoding.md
```diff
@@ -175,13 +182,21 @@ A writer:
...
   - if a writer offers more than one of any field type, it:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
 
 A reader:
   - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h`, or
   `n` fields that do NOT have `data_length`s of 52, 52, or 53, respectively.
+  - if the `9` field contains unknown _odd_ bits that are non-zero:
+    - MUST ignore the bit.
+  - if the `9` field contains unknown _even_ bits that are non-zero:
+    - MUST fail.
```
Roasbeef: What does fail in this context mean? Stop parsing, fail the payment, or both? 

In terms of delivering a better UX (say we want to pay, but the invoice only supports AMP and we don't), we should recommend the parser deliver a well specified error back to the caller, so this can eventually be propagated back to the user.

11-payment-encoding.md
```diff
@@ -509,6 +533,44 @@ Breakdown:
   * `6c6e626332306d0b25fe64570d0e496dbd9f8b0d000dbb44824f751380da37c6dba89b14f6f92047d63f576e304021a00008101820283038404800081018202830384048000810182028303840480810243500c318a1e0a628b34025e8c9019ab6d09b64c2b3c66a693d0dc63194b02481931000` hex of data for signing (prefix + data after separator up to the start of the signature)
   * `399a8b167029fda8564fd2e99912236b0b8017e7d17e416ae17307812c92cf42` hex of SHA256 of the preimage
 
+> ### Please send $30 for coffee beans to the same peer, which supports features 1 and 9
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9qzsze992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq73t7cl
```
cfromknecht: the current test vector (at least in my impl) produces an incomplete group when converting the feature bit data from base32 -> base256, i suspect we're disagreeing on a base 256 <-> base32 encoding somewhere?

for this i get:
```
lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9qyqgpq6w0yzlsl2aeqgg2v60eldr2cl6z59rcmlj255xz5pppewqw0mtpx5tywjnk7jna83jldf85z23k4zx2whl9q7jlzqge48ur3fk5ewcgqn5nkkk
```

with the vectors i arrived at i am able to encode and decode in both directions, but not with the ones in the proposal. @t-bast were you able to reproduce the original vectors?

11-payment-encoding.md
```diff
@@ -509,6 +533,44 @@ Breakdown:
...
+  * `q5`: `data_length` (`q` = 0, `5` = 20; 0 * 32 + 20 == 20)
+  * `vdhkven9v5sxyetpdees`: 'coffee beans'
+* `9`: features
+  * `qz`: `data_length` (`q` = 0, `z` = 2; 0 * 32 + 2 == 2)
+  * `sz`: b1000000010
+* `e992adudgku8p05pstl6zh7av6rx2f297pv89gu5q93a0hf3g7lynl3xq56t23dpvah6u7y9qey9lccrdml3gaqwc6nxsl5ktzm464sq`: signature
+* `73t7cl`: Bech32 checksum
+
+> # Same, but using invalid unknown feature 100
+> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9q4pqqqqqqqqqqqqqqqqqqszk3ed62snp73037h4py4gry05eltlp0uezm2w9ajnerhmxzhzhsu40g9mgyx5v3ad4aqwkmvyftzk4k9zenz90mhjcy9hcevc7r3lx2sphzfxz7
```
cfromknecht: similar here:
```
lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdees9q4zqqqqqqqqqqqqqqqqqpqyusqk43g0w3q2xh4k3uh70xmlz86juxnlffwspjuj2cxudw25yy55zs8xh37q9fk9n9rsvup7t3nms24cseext2u3tdek4g997k84edqq6ncfjf
```

11-payment-encoding.md
t-bast: Yes, with my implementation I end up with the same result as rusty's.
Indeed your vector disagrees on the `9` field, so it looks like this is linked to the bits encoding...could it be because you're not properly truncating the number of bits to a multiple of 5?

11-payment-encoding.md
t-bast: I think `9` was chosen to refer to Bolt 9 (which is where feature bits are usually defined - even though this might create confusion between node feature bits and invoice feature bits).

11-payment-encoding.md
cfromknecht: currently applying the same transformation as all other fields, can you confirm that the raw bytes before converting from base32 to base256 are `[0b10000, 0b10]`?

cfromknecht: Okay I did some digging, here are my findings.

The encoding of feature bits 1 and 9 uses 16 bits in base 256:
```
[0b00000010, 0b00000010]
```

Converting this to base32 requires `ceil(16/5) = 4` characters, and translates to:
```
[0b00000, 0b01000, 0b00001, 0b0====]
```
where `=` are padding bits.

I was able to reproduce this using sipa's [python `convertbits` method](https://paste.ubuntu.com/p/xFK9SsPYvt/).

The input `[0b10000, 0b00010]` fails to convert due to the same invalid group error I encountered. OTOH, `[0b00000, 0b01000, 0b00001, 0b00000]` (the encoding used in the vector I provided) successfully converts to `[0b00000010, 0b00000010]`.

Here's a paste of the experiment https://paste.ubuntu.com/p/xFK9SsPYvt/|

I'm not seeing how a valid vector could only have two base32 characters, i.e. 10 out of the 16 base256 bits.

IINM, this should be:
```
 * `9`: features
    * `qy` `data_length` (`q` = 0, `y` = 4; 0 * 32 + 4 == 4)
    * `qgpq`: 0b00000|01000|00001|00000
```

cfromknecht: FWIW I also reimplemented our convertbits method identically to the python one to rule out any possible bugs in our implementation and I get the same results. I looked about both eclair's and CL's converbit methods and they appear correct to me, so if the vector _is_ correct that means we are disagreeing somewhere else.

cfromknecht: Okay I think I understand what’s happening. I was incorrectly assuming that we would be reusing the wire encoding and converting to base32, rather than encoding to base32 directly. 

IMO this additional complexity and code duplication isn’t really warranted—it saves at most 2 characters and is dwarfed by things like route hints..

It’s also quite a departure from virtually every other complex field that already uses convertbits to get in and out of base32

t-bast: > I was incorrectly assuming that we would be reusing the wire encoding and converting to base32, rather than encoding to base32 directly.

Got it, I understand why your results diverged then. So you're able to reproduce the test vector now?
I think it makes sense to go directly to base32, it's clearer and easier to debug (once you know that's what's done :p).

> IMO this additional complexity and code duplication isn’t really warranted—it saves at most 2 characters and is dwarfed by things like route hints..

It's not code duplication for us, I think this is highly dependent on how that was implemented. Payment request encoding has been separate from message encoding from the start, hasn't it? I'll let rusty provide more details on that as I wasn't there when this was initially defined.

11-payment-encoding.md
rustyrussell: No, it seemed the least desirable "real estate", character-wise, and since it doesn't have a specific meaning, I thought I'd use it. `f` was already taken, of course.

11-payment-encoding.md
rustyrussell: Fail parsing, yes, and by implication not try to make the payment.

rustyrussell: We use the phrase `MUST fail the payment` elsewhere, so I've reused that:
```
   - MUST fail the payment
   - SHOULD indicate the offending bit index to the user.
```

11-payment-encoding.md
```diff
@@ -175,13 +182,21 @@ A writer:
         `fee_base_msat`, `fee_proportional_millionths`, and `cltv_expiry_delta` are as
         specified in [BOLT #7](07-routing-gossip.md#the-channel_update-message).
     - MAY include more than one `r` field to provide multiple routing options.
+  - if `9` contains non-zero bits:
+    - SHOULD use the minimum `data_length` possible.
```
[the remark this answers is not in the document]
rustyrussell: ... this same language is used elsewhere for minimal encoding requirements.

But I did change:
```diff
-* `9` (5): `data_length` variable. One or more bytes containing features
+* `9` (5): `data_length` variable. One or more 5-bit values containing features
```

11-payment-encoding.md
```diff
@@ -175,13 +182,22 @@ A writer:
         `fee_base_msat`, `fee_proportional_millionths`, and `cltv_expiry_delta` are as
         specified in [BOLT #7](07-routing-gossip.md#the-channel_update-message).
     - MAY include more than one `r` field to provide multiple routing options.
+  - if `9` contains non-zero bits:
```
halseth: Should be on the same format as the other fields? `MAY include one `9` field.`

11-payment-encoding.md
```diff
@@ -175,13 +182,22 @@ A writer:
...
   - if a writer offers more than one of any field type, it:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
 
 A reader:
   - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h`, or
   `n` fields that do NOT have `data_length`s of 52, 52, or 53, respectively.
+  - if the `9` field contains unknown _odd_ bits that are non-zero:
+    - MUST ignore the bit.
+  - if the `9` field contains unknown _even_ bits that are non-zero:
+    - MUST fail the payment.
```
halseth: "fail the payment" is perhaps ambiguous here. Instead maybe "MUST NOT pay the invoice"?

rustyrussell: It's the same language used elsewhere, so I think this is a separate issue?

11-payment-encoding.md
rustyrussell: No, it either MUST or MUST NOT.

PR #619, opened 2019-06-06
bolt04: Variable `hop_payload` for the sphinx onion

cdecker:
This is the variable payload length variant of #593. It removes the concept of fixed-length frames, in favor of just prefixing the payload with its varint-encoded length, followed by the HMAC.

The varint is limited to 1 and 3 byte encodings, since the maximum length of the payload is `1300 - 32 - 3` bytes, and the 5 and 9 byte varints are not useful. The HMAC is not part of the payload since it allows us to fit more of the shorter lengths in the single byte varint.

I didn't update the example code yet, since the Go version was not migrated yet, but it should not be too much work to update that later.

PR #658, opened 2019-08-16
BOLT04: Atomic Multi-path Payments [feature 30/31]

cfromknecht:
Hi all, decided to submit this draft concurrently with @rustyrussell's [Base AMP](#643) since there is likely a bit of overlap, and comments can likely be shared between the two. The structure of the proposal and requirements is based on #643 to make that process simpler. The chosen feature bits are just temporary.

## Background

This proposal outlines the requirements for Atomic Multi-path Payments originally outlined in @Roasbeef and I's [email to lightning-dev](https://lists.linuxfoundation.org/pipermail/lightning-dev/2018-February/000993.html). The proposal has also been referred to as OG AMP, or Moon AMP due to the moon-math involved :)

The core of the proposal is mostly unchanged, though I will highlight some distinctions:
 - The Extra Onion Blob (EOB) concept has been ditched in favor of the variable-sized hop-payloads via #619
 - Termination is signaled via `total_msat` (as in [Base AMP](#643)) instead of a transmitting the number of shares. This is more flexible when sending partial payments adaptively, since the required number of partial payments may not be known upfront, while the total amount being paid should be known. Committing to the total amount being paid also eliminates weaknesses in zero-value invoices, or any scenario involving overpayment.
 - The 11-byte payment ID is replaced with a full 32-byte `set_id`
 - Various modifications in nomenclature

## Summary

One of the biggest distinctions (some would argue drawback) from [Base AMP](#643) is that the sender does not receive a preimage as a result of a successful payment. However, the lack of a "proof" of payment unlocks a range of novel features:
 - Partial Payment Decorrelation: each subpayment uses a different payment hash, offering stronger privacy from intermediaries
 - Concurrent-Safe Reusable Invoices: As a by-product of the sender generating the preimages and payment hashes, an invoice is no longer single use. The payer conveys to the payee which invoice is being settled by including the invoice's payment hash. All payments made to the invoice can be tracked as a single subscription or account, and each payment receives it's own unique `set_id`. Multiple, concurrent payments can be made to the same invoice and properly reconstructed using the enclosed `set_id`.
    - A subscription invoice might be generated as an invoice with the desired subscription amount, which will be verified each time the user pays.
    - An account invoice, e.g. a recurring invoice for funding an exchange, would use a zero-value invoice, allowing deposits (or withdrawals) of arbitrary amounts.
 - Spontaneous AMPs: More often discussed in the single-shot case, this describes users making payments to other nodes using just their public key (handwaves hop hints). However, AMP is really a generalization of the single-shot case, so we propose to merge these two concepts under a single, unified proposal. All of the tangential benefits of AMP are carried over to spontaneous single-shot payments (SSSP?), without introducing any additional code complexity.

## Open Questions

 - Some of the fields included in `option_amp` are to be included in other proposals as well. Should we reuse the `total_msat` field from `option_mpp`, which can make use of the truncated encoding? Similarly, the `set_id` is analogous to the proposal to generically include a nonce in all payments, should that be it's own record? The signaling of which feature you are attempting to use may be more complex, but it's RISC-ier in some sense.

Feedback appreciated!

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -251,7 +251,7 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 10 (`option_amp`)
     2. data:
         * [`32*byte`:`payment_id`]
-        * [`32*byte`:`stream_id`]
+        * [`32*byte`:`set_id`]
         * [`32*byte`:`share`]
         * [`u16`:`child_index`]
         * [`u64`:`total_msat`]
@@ -272,12 +272,12 @@ The requirements for the contents of these fields are specified [above](#legacy-
 
 If the final node receives an onion packet with `option_amp` field,
 the payment MAY be an atomic multi-path payment. Such atomic multi-path payments
-MAY use a _distinct_ payment hash for each path.
+SHOULD use a _distinct_ payment hash for each path.
 
 The `amt_to_forward` value will be the amount for this partial payment only. The
-`option_amp` flag flag is a promise by the sender that the rest of the payment
-will follow in succeeding HTLCs with the same `stream_id`; we call these HTLCs,
-which that the same `stream_id`, an "HTLC set".
+`option_amp` flag is a promise by the sender that the rest of the payment will
+follow in succeeding HTLCs with the same `set_id`; we call these HTLCs, which
+share the same `set_id`, an "HTLC set".
 
 One key distinction with `option_amp` is that the sender generates all
 preimages, and only reveals them to the final hop if all partial payments arrive
@@ -293,10 +293,11 @@ The writer:
  - otherwise:
    - MAY include `option_amp` for the final node.
    - if it does include `option_amp`:
-     - MUST generate a random `stream_id` to be used on all HTLCs in the set.
-     - MAY send more than one HTLC using the same `stream_id`.
-     - MUST set the `share` values of all HTLCs such that their xor is a random
+     - MUST generate a random `set_id` to be used on all HTLCs in the set.
+     - MAY send more than one HTLC using the same `set_id`.
+     - MUST set the `share` values of all HTLCs such that their XOR is a random
        root seed `r`.
+       - SHOULD ensure all `share` values are unique and random.
      - SHOULD choose a unique child_index_i for each HTLC.
      - MUST derive the `payment_hash` for an HTLC using `amp_child(r, child_index_i)`.
      - if the invoice specifies a non-zero `amount`:
@@ -328,7 +329,7 @@ The reader:
   - MUST fail the entire HTLC set if `total_msat` is not the same for all HTLCs
     in the set.
   - if the total `amount_to_forward` of the HTLC set is equal `total_msat`:
-    - MUST reconstructs `r` as the xor of all `share`s in the HTLC set.
+    - MUST reconstructs `r` as the XOR of all `share`s in the HTLC set.
     - MUST compute `p_i, h_i = amp_child(r, child_index_i)` for all HTLCs in the
       set.
     - if any `i-th` HTLC's `payment_hash` differs from `h_i`:
@@ -336,7 +337,7 @@ The reader:
     - otherwise:
       - MAY fulfill the `i-th` HTLC in the set using `p_i`.
   - otherwise:
-    - MUST fail an HTLC in set if its `cltv_expiry` elapses.
+    - MUST fail an HTLC in the set if its `cltv_expiry` elapses.
     - MAY fail all HTLCs in the set after a reasonable timeout.
 
 ### Atomic Multi-path Payment Derivation
@@ -352,7 +353,8 @@ func amp_child(root_seed [32]byte, child_index uint16) ([32]byte, [32]byte) {
         return preimage, hash
 }
 ```
-where `child_index` is serialized using big-endian byte order.
+where `child_index` is serialized using big-endian byte order, and `||` denotes
+the concatenation of the left-hand side with the right-hand side.
 
 The sender will use `amp_child` to derive a child hash for each HTLC it sends
 out, and includes the `child_index` used in the derivation in the final hop's
@@ -368,12 +370,14 @@ represents an n-of-n secret sharing of `r`, such that:
     r = s_1 ^ ... ^ s_n
 ```
 
+where `^` denotes the XOR operation.
+
 If `n` is known upfront, satisfying this equation can be done simply by
 generating all `s_i` randomly.
 
 Otherwise, the sender can generate the shares _adaptively_ by first generating a
 random `r`. For all but the last outgoing HTLC, a random `s_i` is generated and
-included directly. The final HTLC then computes `s_n` as the xor of all other
+included directly. The final HTLC then computes `s_n` as the XOR of all other
 shares and `r`:
 
 ```
@@ -381,7 +385,7 @@ shares and `r`:
 ```
 
 If a partial payment fails, this process can be applied recursively to generate
-smaller partial payments, at the same time guaranteeing that the xor of all
+smaller partial payments, at the same time guaranteeing that the XOR of all
 shares results in `r`.
 
 This construction prevents the receiver from learning `r` until all `s_i` have
@@ -412,7 +416,7 @@ dependent variables. All independent variables are chosen upfront by the sender.
      |    |    |       H_i = SHA256(P_i)
      V    V    V     
                    
-    H_1  H_2  H_3    CHILD HASHES
+    H_1  H_2  H_3    CHILD PAYMENT HASHES
 ```
 
 ### Rationale
@@ -423,19 +427,32 @@ incoming AMP payment to a particular invoice. The sender should set the
 permitting the receiver to enforce custom parameters, e.g. CLTV deltas,  and
 unify tracking of AMP payments with the existing invoicing system.
 
-At the same time, AMP payments can be made spontaneously (without and invoice),
+At the same time, AMP payments can be made spontaneously (without an invoice),
 since the sender generates all of the necessary secrets. To do so, the sender
 leaves the `payment_id` blank, which can be used to facilitate secure donations.
 
 In the event that two payments are made with the same `payment_id`, either to
-the same invoice for both are spontaneous, a second identifier is introduced
-called the `stream_id`. The `stream_id` should be unique to each HTLC set sent
-by the sender, and allows the receiver to distinguish concurrent payments that
-collide on `payment_id`.
+the same invoice or both are spontaneous, a second identifier is introduced
+called the `set_id`. The `set_id` should be unique to each HTLC set sent by the
+sender, and allows the receiver to distinguish concurrent payments that collide
+on `payment_id`.
 
-Both the `stream_id` and `payment_id` are only known to sender and receiver,
+All payments made to the invoice can be tracked as a single subscription or
+account, and each payment receives it's own unique `set_id`. Multiple,
+concurrent payments can be made to the same invoice and properly reconstructed
+using the enclosed `set_id`. Two such use cases might involved "subscription" or
+"account" invoices:
+ * A subscription invoice might be generated as an invoice with the desired
+   subscription amount, which will be verified each time the user pays.
+ * An account invoice, e.g. a recurring invoice for funding an exchange, would
+   use a zero-value invoice, allowing deposits (or withdrawals) of arbitrary
+   amounts.
+
+Both the `set_id` and `payment_id` are only known to sender and receiver,
 preventing intermediaries from introducing griefing via colliding payment
-identifiers with high probability.
+identifiers with high probability. In addition, the `set_id` allows the receiver
+to create "subscription" or "account" invoices that can be paid repeatedly using
+different `set_id`s.
 
 The `child_index` is included in the final payload so that the receiver can
 gracefully tolerate reordering of the partial payments. When each `child_index`
@@ -446,7 +463,7 @@ The `total_msat` field is used to determine when all partial payments have been
 received. If the AMP is paying an invoice, this also allows the sender to
 securely overpay an invoice, for instance, if the invoice's `amount` is
 unspecified. If the AMP is spontaneous, this allows the sender to communicate
-the exact value to be received in a end-to-end authenticated manner, preventing
+the exact value to be received in an end-to-end authenticated manner, preventing
 certain classes of attacks where intermediaries can steal up to the overpaid
 amount.
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -248,6 +248,13 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 10 (`option_amp`)
+    2. data:
+        * [`32*byte`:`payment_id`]
+        * [`32*byte`:`stream_id`]
+        * [`32*byte`:`share`]
+        * [`u16`:`child_index`]
+        * [`u64`:`total_msat`]
 
 ### Requirements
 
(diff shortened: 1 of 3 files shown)


Roasbeef:
> Some of the fields included in option_amp are to be included in other proposals as well. Should we reuse the total_msat field from option_mpp, which can make use of the truncated encoding? 

I think that makes sense, as you can consider this to be a super set of the basic mpp in a sense. If we go in this direction, then should we switch to more of an "open coded" format for the AMP payload? This direction allows different use protocols to share common fields, but then create a blurry distinction between the allocated onion space amongst the various protocols. The other down side is that it consumes more type space, but with 64-bits available, that really isn't a concern IMO.

Roasbeef:
> Similarly, the stream_id is analogous to the proposal to generically include a nonce in all payments, should that be it's own record? The signaling of which feature you are attempting to use may be more complex, but it's RISC-ier in some sense.

If we go this route, then we may need to allocate an additional tag to indicate the context, so receivers know how to interpret this field. In the MPP case, it's just to be matched against the invoice, while in the AMP case, it may need to be passed to the receiver via some sort of hook for additional validation (the account ID had actually authorized a deposit, etc).

09-features.md
```diff
@@ -31,9 +31,10 @@ These flags may only be used in the `init` message:
 
 The following `globalfeatures` bits are currently assigned by this specification:
 
-| Bits | Name              | Description                                                        | Link                                  |
-|------|-------------------|--------------------------------------------------------------------|---------------------------------------|
-| 8/9  | `var_onion_optin` | This node requires/supports variable-length routing onion payloads | [Routing Onion Specification][bolt04] |
+| Bits  | Name              | Description                                                        | Link                                  |
+|-------|-------------------|--------------------------------------------------------------------|---------------------------------------|
+| 8/9   | `var_onion_optin` | This node requires/supports variable-length routing onion payloads | [Routing Onion Specification][bolt04] |
+| 12/13 | `option_amp`      | Payee supports AMP                                                 | [BOLT #4](04-onion-routing#atomic-multi-path-payments)
```
Roasbeef: I wonder if we need a way to signal or communicate interdependent feature bits, as the `option_amp` bits are dependent on the node also supporting the `var_onion_optin` bit as well. On the other-hand, one could say that `option_amp` implies the other, though some clients may be stricter and only send an AMP payment if the var onion bit is set as well. 

This also needs to define new feature bits and possibly tags in the BOLT 11 invoices as well.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
 
 The requirements for the contents of these fields are specified [above](#legacy-hop_data-payload-format).
 
+## Atomic Multi-path Payments
+
+If the final node receives an onion packet with `option_amp` field,
+the payment MAY be an atomic multi-path payment. Such atomic multi-path payments
+MAY use a _distinct_ payment hash for each path.
```
Roasbeef: Upgrade to SHOULD, in order to promote the gains from cross-path decorrelation?

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+One key distinction with `option_amp` is that the sender generates all
+preimages, and only reveals them to the final hop if all partial payments arrive
+successfully. As such, the payer _will not_ learn a new preimage as it would in
+the regular payment flow. For accounting purposes, however, `option_amp` can be
+used to fulfill an invoice akin to the regular payment flow, and also enforce
+additional constraints such as amounts and timelocks.
+
+The writer:
+ - MUST NOT include `option_amp` for any non-final node.
+ - if the sender has an invoice and `option_amp` feature was not set in the invoice:
```
Roasbeef: Should specify how this is to be communicated after #656 is merged.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+   - MUST NOT include `option_amp` for the final node.
+ - otherwise:
+   - MAY include `option_amp` for the final node.
+   - if it does include `option_amp`:
+     - MUST generate a random `stream_id` to be used on all HTLCs in the set.
+     - MAY send more than one HTLC using the same `stream_id`.
+     - MUST set the `share` values of all HTLCs such that their xor is a random
```
Roasbeef: MUST ensure all `share` values are unique?

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+function:
+
+```golang
+func amp_child(root_seed [32]byte, child_index uint16) ([32]byte, [32]byte) {
+        preimage := SHA256(root_seed || child_index)
+        hash := SHA256(preimage)
+        return preimage, hash
+}
+```
+where `child_index` is serialized using big-endian byte order.
```
Roasbeef: > and `||` denotes the concatenation of the left-hand side with the right-hand side

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+
+The sender will use `amp_child` to derive a child hash for each HTLC it sends
+out, and includes the `child_index` used in the derivation in the final hop's
+payload. The receiver will use `amp_child` to settle each HTLC with its
+corresponding child preimage, and also to verify that the correct child hash was
+set by the sender.
+
+In order to provide cryptographic atomicity over the fulfillment of an HTLC set,
+each partial payment `i` also transmits a 32-byte `share`.  Each share `s_i`
+represents an n-of-n secret sharing of `r`, such that:
```
Roasbeef: Should mention that `^` denotes the XOR operation.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+Otherwise, the sender can generate the shares _adaptively_ by first generating a
+random `r`. For all but the last outgoing HTLC, a random `s_i` is generated and
+included directly. The final HTLC then computes `s_n` as the xor of all other
+shares and `r`:
+
+```
+    s_n = s_1 ^ ... ^ s_n-1 ^ r
+```
+
+If a partial payment fails, this process can be applied recursively to generate
```
Roasbeef: Def an underrated feature of this scheme!

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+    P_1  P_2  P_3    CHILD PREIMAGES
+                   
+     |    |    |     
+     |    |    |       H_i = SHA256(P_i)
+     V    V    V     
+                   
+    H_1  H_2  H_3    CHILD HASHES
+```
+
+### Rationale
```
Roasbeef: Should we also briefly outline the re-occurring payment use case? Or do we consider this to be more of an application level protocol ,which should be defined in a distinct BIP?

Roasbeef (changes requested):

04-onion-routing.md
joostjager: Is there a problem with the receiver reading the share value, but then canceling the htlc with for example an invalid onion key error? To the sender is looks like a failure that may even have been caused by the second last node, but in reality the receiver has already obtained the root seed.

joostjager:
Could a proof of payment for atomic multi-path payments be achieved by locking the htlcs to two hashes (logical AND)? In order to settle such an htlc, the receiver needs to reveal both the preimage of the payment hash in the invoice and the preimage that was generated by the sender and embedded in the final onion payload.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+## Atomic Multi-path Payments
+
+If the final node receives an onion packet with `option_amp` field,
+the payment MAY be an atomic multi-path payment. Such atomic multi-path payments
+MAY use a _distinct_ payment hash for each path.
+
+The `amt_to_forward` value will be the amount for this partial payment only. The
+`option_amp` flag flag is a promise by the sender that the rest of the payment
+will follow in succeeding HTLCs with the same `stream_id`; we call these HTLCs,
+which that the same `stream_id`, an "HTLC set".
```
joostjager: I found `set_id` more descriptive than `stream_id`.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+       - MUST set `total_msat` to `amount`.
+     - otherwise:
+       - MUST set `total_msat` to the amount it wishes to pay.
+     - MUST ensure the total `amount_to_forward` in the HTLC set which arrives at
+       the payee is equal to `total_msat`.
+     - if the sender has an invoice:
+       - MUST set the `payment_id` of each HTLC to the `payment_hash` in the
+         invoice.
+     - otherwise:
+       - MUST set the `payment_id` of each HTLC to zero.
```
joostjager: Or define it as a separate tlv type? There may be some overlap with the random identifier generated by the receiver that was discussed before for regular payments and mpp. Both are ids generated by the receiver and both are not used to lock the htlc onto.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+  - if the total `amount_to_forward` of the HTLC set is equal `total_msat`:
+    - MUST reconstructs `r` as the xor of all `share`s in the HTLC set.
+    - MUST compute `p_i, h_i = amp_child(r, child_index_i)` for all HTLCs in the
+      set.
+    - if any `i-th` HTLC's `payment_hash` differs from `h_i`:
+      - MUST fail the HTLC set.
+    - otherwise:
+      - MAY fulfill the `i-th` HTLC in the set using `p_i`.
+  - otherwise:
+    - MUST fail an HTLC in set if its `cltv_expiry` elapses.
```
joostjager: Shouldn't it be failed earlier than that? In case of an (amp) hodl invoice, an application assumes that when the invoice is marked as accepted (we know the root seed, but haven't pulled yet), the invoice cltv expiry condition holds.

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+     ┌──┘ | └──┐       P_i = SHA256(R || child_index_i)
+     V    V    V     
+                   
+    P_1  P_2  P_3    CHILD PREIMAGES
+                   
+     |    |    |     
+     |    |    |       H_i = SHA256(P_i)
+     V    V    V     
+                   
+    H_1  H_2  H_3    CHILD HASHES
```
t-bast:
```suggestion
    H_1  H_2  H_3    CHILD PAYMENT HASHES
```

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+arrived. If `r` is successfully reconstructed, the receiver can verify the
+correctness of child hashes used in the HTLC set, and settle use the child
+preimages if they were offered correctly.
+
+The diagram below depicts the relationship between shares, the root seed, child
+preimages, and child hashes in the non-adaptive case. Lowercase variables are
+used to signal independent variables, while capital letters are used to describe
+dependent variables. All independent variables are chosen upfront by the sender.
+```
+    s_1  s_2  s_3    SHARES OF ROOT SEED
```
t-bast: This ascii art is great, it very neatly summarizes the scheme! You should make a t-shirt of it ;)

04-onion-routing.md
t-bast: I think it would be a good idea to briefly mention it (that would help application developers pick it up).

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+
+### Rationale
+
+The inclusion of a `payment_id` in `option_amp` allows the receiver to map an
+incoming AMP payment to a particular invoice. The sender should set the
+`payment_id` to the `payment_hash` of the invoice they are trying to pay,
+permitting the receiver to enforce custom parameters, e.g. CLTV deltas,  and
+unify tracking of AMP payments with the existing invoicing system.
+
+At the same time, AMP payments can be made spontaneously (without and invoice),
```
t-bast:
```suggestion
At the same time, AMP payments can be made spontaneously (without an invoice),
```

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+since the sender generates all of the necessary secrets. To do so, the sender
+leaves the `payment_id` blank, which can be used to facilitate secure donations.
+
+In the event that two payments are made with the same `payment_id`, either to
+the same invoice for both are spontaneous, a second identifier is introduced
```
t-bast:
```suggestion
the same invoice or both are spontaneous, a second identifier is introduced
```

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+The `child_index` is included in the final payload so that the receiver can
+gracefully tolerate reordering of the partial payments. When each `child_index`
+is unique, this offers decorrelation of the partial payments, since they bear
+different child payment hashes while traversing the network.
+
+The `total_msat` field is used to determine when all partial payments have been
+received. If the AMP is paying an invoice, this also allows the sender to
+securely overpay an invoice, for instance, if the invoice's `amount` is
+unspecified. If the AMP is spontaneous, this allows the sender to communicate
+the exact value to be received in a end-to-end authenticated manner, preventing
```
t-bast:
```suggestion
the exact value to be received in an end-to-end authenticated manner, preventing
```

04-onion-routing.md
```diff
@@ -261,6 +268,197 @@ The reader:
...
+amount.
+
+The entire payment is contingent on the receiver being able to reconstruct the
+root seed `r`, which prevents the receiver from pulling any of the partial
+payments until the entirety of the HTLC set arrives. This is enforced by the
+n-of-n secret shares provided in the final hop payload of each arriving partial
+payment.
+
+None of the requirements enforce that more than one HTLC is sent, permitting the
+base case of 1 HTLC to function as a standalone spontaneous payment.
```
t-bast: :clap: I think this is important and this is why I view this work more as a "spontaneous payments" overall feature

t-bast:
Great stuff @cfromknecht I think this adds very valuable features.

One quick thought about share failures.
Imagine the sender uses 3 shares and one fails. In order to preserve `r`, the sender has to either re-use the same share (with a different index of course) or split that share into multiple payments. If it's the recipient that failed one of the shares, the recipient will have learnt `r`. Could this be an issue (if for example the recipient has other nodes that may end up being intermediate nodes that route shares)?
 
Not sure it can be exploited, but just putting it out there to make sure it's obvious to everyone.

04-onion-routing.md
joostjager: Example:
* current height: 100
* (hodl) invoice cltv delta: 40, amt: 100
* first htlc for 50 sats comes in with expiry 140 -> accepted as partial payment
* new block comes in
* second htlc for 50 sats comes in with expiry 141 -> accepted as partial payment

The hodl invoice should now move to the accepted state, because enough has been paid. We'd send the `accepted` event to rpc subscribers. But at that point, there are only 39 blocks left before the first htlc expires. The subscriber will probably assume that when the event comes in, the final cltv delta of 40 is still met.

cfromknecht:
Based on discussion in #643, mpp will include a receiver-generated nonce rather than a sender-generated one. Therefore the prior split would be better defined as:

```
    1. type 8 (`option_mpp`)
    2. data:
        * [`32*byte`:`payment_addr`]
        * [`u64`:`total_msat`]
    1. type: 10 (`option_amp`)
    2. data:
        * [`32*byte`:`set_id`]
        * [`32*byte`:`share`]
        * [`u16`:`child_index`]
```

04-onion-routing.md
```diff
@@ -248,6 +248,13 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
     1. type: 6 (`short_channel_id`)
     2. data:
         * [`short_channel_id`:`short_channel_id`]
+    1. type: 10 (`option_amp`)
+    2. data:
+        * [`32*byte`:`payment_id`]
```
joostjager: For recurring payments, the idea with AMP is that `payment_id` can be used as a type of 'deposit box', similar to a traditional bank account number. 

But this field is also used to prevent multiple people trying to pay to the same invoice simultaneously from disturbing each others sets.

It seems that both uses of the same field are mutually exclusive?

joostjager: Ok, clarified offline. That is what `set_id` is for

09-features.md
cfromknecht: resolved in #719

PR #660, opened 2019-08-19
BOLT 4: Define custom tlv type range

joostjager:
This PR defines a type range for custom tlv records. The goal is to prevent unofficial fields from allocating identifiers in the official range. This could cause problems when new fields are added to this specification and turn out to collide with custom records that are used in the wild.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -108,7 +108,11 @@ messages, a `tlv_stream` is typically placed after all currently defined fields.
 
 The `type` is a varint encoded using the BigSize format. It functions as a
 message-specific, 64-bit identifier for the `tlv_record` determining how the
-contents of `value` should be decoded.
+contents of `value` should be decoded. `type` identifiers below 2^16 are
+reserved for use in this specification. `type` identifiers greater than or equal
+to 2^16 are available for custom records. Any record not defined in this
+specification is considered a custom record. This includes experimental and
+application-specific messages.
 
 The `length` is a varint encoded using the BigSize format signaling the size of
 `value` in bytes.
@@ -121,6 +125,13 @@ according to the message-specific format determined by `type`.
 The sending node:
  - MUST order `tlv_record`s in a `tlv_stream` by monotonically-increasing `type`.
  - MUST minimally encode `type` and `length`.
+ - When defining custom record `type` identifiers:
+   - SHOULD pick random `type` identifiers to avoid collision with other
+     custom types.
+   - SHOULD pick odd `type` identifiers when regular nodes should ignore the
+     additional data.
+   - SHOULD pick even `type` identifiers when regular nodes should reject the
+     full tlv stream containing the custom record.
  - SHOULD NOT use redundant, variable-length encodings in a `tlv_record`.
 
 The receiving node:


t-bast:
I think what you suggest is similar to #623 but less general, WDYT?
I think that if we want to explicitly allow something like that in the spec, we'd want it more generally than only for Bolt 4 so #623 makes more sense.

joostjager:
I don't think it is the same. Afaik #623 is about extensions of the node software itself, while custom tlv fields can be used with unmodified nodes to communicate application specific values from sender to receiver. For example an account number to deposit the funds onto. This isn't something that would be signaled on the p2p message level.

The point is that we already allow custom fields in the spec. As long as they are odd-numbered, they are accepted by receivers. The goal of this PR is to prevent informal pollution of the official range before it is too late.

t-bast:
> custom tlv fields can be used with unmodified nodes to communicate application specific values from sender to receiver

Can you detail how you achieve that? This isn't clear at all to me, if the node doesn't explicitly add support I don't see how this would happen with today's spec and implementations.

I agree we need to avoid polluting tlv namespaces, but from my point of view right now anything that uses a tlv not registered in the spec is doomed to be overridden at some point and "split away" from the network. I'm not sure what we should do to make this better.

joostjager:
The way to achieve this is to accept a map of <int, []byte> on the send payment call. Those bytes will be encoded literally in the final hop payload with the given (custom) keys. At the receiver, the same map will be extracted and signaled to an application in the 'invoice settled' event.

The node itself doesn't need to know what those values are. It is just passing them on along with the payment.

Alternatively, a single official type `user_data` could be defined that contains an opaque byte array with whatever custom data the application requires.

I wouldn't want to generalize this to features to allow experimentation on every level.

t-bast:
> Alternatively, a single official type user_data could be defined that contains an opaque byte array with whatever custom data the application requires.

I think that would make more sense (if I understand your use-case correctly) than encouraging people to use new tlv types that they would choose and which wouldn't be spec-ed.

t-bast:
Summarizing feedback from the IRC meeting: this should probably move to Bolt 1 to apply to tlv types in all namespaces.

niftynei:
Thinking about this further, opening up the TLV's of every message to any inserted format basically allows the passing of arbitrary data over the network. Restricting the 'anything goes' to just the onion payload means that these extensions are limited to 'paid' messages, which seems like a reasonable tradeoff.  The space is also a bit more limited by nature.

Allowing 3rd party devs to add optional information to opening messages and the like seems like it'd offer a wider set of applications to use lightning as rails for their projects. I can see advantages and disadvantages to this level of extensibility.

Strongly in favor of adopting this for the onion TLV's, however. Which seems more short-term practicable anyway given that these already have TLVs.

joostjager:
Moved definition of custom type range to BOLT 1.

Also changed range from >=2^32 to >=2^16. It seems that 65536 official types is enough (...) and prevents annoyances in dealing with long integers for custom types.

t-bast:
> Allowing 3rd party devs to add optional information to opening messages and the like seems like it'd offer a wider set of applications to use lightning as rails for their projects. I can see advantages and disadvantages to this level of extensibility.

TBH we are already using this outside of the onion payload (to experiment with gossip, routing and static backup improvements). So if we're doing it, it's likely that LND and CL will be doing it as well, and other devs too. So I'd like to define this range for all tlvs to make our experimentations spec-compliant ;)

01-messaging.md
```diff
@@ -108,7 +108,8 @@ messages, a `tlv_stream` is typically placed after all currently defined fields.
 
 The `type` is a varint encoded using the BigSize format. It functions as a
 message-specific, 64-bit identifier for the `tlv_record` determining how the
-contents of `value` should be decoded.
+contents of `value` should be decoded. A custom type is defined as any type
```
t-bast: Probably nitpicking but I don't find this paragraph clear enough...
What do you think of:

```
`type` identifiers below 2^16 are reserved for use in this specification.
`type` identifiers above 2^16 can be used for experiments and application-specific messages.
```

And in the sending node requirements:

```
- When defining `type` identifiers for experiments or application-specific messages:
  - MUST use `type` identifiers >= 2^16
  - SHOULD pick random `type` identifiers to avoid collision with other experiments
  - SHOULD pick odd `type` identifiers to allow normal nodes to safely ignore them
```

01-messaging.md
```diff
@@ -143,6 +145,7 @@ The receiving node:
      - MUST fail to parse the `tlv_stream`.
    - otherwise, if `type` is odd:
      - MUST discard the next `length` bytes.
+   - the odd/even rule remains in effect for custom types.
```
t-bast: I don't think this is necessary. This is already implied by the current spec imo.

04-onion-routing.md
```diff
@@ -255,9 +255,12 @@ The writer:
   - MUST include `amt_to_forward` and `outgoing_cltv_value` for every node.
   - MUST include `short_channel_id` for every non-final node.
   - MUST NOT include `short_channel_id` for the final node.
+  - MAY include application-specific [custom tlv records](01-messaging.md#type-length-value-format).
```
t-bast: I don't think that the changes in Bolt 4 are necessary.
The changes you made in Bolt 1 already imply that.
I would remove all changes to this file, wdyt?

01-messaging.md
joostjager: It is implied, but I added it explicitly because the question was raised during the irc meeting. I was wondering about this myself too and it could be that the reader of the spec does the same.

04-onion-routing.md
joostjager: Bolt 1 defines the range, but I think that we should still detail the exact behavior everywhere we use tlv.

t-bast: Meh, I think this applies to everywhere we use TLV so I think it would be weird to add this line to every message requirement (which is what we'll end up doing since at some point every message will contain a tlv stream). But it's nit-picking :)

01-messaging.md
joostjager: Moved definition of the range into this paragraph as you suggested. I did keep the definition of the term 'custom record' to avoid referring to experiments, app-specific message, etc in multiple places.

Also rewrote it so that the range (> 2^16) is only defined in one place.

And added statement about even `type` ids, which could also be what should be enforced for the custom record.

01-messaging.md
joostjager: Removed this, as it is also explained in the section 'when defining custom types'

04-onion-routing.md
joostjager: I've come round to removing all changes in this file as you suggested. I agree it is clear enough with only the BOLT1 changes.

cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/660/commits/e05e34ab4b99a4fb44d89150ee2eae5821a67d0e

PR #663, opened 2019-08-30
BOLT04: use truncated u64 encoding for next_hop_id

cfromknecht:
As per [discussion in this thread](https://github.com/lightningnetwork/lightning-rfc/pull/619#discussion_r318875465), this PR changes the `short_channel_id` onion record (type 8) from `short_channel_id` to `tu64`, which shaves 8 bytes for the final hop when the `short_channel_id` is always 0.

@cdecker do you have any thoughts on [how do to this mapping](https://github.com/lightningnetwork/lightning-rfc/pull/619#discussion_r319075787)? Is that something that should be in this PR?

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -247,7 +247,7 @@ This is a more flexible format, which avoids the redundant `short_channel_id` fi
         * [`tu32`:`outgoing_cltv_value`]
     1. type: 6 (`short_channel_id`)
     2. data:
-        * [`short_channel_id`:`short_channel_id`]
+        * [`tu64`:`short_channel_id`]
 
 ### Requirements
 


rustyrussell:
Huh?  Why would this field appear at all in the final hop?

t-bast:
Doh, rusty's totally right. Currently our implementation simply omits this tlv field entirely for the last hop. And that should settle it :)

cdecker:
During the IRC spec meeting on 2019-09-02 we found out that any current `short_channel_id` would already require 59 bytes at least to encode, which would result in us always using 8 bytes for any realistic `short_channel_id`, except for the last hop.

As @rustyrussell points out the [`short_channel_id` MUST be omitted for the last hop](https://github.com/lightningnetwork/lightning-rfc/blob/6729755f0c0228df0686176e99032fb368ca056a/04-onion-routing.md#requirements). So this is not really a saving after all.

cfromknecht:
> As @rustyrussell points out the short_channel_id MUST be omitted for the last hop. So this is not really a saving after all.

Ah thanks @rustyrussell! Looks instead like we need to fix this on our end, as we were including it for the final hop.

PR #664, opened 2019-09-02
BOLT #3: Correct a typo in 'revocationpubkey derivation'

darosior:
I think `pubkey` was meant instead of `privkey`.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -481,7 +481,7 @@ And this is used to derive the revocation pubkey from the remote node's
 
 This construction ensures that neither the node providing the
 basepoint nor the node providing the `per_commitment_point` can know the
-private key without the other node's secret.
+public key without the other node's secret.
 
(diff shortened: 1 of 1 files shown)


t-bast:
I think the current wording is correct.
We really mean private key here.
Both peers know the public key (because it's computed from basepoints) but in order to know the corresponding private key (to spend a revoked commitment) you need the private part of either the revocation part of the per-commitment part.

t-bast:
The key derivation part is tricky and can be confusing at times ;)

darosior:
> I think the current wording is correct.
> We really mean private key here.
> Both peers know the public key (because it's computed from basepoints) but in order to know the corresponding private key (to spend a revoked commitment) you need the private part of either the revocation part of the per-commitment part.

Ok, reading it again it seems like I've got confused ^^. Thanks

PR #665, opened 2019-09-02
BOLT #3: Indentation consistency in key derivation algorithms

darosior:
NIT but it rendered poorly when tabs where used (too much spaced) whereas spaces were previously used.


t-bast:
Merging per typo-rule.

PR #666, opened 2019-09-04
Flat features: far simpler solution for feature bits.

rustyrussell:
We simply specify how they're presented in the three different contexts.

Consider these theoretical future features:

`opt_dlog_chan`: a new channel type which uses a new discrete log HTLC
type, but can't support traditional HTLC:

* `init`: presents as odd (optional) or even (if traditional channels
  not supported)
* `node_announcement`: the same as above, so you can seek suitable peers.
* `channel_announcement`: presents as even (compulsory), since users need
  to use the new HTLCs.

`opt_wumbochan`: a node which allows channels > 2^24 satoshis:

* `init`: presents as odd (optional), or maybe even (if you only want
  giant channels)
* `node_announcement`: the same as above, so you can seek suitable peers.
* `channel_announcement`: not present, since size of channel indicates
  capacity.

`opt_wumbohtlc`: a channel which allows HTLCs > 2^32 millisatoshis:

* `init`: presents as odd (optional), or even (compulsory)
* `node_announcement`: the same as above, so you can seek suitable peers.
* `channel_announcement`: odd (optional) since you can use the channel
  without understanding what this option means.

Closes: #571

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -226,18 +226,17 @@ The following convenience types are also defined:
 
 Once authentication is complete, the first message reveals the features supported or required by this node, even if this is a reconnection.
 
-[BOLT #9](09-features.md) specifies lists of global and local features. Each feature is generally represented in `globalfeatures` or `localfeatures` by 2 bits. The least-significant bit is numbered 0, which is _even_, and the next most significant bit is numbered 1, which is _odd_.
+[BOLT #9](09-features.md) specifies lists of features. Each feature is generally represented by 2 bits. The least-significant bit is numbered 0, which is _even_, and the next most significant bit is numbered 1, which is _odd_.  For historical reasons, features are divided into global and local feature bitmasks.
 
-Both fields `globalfeatures` and `localfeatures` MUST be padded to bytes with 0s.
+The `features` field MUST be padded to bytes with 0s.
 
 1. type: 16 (`init`)
 2. data:
    * [`u16`:`gflen`]
    * [`gflen*byte`:`globalfeatures`]
-   * [`u16`:`lflen`]
-   * [`lflen*byte`:`localfeatures`]
+   * [`u16`:`flen`]
+   * [`flen*byte`:`features`]
 
-The 2-byte `gflen` and `lflen` fields indicate the number of bytes in the immediately following field.
 
 #### Requirements
 
@@ -245,10 +244,12 @@ The sending node:
   - MUST send `init` as the first Lightning message for any connection.
   - MUST set feature bits as defined in [BOLT #9](09-features.md).
   - MUST set any undefined feature bits to 0.
-  - SHOULD use the minimum lengths required to represent the feature fields.
+  - SHOULD NOT set features greater than 13 in `globalfeatures`.
+  - SHOULD use the minimum length required to represent the `features` field.
 
 The receiving node:
   - MUST wait to receive `init` before sending any other messages.
+  - MUST combine (logical OR) the two feature bitmaps into one logical `features` map.
   - MUST respond to known feature bits as specified in [BOLT #9](09-features.md).
   - upon receiving unknown _odd_ feature bits that are non-zero:
     - MUST ignore the bit.
@@ -257,14 +258,13 @@ The receiving node:
 
 #### Rationale
 
+There used to be two feature bitfields here, but for backwards compatibility they're now
+combined into one.
+
 This semantic allows both future incompatible changes and future backward compatible changes. Bits should generally be assigned in pairs, in order that optional features may later become compulsory.
 
 Nodes wait for receipt of the other's features to simplify error
 diagnosis when features are incompatible.
-
-The feature masks are split into local features (which only affect the
-protocol between these two nodes) and global features (which can affect
-HTLCs and are thus also advertised to other nodes).
 
 ### The `error` Message
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -188,9 +188,7 @@ The receiving node:
   - MUST verify the integrity AND authenticity of the message by verifying the
   signatures.
   - if there is an unknown even bit in the `features` field:
-    - MUST NOT parse the remainder of the message.
-    - MUST NOT add the channel to its local network view.
-    - SHOULD NOT forward the announcement.
+    - MUST NOT attempt to route messages through the channel.
   - if the `short_channel_id`'s output does NOT correspond to a P2WSH (using
     `bitcoin_key_1` and `bitcoin_key_2`, as specified in
     [BOLT #3](03-transactions.md#funding-transaction-output)) OR the output is
@@ -244,8 +242,6 @@ New channel features are possible in the future: backwards compatible (or
 optional) features will have _odd_ feature bits, while incompatible features
 will have _even_ feature bits
 (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide)).
-Incompatible features will result in the announcement not being forwarded by
-nodes that do not understand them.
 
 ## The `node_announcement` Message
 
@@ -326,11 +322,7 @@ any future fields appended to the end):
     - SHOULD fail the connection.
     - MUST NOT process the message further.
   - if `features` field contains _unknown even bits_:
-    - MUST NOT parse the remainder of the message.
-    - MAY discard the message altogether.
     - SHOULD NOT connect to the node.
-  - MAY forward `node_announcement`s that contain an _unknown_ `features` _bit_,
-  regardless of if it has parsed the announcement or not.
   - SHOULD ignore the first `address descriptor` that does NOT match the types
   defined above.
   - if `addrlen` is insufficient to hold the address descriptors of the
@@ -354,8 +346,9 @@ any future fields appended to the end):
 
 New node features are possible in the future: backwards compatible (or
 optional) ones will have _odd_ `feature` _bits_, incompatible ones will have
-_even_ `feature` _bits_. These may be propagated by nodes even if they
-cannot process the announcements themselves.
+_even_ `feature` _bits_. These will be propagated normally; incompatible
+feature bits here refer to the nodes, not the `node_announcement` message
+itself.
 
 New address types may be added in the future; as address descriptors have
 to be ordered in ascending order, unknown ones can be safely ignored.
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -180,7 +180,8 @@ The origin node:
   - MUST set `bitcoin_signature_1` and `bitcoin_signature_2` to valid
   signatures of the hash `h` (using `bitcoin_key_1` and `bitcoin_key_2`'s
   respective secrets).
-  - SHOULD set `len` to the minimum length required to hold the `features` bits
+  - MUST set `features` based on what features were negotiated for this channel, according to [BOLT #9](09-features.md#assigned-features-flags)
+  - MUST set `len` to the minimum length required to hold the `features` bits
   it sets.
 
 The receiving node:
@@ -311,6 +312,7 @@ The origin node:
   to 0.
   - SHOULD ensure `ipv4_addr` AND `ipv6_addr` are routable addresses.
   - MUST NOT include more than one `address descriptor` of the same type.
+  - MUST set `features` according to [BOLT #9](09-features.md#assigned-features-flags)
   - SHOULD set `flen` to the minimum length required to hold the `features`
   bits it sets.
 
--- a/09-features.md
+++ b/09-features.md
@@ -24,6 +24,7 @@ The Context column decodes as follows:
 * `C`: presented in the `channel_announcement` message.
 * `C-`: presented in the `channel_announcement` message, but always odd (optional).
 * `C+`: presented in the `channel_announcement` message, but always even (required).
+* `9`: presented in [BOLT 11](11-payment-encoding.md) invoices.
 
 | Bits  | Name                             | Description                                               | Context  | Link                                  |
 |-------|----------------------------------|-----------------------------------------------------------|----------|---------------------------------------|
--- a/09-features.md
+++ b/09-features.md
@@ -1,14 +1,10 @@
 # BOLT #9: Assigned Feature Flags
 
-This document tracks the assignment of `localfeatures` and `globalfeatures`
-flags in the `init` message ([BOLT #1](01-messaging.md)) along with the
-`features` flag fields in the `channel_announcement` and `node_announcement`
-messages ([BOLT #7](07-routing-gossip.md)).
-The flags are tracked separately, since new flags will likely be added over time.
-
-The `features` flags in the routing messages are a subset of the
-`globalfeatures` flags, as `localfeatures`, by definition, are only of interest
-to direct peers.
+This document tracks the assignment of `features` flags in the `init`
+message ([BOLT #1](01-messaging.md)), as well as `features` fields in
+the `channel_announcement` and `node_announcement` messages ([BOLT
+#7](07-routing-gossip.md)).  The flags are tracked separately, since
+new flags will likely be added over time.
 
 Flags are numbered from the least-significant bit, at bit 0 (i.e. 0x1,
 an _even_ bit). They are generally assigned in pairs so that features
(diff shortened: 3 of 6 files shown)


09-features.md
```diff
@@ -1,43 +1,53 @@
...
+* `C+`: presented in the `channel_announcement` message, but always even (required).
 
+| Bits | Name                             | Description                                               | Context  | Link                               |
+|------|----------------------------------|-----------------------------------------------------------|----------|------------------------------------|
+| 0/1  | `option_data_loss_protect`       | Requires or supports extra `channel_reestablish` fields   | IN       | [BOLT #2][bolt02-retransmit]          |
+| 3    | `initial_routing_sync`           | Sending node needs a complete routing information dump    | I        | [BOLT #7][bolt07-sync]               |
+| 4/5  | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel   | IN       | [BOLT #2][bolt02-open]               |
+| 6/7  | `gossip_queries`                 | More sophisticated gossip control                         | IN       | [BOLT #7][bolt07-query]              |
+| 8/9  | `var_onion_optin`                | Requires/supports variable-length routing onion payloads  | IN       | [Routing Onion Specification][bolt04] |
```
t-bast: Shouldn't this be `INC`?
This could allow nodes to force some of their channels to use variable-length onions (and explicitly disallow the legacy fixed-frames) by setting bit 8 in the `channel_update` of such channels.
Not sure whether it's really useful or not, but maybe worth discussing?

07-routing-gossip.md
```diff
@@ -180,16 +180,15 @@ The origin node:
...
   it sets.
 
 The receiving node:
   - MUST verify the integrity AND authenticity of the message by verifying the
   signatures.
   - if there is an unknown even bit in the `features` field:
-    - MUST NOT parse the remainder of the message.
-    - MUST NOT add the channel to its local network view.
-    - SHOULD NOT forward the announcement.
+    - MUST NOT attempt to route messages through the channel.
```
t-bast: Are you removing `SHOULD NOT forward the announcement` because we now should forward it, or because the flow of `if` statements below implies that we still shouldn't forward it in this case (and it was redundant to mention it here)?

01-messaging.md
```diff
@@ -215,26 +215,25 @@ The following convenience types are also defined:
...
+   * [`u16`:`flen`]
+   * [`flen*byte`:`features`]
 
-The 2-byte `gflen` and `lflen` fields indicate the number of bytes in the immediately following field.
 
 #### Requirements
 
 The sending node:
   - MUST send `init` as the first Lightning message for any connection.
+  - MUST set `mbz` to zero.
```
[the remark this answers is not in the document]
rustyrussell: Not backward compatible.  Currently we have no deployed global bits, so we can do the mbz trick.

09-features.md
rustyrussell: It's tempting, since node_announcement propagation is currently less reliable than channel_announcements.  But channel_announcements can't be updated, so current channels would never set this bit, even though they'd be perfectly fine.

So I think it needs to be a Node bit.

07-routing-gossip.md
rustyrussell: We now forward it (see commit msg).

01-messaging.md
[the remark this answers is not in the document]
t-bast: We do have a global feature bit spec-ed (variable-length onion).
For eclair we haven't made a release that advertises it yet, but if people deployed nodes from master then they will be sending something in the global features field.
So our code would need to handle both the previous format and the new with `mbz` (at least for a while)...this isn't too painful to implement though (just branch on `mbz` being `0` or not).

07-routing-gossip.md
t-bast: Got it, I think it makes sense to forward.

01-messaging.md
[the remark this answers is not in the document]
rustyrussell: Yes, since it's not deployed I was hoping we could sidestep this... but perhaps not.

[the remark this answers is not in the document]
Roasbeef: What does `mbz` mean here? "Must be zero"?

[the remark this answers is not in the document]
Roasbeef: We haven't cut a release either using the new global feature bits, but are very close to doing so....

As a result, would be nice to have all this clarified as otherwise with our target release schedule, we may cut a version that still uses the existing global feature bits. 

Stepping back for a second, this `mbz` change doesn't really strike me as being backwards compatible. We're potentially fracturing the network here (let's say people are using a fork of some implementation in the wild and are using global feature bits to gate their protocol). Instead we can simply _rename_ the field in the spec: global isn't bound to anything (don't set a name), and local becomes the `features` as is here. For some period of time, we should still also merge the two fields together.

Roasbeef:
Why aren't the three scenarios in this PR body expressible using the current set of feature bit fields and names? We can just leave the existing local feature bits as is, but then start to define all new bits as a global feature bit. The node and channel announcement fields would then only inherit the name space from the global feature bits.

Roasbeef:
(also this PR is evil, just saying)

rustyrussell:
> Why aren't the three scenarios in this PR body expressible using the current set of feature bit fields and names? We can just leave the existing local feature bits as is, but then start to define all new bits as a global feature bit. The node and channel announcement fields would then only inherit the name space from the global feature bits. Even in this latest iteration, I still have a nagging feeling that this isn't really fixing anything.

Because everyone got confused about the two schemes, yet a single scheme also sucks, because optional and compulsory are different in different contexts:

1. My new node insists a new crypto peer comms (EVEN bit in init and node_announcement).  But it doesn't matter for channels, so why would I set this to channel_announce?  That would stop old peers from routing through them.

2. My new node allows funky log-based channels (ODD bit in init and node_announcement).  You do too, and we negotiate it.  For that channel, it has to be EVEN in channel_announce, since you can't use the channel unless you understand the log-based scheme.

3. My node only allows log-based channels except for existing channels (EVEN bit in node_announcment, odd bit in init messages to legacy peers).

TLDR: much easier to think about what each feature needs, and specify it at that level.

rustyrussell:
Rebased, restored gflen field.

01-messaging.md
```diff
@@ -215,26 +215,26 @@ The following convenience types are also defined:
...
 
 #### Requirements
 
 The sending node:
   - MUST send `init` as the first Lightning message for any connection.
+  - MUST set `ignorelen` to zero.
   - MUST set feature bits as defined in [BOLT #9](09-features.md).
   - MUST set any undefined feature bits to 0.
-  - SHOULD use the minimum lengths required to represent the feature fields.
+  - SHOULD use the minimum length required to represent the `feature` field.
```
t-bast: nit:
```suggestion
  - SHOULD use the minimum length required to represent the `features` field.
```

t-bast:
It would be nice to make quick progress there; this way we're not slowing down releases.
ACK ` 2f5176e`, as long as no-one made an explicit release advertizing `var_onion_optin` in the global flags we can safely start ignoring those (otherwise if lnd does release before we merge this PR, we'll need our code to temporarily support both schemes which is a bit awkward).

09-features.md
```diff
@@ -1,43 +1,54 @@
...
+each feature defines how it is presented in those contexts.  Some
+features may be required for opening a channel, but not a requirement
+for use of the channel, so the presentation of those features depends
+on the feature itself.
 
-The following `globalfeatures` bits are currently assigned by this specification:
+The Context column decodes as follows:
+* `I`: presented in the `init` message.
+* `N`: presented in the `node_announcement` messages
+* `C`: presented in the `channel_announcement` message.
```
Roasbeef: What about the channel update message? We also need possibly extend the existing hop hints to also encompass this information. One could say the current set of feature bits handles that, however it's possible to provide multiple hop hints, and each of those channels could themselves advertise/require distinct sets of hop hints.

Roasbeef:
> It would be nice to make quick progress there; this way we're not slowing down releases.

Our release is late as is, and it's now being held up by this PR as it creates a dependency with the static remote key PR. We're now schedule to tag next week. As for the current state of the global feature bits: the static key PR is more or less finalized/ready but is now blocked on the inclusion of this PR, the onion tlv feature has itself long been included.

09-features.md
```diff
@@ -1,43 +1,54 @@
...
+| Bits  | Name                             | Description                                               | Context  | Link                                  |
+|-------|----------------------------------|-----------------------------------------------------------|----------|---------------------------------------|
+| 0/1   | `option_data_loss_protect`       | Requires or supports extra `channel_reestablish` fields   | IN       | [BOLT #2][bolt02-retransmit]          |
+| 3     | `initial_routing_sync`           | Sending node needs a complete routing information dump    | I        | [BOLT #7][bolt07-sync]                |
+| 4/5   | `option_upfront_shutdown_script` | Commits to a shutdown scriptpubkey when opening channel   | IN       | [BOLT #2][bolt02-open]                |
+| 6/7   | `gossip_queries`                 | More sophisticated gossip control                         | IN       | [BOLT #7][bolt07-query]               |
+| 8/9   | `var_onion_optin`                | Requires/supports variable-length routing onion payloads  | IN       | [Routing Onion Specification][bolt04] |
```
Roasbeef: Why should a node level feature gate if two peers can/will-continue-to communicate on the connection level? As is, if I went to open a channel with a peer that doesn't support the new onion payload, but they _required_ this bit, then the connection would fail upon feature bit comparison.

Roasbeef:
At this point, we may need to just accept the existing global feature bit namespace, then once this is ready and tested for interop all around, copy over the existing global feature bit namespace to what is currently known as the local feature bit namespace. However, re the combination of the namespaces, see my latest comment re cross-layer feature bit awareness comparisons.

Roasbeef:
To remedy that comparison issue, we may need to add _additional_ feature bit vectors maybe? Not sure this is really doable though, since the `innit` message can be padded like the rest of the messages, so it's possible that a larger preceding feature vectors renders another unable to be coded in the message size limit.

09-features.md
rustyrussell: We don't have a general feature field in channel_update (yet?). But we can add a field if we need to.

09-features.md
rustyrussell: Yes, it's overbroad: this is the tradeoff.  Previously you could in theory gossip with a node which had an even global feature you didn't like, too.

Putting it in the channel_announce would also be overbroad (since both nodes might not need it).  We could add a feature bitfield (TLV-style) to channel_update, but it's marginally more efficient to use node_announce, since it's really per-node not per channel in any sane implementation.

Did we want to go there?

09-features.md
Roasbeef: I think the main thing here is the inability to express distinct feature sets for each channel with the current BOLT 11 feature bit vector extension.

t-bast:
> Our release is late as is, and it's now being held up by this PR as it creates a dependency with the static remote key PR.

Yeah, I might have been too optimistic by hoping we could merge this soon to unblock your release from further delay :).

I quite like the current proposal. Updated writers will only write to the currently-named `localfeatures` field all of their feature bits.

Legacy readers will not be impacted because right now the only global feature bit (variable length onion) will only be set as optional, not mandatory (they aren't missing out on anything since we don't have real features like AMP that actually leverage the variable length onion).

We can have our implementations handle both the legacy case (if we see that `glen` isn't 0) and the updated case (when `glen` is 0) when reading the other peer's features so that we play nicely with legacy nodes. Does that sound like a reasonable compromise?

rustyrussell:
There's definitely no harm in advertizing var_onion_optin in local features, and I'd definitely recommend it.

The other questions are:
1. Adding features to channel_update.
2. Adding features to bolt11 routing hints.

09-features.md
```diff
@@ -1,43 +1,54 @@
...
-## Assigned `globalfeatures` flags
+Some features don't make sense on a per-channels or per-node basis, so
+each feature defines how it is presented in those contexts.  Some
+features may be required for opening a channel, but not a requirement
+for use of the channel, so the presentation of those features depends
+on the feature itself.
 
-The following `globalfeatures` bits are currently assigned by this specification:
+The Context column decodes as follows:
+* `I`: presented in the `init` message.
```
cfromknecht: If this is `I`, what will our letter for invoice feature bits be? Maybe `P` for payment request?

Should we carve out the space for invoice feature bits as well in this PR?

cfromknecht: from the candidate features, we could start to specify that `var_onion_option` should be in invoices

t-bast: I think your first comment raises an interesting point to discuss: do we want to merge invoice feature bits with node feature bits? If we do #643 can't use bits 0-3 (which it currently does) since it clashes with node features.

I think it may be useful to merge invoice feature bits with node feature bits as it sounds more future-proof. The only downside I can see is that it uses more space in the invoice, so QR codes might end up being big. But we can probably compress the features bitfield quite efficiently if it becomes large but we only want to set a few bits in it. WDYT?

cfromknecht: at this point I think we should forbid colliding feature bits altogether, so all feature bit spaces should share the same namespace. however there are some node features that don’t make sense as invoice features, e.g. support for gossip sync protocols, and those don’t need to be present. 

the space for routing hints will probably dwarf the feature bits in the invoice for some time, but agree that a more space efficient representation for feature bits is worth exploring. I’ve been looking into a combination of RLE and/or sparse encodings, tho the more difficult task will be signaling support for the newer format without parsing any feature bits, but we have options :P

01-messaging.md
```diff
@@ -215,26 +215,26 @@ The following convenience types are also defined:
...
-Both fields `globalfeatures` and `localfeatures` MUST be padded to bytes with 0s.
+The `features` field MUST be padded to bytes with 0s.
 
 1. type: 16 (`init`)
 2. data:
-   * [`u16`:`gflen`]
-   * [`gflen*byte`:`globalfeatures`]
-   * [`u16`:`lflen`]
-   * [`lflen*byte`:`localfeatures`]
+   * [`u16`:`ignorelen`]
```
Roasbeef: Reminder to revert back to existing.

rustyrussell: Done

rustyrussell:
(Note: I'm strictly append-only for easy ongoing review, but will rebase and collapse once we're all happy)

09-features.md
```diff
@@ -1,43 +1,55 @@
...
+* `N`: presented in the `node_announcement` messages
+* `C`: presented in the `channel_announcement` message.
+* `C-`: presented in the `channel_announcement` message, but always odd (optional).
+* `C+`: presented in the `channel_announcement` message, but always even (required).
+* `9`: presented in [BOLT 11](11-payment-encoding.md) invoices.
```
cfromknecht: does this mean that invoice features will share the same namespace as everything else? i think that's the right call, should `var_onion_optin` be changed to `N9`?

01-messaging.md
```diff
@@ -215,29 +215,30 @@ The following convenience types are also defined:
...
-The 2-byte `gflen` and `lflen` fields indicate the number of bytes in the immediately following field.
 
 #### Requirements
 
 The sending node:
   - MUST send `init` as the first Lightning message for any connection.
   - MUST set feature bits as defined in [BOLT #9](09-features.md).
   - MUST set any undefined feature bits to 0.
-  - SHOULD use the minimum lengths required to represent the feature fields.
+  - SHOULD NOT set features greater than 11 in `globalfeatures`.
```
cfromknecht: lnd currently sets 13 (`option_static_remotekey`) in global features, i think eclair might as well

09-features.md
cfromknecht: `var_onion_optin` does not affect the p2p layer in lnd at all, does it in other implementations? is the rationale is it mainly to decide whether the node supports TLV onions during channel creation?

cfromknecht:
finished up an lnd impl of flat features, if people want to do interop testing lmk https://github.com/lightningnetwork/lnd/pull/3685

01-messaging.md
```diff
@@ -215,29 +215,30 @@ The following convenience types are also defined:
...
-Both fields `globalfeatures` and `localfeatures` MUST be padded to bytes with 0s.
+The `features` field MUST be padded to bytes with 0s.
 
 1. type: 16 (`init`)
 2. data:
    * [`u16`:`gflen`]
    * [`gflen*byte`:`globalfeatures`]
    * [`u16`:`lflen`]
-   * [`lflen*byte`:`localfeatures`]
+   * [`flen*byte`:`localfeatures`]
```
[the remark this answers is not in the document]
t-bast: And the line before should be:
```
* [`u16`:`flen`]
```

01-messaging.md
t-bast: It does feel safer indeed to use 13 here.

09-features.md
t-bast: We're not using it either for p2p, it will currently only be useful in the Bolt 11 invoice or inferred by other feature bits (MPP, Trampoline, etc).

cfromknecht:
> Looking good! I haven't implemented it in Eclair yet but it's going to be a fairly easy change.

Nice yeah it's not too bad, been running the version above on my node for the last day or so w/o issue

09-features.md
rustyrussell: Indeed, #643 does this (or will, once rebased on this)

rustyrussell:
> var_onion_optin does not affect the p2p layer in lnd at all, does it in other implementations? is the rationale is it mainly to decide whether the node supports TLV onions during channel creation?

My feeling is that (almost?) everything should be in init msgs, since it's more current than node_announcement.  Especially since there's no node_announcement if the node has no channels yet?

rustyrussell:
Here's the same version, with commits cleaned up into 3 distinct commits and rebased onto master (which includes option_static_remotekey): this is the one I actually plan on merging:

https://github.com/rustyrussell/lightning-rfc/commits/flat-features-flattened

t-bast:
> this is the one I actually plan on merging:
> https://github.com/rustyrussell/lightning-rfc/commits/flat-features-flattened

ACK, looks good!
We'll get started on implementing in Eclair.

rustyrussell:
Note this is already implemented on c-lightning 0.7.3, so you can test via BLUEIRON (mainnet) or SLICKERMASTER (testnet):

Mainnet: 024b9a1fa8e006f1e3937f65f66c408e6da8e1ca728ea43222a7381df1cc449605@128.199.202.168

Testnet: 031a3478d481b92e3c28810228252898c5f0d82fc4d07f5210c4f34d4aba56b769@165.227.30.200

Roasbeef:
Needs a rebase then can be merged at will! `lnd` master does this, and it'll be packaged in our next major release.

PR #670, opened 2019-09-12
Clarify origin of  chaining key for rotation

ysangkok:
It is not always directly from the handshake.


rustyrussell:
Oh, I did a similar change based on your comment on IRC!

https://github.com/rustyrussell/lightning-rfc/commit/9b2d5dd78328d401e23c1e7aefd645c7698949e6#diff-999915f2f34badaec50cf54005282740

What do you prefer?

ysangkok:
Yours is simpler, I'll close this.

Issue #659, opened 2019-08-18
bolt11 encoding capital letters


sbtc1:
The reference documentation does not mention the possibility of using a capitalised invoice (eg LNBC..), still most implementations in the wild seem to accept them.

There are good reasons to support capitalised invoices, like smaller QR codes. 

My suggestion is to explicitely mention in the documentation that invoices should be case-independant so that implementations must allow both forms (or even mixed).

t-bast:
That sounds reasonable. Can you open a pull request to propose that change?

rustyrussell:
Note that this is already specified in BIP-0173 which is heavily referenced, however I agree it's good to highlight it.

PR #661, opened 2019-08-19
Added explicit mention of uppercase invoices

sbtc1:
#659 
Added explicit mention that readers must accept uppercase invoices which are generated by some providers, especially for the purpose of smaller QR codes.

PR #677, opened 2019-09-30
BOLT 11: make UPPERCASE explicit.

rustyrussell:
Explicitly mirror the BIP-173 advice for QR codes, and note the reader
requirements.  In addition, we change the case of one of the examples
to ensure testing.

Fixes: #659
Alternative-to: #661

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -35,10 +35,12 @@ encoding.
 
 A writer:
    - MUST encode the payment request in Bech32 (see BIP-0173)
+   - SHOULD use upper case for QR codes (see BIP-0173)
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightningnetwork/lightning-rfc/commit/9a84c661b110dbc64c221dfc28fd93d33a12fcd5

cdecker:
> ACK, just need to validate the new test vectors on eclair.

Aren't they just uppercased versions of the previous vectors?

rustyrussell:
Rebased without test vector change, since spec has since included an upper-case vector.

Issue #675, opened 2019-09-28
random short channel id for private channels


renepickhardt:
Mark H pointed in this [stackexchange comment](https://bitcoin.stackexchange.com/questions/90694/is-it-possible-to-watch-the-chain-to-count-the-number-of-lightning-channels/90695#90695) out that:

> It is best to assume that the `short_channel_id` specified in an invoice is just some arbitrary identifier which tells the last hop how to forward the payment.

I think for private channels we should stardatize the process of creating a short channel id that does not give the possability later to reveal the funding tx in the routing hints. 

One idea - not sure if it is viable - that comes to my mind take the ECDH shared secrete and use it to encrypt the funding txid. use the necessary bytes from the encrypted result. This is deterministic for channel partners and just random junk for outsiders.

btcontract:
This is a great idea, I like that peers are allowed to invent literally anything as long as it only concerns their  private channels (turbo channels, channels without any on-chain funding at all and so on), there is no need to constrain a `short_channel_id` to some on-chain tx in this case.

btcontract:
Another consideration: perhaps it makes sense to constrain random short ids by some max value so that node software can never conflate them with on-chain based short ids. For example, upper block height limit could be set to `500000` (a block height before LN went mainnet) so random short id would always be something like `500000x_x_`.

renepickhardt:
This has been exploited at https://github.com/BitcoinDevShop/hidden-lightning-network as announced on the mailinglist https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-June/003603.html

t-bast:
Fixed by #910

PR #681, opened 2019-10-03
option_scid_assign: privacy protection for private channels.

rustyrussell:
This turned out to be a bit more complex than I expected.

Private channels expose their funding transaction in routehints to
receive payments, but this is unnecesary.  Their funding tx can also
be probed via payment attempts.

1. A new feature, so you know when you can use these controls.
2. A channel_flags bit so you can ensure that new channels are
   protected probes from the start.
3. Messages to assign (and unassign) random short_channel_ids,
   with a side-effect of disabling the funding-tx-based one.

Fixes: #675

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1118,59 +1118,78 @@ it's simplest to only allow it to set fee levels; however, as the same
 fee rate applies to HTLC transactions, the receiving node must also
 care about the reasonableness of the fee.
 
-### Assigning Random `short_channel_id`
+### Assigning Random `short_channel_id`s
 
-To facilitate privacy of unannounced channels, peers can request a random
-short_channel_id to use which does not have any relationship to the actual
+To facilitate privacy of unannounced channels, peers can request random
+short_channel_ids to use which does not have any relationship to the actual
 channel funding transaction.  Note that these messages are all even, so it
 is illegal to send them to peers which do not offer `option_scid_assign`.
 
-1. type: 260 (`assign_scid`) (`option_scid_assign`)
+1. type: 260 (`assign_scids`) (`option_scid_assign`)
 2. data:
    * [`channel_id`:`channel_id`]
+   * [`u16`:`num`]
 
-1. type: 266 (`assign_scid_reply`) (`option_scid_assign`)
+1. type: 266 (`assign_scids_reply`) (`option_scid_assign`)
    * [`channel_id`:`channel_id`]
-   * [`short_channel_id`:`short_channel_id`]
+   * [`u16`:`num`]
+   * [`num*short_channel_id`:`scid_series`]
 
-1. type: 268 (`unassign_scid`) (`option_scid_assign`)
+1. type: 268 (`unassign_scids`) (`option_scid_assign`)
 2. data:
    * [`channel_id`:`channel_id`]
-
-1. type: 270 (`unassign_scid_reply`) (`option_scid_assign`)
-   * [`channel_id`:`channel_id`]
+   * [`u16`:`num`]
+   * [`num*short_channel_id`:`short_channel_ids`]
 
 #### Requirements
 
 A sending node:
   - if it did not offer `option_scid_assign`:
-    - MUST NOT send `assign_scid` or `unassign_scid`
-  - SHOULD remember the last `assign_scid_reply` `short_channel_id` for use in
-    receiving payments.
-  - MAY send `unassign_scid` when no invoices are pending, and send `assign_scid` when
+    - MUST NOT send `assign_scids` or `unassign_scids`
+  - MUST NOT set `assign_scids` `num` to more than 256.
+  - MUST NOT set `assign_scids` `num` to less than the previous number of assigned short_channel_ids.
+  - SHOULD use a unique node_id in invoices for each of the `short_channel_ids`
+    in `assign_scids_reply`.
+  - MAY send `unassign_scids` to remove some or all ids if no invoices are pending, and send `assign_scids` when
     it wants to create a new invoice and none are pending.
 
 A receiving node:
-  - on receiving `assign_scid`:
-    - MUST assign an unpredictable unique 64-bit number to this channel
-      (see [Assigned `short_channel_id` Forwarding](04-onion-routing.md#assigned-short_channel_id-forwarding)).
-	  and return it in an `assign_scid_reply` message.
-	  - SHOULD ensure this does not conflict with future `short_channel_id`s.
-    - MUST discard any previous assigned `short_channel_id` for this channel
-  - on receiving `unassign_scid`:
-    - MUST discard any previous assigned `short_channel_id` for this channel
-	  - Note: it's not an error if there are no previous assignments.
-	- MUST reply with `unassign_scid_reply`.
+  - if it offers `option_scid_assign`:
+    - MUST keep an `scid_series` of up to 256 short_channel_ids for each peer.
+    - if either peer opened the channel with `disable_funding_scid`:
+      - `scid_series` begins empty
+    - otherwise:
+	  - `scid_series` begins with the funding-transaction-based `short_channel_id`
+    - on receiving `assign_scids`:
+      - if `num` is greater than 256, or `num` is less than the size of `scid_series`:
+        - MAY fail the connection with an error.
+      - MUST append additional new unpredictable unique 64-bit numbers to `scid_series` for this channel
+        (see [Assigned `short_channel_id` Forwarding](04-onion-routing.md#assigned-short_channel_id-forwarding)).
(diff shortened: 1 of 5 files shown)


02-peer-protocol.md
```diff
@@ -1106,6 +1118,75 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+adversary to probe for them over time, using payment attempts.  Thus it's
+recommended that they only be assigned when they're actually likely to be
+used, and unassigned afterwards.
+
+If it were to conflict with a future `short_channel_id` for a real
+channel, misrouting may occur.  The simplest way to minimize this is
+to use old block numbers, but that also reduces the search space.
+Simpler is to select a random value which isn't already used, and
+close (before funding_locked) any future channel unfortunate enough to
+clash.
```
t-bast: I'm more in favor of restricting to old blocks, this makes it a lot simpler to manage.
Restricting to old blocks means we have ~500.000*2^40 possibilities ~= 10^17
Assuming 100.000 attempts per second (most nodes probably won't be able to forward that many payment attempts per second, so it feels conservative) it will take more than 150.000 years to probe.
Should be a big enough search space, isn't it? ;)

rustyrussell: In practice, you can just pick a random 64 you will never clash. I don't want to overly restrict implementations: you can choose to use a subset of believable unspent outputs, etc.

rustyrussell:
> Great proposal, thanks for putting this together.
I was thinking about that a couple of weeks ago and tinkering with the idea of going even further and allowing "private/unannounced" nodes to use multiple node IDs (node_id_alias maybe?).

This is always possible, but makes routing loops worse, and can always be done on a separate layer.
I prefer rendezvous routing, which doesn't require registration.

02-peer-protocol.md
t-bast: Agreed, I just think we can remove the statement that choosing old blocks reduces the search space because in practice it's irrelevant.

t-bast:
> I prefer rendezvous routing, which doesn't require registration.

Isn't that true as well for the `scid` trick you're proposing in this PR? It's obsolete once we have rendezvous routing (which may also make the invoice route hints obsolete altogether).

However I believe that rendezvous routing is potentially more costly in terms of bytes used in the invoice, but maybe we should spend more time spec-ing a proposal to figure that out. I can work on that.

t-bast:
Rendezvous routing will not play nicely with MPP...if a recipient wants to use rendezvous for multi-part payments, it will have to generate multiple onions (for each partial payment) which will be too big to put in the invoice.

This `scid` proposal is more lightweight so it will add value even if we have rendezvous support at some point.

> This is always possible, but makes routing loops worse

I don't understand your point about routing loops...for "private" nodes these alias node ids will not be announced to the network and will only be included in invoices, with details of the (unannounced channel) leading to it. So it won't mess anything routing-related. And those alias node ids can be BIP32-derived from a main nodeId, so they won't even clash with any other nodeId in the network. Maybe I should draft something more detailed?

btcontract:
Should random `scid` be the same for both peers? I suppose there's no reason for it to be different from each side as that would be easier to apply to current implementations where same `scid` is already assumed. But that would mean one side should ask, and the other one should update a `scid` for itsef and peer, not both at once, or am I missing something?

btcontract:
https://github.com/lightningnetwork/lightning-rfc/pull/681/files#diff-ad219ee50fcd8926614ca5a5d5f1a1f7R31

Should not `114/115` be `14/15`?

rustyrussell:
> My high-level idea is that Eve (not Alice, Alice is nice and has nothing to hide) opens an unannounced channel with some public node (why not ACINQ, I hear it's a great node?). She then registers several node ids with the public node to use as alias. When the public node receives a forwarding request, he uses the real `node_id` and `short_channel_id`. WDYT?

I thought about this some more: you can do this *already* without revealing to the public node what transient key you're using.  Which is even more win!

02-peer-protocol.md
rustyrussell: > Should random `scid` be the same for both peers? I suppose there's no reason for it to be different from each side as that would be easier to apply to current implementations where same `scid` is already assumed.

No, since it might clash for one side, and one side might tear it down while you were trying to use it.  And why make it the same?  That just leaks information.

rustyrussell:
> https://github.com/lightningnetwork/lightning-rfc/pull/681/files#diff-ad219ee50fcd8926614ca5a5d5f1a1f7R31
> 
> Should not `114/115` be `14/15`?

No, I'm using the experimental range until this gets approved.  THen it will get the next available.

rustyrussell:
> > I prefer rendezvous routing, which doesn't require registration.
> 
> Isn't that true as well for the `scid` trick you're proposing in this PR? It's obsolete once we have rendezvous routing (which may also make the invoice route hints obsolete altogether).

Good point.

> However I believe that rendezvous routing is potentially more costly in terms of bytes used in the invoice, but maybe we should spend more time spec-ing a proposal to figure that out. I can work on that.

So many things to do...!

02-peer-protocol.md
btcontract: > And why make it the same? That just leaks information.

Makes implementation easier at least in Eclair where I'm trying to do it: currently graph engine determines that two `ChannelUpdate`s belong to the same channel by them having a same `scid`. But of course this should not dictate the spec.

Just curious: what kind of info is specifically leaked by same `scid`?

btcontract:
Should a peer also send a fresh `channel_update` after `assign_scid_reply`? An old one would have an incorrect signature (but not sure how important that would be for a privat channel).

02-peer-protocol.md
```diff
@@ -1106,6 +1118,71 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+  - SHOULD remember the last `assign_scid_reply` `short_channel_id` for use in
+    receiving payments.
+  - MAY send `unassign_scid` when no invoices are pending, and send `assign_scid` when
+    it wants to create a new invoice and none are pending.
+
+A receiving node:
+  - on receiving `assign_scid`:
+    - MUST assign an unpredictable unique 64-bit number to this channel
+      (see [Assigned `short_channel_id` Forwarding](04-onion-routing.md#assigned-short_channel_id-forwarding)).
+	  and return it in an ``assign_scid_reply` message.
```
```suggestion
	  and return it in an `assign_scid_reply` message.
```

02-peer-protocol.md
```diff
@@ -1106,6 +1118,71 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+	  - SHOULD ensure this does not conflict with future `short_channel_id`s.
+    - MUST discard any previous assigned `short_channel_id` for this channel
```
pm47: With this requirement, I don't think we need the `unassign_*` messages. If they care about privacy, then they will request new scids regularly. (NB: with this proposal we can't have two pending invoices simultaneously with two different scids, I think that's fine).

rustyrussell: I think we still want unassign to reduce the chance for probing when you're offline?

rustyrussell:
> Should a peer also send a fresh `channel_update` after `assign_scid_reply`? An old one would have an incorrect signature (but not sure how important that would be for a privat channel).

You *can* send channel_update for a private channel; we should mention that will always use the original "real" short_channel_id.

t-bast:
> I thought about this some more: you can do this already without revealing to the public node what transient key you're using. Which is even more win!

Can you detail that @rustyrussell? Or is this something you'll add to that PR once you have something ready?

02-peer-protocol.md
rustyrussell: The leak us:

1. Alice has a private channel with Bob: they call it scid-tmp.
2. Mallory asks Alice's website for an invoice, Alice gives Mallory a fake pubkeyA and scid-tmp from Bob.
3. Mallory asks Bob's website for an invoice, Bob gives a fake pubkeyB and the same scid-tmp from Alice.

Mallory has learned that Alice == pubkeyA and Bob == pubkeyB.

rustyrussell:
> > I thought about this some more: you can do this already without revealing to the public node what transient key you're using. Which is even more win!
> 
> Can you detail that @rustyrussell? Or is this something you'll add to that PR once you have something ready?

The key Alice uses to decrypt the onion is unknown to her peer Bob anyway.  So Alice just tells Mallory an invoice with a made up transient "nodeid".

t-bast:
I have an alternative proposal here: https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-January/002435.html

It has a different set of trade-offs, which I personally prefer :)
I'm interested in getting feedback on whether it makes sense to go down that path or if people prefer the `option_scid_assign` proposal.

rustyrussell:
OK, I'm closing this proposal, in favor of a more powerful blinded paths proposal that @t-bast is working on.  This will cover the private channels as well as more general cases.

PR #691, opened 2019-10-30
BOLT 4: add a requirement of `tlv_payload payload format`

nayuta-gondo:


04-onion-routing.md
```diff
@@ -258,6 +258,7 @@ The writer:
 
 The reader:
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
+  - MUST return an error if it is not the final node and `short_channel_id` is not present.
```
cfromknecht: should we also return an error if a node is the final hop and `short_channel_id` is present?

t-bast:
ACK.
However, #643 also includes that fix. I'd rather wait for #643 to be merged instead of creating conflicts.

Issue #692, opened 2019-10-31
Add pico amount example to Bolt11 test vectors


Sword-Smith:
Several libraries (at least two libraries and one web service) have problems with the `p` numbers representing 10^(-12) BTC. So an example with this encoding should be included in the test vectors. You can use this one if you like:

```
lnbc9678785340p1pwmna7lpp5gc3xfm08u9qy06djf8dfflhugl6p7lgza6dsjxq454gxhj9t7a0sd8dgfkx7cmtwd68yetpd5s9xar0wfjn5gpc8qhrsdfq24f5ggrxdaezqsnvda3kkum5wfjkzmfqf3jkgem9wgsyuctwdus9xgrcyqcjcgpzgfskx6eqf9hzqnteypzxz7fzypfhg6trddjhygrcyqezcgpzfysywmm5ypxxjemgw3hxjmn8yptk7untd9hxwg3q2d6xjcmtv4ezq7pqxgsxzmnyyqcjqmt0wfjjq6t5v4khxxqyjw5qcqp2rzjq0gxwkzc8w6323m55m4jyxcjwmy7stt9hwkwe2qxmy8zpsgg7jcuwz87fcqqeuqqqyqqqqlgqqqqn3qq9qs4x9qlmd57lq7wwr23n3a6pkayy3jpfucyptlncs2maswe3dnnjy3ce2cgrvykmxlfpvn6ptqfqz4df5uaulvd4hjkckuqxrqqkz8jgphputwh
```

Sword-Smith:
Libraries and web services unable to decode this invoice

0: https://github.com/btcpayserver/BTCPayServer.Lightning/issues/25
1: https://github.com/bitcoinjs/bolt11/issues/27
2: https://lightningdecoder.com/ (reported `Amount is outside of valid range` on 2019.10.31)

NicolasDorier:
```
m (milli): multiply by 0.001
u (micro): multiply by 0.000001
n (nano): multiply by 0.000000001
p (pico): multiply by 0.000000000001
```

In BTCPay our conversion in Millisatoshi is the following:

```
    public enum LightMoneyUnit : ulong
    {
        BTC = 100_000_000_000,
        MilliBTC = 100_000_000,
        Bit = 100_000,
        Micro = 100_000,
        Satoshi = 1000,
        Nano = 100,
        MilliSatoshi = 1,
        Pico = 1
    }
```

[...]

PR #693, opened 2019-11-01
Add bolt11 example with amount in pico-BTC

Sword-Smith:
This addresses #692


NicolasDorier:
Concept ACK, but I think on top of it the RFC should be modified so if the pico denomination is not a multiple of 10, the BOLT11 should be invalid. The lightning protocol operates at the millisatoshi level which is 10 pico.

Sword-Smith:
Replacing this with #699

PR #694, opened 2019-11-05
BOLT 11: add optional vendor field.

rustyrussell:
It was pointed out to me at thelightningconference.com that it's often
a legal requirement to list the vendor on a receipt.  It also makes
perfect sense.

It can be done in the description field, but that's really supposed to
be a description of the *items*.  Dividing it also lets wallets have
much better UX.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -141,6 +141,7 @@ Currently defined tagged fields are:
    * `fee_base_msat` (32 bits, big-endian)
    * `fee_proportional_millionths` (32 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `v` (12): `data_length` variable.  Optional name of vendor/supplier (UTF-8).
 * `9` (5): `data_length` variable. One or more 5-bit values containing features
   supported or required for receiving this payment.
   See [Feature Bits](#feature-bits).


yaslama:
@rustyrussell What do you think about adding a field containing data in an already standardized format like https://en.wikipedia.org/wiki/EDIFACT ? I feel that we are reinventing the wheel.

BitcoinErrorLog:
Bitefill CFO says a vendor field would be very useful :)

shesek:
Isn't there a risk that showing an unauthenticated vendor name could give users the wrong impression that the vendor name is somehow verified?

If the vendor name is added as part of the description then its pretty clear that its just some free text added by the invoicer, but a special UI element showing the vendor name separately might make it appear like it is somehow verified, and could potentially be abused by scammers.

renepickhardt:
To me it feels like this is a half baked solution. either go with a standard - like @yaslama suggested - in the invoice because you expect that lightning invoices will replace regular invoices or keep it the way it is. I would suggest to go the full path instead of just adding one field.

hsjoberg:
@rustyrussell I do fully agree that we should have a field that is the name of the recipient, I am not sure about the field name `vendor` though, as the payment could be from person to person as well.

> It can be done in the description field, but that's really supposed to
be a description of the items. Dividing it also lets wallets have
much better UX.

Yes, I have been working on a simple protocol for doing this in the description field, but separating them is a much better approach.

> Isn't there a risk that showing an unauthenticated vendor name could give users the wrong impression that the vendor name is somehow verified?

@shesek This risk exists, the wallet should probably check if the name has already been used by another pubkey.

> Wouldn't vendors typically also issue a standard tax receipt/invoice, in addition to the lightning invoice? Wouldn't it be legally sufficient to list the vendor name on the regular receipt (which can be sent as pdf over email, shown on a website, etc), without being listed on the lightning receipt too?

This is off-channel though and (from a UX standpoint) doesn't reflect nor improve the transaction log inside the wallet. I think it makes sense to improve and integrate information to the actual wallet as this essentially makes lightning network wallets more intuitive than using the current banking system.

In the future, perhaps even the receipt could be integrated to the wallet.

BitcoinErrorLog:
Further comments from our team, mirror Rene's sentiments:

_"Receipts usually have very complex data structures, basically everything an invoice of ours has, not just the vendor name. "
"If the goal is to fulfill all receipt needs then there’s a lot more that needs to be done, such as one line for each item in the basket, terms and conditions, etc. Queue the CVS long receipts meme."_

The scope starts blowing out to a full accounting solution... but generally some threshold of comprehensiveness may need to be achieved to provide utility.

hsjoberg:
I think the scope is very clear. Add a recipient field for BOLT11-invoices. An invoice protocol could be worked on later on, separately.

esneider:
This would be very useful for wallet implementors to do things like this: https://twitter.com/MuunWallet/status/1186738141731983361

While doing research for this feature we found that there are many nodes being used for multiple lapps each, so this would be really useful, as we can't determine the lapp from just its node public key.

We should be careful as to how this field is used in applications though, since vendor impersonation is a real possibility, as @shesek points out.

rustyrussell:
> EDIT: To clarify, I do not think that this PR satisfies legal requirements of a receipt. I am mainly positive for this new field as a UX/UI improvement.

Well, it does in Australia (technically, a "proof of purchase", see https://www.business.gov.au/finance/payments-and-invoicing/receipts-and-proof-of-purchase).

While you can *claim* to be anyone, you can also claim that in the description, so I don't think this is much of an issue.

Alex-CodeLab:
@yaslama I've worked with EDIFACT for years, and all companies that use it implement their own version. It is a standard, but not a protocol, so there are many, many variations. (For example, think about how many address+zipcode variations there are and how they are different per country.)

A far easier solution would be that all this is handled on the wallet level: wallet creators could design their invoice document any way they want (with vendorname, address, goodslines, etc), and add a hash that matches a hash on the lightning invoice.

fiatjaf:
In my opinion the protocol should be left as open and as generic as possible, without trying to include every possible _current_ vendor requirement. We don't know what people may be using Lightning for in the future, and including a lot of fields with a very limited use-case creates complexity for all others.

Also, it's not as if big and law-abiding vendors are using Lightning today, anyway, and what if the law changes? What if the law requires other types of fields in other countries?

In Brazil, for example, a "valid receipt" must include numbers from our national IRS and many other quibbles. What happens is that small merchants don't give one unless explicitly requested and after a long waiting time, big merchants who have automated systems for everything do give, but still it's a separate receipt than the one provided by credit card machines and so on. As @shesek said.

yaslama:
> A far easier solution would be that all this is handled on the wallet level: wallet creators could design their invoice document any way they want (with vendorname, address, goodslines, etc), and add a hash that matches a hash on the lightning invoice.
But in this case, the ln invoice is not self-contained. This is the reason I proposed to add a field with an "opaque" blob which is supposed to follow a recognized and used standard, and each wallet will handle the content of this blob depending on the situation/country etc..

esneider:
While using some comprehensive standard like UBL would be great for the application/wallet layer, I think the length of the invoice should also be taken into account. Invoices are already pretty long, which causes QR codes to be super dense.

Older phones have difficulties scanning these codes, due to having cameras with low(er) resolutions. It was common in the lightning conference to see people try to scan lightning invoices 4/5 times until the QRs were recognized.

It’d be great to keep the standard usable for non-last-gen phones.

nayuta-ueno:
If there is a vendor field in an invoice text, I would like to display vendor information (URL, icon, etc.) without decoding the vendor field.

yaslama:
> While using some comprehensive standard like UBL would be great for the application/wallet layer, I think the length of the invoice should also be taken into account. Invoices are already pretty long, which causes QR codes to be super dense.
> 
> Older phones have difficulties scanning these codes, due to having cameras with low(er) resolutions. It was common in the lightning conference to see people try to scan lightning invoices 4/5 times until the QRs were recognized.
> 
> It’d be great to keep the standard usable for non-last-gen phones.

A point is made here. But if/when BOLT 12 is implemented, the invoice will be retrieved using LN and only the "offer" will be QR encoded. And the offer is a lot shorter than the invoice.
cf  https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002276.html about BOLT 12

esneider:
> A point is made here. But if/when BOLT 12 is implemented, the invoice will be retrieved using LN and only the "offer" will be QR encoded. And the offer is a lot shorter than the invoice.
> cf https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002276.html about BOLT 12

Wouldn't we want to include the vendor info directly in the "offer" though?
* As I user, knowing who I'm paying to is a key piece of information when deciding whether I want to share my address/phone number.
* As a wallet, I want to be able to render the payment info before having to retrieve the final invoice, to avoid delays due to unreliable nodes in the communication path to the destination node.
* A nice side effect of including the vendor in the offer is that it encodes the fact that the vendor cannot change for recurring payments, which would be pretty unintuitive for the user.

Finally, notice that `lno...` URIs aren't that much shorter than `ln...` URIs, and in fact might end up being longer.

Alex-CodeLab:
@esneider 
> * As I user, knowing who I'm paying to is a key piece of information when deciding whether I want to share my address/phone number.

as @shesek pointed out, vendor names should not be trusted and are not unique identifiers. Keys and URI are used for authentication. 

> Wouldn't we want to include the vendor info directly in the "offer" though?

Would that mean broadcasting name/address info ? 

----
I think this is all going in the wrong direction. It started with just the _vendor name_, but now it becomes some business application with (purchase) orders, crm, accounting, etc..
Once we go down that route, the options and requirements will be endless. I can guarantee it won't end with just adding a few fields. 
_(Trying to build SAP into LN is not the way to go.)_ 
  
Imho, *all* personal details should be kept out of the LN network. It would be a goldmine for data-collection and surveillance. 
While one group of developers is working on privacy, others are talking about adding address and phonenumbers... wth is going on??

hsjoberg:
> Wouldn't we want to include the vendor info directly in the "offer" though?

@esneider I don't see the harm of adding a optional field that will probably be around 10 characters or less. It will have a minimal effect on the invoice text length, while greatly improving UX.

> Once we go down that route, the options and requirements will be endless. I can guarantee it won't end with just adding a few fields.

This is a slippery slope. The PR is talking about an optinal vendor name, not anything else.

> Imho, all personal details should be kept out of the LN network. It would be a goldmine for data-collection and surveillance.
> While one group of developers is working on privacy, others are talking about adding address and phonenumbers... wth is going on??

How would the data collection be done? The QR code is hopefully displayed on an HTTPS-secured site, only visible for the actual user buying something.

Adding phone numbers, address etc for  presumably _known party_, like  an e-shop, does nothing to harm privacy.

tkijewski:
This addition to the spec seems out of place. Protocol layers are supposed to be as application agnostic as possible. This is an addition for a specific application: b2c payments. It is completely irrelevant for p2p payments. 

Looking at the existing [Tagged Fields](https://github.com/lightningnetwork/lightning-rfc/blob/a616002b3551e18fea09b790e7e4b032a5ebb3e6/11-payment-encoding.md#tagged-fields), `Optional name of vendor/supplier (UTF-8).` really does not fit in

Alex-CodeLab:
@rustyrussell 
What is the problem you're trying to solve here?   

Some mention _UX_, others are talking about _legal requirements_, others about _business requirements._  So we're all talking about something else. I think a clear problem-statement would improve this conversation.

hsjoberg:
> This addition to the spec seems out of place. Protocol layers are supposed to be as application agnostic as possible. This is an addition for a specific application: b2c payments. It is completely irrelevant for p2p payments.

I suggested to change the name of the field to something more general than `vendor`, perhaps `recipient`.

> Looking at the existing Tagged Fields, Optional name of vendor/supplier (UTF-8). really does not fit in

It fits very well with the Description tag `d`.

11-payment-encoding.md
```diff
@@ -141,6 +141,7 @@ Currently defined tagged fields are:
    * `fee_base_msat` (32 bits, big-endian)
    * `fee_proportional_millionths` (32 bits, big-endian)
    * `cltv_expiry_delta` (16 bits, big-endian)
+* `v` (12): `data_length` variable.  Optional name of vendor/supplier (UTF-8).
```
hsjoberg: The tag could be used for both B2C and P2P payments, so it makes sense to keep this generic.

```suggestion
* `v` (12): `data_length` variable.  Optional name of recipient (UTF-8).
```

(The `r` tag is already used for route hints.)

hsjoberg:
What do people think about this addition?  
This feature will either be a part of the BOLT specification or it will eventually find it's way to secondary protocols such as LNURL. I think it makes the most sense to keep this as a part of the standard protocol, especially since we already have the description tag, and let LNURL leverage more advanced features such as channel openings (e.g Thor) and pubkey login protocols. 

I am arguing mostly from a UX improvement standpoint. I would like to have this tag to improve the user experience, both when doing B2C and P2P transactions.  
We have an opportunity here to improve the UX of Bitcoin greatly compared to on-chain payments, which still uses bitcoin addresses with low/non-existent adoption of the `label` param in [BIP21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki) URIs.

The counter-arguments of this proposal that I've seen is:

- *The tag is out of place, BOLT11 is supposed to cover the technical parts of paying an invoice*
  - While it's true that the tag does not help regarding the actual process of paying an invoice, if we look at the specification right now, we can see that both `description` and `description_hash` exists, with the sole purpose of improving the UX. Thus one would need to explain those fields before criticizing this proposal.
- *If we were to do this, we should make it more fully featured and/or use a standardized format*
  - I think we can do both. A more sophisticated receipt is a much larger task and should probably be done once/if the Offer protocol is done. The Offer protocol will most likely be used when paying to a business and have might not have any purpose when paying to an person, thus a `recipient` tag could be used.
- *Having a `vendor` field opens up the possibility to impersonate someone or some business*
  - I think this is the strongest counter-argument. One thing to note is that BOLT11-invoices will most likely be retrieved via an HTTPS secured website, and if someone would want to hijack an payment, one could already do that right now and the user would have no idea of knowing if he/she paid to the right recipient. This proposal does not change that.
- * Adding another tag will increase the text length of the invoice
  - I don't know the impact of this. The tag will probably on average be 10 UTF-8 characters long. I don't think it will change the resolution of the invoice much, if at all.
 - * Adding a recipient tag and other identifiable information harms privacy and helpsfor surveillance and data-collecting
  - * This tag will most likely be used for known parties, such as known internet stores. If you want to sell something anonymously, this proposal does nothing to change that.

Please tell me if I've missed something or if I've not fully understood an argument.

Best
Hampus

fiatjaf:
I believe `description` and `description_hash` are not there for improving the UX, but to be signed along with the rest of the invoice so the proof of payment refers specifically to some purchased good or service, and it's not just a floating payment with no purpose.

hsjoberg:
Thanks @fiatjaf, makes sense. I've updated my post.

PR #703, opened 2019-11-18
Reflect legacy payload type handling implementation in BOLT

arik-so:
In sphinx.go, legacy payloads are not given a length prefix byte.


t-bast:
@arik-so do you think this clarification is still needed? If so can you rebase your PR and we'll include it?

t-bast:
We got rid of the legacy format in #962 so we don't need this PR anymore.

PR #710, opened 2019-12-02
BOLT-00: minor grammar and format fixes

yyforyongyu:


cdecker:
These do not seem to be wrong to me?

t-bast:
Same here, the initial wording seemed correct...

yyforyongyu:
@cdecker @t-bast I had the same doubts as I was reading it, so I googled...
https://www.google.com/search?q=basis+plural
https://www.google.com/search?q=among+or+between

I think the plural form of "basis" is "bases", and "between" is only used when referring two things.

cdecker:
But it's the singular basis of the lightning technology. I guess this one can go both ways.

yyforyongyu:
@cdecker I think you are right. I paused the sentence at the wrong place😂
Actually I'll close this PR. As I'm reading the BOLTs these days, more minor grammar flaws were found. I will put them together to create a new PR.

cdecker:
No problem @yyforyongyu I stumbled on it as well, and the between vs among I think you're right after thinking about it a bit longer :+1:

PR #711, opened 2019-12-03
BOLT 4: don't allow a "fee" for the final node.

rustyrussell:
I recently made a cut & paste bug with the protocol tests, and
paid an HTLC of amount 100M msat, but with only a 1M msat `amt_to_forward`
in the hop_data.  To my surprise, it was accepted.

This is because we allow overpaying the routing fee (considered 0
for the final hop).  This doesn't make sense for the final hop: anything
but exact equality implies a bug, or that the previous node took the
wrong amount from the payment.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -198,7 +198,7 @@ Field descriptions:
    * `amt_to_forward`: The amount, in millisatoshis, to forward to the next
      receiving peer specified within the routing information.
 
-     This value amount MUST include the origin node's computed _fee_ for the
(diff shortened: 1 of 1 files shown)


t-bast:
We're also currently using `<=` as a receiver, but as a sender we make sure the HTLC `amountMsat` equals the onion's `amountToForward`.

Nothing blocking on our side to merge this change, ACK `aabf4e6`.

cdecker:
ACK

rustyrussell:
Given all three acks, and the meeting comment that this is a "no-brainer", I'm merging now.

PR #717, opened 2019-12-13
minor grammar and format fixes

yyforyongyu:


02-peer-protocol.md
```diff
@@ -1272,7 +1272,7 @@ The handling of updates is similarly atomic: if the commit is not
 acknowledged (or wasn't sent) the updates are re-sent. However, it's not
 insisted they be identical: they could be in a different order,
 involve different fees, or even be missing HTLCs which are now too old
-to be added. Requiring they be identical would effectively mean a
+to be added. Requiring them being identical would effectively mean a
```
niftynei: NACK. either `Requiring that they be identical` or `Requiring them to be identical`

05-onchain.md
```diff
@@ -56,8 +56,8 @@ the output is considered to be its own *resolving* transaction.
 
 Outputs that are *resolved* are considered *irrevocably resolved*
 once the remote's *resolving* transaction is included in a block at least 100
-deep, on the most-work blockchain. 100 blocks is far greater than the
-longest known Bitcoin fork and is the same wait time used for
+deep, on the most-work blockchain. 100 blocks are far greater than the
```
niftynei: i actually think `is` is clearer here, as `100 blocks` is a singular entity. (but i'm admittedly bending grammar rules a smidge here)

05-onchain.md
```diff
@@ -410,7 +410,7 @@ HTLC-success transaction.
 The payment preimage either serves to prove payment (when the offering node is
 the originator of the payment) or to redeem the corresponding incoming HTLC from
 another peer (when the offering node is forwarding the payment). After a node has
-extracted the payment, it no longer need be concerned with the fate of the
+extracted the payment, it no longer needs to be concerned with the fate of the
```
niftynei: although not standard, `need be` is colloquially fine grammatically.

07-routing-gossip.md
```diff
@@ -425,7 +425,7 @@ fields in the `channel_update` message:
 Note that the `htlc_maximum_msat` field is static in the current
 protocol over the life of the channel: it is *not* designed to be
 indicative of real-time channel capacity in each direction, which
-would be both a massive data leak and uselessly spam the network (it
+would be both a massive data leak and uselessly spam in the network (it
```
niftynei: NACK, `spam the network` is correct here

05-onchain.md
yyforyongyu: I guess it's been used as a modal verb then.

07-routing-gossip.md
yyforyongyu: From this sentence, `which would be both a massive data leak and uselessly spam the network`.
Shouldn't it mean **both** a massive data leak and useless spam in the network?
Are you suggesting the `spam` here is a verb rather than a noun?

05-onchain.md
yyforyongyu: how about `the length of 100 blocks is far greater than...`?
I read it several times.

> 100 blocks is far greater than the longest known Bitcoin fork

Now I start to think `100 blocks is greater than Bitcoin fork` doesn't make sense. (block greater than fork???)

PR #718, opened 2019-12-16
BOLT4: don't allow a "fee" for the final node(failure messages)

nayuta-ueno:
#711

> For the final node, this value MUST be exactly equal to the incoming htlc amount, otherwise the HTLC should be rejected.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1076,7 +1076,7 @@ An _intermediate hop_ MUST NOT, but the _final node_:
   - if the `outgoing_cltv_value` does NOT correspond with the `cltv_expiry` from
   the final node's HTLC:
     - MUST return `final_incorrect_cltv_expiry` error.
-  - if the `amt_to_forward` is greater than the `incoming_htlc_amt` from the
+  - if the `amt_to_forward` does NOT correspond with the `incoming_htlc_amt` from the
   final node's HTLC:
     - MUST return a `final_incorrect_htlc_amount` error.
 


04-onion-routing.md
```diff
@@ -1076,7 +1076,7 @@ An _intermediate hop_ MUST NOT, but the _final node_:
   - if the `outgoing_cltv_value` does NOT correspond with the `cltv_expiry` from
   the final node's HTLC:
     - MUST return `final_incorrect_cltv_expiry` error.
-  - if the `amt_to_forward` is greater than the `incoming_htlc_amt` from the
+  - if the `amt_to_forward` does NOT correspond with the `incoming_htlc_amt` from the
```
niftynei: what's the rationale for having these match? as long as it the `amt_to_forward` does not exceed the `incoming_htlc_amt` there's no problem.

i'm not 100% sure about this, but i believe a few use cases currently take advantage of this to allow for paying nodes via fees rather than invoice amounts.

nayuta-ueno: Thank you for your comment.
I read the following sentence in BOLT4:

>For the final node, this value MUST be exactly equal to the incoming htlc amount, otherwise the HTLC should be rejected.

and understood as follow:

```
non final node:
  incoming_htlc_amt - fee >= amt_to_forward

final node:
  incoming_htlc_amt == amt_to_forward
```

t-bast: I understood it like you @nayuta-ueno, IIUC the change you propose should have been made at the same time as #711 but was overlooked?

niftynei: oh i see the context now. thanks.

niftynei:
ACK https://github.com/lightningnetwork/lightning-rfc/pull/718/commits/83f77bfbe5c3ab1e5b2927e7848de9854da8b8f9

04-onion-routing.md
niftynei: Oh I missed the link to #711, thanks for the clarification.

nayuta-ueno: @t-bast 
> the change you propose should have been made at the same time as #711 but was overlooked?

I think so too.

PR #722, opened 2019-12-23
04-onion-routing.md: Fix factual error about `final_expiry_too_soon`.

ZmnSCPxj:
As reading of commit 6729755f shows, `final_expiry_too_soon` was
17, not PERM|17.

Note that because we folded a previously non-permanent failure into
the now-permanent PERM|15 failure code, modifications to payment
algorithms may now be needed to specificalyl detect this case,
otherwise payment algorithms may give up in some edge cases where
blocks are mined while payments are in-transit between sender and
receiver.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -944,12 +944,15 @@ between sending a payment with the wrong final CLTV expiry and an intermediate
 hop delaying the payment so that the receiver's invoice CLTV delta requirement
 is no longer met.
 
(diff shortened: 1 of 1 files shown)


ZmnSCPxj:
What needs to be done to get this merged?

t-bast:
An ack from someone from Lightning Labs, maybe @joostjager who touched this part of the spec recently?

PR #560, opened 2019-01-29
BOLT7: reply_channel_range parameter

nayuta-ueno:
[BOLT7 `reply_channel_range` requirements](https://github.com/lightningnetwork/lightning-rfc/blob/064d6feed036de192f71cf6854e00f33361b6090/07-routing-gossip.md#requirements-5)

> The receiver of query_channel_range:
>  - For each reply_channel_range: 
>     - MUST encode a `short_channel_id` for every open channel it knows in blocks `first_blocknum` to `first_blocknum` plus `number_of_blocks` minus one.

so,

`first_blocknum` = short_ids[0].block
`number_of_blocks` = short_ids[LAST].block - `first_blocknum` + 1

PR #729, opened 2020-01-20
BOLT 7: clarify how to decode empty arrays

sstone:
Channel queries use arrays (of short channel ids, timestamps, ...) encoded with a single byte for the encoding type followed by encoded data.
If the encoded data is empty, it always decodes to an empty array, no matter what the encoding type is set to.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -575,6 +575,8 @@ Encoding types:
 
 This encoding is also used for arrays of other types (timestamps, flags, ...), and specified with an `encoded_` prefix. For example, `encoded_timestamps` is an array of timestamps than can be either compressed (with a `1` prefix) or uncompressed (with a `0` prefix).
 
+An empty encoded data buffer always decodes to an empty array, no matter what the encoding type is set to.
+
 Note that a 65535-byte zlib message can decompress into 67632120
 bytes<sup>[2](#reference-2)</sup>, but since the only valid contents
 are unique 8-byte values, no more than 14 bytes can be duplicated


t-bast:
Actually there's already a pending PR with the same clarification: #560

sstone:
Good point! Here I'm mostly focused on how to specifically deal with empty arrays, which goes a bit further than #560 (it's not explicit on that point), the biggest issue being: are empty arrays sent wit a `zlib` prefix valid ? it's ok for eclair and lnd but not c-lightning, and the answer is not obvious yet.

cdecker:
I don't think #560 touches this at all, and no an empty byte string is not a valid zlib compressed value:

```python
>>> import zlib
>>> zlib.compress(b'')
'x\x9c\x03\x00\x00\x00\x00\x01'
>>> zlib.decompress(b'')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
zlib.error: Error -5 while decompressing data: incomplete or truncated stream
```

As such I think it is invalid to specify an encoding and then providing an invalid value for that encoding, since it might mask other issues internally. We could special-case the empty byte string, but what good would that be? I think the correct way to handle this is https://github.com/ACINQ/eclair/pull/1280, i.e., omitting the fields altogether or specifying the uncompressed encoding if the intention is to communicate an empty byte string.

sstone:
Ok, I still like the idea of "nothing in means nothing out" but it seems that every zlib encoder out there disagree with me except for the one we use :) so I'll agree that it was rather a bug in our code than a uncertainty in the specs
ps. w'eve chosen to encode empty lists in uncompressed format, it works against c-lighting and lnd

PR #730, opened 2020-01-20
BOLT 7: clarify how to encode multiple `reply_channel_range` messages

sstone:
When multiple `reply_channel_range` messages are sent back for a single `query_channel_range` query, they must cover the requested block range without gaps or overlaps:
`first_blocknum` must be set to the previous `reply_channel_range`'s `first_blocknum` plus `number_of_blocks`.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -783,6 +783,7 @@ The receiver of `query_channel_range`:
     - MUST encode a `short_channel_id` for every open channel it knows in blocks `first_blocknum` to `first_blocknum` plus `number_of_blocks` minus one.
     - MUST limit `number_of_blocks` to the maximum number of blocks whose
       results could fit in `encoded_short_ids`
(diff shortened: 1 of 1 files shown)


t-bast:
Actually there's already a pending PR with the same clarification: #560

sstone:
Not exactly. #560 specifies that replies must cover the block range requested in a query, without gaps at the beginning or the end, which was implied in the description of channel queries.

This PR adds a constraint for chaining multiple replies to a single query: the next  `reply_channel_range` must start right after the last  `reply_channel_range` that was sent. This is what lnd wants and does, but not eclair or c-lightning for example, which will handle gaps between consecutive replies (but c-lightning will not allow replies to overlap).
Different impls do different things here because the specs are not specific enough, which is what I'm trying to fix: do we allow gaps between consecutive replies ?
This PR says 'no we don't', which imho is the best option as it allows the recipient to validate what it receives.

t-bast:
Gotcha, that means this PR should supercede #560 because it adds something more restrictive.

t-bast:
Superceded by #737

PR #734, opened 2020-01-31
BOLT 4: `tlv_payload` writer requirements indent fix.

rustyrussell:
I changed the tabs to spaces, but more importantly, the
` - MUST set `total_msat` to the total amount it will send` requirement
is *NOT* dependent on `- if the recipient provided `payment_secret`:`
so must not be indented underneath it.  Moved it above that for extra
clarity.


rustyrussell:
Nope, misread.  Current spec is correct.

PR #736, opened 2020-02-02
BOLT11 additional and negative tests

rustyrussell:
As per last meeting, this expands the BOLT11 test vectors.  I've reproduced them with the c-lightning unit tests.


11-payment-encoding.md
```diff
@@ -609,6 +668,27 @@ Breakdown:
...
+> ### Signature is not recoverable.
+> lnbc2500u1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5xysxxatsyp3k7enxv4jsxqzpuaxtrnwngzn3kdzw5hydlzf03qdgm2hdq27cqv3agm2awhz5se903vruatfhq77w3ls4evs3ch9zw97j25emudupq63nyw24cg27h2rspk28uwq
+
+> ### String is too short.
+> lnbc1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdpl2pkx2ctnv5sxxmmwwd5kgetjypeh2ursdae8g6na6hlh
+
+> ### Invalid multiplier
+> lnbc2500x1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5xysxxatsyp3k7enxv4jsxqzpujr6jxr9gq9pv6g46y7d20jfkegkg4gljz2ea2a3m9lmvvr95tq2s0kvu70u3axgelz3kyvtp2ywwt0y8hkx2869zq5dll9nelr83zzqqpgl2zg
+
+> ### Invalid sub-millisatoshi precision.
```
t-bast: Just a small clarification on that one: IIUC there's no requirement on the reader in the spec.
So it's ok if a reader just rounds this down to millisatoshi without rejecting the payment?
Or should we make the spec stricter on readers in that case?

rustyrussell: Good q, I was undecided.  I would think reader would have to round *up*, but it's probably best to just reject as nonsensical.

rustyrussell: Actually, already there:

```
      - if multiplier is `p` and the last decimal of `amount` is not 0:
	    - MUST fail the payment.
```

t-bast: There's a requirement to fail that payment, so it's clear.

t-bast:
@cfromknecht or @Roasbeef, do we have an ACK to merge this?

rustyrussell:
OK, rebased: had to hoist the new pico-btc test higher, so we could put the new variants underneath the existing ones.

rustyrussell:
Merging as per meeting agreement: http://www.erisian.com.au/meetbot/lightning-dev/2020/lightning-dev.2020-05-11-20.08.html

PR #737, opened 2020-02-03
Reply channel range simplification

rustyrussell:
First a trivial renaming of the confusing `complete` field.

Then, we insist that query_channel_range replies be in order (everyone now does this anyway) and non-overlapping (ditto: c-lightning had strict enforcement on this, which was overzealous but means everyone now does it that way).  Currently c-lightning maintains a bitmap for every block it requested, but it's much simpler to assume ordered replies, and simply check if we received the final one.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -788,7 +788,7 @@ The receiver of `query_channel_range`:
 	  - MUST set `first_blocknum` less than or equal to the `first_blocknum` in `query_channel_range`
 	  - MUST set `first_blocknum` plus `number_of_blocks` greater than `first_blocknum` in `query_channel_range`.
 	- successive `reply_channel_range` message:
-	  - MUST set `first_blocknum` to the previous `first_blocknum` plus `number_of_blocks` minus one.
+	  - MUST set `first_blocknum` to the previous `first_blocknum` plus `number_of_blocks`.
 	- the final `reply_channel_range` message:
 	  - MUST have `first_blocknum` plus `number_of_blocks` equal or greater than the `query_channel_range` `first_blocknum` plus `number_of_blocks`.
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -776,17 +776,21 @@ The sender of `query_channel_range`:
 The receiver of `query_channel_range`:
   - if it has not sent all `reply_channel_range` to a previously received `query_channel_range` from this sender:
     - MAY fail the connection.
-  - MUST respond with one or more `reply_channel_range` whose combined range
-  cover the requested `first_blocknum` to `first_blocknum` plus `number_of_blocks` minus one.
-  - For each `reply_channel_range`:
+  - MUST respond with one or more `reply_channel_range`:
     - MUST set with `chain_hash` equal to that of `query_channel_range`,
-    - MUST encode a `short_channel_id` for every open channel it knows in blocks `first_blocknum` to `first_blocknum` plus `number_of_blocks` minus one.
     - MUST limit `number_of_blocks` to the maximum number of blocks whose
       results could fit in `encoded_short_ids`
     - if does not maintain up-to-date channel information for `chain_hash`:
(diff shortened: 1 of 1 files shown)


07-routing-gossip.md
```diff
@@ -795,9 +799,15 @@ If the incoming message includes `query_option`, the receiver MAY append additio
...
-A single response might be too large for a single packet, and also a peer can
-store canned results for (say) 1000-block ranges, and simply offer each reply
-which overlaps the ranges of the request.
+A single response might be too large for a single packet, so multiple replies
+may be required.  We want to allow a peer to store canned results for (say)
+1000-block ranges, so replies can exceed the requested range.  However, we
+require that each reply be relevant (overlapping the requested range).
+
+By insisting that replies be in increasing order, the receiver can easily
+determine if replies are done: simply check if `first_blocknum` plus
```
t-bast: I think this doesn't work in all cases...
For example when we split the last requested block between multiple responses:

```
* query: first_blocknum = 10, number_of_blocks = 2
* response 1: first_blocknum = 10, number_of_blocks = 2, short_channel_ids = {10x1x1, 11x0x2}
* response 2: first_blocknum = 11, number_of_blocks = 1, short_channel_ids = {11x1x1}
```

You will consider it done after response 1.

rustyrussell: No, we assume that a single block's channels are always in the same response.  Otherwise you'd never know when to terminate.

t-bast: Then why do you need the "minus 1" line 791?

07-routing-gossip.md
```diff
@@ -776,17 +776,21 @@ The sender of `query_channel_range`:
...
-      - MUST set `complete` to 0.
+      - MUST set `full_information` to 0.
     - otherwise:
-      - SHOULD set `complete` to 1.
+      - SHOULD set `full_information` to 1.
+    - the first `reply_channel_range` message:
+	  - MUST set `first_blocknum` less than or equal to the `first_blocknum` in `query_channel_range`
+	  - MUST set `first_blocknum` plus `number_of_blocks` greater than `first_blocknum` in `query_channel_range`.
+	- successive `reply_channel_range` message:
+	  - MUST set `first_blocknum` to the previous `first_blocknum` plus `number_of_blocks` minus one.
```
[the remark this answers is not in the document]
rustyrussell: Note: @t-bast points out the "minus one" here is incorrect.

t-bast:
@wpaulino could you take a look at this? And share your thoughts on the fact that this PR disallows splitting a block's content over two replies?

07-routing-gossip.md
```diff
@@ -776,17 +776,21 @@ The sender of `query_channel_range`:
...
     - MUST limit `number_of_blocks` to the maximum number of blocks whose
       results could fit in `encoded_short_ids`
     - if does not maintain up-to-date channel information for `chain_hash`:
-      - MUST set `complete` to 0.
+      - MUST set `full_information` to 0.
     - otherwise:
-      - SHOULD set `complete` to 1.
+      - SHOULD set `full_information` to 1.
+    - the first `reply_channel_range` message:
+	  - MUST set `first_blocknum` less than or equal to the `first_blocknum` in `query_channel_range`
```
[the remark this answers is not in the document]
t-bast: I believe the rationale between this "less than or equal" is to allow pre-generated responses that may start a few blocks before the `first_blocknum`.

Imagine for example that blocks 3, 4, 5 and 6 fit in a pre-generated response. If you ask for `first_blocknum = 5` instead of generating a response that matches exactly that, I can give you my pre-generated response that contains blocks 3, 4, 5 and 6 and save some computing power. And you have the results you wanted so you should be happy with that.

This is extra flexibility that doesn't cost much, so it's worth having IMHO.

t-bast:
Merging as agreed during yesterday's meeting: http://www.erisian.com.au/meetbot/lightning-dev/2020/lightning-dev.2020-03-30-19.06.log.html

PR #740, opened 2020-02-11
Avoid stuck channels after fee increase with additional reserve

### LINKED ISSUE #728: Stuck channels because of small fee increase
I've been banding my head because of fee management and stuck channels. I've found a not so unreasonable situation where channels can become _completely unusable_ and I don't see a way to unblock them. I'm probably re-discovering something people already know, and I'm interested in your feedback and potential mitigations @cdecker @rustyrussell @pm47 @Roasbeef @cfromknecht @joostjager. I've tested eclair, c-lighting and lnd and they all behave the same (it seems **lnd goes into an infinite loop though**, might be worth investigating).

## Problem statement

Imagine we have Alice open a 150.000 sat channel to Bob, a feerateperkw of 10.000 sat/kw and a 1% reserve.
The CommitTx fee is `724*10 = 7240 sat` when there's no pending HTLC.
The additional fee per HTLC is `172*10 = 1720 sat`.

The balance of the channel can then be:

* `7240+1500 = 8740 sat` on Alice's side
* `141260 sat` on Bob's side

At that point the channel is completely valid but is now unusable.
Alice can't send any HTLC to Bob: that's expected, it's not an issue.
However Bob can't send any HTLC to Alice either: when Bob prepares adding this HTLC, the CommitTx weight becomes `724+172`. Bob notices that sending this HTLC would make Alice unable to afford the fee for this updated CommitTx, so Bob aborts and doesn't send the HTLC (regardless of the HTLC amount, unless it's below dust). The only way to unblock the channel is to wait for a feerate decrease of at least 19.2% (`172/(172+724)`). And when that feerate decrease happens, you need to send an HTLC Bob -> Alice before the fee increases again, otherwise you missed your fix window.

## Proposed solution

Now why doesn't Bob send the HTLC? In my opinion Bob can safely send the HTLC; however Alice takes a risk accepting it (and currently Alice would reject the HTLC in all 3 implementations). The only risk I see is that if Bob broadcasts the updated CommitTx while the HTLC is pending, the fee will be lower than the channel's feerate, so the transaction may take a while to confirm.

If the HTLC is fulfilled, but the CommitTx doesn't confirm, there will be an on-chain race condition between Alice's HTLC-success and Bob's HTLC timeout; if Bob wins the race, Bob will have stolen the HTLC amount from Alice.

It seems to me that Alice could easily take this risk; the estimated fee is supposed to be quite high already (the spec says `the current fee rate is sufficient (by a significant margin) for timely processing of the commitment transaction`), so even with an added 23.7% weight to the commit tx it should confirm in time to avoid the race condition. But maybe we should restrict that case to a single added HTLC (otherwise the weight can grow unbounded); that would be sufficient to unblock the channel. **Would that be too reckless in your opinion? Or is it a reasonable trade-off to avoid this problem?**

### CPFP / Anchor outputs

With CPFP on the CommitTx (anchor outputs) we should allow such HTLCs; if Bob broadcasts the CommitTx and it doesn't confirm, just CPFP it before the race condition is reached.

But I'd like to fix this in the shorter term too if possible :).

### Reaching the faulty state

It's easy to reach that faulty state for testing/repro by using the `push_msat` in the `open_channel`.
Just set `push_msat` to `141260000` (with the values used in this issue) and you should be in that state.

In real life though that's not how channels will end up in this state. Alice won't be able to send this amount via HTLCs because before sending, we take into account the weight this HTLC adds to the CommitTx (`172*feerate`). 

However imagine the feerate decreases to 8.000 sat/kw. Alice can now send an HTLC of `141260 sat` to Bob. Then the fee increases back to 10.000 sat/kw after the HTLC has been fulfilled and removed from the CommitTx: the `update_fee` is thus valid, and now Alice and Bob have a stuck channel.

t-bast:
Add an additional "reserve" for funders on top of the real reserve to
avoid getting in a state where the channel is unusable because
of the increased commit tx cost of a new HTLC.

Requirements are only added for the funder sending an HTLC.
Fundee receiving HTLCs may choose to verify that funders apply
this, but it may lead to an unusable UX.

@halseth do you want to chime in since you've been confronted
with this issue on the lnd side?

Fixes #728. (see the issue for a longer discussion)

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -787,9 +787,20 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
 #### Requirements
 
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see "Updating Fees").
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the fee for a
+    future additional non-dust HTLC at `2*feerate_per_kw` while maintaining its
+    channel reserve ("fee spike buffer").
+  - if it is _not responsible_ for paying the Bitcoin fee:
+    - SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    its commitment transaction, it cannot pay the fee for the updated local or
+    remote transaction at the current `feerate_per_kw` while maintaining its
+    channel reserve.
   - MUST offer `amount_msat` greater than 0.
   - MUST NOT offer `amount_msat` below the receiving node's `htlc_minimum_msat`
   - MUST set `cltv_expiry` less than 500000000.
@@ -855,6 +866,15 @@ seconds, and the protocol only supports an expiry in blocks.
 `amount_msat` is deliberately limited for this version of the
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
+
+The node _responsible_ for paying the Bitcoin fee has to maintain a small "fee
+spike buffer" on top of its reserve to accommodate a future fee increase.
+Without this buffer, the node _responsible_ for paying the Bitcoin fee may
+reach a state where it is unable to receive any HTLC while maintaining its
+channel reserve (because of the increased weight of the commitment transaction),
+resulting in an unusable channel.
+See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728) for
+more details.
 
 ### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc`, and `update_fail_malformed_htlc`
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -794,11 +794,12 @@ A sending node:
     reserve (see [Updating Fees](#updating-fees-update_fee)).
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, its remaining balance doesn't allow it to pay the commitment
-    transaction fee when receiving a future additional non-dust HTLC while maintaining
-    its channel reserve. It is recommended that this "fee spike buffer" can handle
-    twice the current `feerate_per_kw` to ensure predictability between implementations.
+    transaction fee when receiving or sending a future additional non-dust HTLC
+    while maintaining its channel reserve. It is recommended that this "fee spike
+    buffer" can handle twice the current `feerate_per_kw` to ensure predictability
+    between implementations.
   - if it is _not responsible_ for paying the Bitcoin fee:
-    - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    - SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or
     remote transaction at the current `feerate_per_kw` while maintaining its
     channel reserve.
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -872,9 +872,8 @@ spike buffer" on top of its reserve to accommodate a future fee increase.
 Without this buffer, the node _responsible_ for paying the Bitcoin fee may
 reach a state where it is unable to receive any HTLC while maintaining its
 channel reserve (because of the increased weight of the commitment transaction),
-resulting in an unusable channel.
-See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728) for
-more details.
+resulting in an unusable channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
+for more details.
 
 ### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc`, and `update_fail_malformed_htlc`
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -793,10 +793,10 @@ A sending node:
     transaction at the current `feerate_per_kw` while maintaining its channel
     reserve (see [Updating Fees](#updating-fees-update_fee)).
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
-    transaction, its remaining balance doesn't allow it to pay the fee for a
-    future additional non-dust HTLC at a higher feerate while maintaining its
-    channel reserve ("fee spike buffer"). A buffer of `2*feerate_per_kw` is
-    recommended to ensure predictability.
+    transaction, its remaining balance doesn't allow it to pay the fee to receive a
+    future additional non-dust HTLC while maintaining its channel reserve ("fee
+    spike buffer"). A buffer that expects the current `feerate_per_kw` to potentially
+    double is recommended to ensure predictability between implementations.
   - if it is _not responsible_ for paying the Bitcoin fee:
     - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -795,9 +795,8 @@ A sending node:
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, its remaining balance doesn't allow it to pay the commitment
     transaction fee when receiving a future additional non-dust HTLC while maintaining
-    its channel reserve ("fee spike buffer"). A buffer that expects the current
-    `feerate_per_kw` to potentially double is recommended to ensure predictability
-    between implementations.
+    its channel reserve. It is recommended that this "fee spike buffer" can handle
+    twice the current `feerate_per_kw` to ensure predictability between implementations.
   - if it is _not responsible_ for paying the Bitcoin fee:
     - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or
@@ -872,9 +871,9 @@ bootstrap phase of the network.
 The node _responsible_ for paying the Bitcoin fee should maintain a "fee
 spike buffer" on top of its reserve to accommodate a future fee increase.
 Without this buffer, the node _responsible_ for paying the Bitcoin fee may
-reach a state where it is unable to receive any HTLC while maintaining its
-channel reserve (because of the increased weight of the commitment transaction),
-resulting in an unusable channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
+reach a state where it is unable to receive any non-dust HTLC while maintaining
+its channel reserve (because of the increased weight of the commitment transaction),
+resulting in a degraded channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
 for more details.
 
 ### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc`, and `update_fail_malformed_htlc`
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -791,12 +791,12 @@ A sending node:
     - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, it cannot pay the fee for either the local or remote commitment
     transaction at the current `feerate_per_kw` while maintaining its channel
-    reserve (see "Updating Fees").
+    reserve (see [Updating Fees](#updating-fees-update_fee)).
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, its remaining balance doesn't allow it to pay the fee for a
-    future additional non-dust HTLC at `N*feerate_per_kw` while maintaining its
-    channel reserve ("fee spike buffer"), where `N` is a parameter chosen by the
-    implementation (`N = 2` is recommended to ensure predictability).
+    future additional non-dust HTLC at a higher feerate while maintaining its
+    channel reserve ("fee spike buffer"). A buffer of `2*feerate_per_kw` is
+    recommended to ensure predictability.
   - if it is _not responsible_ for paying the Bitcoin fee:
     - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or
@@ -868,7 +868,7 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
-The node _responsible_ for paying the Bitcoin fee should maintain a small "fee
+The node _responsible_ for paying the Bitcoin fee should maintain a "fee
 spike buffer" on top of its reserve to accommodate a future fee increase.
 Without this buffer, the node _responsible_ for paying the Bitcoin fee may
 reach a state where it is unable to receive any HTLC while maintaining its
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -872,9 +872,9 @@ bootstrap phase of the network.
 The node _responsible_ for paying the Bitcoin fee should maintain a "fee
 spike buffer" on top of its reserve to accommodate a future fee increase.
 Without this buffer, the node _responsible_ for paying the Bitcoin fee may
-reach a state where it is unable to receive any non-dust HTLC while maintaining
-its channel reserve (because of the increased weight of the commitment transaction),
-resulting in a degraded channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
+reach a state where it is unable to send or receive any non-dust HTLC while
+maintaining its channel reserve (because of the increased weight of the
+commitment transaction), resulting in a degraded channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
 for more details.
 
 ### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc`, and `update_fail_malformed_htlc`
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -794,10 +794,11 @@ A sending node:
     reserve (see "Updating Fees").
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, its remaining balance doesn't allow it to pay the fee for a
-    future additional non-dust HTLC at `2*feerate_per_kw` while maintaining its
-    channel reserve ("fee spike buffer").
+    future additional non-dust HTLC at `N*feerate_per_kw` while maintaining its
+    channel reserve ("fee spike buffer"), where `N` is a parameter chosen by the
+    implementation (`N = 2` is recommended to ensure predictability).
   - if it is _not responsible_ for paying the Bitcoin fee:
-    - SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or
     remote transaction at the current `feerate_per_kw` while maintaining its
     channel reserve.
@@ -867,7 +868,7 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
-The node _responsible_ for paying the Bitcoin fee has to maintain a small "fee
+The node _responsible_ for paying the Bitcoin fee should maintain a small "fee
 spike buffer" on top of its reserve to accommodate a future fee increase.
 Without this buffer, the node _responsible_ for paying the Bitcoin fee may
 reach a state where it is unable to receive any HTLC while maintaining its
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -793,10 +793,11 @@ A sending node:
     transaction at the current `feerate_per_kw` while maintaining its channel
     reserve (see [Updating Fees](#updating-fees-update_fee)).
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
-    transaction, its remaining balance doesn't allow it to pay the fee to receive a
-    future additional non-dust HTLC while maintaining its channel reserve ("fee
-    spike buffer"). A buffer that expects the current `feerate_per_kw` to potentially
-    double is recommended to ensure predictability between implementations.
+    transaction, its remaining balance doesn't allow it to pay the commitment
+    transaction fee when receiving a future additional non-dust HTLC while maintaining
+    its channel reserve ("fee spike buffer"). A buffer that expects the current
+    `feerate_per_kw` to potentially double is recommended to ensure predictability
+    between implementations.
   - if it is _not responsible_ for paying the Bitcoin fee:
     - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
     its commitment transaction, it cannot pay the fee for the updated local or


02-peer-protocol.md
```diff
@@ -787,9 +787,17 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` it cannot pay for in the remote commitment
+    transaction at the current `feerate_per_kw` (see "Updating Fees") while
+    maintaining its channel reserve.
+    - MUST NOT offer `amount_msat` if the resulting balance doesn't allow it to
+    pay the fee for a future additional HTLC at `2*feerate_per_kw` while
```
t-bast: This factor 2 should be challenged. Matt proposed 5, Rusty implemented 1.5, meeting in the middle was an ugly 3.25 so I chose 2 completely randomly.

02-peer-protocol.md
```diff
@@ -787,9 +787,17 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
 
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` it cannot pay for in the remote commitment
+    transaction at the current `feerate_per_kw` (see "Updating Fees") while
+    maintaining its channel reserve.
+    - MUST NOT offer `amount_msat` if the resulting balance doesn't allow it to
```
halseth: is "resulting balance" here referring to balance after the HTLC is settled or while it is still on the commitment?

t-bast: This is where I struggled the most with the wording...

It's both in my opinion, but unless I'm missing something (correct me if I'm wrong there, this really needs some challenging) the more restrictive case is the one where the HTLC is still in the commitment tx (because the commit tx fee is higher in that case). If you're able to satisfy the condition while the HTLC is in the commitment, you're able to satisfy it once it's resolved.

If we agree that's what we expect, I can try to find a more accurate wording to reflect that. Maybe "MUST NOT offer `amount_msat` if, once that HTLC is added to the remote commitment transaction, the sender cannot pay the fee for a future additional HTLC at `2*feerate_per_kw` while maintaining its channel reserve."?

halseth: Mhm, maybe we have to be this restrictive, otherwise the other node possibly cannot add more HTLCs while this HTLC is still pending. If we only required that the balance was high enough after the HTLC had been settled, then the channel would essentially be "locked" while this HTLC was still pending. But it would also make less of you balance available to send of course...

02-peer-protocol.md
joostjager: I think hard-coding the `2*` via MUST (and later also making this a hard requirement on the receiver side) is not necessary and could be downgraded to SHOULD.

joostjager: Specify that it is a 'future non-dust htlc'?

02-peer-protocol.md
```diff
@@ -787,9 +787,17 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
 #### Requirements
 
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` it cannot pay for in the remote commitment
+    transaction at the current `feerate_per_kw` (see "Updating Fees") while
+    maintaining its channel reserve.
```
joostjager: Must also not offer an htlc if it cannot be paid for when the remote adds this htlc to the funder's local commitment tx. There can be a different dust limit and it therefore needs to be checked as well.

02-peer-protocol.md
t-bast: I'm ok with making that one a SHOULD. However please note that if we go down the road of this "extra reserve", I think it's important that implementations agree on the factor (current `2`). That allows a node to compute exactly how much the remote is able to send or receive without having to try and fail. Does that make sense?

t-bast: Good point, I'll add that.

02-peer-protocol.md
t-bast: True, this sentence also needs clarifying, it seems that lnd, c-lightning and eclair all understood it the same way but the spec isn't clear enough. I'll give a shot at clarifying that.

t-bast: Can we simply change `in the remote commitment transaction` to `in either the local or remote commitment transaction`?
Does that correctly account for the case you're mentioning?

02-peer-protocol.md
m-schmoock: suggestion: 

```
pay the fee for a future additional untrimmed HTLC at `2*feerate_per_kw` while
```

m-schmoock:
@t-bast note: lets think about if we need signaling.

A very nice feature is that we currently know **exactly** how much msat a channel can receive or send, if we don't add signaling for this change, we can't know anymore, since a remote might or might not support this (yet), which kinda eliminates the reason I started this discussion with my little circular routing remote channel lockup POC...

02-peer-protocol.md
```diff
@@ -856,6 +864,11 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee has to maintain a small
+"additional reserve" on top of its reserve to accommodate a fee increase,
+otherwise the channel may end up in an unusable state
```
ariard: Explain how this fix the issue: "Balance may reach the state, where any committed HTLC can't be added from the fundee-side without decreasing funder balance under the security reserve to pay fees for commitment tx. By keeping a _fee-variance buffer_, funder guarantee to the fundee at least one HTLC to rebalance channel."

02-peer-protocol.md
ariard: Do we have other cases where a remote can't reliably predict if a HTLC is going to be rejected, knowing channel parameters and balance ? If yes, SHOULD is fine.

t-bast: We can never really predict that, because a node can implement any policy it wants on HTLC filtering and reject HTLCs arbitrarily...there's always a chance that a valid HTLC you're offering ends up being rejecting by the remote.

There's also always the risk of a concurrent fee update, which makes your valid-looking HTLC invalid but you couldn't know that because you didn't yet receive information about the new fee.

t-bast:
> A very nice feature is that we currently know exactly how much msat a channel can receive or send, if we don't add signaling for this change, we can't know anymore, since a remote might or might not support this (yet), which kinda eliminates the reason I started this discussion with my little circular routing remote channel lockup POC...

Estimating how much you can *receive* will always be only a best effort, not a guarantee (see some of the reasons in the comments on this PR). After thinking about it I think that signalling this explicitly would be overkill. It would be better to just assume that the remote (if he's funder) uses this extra reserve. If he doesn't, that means you underestimated a bit how much you were able to receive, but that's probably not a big deal. If you use that value in invoices you create, that's even better because you help the remote artificially keep this additional reserve even if he doesn't explicitly support it, so you're helping him avoid getting stuck.

m-schmoock:
> After thinking about it I think that signalling this explicitly would be overkill. It would be better to just assume that the remote (if he's funder) uses this extra reserve.

I agree

02-peer-protocol.md
t-bast: Done.

02-peer-protocol.md
t-bast: I applied that change, let me know if you like the new wording.

02-peer-protocol.md
t-bast: Done, let me know if the current wording is better.

02-peer-protocol.md
t-bast: I updated this to a SHOULD.

02-peer-protocol.md
```diff
@@ -787,9 +787,20 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
 
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see "Updating Fees").
```
joostjager: Add note that this is not entirely preventable in certain scenarios (concurrency)?

02-peer-protocol.md
```diff
@@ -856,6 +867,15 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee has to maintain a small "fee
+spike buffer" on top of its reserve to accommodate a future fee increase.
+Without this buffer, the node _responsible_ for paying the Bitcoin fee may
+reach a state where it is unable to receive any HTLC while maintaining its
+channel reserve (because of the increased weight of the commitment transaction),
+resulting in an unusable channel.
+See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728) for
```
joostjager: Merge this paragraph with the one above?

02-peer-protocol.md
t-bast: I think this is already explained in the "Updating Fees" section that's linked.
Do you think it's worth replicating this information here?

02-peer-protocol.md
```diff
@@ -787,9 +787,20 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+    reserve (see "Updating Fees").
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the fee for a
+    future additional non-dust HTLC at `2*feerate_per_kw` while maintaining its
+    channel reserve ("fee spike buffer").
+  - if it is _not responsible_ for paying the Bitcoin fee:
+    - SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    its commitment transaction, it cannot pay the fee for the updated local or
+    remote transaction at the current `feerate_per_kw` while maintaining its
+    channel reserve.
```
```suggestion
    - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
    its commitment transaction, it cannot pay the fee for the updated local or
    remote transaction at the current `feerate_per_kw` while maintaining its
    channel reserve. Note that there is a possibility of a race with HTLCs being
    simultaneously sent by the remote node; if that happens the channel will fail.
```

Just because the fundee cannot guarantee (because of the race) that the reserve will be maintained doesn't mean that this should be a `SHOULD` imo.

02-peer-protocol.md
```diff
@@ -856,6 +867,14 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee has to maintain a small "fee
```
halseth: *should maintain

02-peer-protocol.md
```diff
@@ -787,9 +787,20 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see "Updating Fees").
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the fee for a
+    future additional non-dust HTLC at `2*feerate_per_kw` while maintaining its
```
halseth: Now that we make this `SHOULD`, we can make it clear that a node can make this fee buffer as small/large as it want, no need to choose a constant (although 2 can be recommended)

m-schmoock: No, If we have no signaling of this value/feature (currently not planned) we should really not encourage to set other values than 2 because we loose calculatability how much we can receive on any channel.

halseth: Reading this latest iteration of the PR, it seems that B (In the channel A->B) doesn't have to take this constant into account at all:

```
    SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
    channel reserve.
```

which is the same as was previously specced.

t-bast: I agree with @halseth there. I removed the requirement on the fundee, because we can't force fundees to respect this additional reserve (since race conditions exist, there will always be cases where this requirement will not hold). And going in that direction, it makes sense not to hardcode that factor 2 in the spec (otherwise this will simply encourage implementations to *not* be spec-compliant if at some point in the future a factor 2 doesn't seem to be enough).

I understand that from a UX perspective you want to be able to compute how much you can receive, but that calculation will always be flawed. You can't restrict a remote peer's behaviour; your peer could decide to never let the channel dip below 30% on his side, except in some specific cases. Or he could have any other kind of weird behaviour.
That doesn't prevent you from calculating an indication of how much you could receive, if the remote peer behaved in a specific way. That may be slightly different from reality, but it's the best you can get.

joostjager:
@TheBlueMatt about your irc comment on commit fee and packaged relays: Suppose the packaged relays are there, wouldn't you rather use 0 sat/b than any other fixed fee rate on the commitment txes? With zero fees, this pr wouldn't be necessary. Htlcs can always be sent.

m-schmoock:
ACK 95c74fe

From my side, I can live this PR as a mitigation. Not perfect, but adding signaling on the applied factor seems overkill. It mostly 'fixes the symptom' and makes it hard to exploit until we have eltoo.

02-peer-protocol.md
joostjager: Ok, no then I think it is good as is. Could make "Updating Fees" a link.

t-bast: True, that would be helpful. I'll make this a link to the relevant section.

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see "Updating Fees").
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the fee for a
+    future additional non-dust HTLC at `N*feerate_per_kw` while maintaining its
+    channel reserve ("fee spike buffer"), where `N` is a parameter chosen by the
+    implementation (`N = 2` is recommended to ensure predictability).
```
joostjager: I am wondering now, does it make sense to include any formula at all in the spec? This `N*feerate` is just an idea. A different implementation may for example just use a fixed buffer regardless of what the current fee rate is.

02-peer-protocol.md
```diff
@@ -856,6 +868,14 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee should maintain a small "fee
```
joostjager: What does the word 'small' add here?

t-bast: Good catch, it should be removed.

02-peer-protocol.md
t-bast: You're right, we can give more leeway here and just mention this as a suggestion. I think it's still worth having as a suggestion as it's the most natural way we came up with, but it's not a hard requirement.

m-schmoock: @joostjager  Kinda agree, but we should definitively mention `2*feerate_per_kw` as a suggestion to maintain some kind of predictability.

02-peer-protocol.md
t-bast: Done in f5490f17d17ff49dc26ee459432b3c9db4fda8a9

02-peer-protocol.md
t-bast: Done in f5490f17d17ff49dc26ee459432b3c9db4fda8a9

02-peer-protocol.md
t-bast: Done in f5490f17d17ff49dc26ee459432b3c9db4fda8a9

t-bast:
@joostjager @halseth could you please do a last pass on that before the spec meeting?

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see [Updating Fees](#updating-fees-update_fee)).
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the fee for a
+    future additional non-dust HTLC at a higher feerate while maintaining its
```
halseth: can remove "higher feerate"? It is kinda implied from the recommended buffer below.

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
 A sending node:
-  - MUST NOT offer `amount_msat` it cannot pay for in the
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see [Updating Fees](#updating-fees-update_fee)).
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
```
halseth: Now this sounds like the sender must be able to send another HTLC before this one settles. I think we should say just "the remaining balance after it has settled doesn't..."

t-bast: On the contrary, I think that "after it has settled" would be incorrect for two reasons:

* this paragraph shouldn't prevent you from sending multiple HTLCs (not settled) as long as your fee buffer allows it
* when you send an outgoing htlc, you should be able to receive an incoming htlc: this way even if your outgoing htlc is stuck in your commit tx, you can receive an incoming to increase your balance and prevent your channel from being stuck. The case where the first outgoing htlc is still in the commit tx costs more fees than the case where it has settled, so that should be handled.

Does this make sense or am I missing something?

halseth: I think it comes down to how strict you want to be. The channel being stuck because you have a pending HTLC on it is arguably not stuck, since it will eventually be unstuck.

> even if your outgoing htlc is stuck in your commit tx, you can receive an incoming

Not sure how important this is?

t-bast: > Now this sounds like the sender must be able to send another HTLC before this one settles.

It's more that the current wording _allows_ the sender to send another HTLC before this one settles (if he has the balance to do so). IIUC your proposal disallows that, and I don't see why we'd need to be that restrictive.

> Not sure how important this is?

If you disallow that, that means you potentially need to wait for a cltv-timeout before you can unblock your channel. True, it will _eventually_ settle, but there's no reason to wait for that and be penalized by a stuck HTLC. If you can accept an incoming HTLC even though the outgoing is still pending, that allows you to increase your balance and then send other outgoing HTLCs, which increases your relay throughput.

I think the current wording is less strict than what you propose: you just deal with the present situation instead of predicting the situation after the htlc is settled. Is this correct or am I misunderstanding your point?

halseth: > It's more that the current wording allows the sender to send another HTLC before this one settles (if he has the balance to do so).

No, if the sender has some balance left, he must reserve this for fees to receive another HTLC. What I suggested allows using more of this balance, since only after it settles we require an HTLC to be received.

> If you can accept an incoming HTLC even though the outgoing is still pending, that allows you to increase your balance and then send other outgoing HTLCs, which increases your relay throughput.

This is true, the current wording makes sure the channel is stuck "less" 😛  What I meant with the current working being more strict is that it also disallows the "stuck-while-HTLC-pending" case.

To summarize the way I interpret this: the current wording requires the initiator to always keep enough balance to receive another HTLC. My suggestion was to loosen this to require the initiator to keep enough balance to receive another HTLC when there are no HTLCs pending.

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+    transaction, its remaining balance doesn't allow it to pay the fee for a
```
halseth: I also realized this can be interpreted as the remaining balance must be enough for (non-dust HTLC value + fee for this HTLC). Maybe we should just explicitly state that it should be enough to pay the fee to receive another HTLC.

02-peer-protocol.md
t-bast: > To summarize the way I interpret this: the current wording requires the initiator to always keep enough balance to receive another HTLC. My suggestion was to loosen this to require the initiator to keep enough balance to receive another HTLC when there are no HTLCs pending.

Thanks, this is a lot clearer with that explanation. It is an alternative that makes sense, but it has a small additional trade-off and I prefer the current proposal :smile:.

This issue happens when the balance on your side is low (very close to the reserve). When that happens, I think you should prioritize incoming HTLCs (that will increase your balance), not outgoing HTLCs (that keep getting you closer to a stuck channel).

I see two reasons for that. The first one is that the additional outgoing HTLC that you could send is tiny (it's on the order of an htlc fee), so it isn't a game-changer to allow it to go through. The second reason is that you're reducing the window of time where you allow incoming HTLCs to restore balance in your channel. While you have that last outgoing HTLC in your commitment, you are rejecting incoming HTLCs (which could have helped your channel balance). When that last outgoing HTLC settles, if you re-add another outgoing one you missed your opportunity window to receive an incoming HTLC and you need to wait for another outgoing HTLC to settle. And I think that incoming HTLCs should be prioritized because they are likely to be bigger (since the only outgoing HTLC you can add is very tiny).

Thanks for taking the time to detail your point, I think it's really a matter of preference and both could apply, so it's an interesting discussion to have.

02-peer-protocol.md
t-bast: True, I'll try re-wording that a bit.

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+    future additional non-dust HTLC at a higher feerate while maintaining its
+    channel reserve ("fee spike buffer"). A buffer of `2*feerate_per_kw` is
+    recommended to ensure predictability.
```
SomberNight: I suppose the wording is not clear enough, given both c-lightning and eclair now implemented this, but in incompatible ways due to different interpretations.

- The wording here is basically "ctx plus 1 more HTLC at twice the current feerate".
- [In eclair](https://github.com/ACINQ/eclair/blame/eae113f0988069044dc268d8a7f16728563baa5f/eclair-core/src/main/scala/fr/acinq/eclair/channel/Commitments.scala#L192-L198), the implementation checks "ctx plus (1 more HTLC at twice the current feerate)":
there is an "additional reserve", which is the on-chain fee cost of 2 more HTLC outputs at the current feerate.
- [In c-lightning](https://github.com/ElementsProject/lightning/blob/af7e87930878117a0633bb7a8cd4488b8e9c5b6d/channeld/full_channel.c#L430-L440) however, the check is "(ctx plus 1 more HTLC) at twice the current feerate":
there is an "additional reserve", set by calculating how much fees the funder would need to pay for the whole commitment transaction (with an extra dummy HTLC).

(to be clear, in both cases, the "ctx" already includes the HTLC being offered, and the "1 more HTLC" is an additional one)

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+  - if it is _not responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    its commitment transaction, it cannot pay the fee for the updated local or
+    remote transaction at the current `feerate_per_kw` while maintaining its
+    channel reserve.
```
SomberNight: So is the sender supposed to check whether the remote funder can pay on-chain fees for the receiver's **next** commitment tx at corresponding feerate1, and also whether the remote funder can pay on-chain fees for the sender's **next** commitment tx at corresponding feerate2? (note that e.g. dust limits might differ so both ctxs need to be inspected)

If so, what is feerate2? The sender (being the fundee) has no way of knowing the actual future feerate for that tx at this point.

Indeed while the sender is running these checks and considering sending an HTLC, if the remote funder sends an update_fee exactly at that time the actual feerate2 value will change.
Note that the feerate changing implies the htlc trim threshold changing, so even dust value HTLCs are not safe to send due to this race (as they might not be dust at that later time).

02-peer-protocol.md
t-bast: Interesting, that means we'll have to discuss this further on IRC tonight to get more bandwidth than discussing it on Github ;)
Feel free to suggest a better wording in the meantime though.

02-peer-protocol.md
t-bast: > If so, what is feerate2? The sender (being the fundee) has no way of knowing the actual future feerate for that tx at this point.

Both peers know the feerate. It's what's currently applied to the commit tx. There is no "feerate2" here, the only thing both know is the current feerate.

Obviously there may be concurrent `update_fee`, but you can't predict those. Depending on what is signed and revoked first (the HTLC or an `update_fee`) one of the two may fail. The faster will succeed, the slower will fail.

02-peer-protocol.md
SomberNight: I don't know what the wording should be because I don't know what the intended meaning is. :)

02-peer-protocol.md
SomberNight: > Depending on what is signed and revoked first (the HTLC or an update_fee) one of the two may fail. The faster will succeed, the slower will fail.

Sure but it might fail in a way that such there will be a commitment tx with the channel reserve violated. For example, local add htlc and commits right away, while remote concurrently updates (increases) fee and commits right away. Now, in the "next" commitment tx of remote, they might not afford the onchain fees (they could have afforded the fees if either one of the updates did not occur).
As remote dipped below channel reserve, and the updates are now committed, depending on implementation, the channel might get force-closed at that point.

Note that if there was also a "fee spike buffer" in this scenario, the fatal race condition would become less likely.

m-schmoock: No, we don't force close when the remote violates this additional margin. Its a choice how much margin the remote calculates though.   2*feerate is just the default suggestion.

m-schmoock: The different interpretations will just lead to `spendable_msat` and `receivable_msat` calculations being wrong, leading to potentially CAPACITY EXCEEDED errors when someone tries to drain a channel.

t-bast: > Sure but it might fail in a way that such there will be a commitment tx with the channel reserve violated.

That concurrency has always existed. It's mentioned in the `Updating Fees` paragraph below. We will continue handling those races as detailed in that paragraph. But the outcome isn't a force-close; on the contrary, the outcome is that the current commitment tx may have a lower feerate than what the funder wishes it to be because it couldn't go through with his `update_fee`.

SomberNight: > Both peers know the feerate. It's what's currently applied to the commit tx. There is no "feerate2" here, the only thing both know is the current feerate.

Also, there are actually two different feerates, if e.g.:
- Remote sends update_fee (changes fee from feerate1 to feerate2), and local receives this. It is not yet committed.
- Local wants to send update_add_htlc. Now, what feerates should local use when running the checks? The next remote ctx will still use feerate1, while the next local ctx will use feerate2.

t-bast: There is only one latest feerate that has been signed by both parties (everything else is just a pending change).

> Local wants to send update_add_htlc

If local chooses to apply and sign the `update_fee` with the new feerate, it uses the new feerate. Otherwise it uses the previous feerate. Choosing which one it applies first (or both at the same time) is an implementation choice and all are valid from the spec's point of view.

I feel this discussion is not related to stuck channels and the fee spike buffer though. If you want to further discuss `update_fee` and its asynchronous nature, can you please open an issue where we'd continue? I'd really like to make progress on stuck channels and I fear this side discussion is going to add more unrelated cognitive load on reviewers.

SomberNight: > That concurrency has always existed. It's mentioned in the `Updating Fees` paragraph below. We will continue handling those races as detailed in that paragraph. 

Ah, you are right, I forgot about that. I guess that contradicts with this other paragraph though?

> The channel reserve is specified by the peer's channel_reserve_satoshis: 1% of the channel total is suggested. Each side of a channel maintains this reserve so it always has something to lose if it were to try to broadcast an old, revoked commitment transaction. Initially, this reserve may not be met, as only one side has funds; **but the protocol ensures that there is always progress toward meeting this reserve, and once met, it is maintained**.

If this race happens, the channel reserve can be violated.
And a malicious node can intentionally keep trying to trigger the race.

SomberNight: > I feel this discussion is not related to stuck channels and the fee spike buffer though. If you want to further discuss update_fee and its asynchronous nature, can you please open an issue where we'd continue? 

This discussion might not be related to "stuck channels" but it is certainly related to the highlighted text at the top of this thread which is new text added by this PR.
I am questioning what current `feerate_per_kw` is.
In any case, as there is no corresponding check by the receiver, I guess it's fine if everyone interprets it their own way. -- it only really matters if you intend to add a corresponding receiver check later.

t-bast: > I guess that contradicts with this other paragraph though?
> If this race happens, the channel reserve can be violated.

No it cannot. I don't understand why you're claiming that...?

Let's ignore the initial period where one side has 0 funds, and jump to the moment where both sides have funds above the reserve. There is no situation where the channel reserve will be violated, and this PR doesn't change that invariant. I really don't see why you think it does.

SomberNight: Well this part (channel reserve) of this conversation is indeed getting off-topic a bit, I tried to find you on IRC but could not.
In any case, as established above, a race can happen, where the funder cannot actually afford the onchain fees, and in that case this paragraph applies:
> Base commitment transaction fees are extracted from the funder's amount; if that amount is insufficient, the entire amount of the funder's output is used.

That is, it could happen that the funder does not have an output in the commitment tx. In the extreme case, due to the race, given appropriate dust_limit and feerate values, it could happen that the funder's output (would it not be pruned) would be 0 sat. If after fees the funder's output would be 0 sat, how could the channel reserve be met?

02-peer-protocol.md
halseth: I think that makes sense to suggest, always being able to receive definitely keeps your channel more healthy more of the time!

02-peer-protocol.md
t-bast: > I'll open a new issue just for this channel reserve part.

Please do, that will help others join and help clarify this without side-tracking this PR.
This topic had already been discussed a lot in the past, it's worth doing a bit of research to dig up those discussions.

02-peer-protocol.md
t-bast: I think c-lighting got it right. An increase of the feerate must be applied to the whole tx, not only an htlc, so my implementation in eclair is a bit weird. I tried rephrasing in https://github.com/lightningnetwork/lightning-rfc/pull/740/commits/afb7313e0f81bfee32aaa9b71fbf6c512407e8d1, I hope it's clearer. Feel free to suggest something if still unclear (I'm still not 100% happy with what I came up with).

02-peer-protocol.md
t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/740/commits/afb7313e0f81bfee32aaa9b71fbf6c512407e8d1

02-peer-protocol.md
t-bast: I made receiving an HTLC more explicitly the goal of this requirement in https://github.com/lightningnetwork/lightning-rfc/pull/740/commits/afb7313e0f81bfee32aaa9b71fbf6c512407e8d1, let me know if that's better.

02-peer-protocol.md
```diff
@@ -793,10 +793,11 @@ A sending node:
     transaction at the current `feerate_per_kw` while maintaining its channel
     reserve (see [Updating Fees](#updating-fees-update_fee)).
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
-    transaction, its remaining balance doesn't allow it to pay the fee to receive a
-    future additional non-dust HTLC while maintaining its channel reserve ("fee
-    spike buffer"). A buffer that expects the current `feerate_per_kw` to potentially
-    double is recommended to ensure predictability between implementations.
+    transaction, its remaining balance doesn't allow it to pay the commitment
+    transaction fee when receiving a future additional non-dust HTLC while maintaining
```
m-schmoock: > `when receiving a future additional non-dust HTLC`

maybe:  `when adding another potential non-dust HTLC`

m-schmoock: or just: `when adding a future additional non-dust HTLC`
Or does the word `receiving` add anything important here?

t-bast: I think `receiving` is important: see that discussion: https://github.com/lightningnetwork/lightning-rfc/pull/740#discussion_r399197774

02-peer-protocol.md
```diff
@@ -787,9 +787,22 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see [Updating Fees](#updating-fees-update_fee)).
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the commitment
+    transaction fee when receiving a future additional non-dust HTLC while maintaining
+    its channel reserve ("fee spike buffer"). A buffer that expects the current
```
joostjager: This makes it sound like "channel reserve" == "fee spike buffer", while afaik "channel reserve" is only the reserve negotiated in open channel?

02-peer-protocol.md
```diff
@@ -787,9 +787,22 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+    `feerate_per_kw` to potentially double is recommended to ensure predictability
+    between implementations.
+  - if it is _not responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, once the remote node adds that HTLC to
+    its commitment transaction, it cannot pay the fee for the updated local or
+    remote transaction at the current `feerate_per_kw` while maintaining its
+    channel reserve.
```
joostjager: This channel reserve is again the reserve excluding the spike buffer, right? (also because it says MUST NOT)

02-peer-protocol.md
```diff
@@ -856,6 +869,14 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee should maintain a "fee
+spike buffer" on top of its reserve to accommodate a future fee increase.
+Without this buffer, the node _responsible_ for paying the Bitcoin fee may
+reach a state where it is unable to receive any HTLC while maintaining its
+channel reserve (because of the increased weight of the commitment transaction),
+resulting in an unusable channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
```
joostjager: Complete unusable, are better to say 'degraded', because dust htlcs are still possible?

t-bast: True, I'll update that, unusable is a bit strong :)

02-peer-protocol.md
t-bast: Yes this is the reserve we know and love. The "fee spike buffer" is only applicable to the paragraph it's in (below).

02-peer-protocol.md
t-bast: I didn't understand what you mean at first, but I get it now. It sounds like what's inside the parenthesis complements what comes right before (channel reserve). I'll try to re-word that to avoid linking the two.

t-bast: I re-phrased in https://github.com/lightningnetwork/lightning-rfc/pull/740/commits/1ff808335dd3dc2f30e713d6ea98ed7297c354f3, let me know if that's better. I'm still not very happy with the last sentence...

02-peer-protocol.md
```diff
@@ -856,6 +868,14 @@ seconds, and the protocol only supports an expiry in blocks.
 specification; larger amounts are not necessary, nor wise, during the
 bootstrap phase of the network.
 
+The node _responsible_ for paying the Bitcoin fee should maintain a "fee
+spike buffer" on top of its reserve to accommodate a future fee increase.
+Without this buffer, the node _responsible_ for paying the Bitcoin fee may
+reach a state where it is unable to receive any non-dust HTLC while maintaining
```
joostjager: unable to _send or receive_ ?

02-peer-protocol.md
```diff
@@ -787,9 +787,21 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-remote commitment transaction at the current `feerate_per_kw` (see "Updating
-Fees") while maintaining its channel reserve.
+  - if it is _responsible_ for paying the Bitcoin fee:
+    - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, it cannot pay the fee for either the local or remote commitment
+    transaction at the current `feerate_per_kw` while maintaining its channel
+    reserve (see [Updating Fees](#updating-fees-update_fee)).
+    - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
+    transaction, its remaining balance doesn't allow it to pay the commitment
+    transaction fee when receiving a future additional non-dust HTLC while maintaining
```
joostjager: "while maintaining" -> "and still maintain" ? (non blocking, English isn't my specialty)

02-peer-protocol.md
t-bast: See this long (but interesting) discussion around that: https://github.com/lightningnetwork/lightning-rfc/pull/740#discussion_r399197774 :)

TL;DR: we probably want to prioritize receiving here. If you still feel it's better to mention both, let me know.

02-peer-protocol.md
joostjager: Yes, that resolves the confusion

02-peer-protocol.md
t-bast: > English isn't my specialty

Me neither, so I decided to stick with the same wording as there was before :)

02-peer-protocol.md
joostjager: If you are describing the limitations of the channel state in this paragraph, there is nothing to prioritize? But I don't mind just mentioning "receive" here. The statement isn't incorrect.

t-bast: I think that what's slightly confusing with sending here is that you also need to reserve the amount you want to send. But it's quite a nit IMO, I could definitely go with `send nor receive` if others also lean towards that.

m-schmoock: I also tend to `send nor receive any non-dust HTLC`

t-bast: Alright let's go for send nor receive then.

02-peer-protocol.md
t-bast: What about now? Do you have a suggestion on how to make it more explicit @SomberNight?

t-bast:
@joostjager @halseth what do you think of the latest version?

02-peer-protocol.md
joostjager: Still want to do `send nor receive`? Because that isn't in the current version of the pr

t-bast: Good catch, I had updated it in the requirements but not here. Done in the latest commit.

PR #742, opened 2020-02-14
Remove reference to DER encoding for public keys in compressed format

real-or-random:
ECDSA signatures in Bitcoin are DER-encoded but public keys are not.

The compressed format for public keys is for example standardized in
Sections 2.3.3 and 2.3.4 of

  Standards for Efficient Cryptography, SEC 1: Elliptic Curve
  Cryptography, Certicom Research, Version 2, 2009,
  https://www.secg.org/sec1-v2.pdf

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -62,7 +62,7 @@ Most transaction outputs used here are pay-to-witness-script-hash<sup>[BIP141](h
 
 `2 <pubkey1> <pubkey2> 2 OP_CHECKMULTISIG`
 
-* Where `pubkey1` is the numerically lesser of the two DER-encoded `funding_pubkey` and where `pubkey2` is the numerically greater of the two.
+* Where `pubkey1` is the numerically lesser of the two `funding_pubkey` in compressed format, and where `pubkey2` is the numerically greater of the two.
 
 ## Commitment Transaction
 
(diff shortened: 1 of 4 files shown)


03-transactions.md
```diff
@@ -62,7 +62,7 @@ Most transaction outputs used here are pay-to-witness-script-hash<sup>[BIP141](h
 
 `2 <pubkey1> <pubkey2> 2 OP_CHECKMULTISIG`
 
-* Where `pubkey1` is the numerically lesser of the two DER-encoded `funding_pubkey` and where `pubkey2` is the numerically greater of the two.
+* Where `pubkey1` is the numerically lesser of the two `funding_pubkey` in compressed format, and where `pubkey2` is the numerically greater of the two.
```
real-or-random: Not related to this PR but how does one compare compare two byte arrays numerically?

t-bast: What we're doing is byte-per-byte (taking the pubkey as a big-endian number):

```scala
  @tailrec
  def isLessThan(a: ByteVector, b: ByteVector): Boolean = {
    if (a.isEmpty && b.isEmpty) false
    else if (a.isEmpty) true
    else if (b.isEmpty) false
    else if (a.head == b.head) isLessThan(a.tail, b.tail)
    else ((a.head & 0xff) < (b.head & 0xff))
  }
```

That follows https://github.com/bitcoin/bips/blob/master/bip-0069.mediawiki

Is that an issue?

real-or-random: The way you're doing this is not an issue. I just wanted to point out that it's not clear from the text how to convert byte arrays to numbers for comparing, i.e., interpreting the entire 33 bytes  as big-endian. Maybe this should be added.

t-bast: True, and probably worth a separate PR. I'll open one.

cfromknecht:
thanks @real-or-random!

ACK 92404c034556c74dec99416d9d17706153ebbd94

PR #747, opened 2020-02-19
[BOLT11] Feature Bits MUST be required

junderw:
requires s

s requires payment_secret bit set
which requires 9 tag

Therefore 9 tag is also required.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -191,7 +191,7 @@ A writer:
   - if `9` contains non-zero bits:
     - SHOULD use the minimum `data_length` possible.
   - otherwise:
-    - MUST omit the `9` and `s` field altogether.
+    - MUST omit the `9` field altogether.
   - MUST pad field data to a multiple of 5 bits, using 0s.
   - if a writer offers more than one of any field type, it:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -191,7 +191,7 @@ A writer:
   - if `9` contains non-zero bits:
     - SHOULD use the minimum `data_length` possible.
   - otherwise:
-    - MUST omit the `9` and `s` field altogether and follow the old BOLT11 protocol.
+    - MUST omit the `9` and `s` field altogether.
   - MUST pad field data to a multiple of 5 bits, using 0s.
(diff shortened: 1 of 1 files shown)


junderw:
var_onion_optin is a dependency of payment_secret.

So maybe we should explicitly require at least the odd bits (supported) for payment_secret and var_onion_optin in BOLT11?

11-payment-encoding.md
```diff
@@ -153,7 +153,7 @@ Currently defined tagged fields are:
 ### Requirements
 
 A writer:
-  - MUST include exactly one `p` and `s` fields.
+  - MUST include exactly one `p`, `9` (with at least the `payment_secret` support bit set), and `s` fields.
```
t-bast: I believe this would be too specific.
What's probably confusing the requirement to include an `s` field without any mention of the feature bits.
Here is what I propose:

```suggestion
  - MUST include exactly one `p` field.
  - If the `payment_secret` feature is set, MUST include exactly one `s` field.
```

t-bast: Note that the requirements around features are in the `Feature Bits` section, so we could even remove that additional requirement which ends up being duplicated.

```suggestion
  - MUST include exactly one `p` field.
```

11-payment-encoding.md
```diff
@@ -190,7 +190,7 @@ A writer:
   - if `9` contains non-zero bits:
     - SHOULD use the minimum `data_length` possible.
   - otherwise:
-    - MUST omit the `9` field altogether.
+    - MUST omit the `9` and `s` field altogether and follow the old BOLT11 protocol.
```
t-bast: I don't think it's a good idea to mention an "old BOLT11 protocol".
The spec should only reflect the latest protocol (which is backwards-compatible).
I would remove this.

junderw:
Made suggested changes.

11-payment-encoding.md
```diff
@@ -190,7 +191,7 @@ A writer:
   - if `9` contains non-zero bits:
     - SHOULD use the minimum `data_length` possible.
   - otherwise:
-    - MUST omit the `9` field altogether.
+    - MUST omit the `9` and `s` field altogether.
```
t-bast: I think you can remove that change. A requirement a bit further down already specifies that: "MUST set an s field if and only if the payment_secret feature is set."

PR #748, opened 2020-02-20
WIP: BOLT 7: Directed message support.

rustyrussell:
These use HTLC-style encoding for simple send-and-reply messaging.

@joostjager should be particularly interested in the tradeoffs here!

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -1,4 +1,4 @@
-# BOLT #7: P2P Node and Channel Discovery
+# BOLT #7: P2P Node and Channel Discovery and Directed Messages
 
 This specification describes simple node discovery, channel discovery, and channel update mechanisms that do not rely on a third-party to disseminate the information.
 
@@ -31,6 +31,7 @@ multiple `node_announcement` messages, in order to update the node information.
   * [HTLC Fees](#htlc-fees)
   * [Pruning the Network View](#pruning-the-network-view)
   * [Recommendations for Routing](#recommendations-for-routing)
+  * [Directed Messages](#directed-messages)
   * [References](#references)
 
 ## Definition of `short_channel_id`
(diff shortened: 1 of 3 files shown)


rustyrussell:
@joostjager Surprised you don't have write access to the repo.  Gave you Triage access for now, so I can assign you issues Bwahahaha!

joostjager:
I see this as an optimization of something that is already possible. Sending htlcs, extracting the message and failing them with a custom message can be done today. Especially if the receiver doesn't even know the preimage, the money can't be taken.

Actually I think we should move in the opposite direction. Instead of promoting free messaging from a hack to a real feature, we should somehow put a price tag on every htlc.

07-routing-gossip.md
```diff
@@ -1103,6 +1104,37 @@ A->D's `update_add_htlc` message would be:
...
+Directed messages allow peers to use existing connections to query for
+invoices (see [BOLT 12](12-offer-encoding.md)).  Like gossip messages,
+they are not associated with a particular local channel.  Like HTLCs,
+they use [BOLT 4](04-onion-routing.md#directed-messages) protocol for
+end-to-end encryption.
+
+Directed messages are unreliable: in particular, they are designed to
+be cheap and not to need to be committed to a database (though an
+implementation may choose to).  Each one has an optional reply, which
+is [onion encoded](04-onion-routing.md#directed-message-replies) 0just
```
t-bast:
```suggestion
is [onion encoded](04-onion-routing.md#directed-message-replies) just
```

07-routing-gossip.md
```diff
@@ -1103,6 +1104,37 @@ A->D's `update_add_htlc` message would be:
...
+like HTLC errors.
+
+## The `directed` and `directed_reply` Messages
+
+1. type: 385 (`directed`) (`option_directed_messages`)
+2. data:
+    * [`1366*byte`:`onion_routing_packet`]
+
+1. type: 386 (`directed_reply`) (`option_directed_messages`)
```
t-bast: Are we still limited to 256 bytes here?

rustyrussell: No, but then we never were (really), we just padded.  Implementations may pad here too, I guess, though I haven't specced what offers would do here.

t-bast: Yes we padded, but you couldn't exceed 256 bytes (otherwise we lose the fixed-size packet).
I meant do we want something potentially bigger than 256 bytes to allow bigger messages (e.g. for offer)?

cdecker:
While I share @joostjager's concerns about enabling spam, I do like the fact that it separates payments and HTLCs from simple communication needs. It acknowledges that there is a need for a simple way to communicate between nodes and, with appropriate featurebits signalling, gives nodes the ability to opt-out of relaying or receiving messages.

That'd address my main concern about nodes being co-opted into relaying when all they care about is sending and receiving payments.

cdecker:
Like I wrote on the mailing list it might be worth considering an onion construction that is closer to the Sphinx paper, and not as highly specialized as the routing packet.

For reference: the routing packet has had the end-to-end payload dropped, and the per-hop payloads have been beefed up to include variable payload sizes. As a side-effect of having only a header, and no body, the entirety of the packet is pre-determined by the sender, as it is fully HMACd, this not giving the recipient the ability to reply with an included reply onion header. If we were to re-introduce the body, we could pass a reply header back to the sender in the payload and we'd free the intermediate hops from having to keep state in case the recipient wants to reply. Keeping state is particularly nasty since we don't know that there will be a reply at all, so when do we timeout and clean up the state?

One issue that'd not be addressed by this scheme though is if there is an error while forwarding, but those should be much rarer given that most of our failures currently are about channel capacities, which become meaningless if we detach the messages from HTLCs.

rustyrussell:
I'm closing this in favor of an even lighter construction (one way, no memory) since we can use a Sphinx end-to-end payload and a compact onion to enable replies.  This means no error messages, but that's actually a feature if you want a really light-weight protocol.

PR #755, opened 2020-03-06
RFC: Onion messages (previously "directed messages")

rustyrussell:
These are simplified messages, in particular they contain a compact onion for replies.

This could be further enhanced for hornet, if we wanted to go that way.  I've implemented and tested this, including rendevous routing for messages and using the reply.


rustyrussell:
Succeeded by #759

PR #762, opened 2020-04-01
Bolt 41: optimized invoice encoding

t-bast:
This is a proposal for a more compact encoding than bech32 for lightning invoices.
It solves the issue of embedding potentially long invoices in tweets.


t-bast:
Closing this proposal; while emojis are trending, memes and gifs are likely to have greater cardinality. A new proposal based on using memes and gifs for invoice encoding will provide better compression and more fun (see you next year!).

Issue #646, opened 2019-07-18
BOLT-08: replace RFC 7539 with RFC 8439


ysangkok:
There are [errata for RFC 7539](https://www.rfc-editor.org/errata_search.php?rfc=7539), and RFC 8439 claims that it is just RFC 7539 with the errata applied, and a new section. If this is indeed the case, all implementations should be conforming to RFC 8439, and the reference in BOLT-02 could be changed with no issues. This would make it easier for implementers since they wouldn't be confused by the errata.

t-bast:
Fixed by https://github.com/lightningnetwork/lightning-rfc/pull/763

PR #763, opened 2020-04-01
Replace RFC7539 with RFC8439

yyforyongyu:
7539 is obsoleted by [8439](https://tools.ietf.org/html/rfc8439), as [mentioned in 7539](https://tools.ietf.org/html/rfc7539).


ysangkok:
This fixes #646

PR #764, opened 2020-04-02
Make it explicit witness must conform to MINIMALIF

ariard:
I guess everyone get it right for its onchain transactions but I found some of my test transactions failing against core so make it explicit spending transaction must conform to MINIMALIF. Currently, it's only a standard rule, not a consensus one. Not conforming transaction will be rejected by core, which can be quite annoying for the timeout case.

If you have a better way to encode this, I will modify.

More context:
https://github.com/bitcoin/bitcoin/pull/8526
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2016-August/013014.html

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -56,6 +56,8 @@ thus the two peers must agree on the canonical ordering for this case.
 
 Most transaction outputs used here are pay-to-witness-script-hash<sup>[BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#witness-program)</sup> (P2WSH) outputs: the Segwit version of P2SH. To spend such outputs, the last item on the witness stack must be the actual script that was used to generate the P2WSH output that is being spent. This last item has been omitted for brevity in the rest of this document.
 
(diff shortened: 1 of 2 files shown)


ariard:
Updated following 04/13/2020 meeting.

03-transactions.md
```diff
@@ -56,6 +56,8 @@ thus the two peers must agree on the canonical ordering for this case.
 
 Most transaction outputs used here are pay-to-witness-script-hash<sup>[BIP141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki#witness-program)</sup> (P2WSH) outputs: the Segwit version of P2SH. To spend such outputs, the last item on the witness stack must be the actual script that was used to generate the P2WSH output that is being spent. This last item has been omitted for brevity in the rest of this document.
 
+A `<>` designs an empty vector as required for compliance with MINIMALIF-standard rule.<sup>[MINIMALIF](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2016-August/013014.html)</sup>
```
cfromknecht, line 59: nit: designs -> designates?

ariard: Thanks, corrected on last version

t-bast:
Looks like we have a quorum, merging this and taking it off the agenda for today (yay!)

PR #777, opened 2020-05-17
Clarification on the occurences of TLV records with the same type

### LINKED ISSUE #776: BOLT 2: Clarifying requirements for multiple TLV record occurrences of the same type
Currenlty nothing in the BOLT-2 specification <https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#type-length-value-format> prevents from using multiple TLV records of the same type inside message TLV stream. However, it is unclear whether this is an intentional feature – or just a bug of underspecification. I propose to clarify this point. I am prepared to write a PR for this, but not sure what is the original design intention so will wait for your clarification. Also, if this is the intentional feature, it is than unclear why [Init](https://github.com/lightningnetwork/lightning-rfc/blob/master/01-messaging.md#the-init-message) message takes an array of `chain_hash` as a single TLV record.

dr-orlovsky:
Closes #776 as discussed within the issue

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -135,7 +135,8 @@ according to the message-specific format determined by `type`.
 ### Requirements
 
 The sending node:
- - MUST order `tlv_record`s in a `tlv_stream` by monotonically-increasing `type`.
+ - MUST order `tlv_record`s in a `tlv_stream` by strictly-increasing `type`,
+   hence MUST not produce more than a single TLV record with the same `type`
  - MUST minimally encode `type` and `length`.
  - When defining custom record `type` identifiers:
    - SHOULD pick random `type` identifiers to avoid collision with other
@@ -151,7 +152,8 @@ The receiving node:
    - MUST stop parsing the `tlv_stream`.
  - if a `type` or `length` is not minimally encoded:
    - MUST fail to parse the `tlv_stream`.
- - if decoded `type`s are not monotonically-increasing:
+ - if decoded `type`s are not strictly-increasing (including situations when
+   two or more occurrences of the same `type` are met):
    - MUST fail to parse the `tlv_stream`.
  - if `length` exceeds the number of bytes remaining in the message:
    - MUST fail to parse the `tlv_stream`.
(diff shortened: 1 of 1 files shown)


01-messaging.md
```diff
@@ -192,6 +194,21 @@ things, enables the following optimizations:
...
+have multiple records. The first defeats the purpose of the proposal, and the
+latter is cumbersome in practice and would further fragment the type space
+beyond the existing even/odd rule.
+
+A minor reasons for strictly-unique types is that permitting multiple of the
+same type is less efficient on the wire due to the overhead of redundant
+type-length pairs. It is also less efficient when allocating memory for the
+decoded objects, e.g. one can't allocate the exact size of a vector upfront and
+will incur more copying overhead when appending each item individually.
+
```
t-bast, line 211: I would remove these paragraphs entirely, they clutter the spec (IMHO).

dr-orlovsky: Well, I personally find this explanation (when was given it) quire useful; it helps understanding the reasons of particular design. The spec is not that large yet, so having two more paragraphs adding clarity is good. But I am fine to remove them LN community thinks they are unnecessary.

t-bast: I'll let @cfromknecht chime in, but I think the general feeling when we discussed this during a spec meeting was to remove or condense these paragraphs.

dr-orlovsky: Done + rebased

dr-orlovsky:
The paragraph, according to @t-bast critics, is removed. Also, have done a rebase onto master

01-messaging.md
```diff
@@ -135,7 +135,8 @@ according to the message-specific format determined by `type`.
 ### Requirements
 
 The sending node:
- - MUST order `tlv_record`s in a `tlv_stream` by monotonically-increasing `type`.
+ - MUST order `tlv_record`s in a `tlv_stream` by strictly-increasing `type`,
```
ysangkok, line 138: strictly-increasing is not the same as "unique", which this sentence implies with "i.e."

t-bast: Strictly increasing does imply unique, doesn't it? (the reverse isn't true)

ariard:
ACK

01-messaging.md
Roasbeef: Below we have: 
> The strict monotonicity constraint ensures that all `type`s are unique and can at most once. Fields that map to complex objects, e.g. vectors, maps, or	appear at most once

ysangkok: indeed, but "i.e." implies that they are equivalent in that, as merriam-webster says "it introduces a rewording or a clarification". An implication is not a rewording.

t-bast: Is your suggestion to change `i.e. MUST not produce` to `hence MUST not produce`?
If so, @dr-orlovsky can you apply this small change and we can merge?

ysangkok: yes, that is a great idea.

dr-orlovsky: done!

PR #778, opened 2020-05-18
Complete the Fundamental Types.

rustyrussell:
1. It's `bigsize` not `varint`: having both is confusing; we chose the name bigsize, so use it explicitly.
2. I accidentally used `u8` (a c-lightning internal alias) instead of byte in BOLT 7.
3. Put `bigsize` in our Fundamental Types section.

### DIFF
--- a/.aspell.en.pws
+++ b/.aspell.en.pws
@@ -355,7 +355,6 @@ verifier
 verifiers
 EOF
 monotonicity
-varint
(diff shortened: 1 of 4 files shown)


t-bast:
Should this be a `spelling` PR?

ariard:
ACK 132a4e4

PR #779, opened 2020-05-19
BOLT 3: fix definition of flip(B) in P.

rustyrussell:
It turns out everyone does `P[B / 8] ^= (1 << (P % 8))`, which is not what the spec says to do (it implies you would treat P as a bitstring numbered 255 to 0).

See this stackoverflow question:
	https://stackoverflow.com/questions/49928131/lightning-secret-generation-from-seed

Reported-by: Janus Troelsen @ysangkok (on Twitter)

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -527,7 +527,9 @@ generate_from_seed(seed, I):
     return P
 ```
 
-Where "flip(B)" alternates the B'th least significant bit in the value P.
+Where "flip(B)" alternates the (B mod 8) bit of the (B div 8)
+byte of the value.  So, "flip(0) in e3b0..." is "e2b0...", and
+"flip(10) in "e3b0..." is "e3b4...".
 
 The receiving node:
   - MAY store all previous per-commitment secrets.


t-bast:
Interesting, I hadn't visited that part of our code much to be honest.
This is quite different from what the spec required, did implementations communicate and coordinate to implement the same thing?

rustyrussell:
The spec says everything should be big-endian, even our bitfields (thought that is, of course, insane).   Here we treat the sha as a little-endian bitfield, which is natural and wholesome, but not in line with the rest of the BOLTs.

03-transactions.md
```diff
@@ -527,7 +527,9 @@ generate_from_seed(seed, I):
     return P
 ```
 
-Where "flip(B)" alternates the B'th least significant bit in the value P.
+Where "flip(B)" alternates the (B mod 8)'th bit of the (B div 8)'th
+byte of the value.  So, "flip(0) in e3b0..." is "e2b0...", and
+"flip(10) in "e3b0..." is "e3b4".
```
ysangkok, line 532: why does the last hex string have no "..."?

rustyrussell: Thanks, fixed!

rustyrussell:
OK, I felt it a little dangerous to add "th" to the word list, so I removed the "'th" instead.

PR #785, opened 2020-06-10
More conservative `cltv_expiry_delta` recommendations

t-bast:
Many channels use a value below 6, which is really insecure (there are more than 2k such channels on mainnet).

While less risky, there are more than 7k channels with a value below 12.

This indicates that the spec should probably make the risks a bit more clear to help guide node operators.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -744,7 +744,7 @@ almost arbitrary fees), `S` should be small during normal operation; although,
 given that block times are irregular, empty blocks still occur, fees may vary
 greatly, and the fees cannot be bumped on HTLC transactions, `S=12` should be
 considered a minimum. `S` is also the parameter that may vary the most under
-attack, so a higher value may be desirable when non negligible amounts are at
+attack, so a higher value may be desirable when non-negligible amounts are at
 risk. The grace period `G` can be low (1 or 2), as nodes are required to timeout
 or fulfill as soon as possible; but if `G` is too low it increases the risk of
 unnecessary channel closure due to networking delays.
@@ -766,7 +766,7 @@ There are four values that need be derived:
 4. the minimum `cltv_expiry` accepted for terminal payments: the
    worst case for the terminal node C is `2R+G+S` blocks (as, again, steps
    1-3 above don't apply). The default in [BOLT #11](11-payment-encoding.md) is
-   9, which is less conservative than the 18 that this calculation suggests.
+   18, which matches this calculation.
 
 #### Requirements
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -740,31 +740,33 @@ the longest possible time to redeem it on-chain:
 Thus, the worst case is `3R+2G+2S`, assuming `R` is at least 1. Note that the
 chances of three reorganizations in which the other node wins all of them is
 low for `R` of 2 or more. Since high fees are used (and HTLC spends can use
-almost arbitrary fees), `S` should be small; although, given that block times are
-irregular and empty blocks still occur, `S=2` should be considered a
-minimum. Similarly, the grace period `G` can be low (1 or 2), as nodes are
-required to timeout or fulfill as soon as possible; but if `G` is too low it increases the
-risk of unnecessary channel closure due to networking delays.
+almost arbitrary fees), `S` should be small during normal operation; although,
+given that block times are irregular, empty blocks still occur, fees may vary
+greatly, and the fees cannot be bumped on HTLC transactions, `S=12` should be
+considered a minimum. `S` is also the parameter that may vary the most under
+attack, so a higher value may be desirable when non negligible amounts are at
+risk. The grace period `G` can be low (1 or 2), as nodes are required to timeout
+or fulfill as soon as possible; but if `G` is too low it increases the risk of
+unnecessary channel closure due to networking delays.
 
 There are four values that need be derived:
 
 1. the `cltv_expiry_delta` for channels, `3R+2G+2S`: if in doubt, a
-   `cltv_expiry_delta` of 12 is reasonable (R=2, G=1, S=2).
+   `cltv_expiry_delta` of at least 34 is reasonable (R=2, G=2, S=12).
 
-2. the deadline for offered HTLCs: the deadline after which the channel has to be failed
-   and timed out on-chain. This is `G` blocks after the HTLC's
-   `cltv_expiry`: 1 block is reasonable.
+2. the deadline for offered HTLCs: the deadline after which the channel has to
+   be failed and timed out on-chain. This is `G` blocks after the HTLC's
+   `cltv_expiry`: 1 or 2 blocks is reasonable.
 
-3. the deadline for received HTLCs this node has fulfilled: the deadline after which
-the channel has to be failed and the HTLC fulfilled on-chain before its
-   `cltv_expiry`. See steps 4-7 above, which imply a deadline of `2R+G+S`
-   blocks before `cltv_expiry`: 7 blocks is reasonable.
+3. the deadline for received HTLCs this node has fulfilled: the deadline after
+   which the channel has to be failed and the HTLC fulfilled on-chain before
+   its `cltv_expiry`. See steps 4-7 above, which imply a deadline of `2R+G+S`
+   blocks before `cltv_expiry`: 18 blocks is reasonable.
 
 4. the minimum `cltv_expiry` accepted for terminal payments: the
    worst case for the terminal node C is `2R+G+S` blocks (as, again, steps
-   1-3 above don't apply). The default in
-   [BOLT #11](11-payment-encoding.md) is 9, which is slightly more
-   conservative than the 7 that this calculation suggests.
+   1-3 above don't apply). The default in [BOLT #11](11-payment-encoding.md) is
+   9, which is less conservative than the 18 that this calculation suggests.
 
 #### Requirements
 
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -170,11 +170,10 @@ A writer:
   - MAY include one `x` field.
     - if `x` is included:
       - SHOULD use the minimum `data_length` possible.
-  - MAY include one `c` field.
+  - MUST include one `c` field (`min_final_cltv_expiry`).
     - MUST set `c` to the minimum `cltv_expiry` it will accept for the last
     HTLC in the route.
-    - if `c` is included:
-      - SHOULD use the minimum `data_length` possible.
+    - SHOULD use the minimum `data_length` possible.
   - MAY include one `n` field. (Otherwise performing signature recovery is required)
     - MUST set `n` to the public key used to create the `signature`.
   - MAY include one or more `f` fields.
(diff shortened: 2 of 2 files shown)


ariard:
Concept ACK on a far more conservative delta for now.

02-peer-protocol.md
```diff
@@ -740,31 +740,33 @@ the longest possible time to redeem it on-chain:
...
-almost arbitrary fees), `S` should be small; although, given that block times are
-irregular and empty blocks still occur, `S=2` should be considered a
-minimum. Similarly, the grace period `G` can be low (1 or 2), as nodes are
-required to timeout or fulfill as soon as possible; but if `G` is too low it increases the
-risk of unnecessary channel closure due to networking delays.
+almost arbitrary fees), `S` should be small during normal operation; although,
+given that block times are irregular, empty blocks still occur, fees may vary
+greatly, and the fees cannot be bumped on HTLC transactions, `S=12` should be
+considered a minimum. `S` is also the parameter that may vary the most under
+attack, so a higher value may be desirable when non negligible amounts are at
```
cfromknecht, line 747:
```suggestion
attack, so a higher value may be desirable when non-negligible amounts are at
```

02-peer-protocol.md
```diff
@@ -740,31 +740,33 @@ the longest possible time to redeem it on-chain:
...
+   its `cltv_expiry`. See steps 4-7 above, which imply a deadline of `2R+G+S`
+   blocks before `cltv_expiry`: 18 blocks is reasonable.
 
 4. the minimum `cltv_expiry` accepted for terminal payments: the
    worst case for the terminal node C is `2R+G+S` blocks (as, again, steps
-   1-3 above don't apply). The default in
-   [BOLT #11](11-payment-encoding.md) is 9, which is slightly more
-   conservative than the 7 that this calculation suggests.
+   1-3 above don't apply). The default in [BOLT #11](11-payment-encoding.md) is
+   9, which is less conservative than the 18 that this calculation suggests.
```
cfromknecht, line 769: should BOLT 11 also be updated with this more conservative value?

t-bast: I hesitated. The fact you're mentioning it shows that it's probably a good idea, I'll update it!

t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/785/commits/791801de63b3ece6dbc20fd83095748d32b742c1

02-peer-protocol.md
t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/785/commits/791801de63b3ece6dbc20fd83095748d32b742c1

02-peer-protocol.md
t-bast: NB: that means that when an invoice is not providing a `min_final_cltv_expiry`, implementations must be updated to use `18` instead of `9`. I'm making that change to eclair right now, but lnd and c-lightning need to implement it as well. On the receiving side we'll need to accept the old value (`9`) for a while to give the network time to update. Does that sound ok? @rustyrussell @cfromknecht

cfromknecht: @t-bast I looked into this as well, I think the changes to BOLT 11 are more complicated than I initially expected. As you mentioned, raising the default value when `min_final_cltv_delta` is not present creates incompatibilities when the sender is assuming a lower default than the receiver. It is a bit tricky since there are several places where it is ambiguous which default to return (sender vs receiver), e.g. when decoding an arbitrary invoice.

In hind sight, we probably shouldn't have used these implicit defaults because it makes it very difficult to upgrade. For a smooth upgrade, perhaps there is a preliminary step where all implementations change to always set the `min_final_cltv_delta` explicitly, and at a later point, fail to parse invoices that omit it. Increasing the default in BOLT 11 could probably be done in tandem with the last stage I think. The same should precautions should probably be taken for `expiry` while we're at it.

cfromknecht: @t-bast actually I think I see what you're saying, we can all change to _set_ 18 on new invoices but still assume 9 when not present?

t-bast: I think the reason we put a default value in the spec and omitted it from invoices was to make invoices fit in a tweet or something like that, but indeed it makes upgrades more painful.

The change I've done in eclair is to start using 18 when *paying* an invoice that doesn't specify `min_final_cltv_delta`.
When we *receive* a payment for an invoice for which we didn't specify a `min_final_cltv_delta`, we still accept 9.

If we all release new versions that do that, we can probably completely phase out the old default value of 9 in 6 months or something, and that will prompt outdated software to update to be able to pay invoices as they start being rejected by up-to-date nodes. How does that sound? I still think it's helpful to update the spec so that new-comers are aware of these safe(r) defaults.

t-bast:
Following yesterday's spec meeting discussion, I updated Bolt 11 requirements in https://github.com/lightningnetwork/lightning-rfc/pull/785/commits/eb946f07bf72ccef431410f99b33007edfa44df9.

Writers now MUST include `min_final_cltv_expiry`: this is more future-proof. If not included (for backwards-compatibility) the default value that should be used is now 18.

11-payment-encoding.md
```diff
@@ -140,7 +140,7 @@ Currently defined tagged fields are:
 * `n` (19): `data_length` 53. 33-byte public key of the payee node
 * `h` (23): `data_length` 52. 256-bit description of purpose of payment (SHA256). This is used to commit to an associated description that is over 639 bytes, but the transport mechanism for the description in that case is transport specific and not defined here.
 * `x` (6): `data_length` variable. `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
-* `c` (24): `data_length` variable. `min_final_cltv_expiry` to use for the last HTLC in the route. Default is 9 if not specified.
+* `c` (24): `data_length` variable. `min_final_cltv_expiry` to use for the last HTLC in the route. Default is 18 if not specified.
```
cfromknecht, line 143: doesn't this need to be conditioned on the context of sender/receiver? i.e., the default is 9 when receiving but 18 when sending? it is also worth asking whether making this distinction is necessary, or just changing to always set explicit values is good enough until we remove the behavior that assumes a CLTV when none is present

02-peer-protocol.md
cfromknecht: it turns out LND has been setting an explicit final CLTV value for several releases, with a default value of 40. i have made the change which now clamps user-chosen final CLTV values to be at least 18 when creating new invoices to be in line with these recommendations.

we still retain the behavior of using 9 if no CLTV is present for backwards compatibility on both sender and receiver side, but it sounds like this is compatible with your changes 👍

11-payment-encoding.md
t-bast: > doesn't this need to be conditioned on the context of sender/receiver?

This is what's done in https://github.com/lightningnetwork/lightning-rfc/pull/785/commits/eb946f07bf72ccef431410f99b33007edfa44df9: the receiver now MUST explicitly specify this value, and the sender should use 18 if it's not specified (backwards-compatibility)

02-peer-protocol.md
t-bast: > it turns out LND has been setting an explicit final CLTV value for several releases

Great!

> we still retain the behavior of using 9 if no CLTV is present for backwards compatibility on both sender and receiver side, but it sounds like this is compatible with your changes :+1:

Awesome!

cdecker:
The change in the suggestions is good, the change in the default if not specified is not. But making the field mandatory further down has the same effect, so keep the default as is to avoid obsoleting existing implementations.

02-peer-protocol.md
```diff
@@ -740,31 +740,33 @@ the longest possible time to redeem it on-chain:
...
-almost arbitrary fees), `S` should be small; although, given that block times are
-irregular and empty blocks still occur, `S=2` should be considered a
-minimum. Similarly, the grace period `G` can be low (1 or 2), as nodes are
-required to timeout or fulfill as soon as possible; but if `G` is too low it increases the
-risk of unnecessary channel closure due to networking delays.
+almost arbitrary fees), `S` should be small during normal operation; although,
+given that block times are irregular, empty blocks still occur, fees may vary
+greatly, and the fees cannot be bumped on HTLC transactions, `S=12` should be
+considered a minimum. `S` is also the parameter that may vary the most under
+attack, so a higher value may be desirable when non-negligible amounts are at
```
ariard, line 747: Note, I think you can keep this line but overall if you're a meaningful routing node an attacker can exploit by opening either 1 big channel or 10 small ones, even batching opening fees. Really most of attacks are "batchable" :/

02-peer-protocol.md
```diff
@@ -740,31 +740,33 @@ the longest possible time to redeem it on-chain:
...
+   its `cltv_expiry`. See steps 4-7 above, which imply a deadline of `2R+G+S`
+   blocks before `cltv_expiry`: 18 blocks is reasonable.
 
 4. the minimum `cltv_expiry` accepted for terminal payments: the
    worst case for the terminal node C is `2R+G+S` blocks (as, again, steps
-   1-3 above don't apply). The default in
-   [BOLT #11](11-payment-encoding.md) is 9, which is slightly more
-   conservative than the 7 that this calculation suggests.
+   1-3 above don't apply). The default in [BOLT #11](11-payment-encoding.md) is
+   18, which matches this calculation.
```
ariard, line 769: Should we subtract one block from deadline for received HTLCs compare to minimum `cltv_expiry` ? You might evaluate both values with different logics/threads and in-between you may process a block thus breaking the channel without letting a bit of "block"-time for a happy resolution ?

02-peer-protocol.md
t-bast: Of course, but that doesn't make these recommendations obsolete, on the contrary! It will need to be coupled with other techniques (e.g. limiting your exposure in the number of channels with nodes you don't trust) but in any case, you'll want high enough deltas.

02-peer-protocol.md
t-bast: I think this is more an implementation concern than a spec concern, and all implementations naturally did it the other way around (sending a higher value than what the spec recommends to accommodate for such things).

rustyrussell:
Approved at http://www.erisian.com.au/meetbot/lightning-dev/2020/lightning-dev.2020-08-17-20.08.html

11-payment-encoding.md
```diff
@@ -212,6 +211,8 @@ A reader:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
   - if there is a valid `s` field:
     - MUST use that as [`payment_secret`](04-onion-routing.md#tlv_payload-payload-format)
+  - if the `c` field (`min_final_cltv_expiry`) is not provided:
+    - MUST use an expiry delta of at least 18 when making the payment
```
OrfeasLitos, line 215: The writer now MUST include a `min_final_cltv_expiry` `c` field ([L173](https://github.com/lightningnetwork/lightning-rfc/pull/785/files#diff-16621ac6a270af29f36f0a4f900955fcR173)), therefore the reader should reject if `c` is not provided. 

On the one hand, this suggestion makes the spec more consistent. On the other, it may be a non-backwards compatible change. But if the two parties use different defaults payments will fail to go through anyway (if I understand correctly).

t-bast: That's exactly why we kept that fallback to 18 when missing, we don't want to make a backwards-incompatible change here.

PR #787, opened 2020-06-15
BOLT 8: add missing MAC check in Act Three

bitonic-cjp:
Presumably, this `decryptWithAD` also has a MAC check that can potentially fail, like the other ones. Unlike the other ones, this one does not describe what MUST happen if the MAC check fails.

Maybe it's just a nitpick. Terminating the connection without any further messages seems sensible to me in this case, but I'm not 100% sure if it's necessary, or even if it could be harmful. So don't take my word for it that this is the right behavior: think about it.

### DIFF
--- a/08-transport.md
+++ b/08-transport.md
@@ -382,6 +382,8 @@ construction, and 16 bytes for a final authenticating tag.
 4. `rs = decryptWithAD(temp_k2, 1, h, c)`
      * At this point, the responder has recovered the static public key of the
        initiator.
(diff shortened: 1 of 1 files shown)


renepickhardt:
in the beginning of the text the spec gives a general statement:
  
> If a MAC check ever fails during the handshake process, then the connection is to be immediately terminated

Thus this seems very logical to me.

t-bast:
Feels to me as well that terminating the connection is the sane choice here.

PR #793, opened 2020-08-11
Add section about private `node_announcement`.

joemphilips:
Imagine a following situation.

We are running lightning channel provider service.
When lots of users connect to our service with a private channel. and our node disconnects from the network for some reason, we want to send `channel_reestablish` from our side. Otherwise force-closing all channels will cause lots of on-chain txs.

But for the reconnection, we must know about their public addresses. Their port might be mapped to a different number for inbound and outbound connection so we need a mechanism to tell private nodes to tell their address for inbound connection.

Thus I added section about `node_announcement` for private node.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -359,6 +359,16 @@ to be ordered in ascending order, unknown ones can be safely ignored.
 Additional fields beyond `addresses` may also be added in the future—with
 optional padding within `addresses`, if they require certain alignment.
 
+### `node_annoucement` for private channels
+
+while other routing messages (`announcement_signatures`, `channel_announcement` and `channel_update` ) are
+for announcing public channels, `node_announcement` can be used for communication between private channels.
+
+This is useful when a peer wants to accept incoming connection only for a reconnection from the same peer.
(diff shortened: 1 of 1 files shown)


t-bast:
Why do you want to be the one reconnecting?
If unannounced nodes connect to you, it's their responsibility to try reconnecting to you regularly.

Most of the time these unannounced nodes won't have a stable address anyway (wallets on mobile phones) so there's no point in *you* trying to reconnect to *them*.

Adding a `node_announcement` here is dangerous, because it needs to be signed by the unannounced node to be valid. Once it's created, there's no guarantee for the unannounced node that you won't broadcast it the whole network, thus revealing them. As an unannounced node, I would not want to give you a `node_announcement`.

joemphilips:
Ah sorry, after thinking carefully, there is no reason for me to establish the connection from the server side.
I first thought that since we cannot expect users to reconnect from their side, there is a chance for DoS (or worse steal our funds). But I realized this is not a problem of a private channel. they can do it with a public channel too. 
Nevermind, sorry for the confusion.

PR #796, opened 2020-08-20
Clarify about byte representaiton of `chain_hash`

joemphilips:
* ref: https://github.com/lightningnetwork/lightning-rfc/issues/795

PR #797, opened 2020-08-21
Clarify endian notation about `chain_hash`

joemphilips:
* ref: https://github.com/lightningnetwork/lightning-rfc/issues/795

previous attempt #796


dr-orlovsky:
There is a difference about human-readable hex representation and actual binary encoding used in networking messages. Block hash is only displayed in reverse order, but not binary encoded; so we need to clarify what type of order we are talking about for `chain_hash` (binary encoding or hexadecimal representation).

PR #803, opened 2020-10-08
Require to claim revoked local output in its own penalty tx post-anchor

ariard:
If `option_anchor_outputs` applies, the cheating node can pin spends of its
HTLC-timeout/HTLC-success outputs thanks to SIGHASH_SINGLE malleability.
Using a single penalty transaction for all revoked outputs is thus unsafe as it
could be blocked to propagate long enough for the `_local node's main output_ 's
relative timelock to expire and the cheating party escaping the penalty on this
output.

I think we may just make this an unconditional requirement, indifferently of anchors.

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -513,7 +513,15 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - MAY use a single transaction to *resolve* all the outputs.
+  - if `option_anchor_outputs` applies:
+    - MAY use a single transaction to *resolve* all the outputs.
+    - if confirmation doesn't happen before reaching `security_delay` blocks from
+  expiry:
+      - SHOULD *resolve* revoked outputs in their own, separate penalty transactions. A previous
+  penalty transaction claiming multiple revoked outputs at once may be blocked from confirming
+  because of a transaction pinning attack.
+  - otherwise:
+    - MAY use a single transaction to *resolve* all the outputs.
   - MUST handle its transactions being invalidated by HTLC transactions.
 
 ## Rationale
@@ -522,10 +530,19 @@ A single transaction that resolves all the outputs will be under the
 standard size limit because of the 483 HTLC-per-party limit (see
 [BOLT #2](02-peer-protocol.md#the-open_channel-message)).
 
-Note: if a single transaction is used, it may be invalidated if the remote node
-refuses to broadcast the HTLC-timeout and HTLC-success transactions in a timely
-manner. Although, the requirement of persistence until all outputs are
-irrevocably resolved, should still protect against this happening. [ FIXME: May have to divide and conquer here, since the remote node may be able to delay the local node long enough to avoid a successful penalty spend? ]
+Note: if `option_anchor_outputs` applies, the cheating node can pin spends of its
+HTLC-timeout/HTLC-success outputs thanks to SIGHASH_SINGLE malleability.
+Using a single penalty transaction for all revoked outputs is thus unsafe as it
+could be blocked to propagate long enough for the _local node's `to_local` output_ 's
+relative locktime to expire and the cheating party escaping the penalty on this
+output. Though this situation doesn't prevent faithful punishment of the second-level
+revoked output if the pinning transaction confirms.
+
+The `security_delay` is a fixed-point relative to the absolute expiration of
+the revoked output at which the punishing node must broadcast a single-spend
+transaction for the revoked output and actively fee-bump it until its confirmation.
+The exact value of `security_delay` is left as a matter of node policy, though we
+recommend 18 blocks (similar to incoming HTLC deadline).
 
 ## Penalty Transactions Weight Calculation
 


halseth:
I think we can allow spending them all in a single tx, but perhaps recommend something like "if confirmation doesn't happen within X blocks from expiry, to_local should be swept separately as it might be pinned etc".

t-bast:
I agree with @halseth, it's worth mentioning this potential attack (good find!) but we should be more permissive with implementers.

05-onchain.md
```diff
@@ -522,10 +525,12 @@ A single transaction that resolves all the outputs will be under the
...
 [BOLT #2](02-peer-protocol.md#the-open_channel-message)).
 
-Note: if a single transaction is used, it may be invalidated if the remote node
-refuses to broadcast the HTLC-timeout and HTLC-success transactions in a timely
-manner. Although, the requirement of persistence until all outputs are
-irrevocably resolved, should still protect against this happening. [ FIXME: May have to divide and conquer here, since the remote node may be able to delay the local node long enough to avoid a successful penalty spend? ]
+Note: if `option_anchor_outputs` applies, the cheating node can pin spends of its
+HTLC-timeout/HTLC-success outputs thanks to SIGHASH_SINGLE malleability.
+Using a single penalty transaction for all revoked outputs is thus unsafe as it
+could be blocked to propagate long enough for the `_local node's main output_ 's
```
Roasbeef, line 531: IMO it's worth nothing more of the subtly here. If they're pinning, then in order to even get into the mempool, they're likely paying a relatively high fee, which means that if they _do_ confirm, then there's the CSV period on the second level. Also the defender can mix things up themselves, by only sweeping _one_ of the HTLC outputs, which would then invalidate the entire "tree" (possibly needing to pay high enough fees?) of the attacker.

05-onchain.md
```diff
@@ -513,7 +513,10 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - MAY use a single transaction to *resolve* all the outputs.
+  - if `option_anchor_outputs` applies:
+    - MUST *resolve* the _local node's main output_ in its own penalty transaction
```
Roasbeef, line 517: Perhaps this should be "should" instead? As there're other ways to defend against this, for example: sweep every output in its own transaction.

ariard: Reworded, it's used now a `MAY use a single transaction to *resolve* all the outputs` which by the negative allow to sweep every output in its own transaction as _first_ broadcast to mitigate. I kept the MUST if the confirmation isn't reached near to CSV expiration as suggested by Johan. Thus you're free to aggregate as you like until some point to save on fees but after it's a matter of security and should you care to confirm.

05-onchain.md
ariard: As usual they might pay a high fee but low-feerate, which isn't a real deterrence. Updated with the reference to CSV period. You can sweep one of the HTLC output thus invalidating the batched pinning of the attacker, but the attacker can pinning each commitment HTLC output at first broadcast as you have full-malleability for each of them. Generally I don't think we should successively "react" compared on guessed attacker moves as our transaction broadcast algorithm is public but not attacker one.

ariard:
Updated following Johan suggestion, making the new requirement conditional on reaching a new `security_delay`, i.e "if confirmation doesn't happen before reaching X blocks from expiry". I poured a recommendation for its default value but defining a variable sounds better as its exact value can be a matter of implementation policy (and the confidence if your fee-bumping logic).

@Roasbeef I didn't introduce a note on a scorched earth approach as we discussed on IRC. I'm still fearful of this being used as a double-edge by an attacker to burn more of the victim _actual_ balance than it would loss by the punishment of its channel reserve.

05-onchain.md
```diff
@@ -513,7 +513,14 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - MAY use a single transaction to *resolve* all the outputs.
+  - if `option_anchor_outputs` applies:
+  - otherwise:
```
t-bast, line 517: Looks like you're missing something there...?

05-onchain.md
```diff
@@ -513,7 +513,14 @@ A local node:
...
+    - MAY use a single transaction to *resolve* all the outputs.
+    - if confirmation doesn't happen before reaching `security_delay` blocks from
+  expiry:
+      - MUST *resolve* the _local node's main output_ in its own penalty transaction
+  as if a single resolving-transaction has been previously used its propagation might be
+  under pinning.
```
t-bast, line 523: nit:

```suggestion
      - MUST *resolve* the _local node's main output_ in its own, separate penalty transaction.
      A previous penalty transaction claiming multiple revoked outputs at once may be blocked
      from confirming because of a transaction pinning attack.
```

05-onchain.md
```diff
@@ -522,10 +529,19 @@ A single transaction that resolves all the outputs will be under the
...
 [BOLT #2](02-peer-protocol.md#the-open_channel-message)).
 
-Note: if a single transaction is used, it may be invalidated if the remote node
-refuses to broadcast the HTLC-timeout and HTLC-success transactions in a timely
-manner. Although, the requirement of persistence until all outputs are
-irrevocably resolved, should still protect against this happening. [ FIXME: May have to divide and conquer here, since the remote node may be able to delay the local node long enough to avoid a successful penalty spend? ]
+Note: if `option_anchor_outputs` applies, the cheating node can pin spends of its
+HTLC-timeout/HTLC-success outputs thanks to SIGHASH_SINGLE malleability.
+Using a single penalty transaction for all revoked outputs is thus unsafe as it
+could be blocked to propagate long enough for the `_local node's main output_ 's
```
t-bast, line 535:
```suggestion
could be blocked to propagate long enough for the `_local node's main output_`s
```

rustyrussell:
OK, lots of good discussion at meeting about this.  The spec says MAY, because if you do make one bit tx, you need to split (as HTLC txs can race you), but you also need to split because your big tx might be pinned.

It's worth noting this, for sure, but I think we can be simpler; also why pick on the to-local output?  In theory any output could be largest, and you want that one.

ariard:
>  also why pick on the to-local output? In theory any output could be largest, and you want that one.

If "you want" designates the cheating party, it's right that you can pin any largest HTLC outputs but those ones must be spent by a 2nd-level HTLC txn sooner or latter. Thus for a profitable attack to makes sense you will target the `to_local` output. That said, a cheating party may take hostage HTLC outputs, according to mempool conditions so better to split-off there too. Extended wording to them.

05-onchain.md
```diff
@@ -513,7 +513,14 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - MAY use a single transaction to *resolve* all the outputs.
+  - if `option_anchor_outputs` applies:
+    - MAY use a single transaction to *resolve* all the outputs.
+    - if confirmation doesn't happen before reaching `security_delay` blocks from
+  expiry:
+      - MUST *resolve* the _local node's `to_local` output_ in its own, separate penalty transaction   A previous penalty transaction claiming multiple revoked outputs at once may be blocked from
```
t-bast, line 520: nit:

```suggestion
      - MUST *resolve* the _local node's `to_local` output_ in its own, separate penalty transaction. A previous penalty transaction claiming multiple revoked outputs at once may be blocked from
```

t-bast (commented):
LGTM

rustyrussell (approved):

ariard:
Updated with nit fix.

05-onchain.md
```diff
@@ -513,7 +513,15 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - MAY use a single transaction to *resolve* all the outputs.
+  - if `option_anchor_outputs` applies:
+    - MAY use a single transaction to *resolve* all the outputs.
+    - if confirmation doesn't happen before reaching `security_delay` blocks from
+  expiry:
+      - MUST *resolve* revoked outputs in their own, separate penalty transactions. A previous
```
halseth, line 520: Feels like `SHOULD` is more correct here, since the sweeping node is aware of the risk and should be allowed to do as it pleases.

halseth (approved):
ACK

t-bast (approved):

05-onchain.md
t-bast: I agree, @ariard I think this is the last pending point before we merge, what do you think?

niftynei: We've got the ACKs on this, should we wait til this is addressed? paging @ariard

ariard: Sounds good to me, it's up to the implementer to gauge risk-vs-fee-saving.

ariard:
Sorry for delay, updated at aeb19de!

t-bast (approved):

t-bast:
Merging as agreed during the last spec meeting.

PR #810, opened 2020-10-23
Clarify relative order of some messages after reestablish

### LINKED ISSUE #794: commit_sig, revoke_and_ack ordering in channel reestablish flow
If we have the following state transitions:

```
A            B
<---add-----
----add---->
<---sig-----
----rev----x
----sig----x
```
In this scenario, Bob's node dies and the revocation and signature don't reach him.
Since Alice just revoked her last commitment, the next sig also covers the add that Bob sent, in addition to the one Alice sent. So the commit_sig has two htlc signatures.

If, on restart, Alice sends the wrong order:
```
----add-->
----sig-->
----rev-->
```
then Bob will interpret the commit_sig that Alice sends as only having one sig, when it has two. Bob should then force close the channel.

Thus, the ordering of these two messages matters on startup. We can't hardcode a specific ordering as the "incorrect" ordering could always happen. This came up because in lnd, we always send the revocation first.  This should be made explicit in the spec, but thought I'd open an issue up first.

t-bast:
The existing requirements were not specifying the case where both a `commitment_signed` and `revoke_and_ack` need to be retransmitted.

This is an important case to specify because if the relative order is not preserved, the channel will close.
See for example the following two scenarios.

## Rev then Sig

```
A            B
 <---add-----
 ----add---->
 <---sig-----
 ----rev----x
 ----sig----x
```

A needs to retransmit:

```
----add-->
----rev-->
----sig-->
```

With some leeway on when she sends the `add`: it can be either before or after `rev`, she can choose to not send it at all, or update the fees. The important point is that `sig` must occur **after** `rev`.

## Sig then Rev

```
A            B
 <---add-----
 ----add---->
 ----sig----x
 <---sig-----
 ----rev----x
```

A needs to retransmit:

```
----add-->
----sig-->
----rev-->
```

With some leeway on the `add`, she may choose to omit it, update the fees, etc (but all of these before the `sig`). The important point is that `sig` must occur **before** `rev`.

Fixes #794

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1256,6 +1256,10 @@ A node:
   the last `revoke_and_ack` the receiving node sent, AND the receiving node
   hasn't already received a `closing_signed`:
     - MUST re-send the `revoke_and_ack`.
+    - if it has previously sent a `commitment_signed` that needs to be
+    retransmitted:
+      - MUST retransmit `revoke_and_ack` and `commitment_signed` in the same
+      relative order as initially transmitted.
   - otherwise:
     - if `next_revocation_number` is not equal to 1 greater than the
     commitment number of the last `revoke_and_ack` the receiving node has sent:
@@ -1325,7 +1329,9 @@ involve different fees, or even be missing HTLCs which are now too old
 to be added. Requiring they be identical would effectively mean a
 write to disk by the sender upon each transmission, whereas the scheme
 here encourages a single persistent write to disk for each
-`commitment_signed` sent or received.
+`commitment_signed` sent or received. But if you need to retransmit both a
+`commitment_signed` and a `revoke_and_ack`, the relative order of these two
+must be preserved, otherwise it will lead to a channel closure.
 
 A re-transmittal of `revoke_and_ack` should never be asked for after a
 `closing_signed` has been received, since that would imply a shutdown has been


Crypt-iQ:
ACK fc07aabd93ff7a6225ad61b1ee897174ea4f17ed 
I don't know the language format of the spec too well, but this would fix the issue.

Roasbeef:
Interestingly, I considered this case when first implementing our retransmission ordering and arbitrarily chose an ordering: https://github.com/lightningnetwork/lnd/blob/master/lnwallet/channel.go#L3810

Roasbeef:
However the ordering may not be the same with how we implement it atm, since we always do revoke->sig if we need to send _both_.

02-peer-protocol.md
```diff
@@ -1256,6 +1256,10 @@ A node:
   the last `revoke_and_ack` the receiving node sent, AND the receiving node
   hasn't already received a `closing_signed`:
     - MUST re-send the `revoke_and_ack`.
+    - if it has previously sent a `commitment_signed` that needs to be
+    retransmitted:
+      - MUST retransmit `revoke_and_ack` and `commitment_signed` in the same
```
Roasbeef, line 1261: Perhaps we should also spell out the _recommended_ persistence requirements here as well? This was something `lnd` did incorrectly in a few places (not storing what we need to in order to properly reconstruct a commitment to validate a retransmission. However np with also leaving this off for a more fundamental path to make things more explicit generally.

Roasbeef: Ah re-reading again, there _is_ a light requirement re naive persistence down below where the diff touches the spec.

Roasbeef (commented):
Nice work! I thought initially that we need to prescribe a _global_ ordering, but reading this I realize that the main requirement is that the sender itself be consistent in their retransmission ordering.

02-peer-protocol.md
t-bast: Yes, I didn't update that paragraph about persistence that rusty wrote a while ago, but I could, let me know what changes you think would make sense there (if any)!

rustyrussell (approved):

Roasbeef:
One question that comes to mind: do we need to care about backwards compatibility for this at all? I'm leaning towards no, as this is the "right" way, although switching to this may create a period of instability (spurious force closes) in the network).

MrManPew:
Fixes https://github.com/lightningnetwork/lnd/issues/4613 ?

t-bast:
@Roasbeef do you think I need to add some changes, or this the last state of the PR ok to you?

TheBlueMatt (approved):

PR #812, opened 2020-11-01
Remove broken structured.py

ysangkok:
This script was previously used in Electrum but no more.

So since this script doesn't even parse the output of
`extract-formats.py` any more, just remove it.


t-bast (commented):
ACK, @cdecker do you know if c-lightning is using this script at all?

cdecker:
> ACK, @cdecker do you know if c-lightning is using this script at all?

No, we use the `extract-formats.py` script, which this script wraps with JSON. So this is good to go from our side :+1:

Roasbeef (approved):
LGTM 🍭

PR #824, opened 2020-12-08
feature: define option_zero_htlc_tx_fee (feature 22/23)

halseth:
Extension to the anchor output format that hinders the fee siphoning attack described by @ariard in https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-September/002796.html

The fix itself, making the HTLC second stage transactions zero fee, was proposed by @ariard in the same thread.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -378,10 +378,14 @@ This message introduces the `channel_id` to identify the channel. It's derived f
 #### Requirements
 
 Both peers:
-  - if `option_static_remotekey` or `option_anchor_outputs` was negotiated:
-    - `option_static_remotekey` or `option_anchor_outputs` applies to all commitment transactions
+  - if `option_static_remotekey`, `option_anchor_outputs` or
+    `option_anchors_zero_fee_htlc_tx` was negotiated:
+    - `option_static_remotekey`, `option_anchor_outputs` or
+      `option_anchors_zero_fee_htlc_tx` applies to all commitment transactions
   - otherwise:
-    - `option_static_remotekey` or `option_anchor_outputs` does not apply to any commitment transactions
+    - `option_static_remotekey`, `option_anchor_outputs` or
+      `option_anchors_zero_fee_htlc_tx` does not apply to any commitment
+      transactions
 
 The sender MUST set:
   - `channel_id` by exclusive-OR of the `funding_txid` and the `funding_output_index` from the `funding_created` message.
@@ -396,10 +400,14 @@ The recipient:
 
 #### Rationale
 
-We decide on `option_static_remotekey` or `option_anchor_outputs` at this point when we first have to generate the commitment
-transaction. The feature bits that were communicated in the `init` message exchange for the current connection determine
-the channel commitment format for the total lifetime of the channel. Even if a later reconnection does not negotiate this parameter,
-this channel will continue to use `option_static_remotekey` or `option_anchor_outputs`; we don't support "downgrading".
+We decide on `option_static_remotekey`, `option_anchor_outputs` or
+`option_anchors_zero_fee_htlc_tx` at this point when we first have to generate
+the commitment transaction. The feature bits that were communicated in the
+`init` message exchange for the current connection determine the channel
+commitment format for the total lifetime of the channel. Even if a later
+reconnection does not negotiate this parameter, this channel will continue to
+use `option_static_remotekey`, `option_anchor_outputs` or
+`option_anchors_zero_fee_htlc_tx`; we don't support "downgrading".
 
 ### The `funding_locked` Message
 
@@ -814,7 +822,7 @@ A sending node:
     transaction, it cannot pay the fee for either the local or remote commitment
     transaction at the current `feerate_per_kw` while maintaining its channel
     reserve (see [Updating Fees](#updating-fees-update_fee)).
-    - if `option_anchor_outputs` applies to this commitment transaction and the sending
+    - if `option_anchors` applies to this commitment transaction and the sending
     node is the funder:
       - MUST be able to additionally pay for `to_local_anchor` and 
       `to_remote_anchor` above its reserve.
@@ -1041,7 +1049,7 @@ output HTLCs are fully resolved.
 
 Note that the `htlc_signature` implicitly enforces the time-lock mechanism in the case of offered HTLCs being timed out or received HTLCs being spent. This is done to reduce fees by creating smaller scripts compared to explicitly stating time-locks on HTLC outputs.
 
-The `option_anchor_outputs` allows HTLC transactions to "bring their own fees" by attaching other inputs and outputs, hence the modified signature flags.
+The `option_anchors` allows HTLC transactions to "bring their own fees" by attaching other inputs and outputs, hence the modified signature flags.
 
 ### Completing the Transition to the Updated State: `revoke_and_ack`
 
@@ -1128,10 +1136,10 @@ A receiving node:
 #### Rationale
 
 Bitcoin fees are required for unilateral closes to be effective.
-With `option_anchor_outputs`, `feerate_per_kw` is not as critical anymore to
-guarantee confirmation as it was in the legacy commitment format, but it still
-needs to be enough to be able to enter the mempool (satisfy min relay fee and
-mempool min fee).
+With `option_anchors`, `feerate_per_kw` is not as critical anymore to guarantee
+confirmation as it was in the legacy commitment format, but it still needs to
+be enough to be able to enter the mempool (satisfy min relay fee and mempool
+min fee).
 
 For the legacy commitment format, there is no general method for the
 broadcasting node to use child-pays-for-parent to increase its effective fee.
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -409,6 +409,11 @@ reconnection does not negotiate this parameter, this channel will continue to
 use `option_static_remotekey`, `option_anchor_outputs` or
 `option_anchors_zero_fee_htlc_tx`; we don't support "downgrading".
 
+`option_anchors_zero_fee_htlc_tx` is considered superior to
+`option_anchor_outputs`, which again is considered superior to
+`option_static_remotekey`, and the superior one is is favored if more than one
+is negotiated.
+
 ### The `funding_locked` Message
 
 This message indicates that the funding transaction has reached the `minimum_depth` asked for in `accept_channel`. Once both nodes have sent this, the channel enters normal operating mode.
@@ -1047,9 +1052,13 @@ offline until after sending `commitment_signed`.  Once
 those HTLCs, and cannot fail the related incoming HTLCs until the
 output HTLCs are fully resolved.
 
-Note that the `htlc_signature` implicitly enforces the time-lock mechanism in the case of offered HTLCs being timed out or received HTLCs being spent. This is done to reduce fees by creating smaller scripts compared to explicitly stating time-locks on HTLC outputs.
+Note that the `htlc_signature` implicitly enforces the time-lock mechanism in
+the case of offered HTLCs being timed out or received HTLCs being spent. This
+is done to reduce fees by creating smaller scripts compared to explicitly
+stating time-locks on HTLC outputs.
 
-The `option_anchors` allows HTLC transactions to "bring their own fees" by attaching other inputs and outputs, hence the modified signature flags.
+The `option_anchors` allows HTLC transactions to "bring their own fees" by
+attaching other inputs and outputs, hence the modified signature flags.
 
 ### Completing the Transition to the Updated State: `revoke_and_ack`
 
@@ -1230,7 +1239,8 @@ The sending node:
   next `commitment_signed` it expects to receive.
   - MUST set `next_revocation_number` to the commitment number of the
   next `revoke_and_ack` message it expects to receive.
-  - if `option_static_remotekey` applies to the commitment transaction:
+  - if `option_static_remotekey` or `option_anchors` applies to the commitment
+    transaction:
     - MUST set `my_current_per_commitment_point` to a valid point.
   - otherwise:
     - MUST set `my_current_per_commitment_point` to its commitment point for
@@ -1277,7 +1287,8 @@ A node:
       - SHOULD fail the channel.
 
  A receiving node:
-  - if `option_static_remotekey` applies to the commitment transaction:
+  - if `option_static_remotekey` or `option_anchors` applies to the commitment
+    transaction:
     - if `next_revocation_number` is greater than expected above, AND
     `your_last_per_commitment_secret` is correct for that
     `next_revocation_number` minus 1:
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -121,7 +121,7 @@ If a revoked commitment transaction is published, the other party can spend this
 
 #### `to_remote` Output
 
-If `option_anchor_outputs` applies to the commitment transaction, the `to_remote` output is encumbered by a one block csv lock.
+If `option_anchors` applies to the commitment transaction, the `to_remote` output is encumbered by a one block csv lock.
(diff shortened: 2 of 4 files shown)


09-features.md
```diff
@@ -39,6 +39,7 @@ The Context column decodes as follows:
 | 16/17 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | `payment_secret`  | [BOLT #4][bolt04-mpp]                 |
 | 18/19 | `option_support_large_channel`   | Can create large channels                                 | IN       |                   | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
 | 20/21 | `option_anchor_outputs`          | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)         |
+| 22/23 | `option_zero_fee_second_stage`   | Anchor commitment type with zero fee second-stage HTLC txs| IN       | `option_anchor_outputs` | [BOLT #3](03-transactions.md)         |
```
halseth, line 42: We can discuss what bit to use, but to me it sounds preferable to just one-up the anchor bit.

t-bast:
You went through all the painful steps of https://github.com/lightningnetwork/lightning-rfc/pull/815 to find the right tx weight, just to multiply that by 0? ¯\_(ツ)_/¯

I'll think about this more thoroughly, thanks for sharing! What feature bit are you currently using in lnd if we want to do some testing?

03-transactions.md
```diff
@@ -417,10 +417,14 @@ Note the reference to the "base fee" for a commitment transaction in the require
 #### Requirements
 
 The fee for an HTLC-timeout transaction:
-  - MUST BE calculated to match:
+  - If `option_zero_fee_second_stage` applies:
+    1. MUST BE 0.
+  - Otherwise, MUST BE calculated to match:
```
ariard, line 422: Should you also update the Requirements in BOLT5 #Generation of HTLC transactions to highlight the enough-fee-contribution with `option_zero_fee_second_stage`. I think it should have been a SHOULD with `option_anchor_output` but now it's obviously a MUST.

Or even make a new requirement to attach a bumping input to second-stage HTLCs, let's avoid zero-fee HTLC transactions ?

09-features.md
ariard: What about adding the original vulnerability mail as a footnote, let's ease understanding of this cryptic option rational by future LN devs ?

ariard (commented):
I prefer this approach over the one proposed in your 11/23 mail :)

Advantages:
1. it doesn't make the number of HTLCs function of channel_reserve, you might have a trusted relation with your counterparty thus not caring at all about reserve but willingly to have high HTLC-throughput
2. it avoids adverserial, congestioned mempools where a commitment tx with 10sat/b doesn't propagate, and confirm, even if the proposed feerate is likely fine for now
3. less channel liquidity provisioned as likely-not-used feerate

Disadvantages:
* increase the cost of any "sad path" channel closure by encumbering HTLC txn, even aggregated, with a bumping utxo spent

For a full-BYOF future, it could be super interesting would be to model when BYOF is a loss compare to `update_fee`. You should take as parameters a channel liquidity usage rate over a block period, a routing fees rate and a % of unilateral channel closure. At some high-level of mempool congestion it will be more interesting to switch back to `update_fee` as the bumping utxo cost you more that the marginal routing fees gain from the non-feerate committed liquidity.

halseth:
> You went through all the painful steps of #815 to find the right tx weight, just to multiply that by 0? ¯_(ツ)_/¯

@t-bast Haha. Yeah, I have to do actual weight estimation now when bumping the fees instead of relying on magic constants, so had to go through to make sure it was correct 😀 

> I'll think about this more thoroughly, thanks for sharing! What feature bit are you currently using in lnd if we want to do some testing?

I have a branch here if you want to test, using feature bits 22/23: https://github.com/lightningnetwork/lnd/pull/4840

09-features.md
halseth: Or, depending on how widely anchors have been rolled out (on mainnet not a lot apart from the devs I would guess) we could just say that `option_anchor_outputs` means also this new zero-fee requirement.

halseth:
> I prefer this approach over the one proposed in your 11/23 mail :)

@ariard I prefer this also :)

Thanks for your suggestions, I will update the PR! 

The main thing to agree on at this point I think, is the feature bit. I see three avenues to take:
1. Amend the spec to just say `anchors == zero_fee_second_stage`. Since anchors haven't been widely rolled out, this should be okay. The implementations should make sure that the few, if any, channels that are already deployed will be closed or continue operate when one of the peers update (because sigs won't match).
2. Assign a new feature bit (as done in the current proposal as of https://github.com/lightningnetwork/lightning-rfc/pull/824/commits/c35e12ef3087bac4e1527fbe0fb105702bf9fb13), and start encouraging all new channels to use this one. Feature bits are cheap (?) 😀 
3. Assign an experimental feature bit.

I prefer 1 or 2, they should be similar in complexity. 1 is perhaps better for avoiding the spec to sprawl too much, but harder to guarantee that it won't have unintended consequences for existing deployments of anchors.

ariard:
From RL-side, I favor 1., our anchor implementation doesn't bump if feerate is still gauged as fast-confirm according to the fee-estimator at first-broadcast so just have to remove this check and also it would let remove some feerate negotiation code.

I don't think anchors outputs has been rolled out on mainet beyond LND so OG anchor channels aren't that much a concern but I would let other implems confirm this point?

rustyrussell:
No, you have to make this a new bit, don't roll it into anchors (though this should require anchors, ofc).  Anchors is already deployed, and this will cause massive confusion.  Plus if there are bugs it's easier to disable a single feature.

Secondly, please call it "option_zero_htlc_tx_fee" since the term "htlc transaction" is used in the spec.

03-transactions.md
halseth: Added!

09-features.md
halseth: Good idea, added!

halseth:
Renamed to `option_zero_htlc_tx_fee`, also added some explanation and requirements to Bolt#5.

cfromknecht:
ACK on the behavioral changes in this PR.

My only concern is with the interaction of the new feature bit with `option_anchor_output`. IMO we should structure this not as an amendment to `option_anchor_outputs`, it should be its own completely distinct commitment format that just happens to share a lot in common with `option_anchor_outputs`. For now we'll call the new format `option_anchors_zero_fee_htlc`.

The primary distinction is that IMO `option_anchors_zero_fee_htlc` should depend directly on `option_static_remotekey` and not `option_anchor_outputs`, which means that nodes aren't forced to advertise support for the broken anchor format in order to signal support for the better one. By extension, this also gives us the flexibility to completely remove `option_anchor_outputs` from code/spec in the future.

If we don't take this approach, consider Alice and Bob advertising the following:
 - Alice: `option_static_remotekey`, `option_anchor_outputs` (reject), `option_anchors_zero_fee_htlc`
 - Bob: `option_static_remotekey`, `option_anchor_outputs` (for real)

where Alice will reject any attempt to make a broken anchor channel.

When Bob is proposing, he will ignore the optional `option_anchors_zero_fee_htlc` and attempt to make a broken anchor channel, which Alice will reject. If Alice is proposing, she _should_ try to make an `option_anchor_outputs` channel but doesn't want to. If she tries to make an `option_static_remotekey` channel, there will be a commitment disagreement since Bob is expecting `option_anchor_outputs`.

So, by forcing Alice to signal `option_anchor_outputs` as a dependency, there is no way for either to downgrade to `option_static_remotekey` which is the correct behavior based on each node's _actual_ preferences and support.

As far as it pertains to the spec, it means we would need to update any existing references of `option_anchor_outputs` to `option_anchor_*` (aka `option_anchor_outputs` or `option_anchor_zero_fee_htlc`) and only keep the divergence as it pertains to the second-level fees.

NOTE: There is a similar discussion to be had as to whether `option_anchor_zero_fee_htlc` should depend on `option_static_remotekey`, but I think that's much less of a practical concern seeing as we aren't looking to phase it out any time soon.

t-bast:
I think @cfromknecht raises a very good point here.

If we want Alice to be able to advertize support for `anchor_output_with_0_fee_htlc` (and not the current `anchor_outputs`) and another commitment format (for example the 1.0 commitment), it's simply not possible with the current state of the PR. Alice is forced to set `option_anchors_zero_fee_htlc` to mandatory which prevents her from connecting to nodes that don't support any flavor of anchor outputs.

halseth:
I think that makes sense. The lnd imlementation has been updated to _only_ set this new bit, and not the "old" anchor bit.  Will update this PR as well.

halseth:
Updated to remove the dependency on previous feature bits, and added definition of `option_anchors`: https://github.com/lightningnetwork/lightning-rfc/pull/824/files#diff-c3e160eeb369e80b50a24ceee86928786ae3b67a0687a7aea80e8cefac0145daR46, which is the `OR` of the two anchor formats.

This is used everywhere in the spec where the two formats now should be handled equally.

halseth:
If anybody wants to test, lnd has merged support for what is outlined in this PR to master.

t-bast (commented):
ACK https://github.com/lightningnetwork/lightning-rfc/pull/824/commits/e4f15794cee1fdca9c8ef4fc4e57f70b19c135e0.

We'll need some time to implement this in eclair and do proper cross-implementation tests, but it looks reasonable.

ariard:
ACK e4f1579

Will jump directly to this new commitment transaction format for RL.

02-peer-protocol.md
```diff
@@ -378,10 +378,14 @@ This message introduces the `channel_id` to identify the channel. It's derived f
 #### Requirements
 
 Both peers:
-  - if `option_static_remotekey` or `option_anchor_outputs` was negotiated:
-    - `option_static_remotekey` or `option_anchor_outputs` applies to all commitment transactions
+  - if `option_static_remotekey`, `option_anchor_outputs` or
```
ysangkok, line 381: since `option_anchors` has been added to cover `_outputs` and `_zero_fee_htlc_tx`, why is it not used here?

ysangkok (changes requested):

09-features.md
```diff
@@ -39,6 +39,11 @@ The Context column decodes as follows:
 | 16/17 | `basic_mpp`                      | Node can receive basic multi-part payments                | IN9      | `payment_secret`  | [BOLT #4][bolt04-mpp]                 |
 | 18/19 | `option_support_large_channel`   | Can create large channels                                 | IN       |                   | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
 | 20/21 | `option_anchor_outputs`          | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)         |
+| 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC txs             | IN       |                   | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-zero-fee-vuln]|
```
rustyrussell, line 42: Depends on option_static_remotekey, please.

02-peer-protocol.md
rustyrussell: Worse, you don't define what the effect is if *both* are negotiated!

```
- if `option_anchors_zero_fee_htlc_tx` was negotiated:
  - `option_anchor_outputs` is ignored.
...
Rationale:
`option_anchors_zero_fee_htlc_tx` is considered superior to `option_anchor_outputs`, so is favored if both are negotiated.
```

rustyrussell (approved):
Ack with minor feedback on details.

09-features.md
SomberNight: I guess this way deliberate because of https://github.com/lightningnetwork/lightning-rfc/pull/824#issuecomment-742926077 :

> On second thought, maybe it is a good idea to make all commitment formats dependency free...

02-peer-protocol.md
halseth: If both are negotiated, then both can apply which I don't think is a problem since the latter is a superset of features.

 I added a rationale and also specified `option_anchors` the places where `option_static_remotekey` defines behavior (since technically `option_anchors_zero_fee_htlc_tx` does no longer depend on `option_static_remotekey`): https://github.com/lightningnetwork/lightning-rfc/pull/824/commits/75a4cde905ad614307b6573ab14109a41bce5a83

09-features.md
rustyrussell: No, it's a terrible idea because it makes it look like you can negotiate a option_anchors_zero_fee_htlc_tx channel without also negotiating `option_static_remotekey` and you'll just fail if you try because we won't be supporting that.  We certainly won't be testing it (and neither will anyone else).

Roasbeef:
I realize now this never was merged in, lnd has been using this feature bit for the last major version (0.13) in the wild. Have other implementations shipped a release that uses this new bit?

t-bast:
Not yet for eclair, but it is on the short term todo list ;)

02-peer-protocol.md
```diff
@@ -396,10 +400,19 @@ The recipient:
...
+the commitment transaction. The feature bits that were communicated in the
+`init` message exchange for the current connection determine the channel
+commitment format for the total lifetime of the channel. Even if a later
+reconnection does not negotiate this parameter, this channel will continue to
+use `option_static_remotekey`, `option_anchor_outputs` or
+`option_anchors_zero_fee_htlc_tx`; we don't support "downgrading".
+
+`option_anchors_zero_fee_htlc_tx` is considered superior to
+`option_anchor_outputs`, which again is considered superior to
+`option_static_remotekey`, and the superior one is is favored if more than one
```
ariard, line 414: nit: s/is//g

05-onchain.md
```diff
@@ -587,16 +587,37 @@ Note: even if the `to_remote` output is not swept, the resulting
...
+A node which broadcasts an HTLC-success or HTLC-timeout transaction for a
+commitment transaction:
+  1. if `option_anchor_outputs` applies:
+    - SHOULD combine it with inputs contributing sufficient fee to ensure
+      timely inclusion in a block.
+    - MAY combine it with other transactions.
+  2. if `option_anchors_zero_fee_htlc_tx` applies:
+    - MUST combine it with inputs contributing sufficient fee to ensure timely
+      inclusion in a block.
+    - MAY combine it with other transactions.
```
ariard, line 615: This MAY is really unsafe, if the combined-with transaction includes an input from a third-party transaction you might have feerate manipulation (like big witness data). Or same if it pays to a third-party (pinning the first-broadcast HTLC-txn and blocking any replacement)

ariard (commented):
See the IRC point on clarifying trimmed HTLC outputs must be computed only against `dust_limit_satoshis`, HTLC-timeout/HTLC-success fees are 0 with `option_zero_htlc_tx_fee`.

09-features.md
```diff
@@ -41,6 +41,10 @@ The Context column decodes as follows:
 | 20/21 | `option_anchor_outputs`          | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)         |
 | 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC transactions   | IN       |                   | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-sighash-single-harmful]|
 
+## Definitions
+
+We define `option_anchors` as `option_anchor_outputs || option_anchors_zero_fee_htlc_tx`.
```
Roasbeef, line 46: Commenting here that we want the table row above to explicitly add the static key dependancy.

Roasbeef:
Will squash this and have it land now, then make a smol follow up PR with the additions.

t-bast:
@halseth enjoy a glass of champagne, it eventually landed! :laughing:

PR #829, opened 2020-12-28
Trampoline Routing (2021 edition)

t-bast:
This proposal allows nodes running on constrained devices to sync only a small portion of the network and leverage trampoline nodes to calculate the missing parts of the payment route while providing the same privacy as fully source-routed payments.

The main idea is to use layered onions: a normal onion contains a smaller onion for the last hop of the route, and that smaller onion contains routing information about the next trampoline hop.

This PR is split in two files. Reviewers should start with `proposals/trampoline.md`, where concepts and designs are presented at a higher level. This document lets reviewers see the big picture and how all the pieces work together. It also contains pretty detailed examples that should give reviewers some intuition about the subtle low-level details.

Then reviewers can move on to `trampoline.md` which contains an aggregate of the spec sections that will be incorporated into existing bolts. This is following the usual spec format, and is where we'll work on the nitty-gritty details.

This PR supercedes #654 based on what we learnt after 1 year running trampoline in production in [Phoenix](https://phoenix.acinq.co/) and many discussions with @ecdsa while Electrum worked on their own trampoline implementation. The important changes are:

* the trampoline onion is now variable-size: it's much more flexible and has no privacy downside since it's not observable at the network layer (which is the reason why the outer onion is constant size)
* trampoline doesn't need any new gossip mechanism and instead relies on the recipient doing a small amount of work to include trampoline hints in invoices

### DIFF
--- a/proposals/trampoline.md
+++ b/proposals/trampoline.md
@@ -1,0 +1,343 @@
+# Trampoline Onion Routing
+
+## Table of Contents
+
+* [Introduction](#introduction)
+* [Overview](#overview)
+* [Trampoline onion](#trampoline-onion)
+* [Invoice flow](#invoice-flow)
+* [Privacy](#privacy)
+* [Trampoline MPP](#trampoline-mpp)
+* [Future gossip extensions](#future-gossip-extensions)
+
+## Introduction
+
+As the network grows, more bandwidth and storage will be required to keep an
+up-to-date view of the whole network. Finding a payment path will also require
+more computing power, making it unsustainable for constrained devices.
+
+Constrained devices should only keep a view of a small part of the network and
+leverage trampoline nodes to route payments.
+
+While a thorough analysis has not been made, the current state of the proposal
+should not result in a loss of privacy; on the contrary, it gives more routing
+flexibility and is likely to improve the anonymity set.
+
+## Overview
+
+Nodes that are able to calculate (partial) routes on behalf of other nodes will
+advertise support for the `trampoline_routing` feature. This is an opportunity
+for them to earn more fees than with the default onion routing (payers choose
+to pay a fee premium in exchange for reliable path-finding-as-a-service from
+trampoline nodes).
+
+A payer selects a few trampoline nodes and builds a corresponding trampoline
+onion, very similar to the normal onion; the only difference is that the next
+nodes are identified by `node_id` instead of `short_channel_id` and may not be
+direct peers.
+
+The payer then embeds that trampoline onion in the last `hop_payload` of an
+`onion_packet` destined to the first trampoline node. Computing routes between
+trampoline nodes is then deferred to the trampoline nodes themselves.
+
+A trampoline payment will thus look like this:
+
+```text
+Trampoline route:
+
+Alice ---> Bob ---> Carol ---> Dave
+
+Complete route:
+
+        +--> Irvin --> Iris --+          +--> ... --+            +--> ... --+
+        |                     |          |          |            |          |
+        |                     |          |          |            |          |
+Alice --+                     +--> Bob --+          +--> Carol --+          +--> Dave
+
+<----------------------------------><----------------------><--------------------->
+            (normal payment)            (normal payment)        (normal payment)
+```
+
+Note that this construction still uses onions created by the payer so it doesn't
+sacrifice privacy (in most cases it may even provide a bigger anonymity set).
+
+## Trampoline onion
+
+A trampoline onion uses the same construction as normal payment onions but with
+a smaller size, to allow embedding it inside a normal payment onion.
+
+For example, Alice may want to pay Carol using trampoline. Alice inserts
+trampoline nodes between her and Carol. For this example we use a single
+trampoline (Bob) but Alice may use more than one trampoline.
+
+Alice then finds a route to the first trampoline Bob. The complete route is:
+
+```text
+        +--> Irvin --> Iris --+          +--> ... --+
+        |                     |          |          |
+        |                     |          |          |
+Alice --+                     +--> Bob --+          +--> Carol
+```
+
+Note again that Alice only needs to find a route to Bob (instead of having to
+find a complete route to Carol). She doesn't know what route will be used
+between Bob and Carol: it will be Bob's responsibility to find one that works.
+
+If Alice wants to send 50 000 msat to Carol, the corresponding `update_add_htlc`
+message sent by Alice to Irvin will look like:
+
+```text
+                        update_add_htlc
++---------------------------------------------------------------------------+
+| channel_id: 1x2x3                                                         |
+| htlc_id: 42                                                               |
+| amount_msat: 55000                                                        |
+| cltv_expiry: 1500                                                         |
+| payment_hash: 0xabcd                                                      |
+| onion_routing_packet:                                                     |
+| +-----------------------------------------------------------------------+ |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward: 54000                      | normal onion payload | | |
+| | | outgoing_cltv_value: 1450                  | for Irvin            | | |
+| | | short_channel_id: 5x3x7                    |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward: 53000                      | normal onion payload | | |
+| | | outgoing_cltv_value: 1400                  | for Iris             | | |
+| | | short_channel_id: 6x3x0                    |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| | +--------------------------------------------+----------------------+ | |
+| | | amt_to_forward: 53000                      |                      | | |
+| | | outgoing_cltv_value: 1400                  | normal onion payload | | |
+| | | payment_secret: 0x1111                     | for Bob              | | |
+| | | total_amount: 53000                        |                      | | |
+| | | trampoline_onion:                          |                      | | |
+| | | +---------------------------+------------+ |                      | | |
+| | | | amt_to_forward: 50000     | trampoline | |                      | | |
+| | | | outgoing_cltv_value: 1000 | payload    | |                      | | |
+| | | | outgoing_node_id: carol   | for Bob    | |                      | | |
+| | | +---------------------------+------------+ |                      | | |
+| | | +---------------------------+------------+ |                      | | |
+| | | | amt_to_forward: 50000     | trampoline | |                      | | |
+| | | | outgoing_cltv_value: 1000 | payload    | |                      | | |
+| | | | payment_secret: 0x6666    | for Carol  | |                      | | |
+| | | | total_amount: 50000       | (final)    | |                      | | |
+| | | +---------------------------+------------+ |                      | | |
+| | +--------------------------------------------+----------------------+ | |
+| +-----------------------------------------------------------------------+ |
++---------------------------------------------------------------------------+
+```
+
+When Bob receives the trampoline onion, he discovers that the next node is Carol.
+Bob finds a route to Carol, builds a new payment onion and includes the peeled
+trampoline onion in the final payload. This process may repeat with multiple
+trampoline hops until we reach the final recipient.
+
+Here is an example flow of `update_add_htlc` messages received at each hop:
+
+```text
+Alice ------------------> Irvin -------------------------------------> Iris -----------------------------------> Bob -----------------------------------------> John ------------------------------------> Jack ------------------------------------> Carol
+
+                      update_add_htlc                            update_add_htlc                            update_add_htlc                                update_add_htlc                            update_add_htlc                            update_add_htlc               
+               +-------------------------------+          +-------------------------------+          +-----------------------------------+          +-------------------------------+          +-------------------------------+          +-----------------------------------+
+               | channel_id: 1x2x3             |          | channel_id: 5x3x7             |          | channel_id: 6x3x0                 |          | channel_id: 2x8x3             |          | channel_id: 1x3x5             |          | channel_id: 6x5x4                 |
+               | htlc_id: 42                   |          | htlc_id: 21                   |          | htlc_id: 17                       |          | htlc_id: 7                    |          | htlc_id: 15                   |          | htlc_id: 3                        |
+               | amount_msat: 55000            |          | amount_msat: 54000            |          | amount_msat: 53000                |          | amount_msat: 51000            |          | amount_msat: 50500            |          | amount_msat: 50000                |
+               | cltv_expiry: 1500             |          | cltv_expiry: 1450             |          | cltv_expiry: 1400                 |          | cltv_expiry: 1200             |          | cltv_expiry: 1100             |          | cltv_expiry: 1000                 |
+               | payment_hash: 0xabcd          |          | payment_hash: 0xabcd          |          | payment_hash: 0xabcd              |          | payment_hash: 0xabcd          |          | payment_hash: 0xabcd          |          | payment_hash: 0xabcd              |
+               | onion_routing_packet:         |          | onion_routing_packet:         |          | onion_routing_packet:             |          | onion_routing_packet:         |          | onion_routing_packet:         |          | onion_routing_packet:             |
+               | +---------------------------+ |          | +---------------------------+ |          | +-------------------------------+ |          | +---------------------------+ |          | +---------------------------+ |          | +-------------------------------+ |
+               | | amt_to_forward: 54000     | |          | | amt_to_forward: 53000     | |          | | amt_to_forward: 53000         | |          | | amt_to_forward: 50500     | |          | | amt_to_forward: 50000     | |          | | amt_to_forward: 50000         | |
+               | | outgoing_cltv_value: 1450 | |          | | outgoing_cltv_value: 1400 | |          | | outgoing_cltv_value: 1400     | |          | | outgoing_cltv_value: 1100 | |          | | outgoing_cltv_value: 1000 | |          | | outgoing_cltv_value: 1000     | |
+               | | short_channel_id: 5x3x7   | |          | | short_channel_id: 6x3x0   | |          | | payment_secret: 0x1111        | |          | | short_channel_id: 1x3x5   | |          | | short_channel_id: 6x5x4   | |          | | payment_secret: 0x2222        | |
+               | +---------------------------+ |          | +---------------------------+ |          | | total_amount: 53000           | |          | +---------------------------+ |          | +---------------------------+ |          | | total_amount: 50000           | |
+               | |       (encrypted)         | |          | |       (encrypted)         | |          | | trampoline_onion:             | |          | |       (encrypted)         | |          | |       (encrypted)         | |          | | trampoline_onion:             | |
+               | +---------------------------+ |          | +---------------------------+ |          | | +---------------------------+ | |          | +---------------------------+ |          | +---------------------------+ |          | | +---------------------------+ | |
+               +-------------------------------+          +-------------------------------+          | | | amt_to_forward: 50000     | | |          +-------------------------------+          +-------------------------------+          | | | amt_to_forward: 50000     | | |
+                                                                                                     | | | outgoing_cltv_value: 1000 | | |                                                                                                | | | outgoing_cltv_value: 1000 | | |
+                                                                                                     | | | outgoing_node_id: carol   | | |                                                                                                | | | payment_secret: 0x6666    | | |
+                                                                                                     | | +---------------------------+ | |                                                                                                | | | total_amount: 50000       | | |
+                                                                                                     | | |       (encrypted)         | | |                                                                                                | | +---------------------------+ | |
+                                                                                                     | | +---------------------------+ | |                                                                                                | | |           EOF             | | |
+                                                                                                     | +-------------------------------+ |                                                                                                | | +---------------------------+ | |
+                                                                                                     | |             EOF               | |                                                                                                | +-------------------------------+ |
+                                                                                                     | +-------------------------------+ |                                                                                                | |             EOF               | |
+                                                                                                     +-----------------------------------+                                                                                                | +-------------------------------+ |
+                                                                                                                                                                                                                                          +-----------------------------------+
+```
+
+## Invoice flow
+
+Now the question is: how does Alice choose the trampoline nodes to reach Carol?
+
+Similarly to what is done for normal payments, we introduce trampoline hints in
+Bolt 11 invoices. Carol can run a simple breadth-first search to select a few
+trampoline nodes that will be able to relay payments to her and calculates the
+fees and cltv for the corresponding routes. If Carol is accessible via public
+channels, she may not even need to include any trampoline hint.
+
+Each trampoline hint contains only the `node_id` of the trampoline node, a
+`cltv_expiry_delta`, `fee_base_msat` and `fee_proportional_millionths`, for
+example:
+
+```json
+{
+  "trampoline_hints": [
+    {
+      "node_id": "036d6caac248af96f6afa7f904f550253a0f3ef3f5aa2fe6838a95b216691468e2",
+      "cltv_expiry_delta": 288,
+      "fee_base_msat": 5,
+      "fee_proportional_millionths": 300
+    },
+    {
+      "node_id": "025f7117a78150fe2ef97db7cfc83bd57b2e2c0d0dd25eaf467a4a1c2a45ce1486",
+      "cltv_expiry_delta": 432,
+      "fee_base_msat": 8,
+      "fee_proportional_millionths": 250
+    }
+  ]
+}
+```
+
+When Alice wants to pay the invoice, she selects a trampoline node close to her
+(let's call it T1) and a trampoline node from Carol's invoice (let's call it
+T2). If Carol doesn't include any routing hint in the invoice, T2 = Carol.
+
+Alice then simply uses the trampoline route `Alice -> T1 -> T2 -> Carol` and
+creates the corresponding trampoline onion. Alice then finds a route to T1, and
+sends the payment through that route.
+
+```text
+     Alice's neighborhood                                                        Carol's neighborhood
++----------------------------------+                                        +--------------------------------+
+|                                  |                                        |                                |
+|      +-----> T4                  |                                        |       T3 <----- N8 <-----+     |
+|      |                           |                                        |                          |     |
+|      |                           |                                        |                          |     |
+|    Alice -----> N1 -----> T1     |-------> (public network graph) ------->|    T2 -----> N7 -----> Carol   |
+|      |                           |                                        |                                |
+|      |                           |                                        +--------------------------------+
+|      +-----> N2 -----> T5        |
+|                                  |
++----------------------------------+
+```
+
+Alice and Carol both assume connectivity between their respective neighborhoods.
+Note that this assumption is also necessary with the default source-routing
+scheme, so trampoline is not adding any new restrictions.
+
+Alice is completely free to add more intermediate trampoline hops to the route
+to improve privacy (by effectively using longer routes) or to ensure a given
+node is able to witness the payment.
+
+## Privacy
+
+Trampoline routing allows constrained devices to send and receive payments
+without sacrificing privacy.
+
+Such nodes are advised to sync only a small portion of the graph (their local
+neighborhood) and to ensure connectivity to a few distinct trampoline nodes.
+
+This allows them to insert normal hops before the first trampoline node, thus
+protecting their privacy with the same guarantees than normal payments. In the
+example graph from the previous section, T1 cannot know that the payment comes
+from Alice because it only sees an HTLC coming from N1 (especially if Alice is
+using an unannounced channel to N1). When T1 receives the HTLC, it cannot know
+if the next node in the route (T2) is the final destination or not. Similarly,
+T2 cannot know that Carol is the recipient nor that the payment comes from
+Alice (T2 cannot even know that the previous trampoline node was T1).
+
+On top of these guarantees, receivers may use route-blinding or rendezvous to
+hide their exact location. Rendezvous is hard to use with fully source-routed
+payments because if a single channel disappears (or simply does not have enough
+liquidity to relay the payment), it invalidates the whole rendezvous onion.
+Trampoline rendezvous does not have this issue because trampoline routes are
+by design more flexible (only the `node_ids` are fixed, but any route between
+these `node_ids` can be used as long as they fit the fee budget allocated to
+each trampoline hop).
+
+## Trampoline MPP
+
+Trampoline routing combines nicely with multi-part payments. When multi-part
+payment is used, we can let trampoline nodes combine all the incoming partial
+payments before forwarding. Once the totality of the payment is received, the
+trampoline node can choose the most efficient way to re-split it to reach the
+next trampoline node.
+
+For example, Alice could split a payment in 3 parts to reach a trampoline node,
+which would then split it in only 2 parts to reach the destination (we use a
+single hop between trampoline nodes for simplicity here, but any number of
+intermediate nodes may be used):
+
+```text
+                                                     HTLC(1500 msat, 600112 cltv)
+                                                    +---------------------------+
+                                                    | amount_fwd: 1500 msat     |
+                                                    | expiry: 600112            |
+            HTLC(1560 msat, 600124 cltv)            | payment_secret: aaaaa     |
+             +-----------------------+              | total_amount: 2800 msat   |
+             | amount_fwd: 1500 msat |              | trampoline_onion:         |                                                                  HTLC(1100 msat, 600000 cltv)
+             | expiry: 600112        |              | +-----------------------+ |                                                                 +-----------------------------+
+        +--> | channel_id: 3         | ---> I1 ---> | | amount_fwd: 2500 msat | | --+                                                             | amount_fwd: 1100 msat       |
+        |    |-----------------------|              | | expiry: 600000        | |   |                                                             | expiry: 600000              |
+        |    |     (encrypted)       |              | | node_id: Bob          | |   |                                                             | payment_secret: xxxxx       |
+        |    +-----------------------+              | +-----------------------+ |   |                     HTLC(1150 msat, 600080 cltv)            | total_amount: 2500 msat     |
+        |                                           | |      (encrypted)      | |   |                      +-----------------------+              | trampoline_onion:           |
+        |                                           | +-----------------------+ |   |                      | amount_fwd: 1100 msat |              | +-------------------------+ |
+        |                                           +---------------------------+   |                      | expiry: 600000        |              | | amount_fwd: 2500 msat   | |
+        |                                           |             EOF           |   |                 +--> | channel_id: 561       | ---> I4 ---> | | expiry: 600000          | | --+
+        |                                           +---------------------------+   |                 |    |-----------------------|              | | total_amount: 2500 msat | |   |
+        |                                            HTLC(800 msat, 600112 cltv)    |                 |    |     (encrypted)       |              | | payment_secret: yyyyy   | |   |
+        |                                           +---------------------------+   |                 |    +-----------------------+              | +-------------------------+ |   |
+        |                                           | amount_fwd: 800 msat      |   |                 |                                           | |         EOF             | |   |
+        |                                           | expiry: 600112            |   |                 |                                           | +-------------------------+ |   |
+        |   HTLC(820 msat, 600130 cltv)             | payment_secret: aaaaa     |   |                 |                                           +-----------------------------+   |
+        |    +-----------------------+              | total_amount: 2800 msat   |   |                 |                                           |             EOF             |   |
+        |    | amount_fwd: 800 msat  |              | trampoline_onion:         |   |                 |                                           +-----------------------------+   |
+        |    | expiry: 600112        |              | +-----------------------+ |   |                 |                                                                             |
+Alice --+--> | channel_id: 5         | ---> I2 ---> | | amount_fwd: 2500 msat | | --+--> Trampoline --+                                                                             +--> Bob
+        |    |-----------------------|              | | expiry: 600000        | |   |  (fee 170 msat) |                                            HTLC(1400 msat, 600000 cltv)     |
+        |    |     (encrypted)       |              | | node_id: Bob          | |   |    (delta 32)   |                                           +-----------------------------+   |
+        |    +-----------------------+              | +-----------------------+ |   |                 |                                           | amount_fwd: 1400 msat       |   |
+        |                                           | |      (encrypted)      | |   |                 |                                           | expiry: 600000              |   |
+        |                                           | +-----------------------+ |   |                 |                                           | payment_secret: xxxxx       |   |
+        |                                           +---------------------------+   |                 |   HTLC(1480 msat, 600065 cltv)            | total_amount: 2500 msat     |   |
+        |                                           |             EOF           |   |                 |    +-----------------------+              | trampoline_onion:           |   |
+        |                                           +---------------------------+   |                 |    | amount_fwd: 1400 msat |              | +-------------------------+ |   |
+        |                                            HTLC(500 msat, 600112 cltv)    |                 |    | expiry: 600000        |              | | amount_fwd: 2500 msat   | |   |
+        |                                           +---------------------------+   |                 +--> | channel_id: 1105      | ---> I5 ---> | | expiry: 600000          | | --+
+        |                                           | amount_fwd: 500 msat      |   |                      |-----------------------|              | | total_amount: 2500 msat | |
+        |                                           | expiry: 600112            |   |                      |     (encrypted)       |              | | payment_secret: yyyyy   | |
+        |   HTLC(510 msat, 600120 cltv)             | payment_secret: aaaaa     |   |                      +-----------------------+              | +-------------------------+ |
+        |    +-----------------------+              | total_amount: 2800 msat   |   |                                                             | |         EOF             | |
+        |    | amount_fwd: 500 msat  |              | trampoline_onion:         |   |                                                             | +-------------------------+ |
+        |    | expiry: 600112        |              | +-----------------------+ |   |                                                             +-----------------------------+
+        +--> | channel_id: 7         | ---> I3 ---> | | amount_fwd: 2500 msat | | --+                                                             |             EOF             |
+             |-----------------------|              | | expiry: 600000        | |                                                                 +-----------------------------+
+             |     (encrypted)       |              | | node_id: Bob          | |
+             +-----------------------+              | +-----------------------+ |
+                                                    | |      (encrypted)      | |
+                                                    | +-----------------------+ |
+                                                    +---------------------------+
+                                                    |             EOF           |
+                                                    +---------------------------+
+```
+
+## Future gossip extensions
+
+With the current gossip mechanisms, some bandwidth will be unnecessarily wasted
+when constrained nodes sync their local neighborhood, because they have no way
+of asking for `channel_update`s within a given distance boundary nor telling
+their peers to *not* relay some updates to them.
+
+That can be easily addressed in the future with new Bolt 7 gossip queries, but
+is not mandatory to the deployment of trampoline routing.
+
+Two mechanisms in particular would be useful:
+
+* `channel_update` filters: constrained nodes can ask their peers to discard
+  updates coming from nodes that are further than N hops instead of relaying
+  them.
+* distance-based gossip queries: constrained nodes can ask their peers for all
+  updates from nodes that are at most N hops far, and combine that with the
+  current gossip query tricks to avoid replaying updates they already have.
--- a/trampoline.md
+++ b/trampoline.md
@@ -1,0 +1,185 @@
+# Trampoline Onion Routing
+
+This file contains all the low-level details about the proposal in one place to
+simplify reviewers life.
+
+Once we make progress towards standardization, we will move these sections and
+include them in the existing bolts.
+
+## Table of Contents
+
+* [Features](#features)
+* [Packet Structure](#packet-structure)
+  * [Trampoline Onion](#trampoline-onion)
+  * [Paying via trampoline nodes](#paying-via-trampoline-nodes)
+  * [Invoice trampoline hints](#invoice-trampoline-hints)
+  * [Failure messages](#failure-messages)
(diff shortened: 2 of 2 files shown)


saubyk:
Hi @t-bast can you please link to `proposals/trampoline.md`? I am not able to locate it.

t-bast:
> Hi @t-bast can you please link to proposals/trampoline.md?

https://github.com/lightningnetwork/lightning-rfc/blob/trampoline-routing-no-gossip/proposals/trampoline.md is the best way to read it (with github's markdown viewer).

ecdsa:
Regarding trampoline invoices: Do you suggest to create a new tagged field for `trampoline_hints`, or are we going to reuse the existing `r` tags, and to consider them as trampoline hints if the invoice has `trampoline_routing_opt` in its features?

fiatjaf:
@ecdsa the proposed spec has a new tagged field `t`: https://github.com/lightningnetwork/lightning-rfc/blob/trampoline-routing-no-gossip/trampoline.md#invoice-trampoline-hints

t-bast:
Yes, I think it's better to introduce a new one, tailored for trampoline.
But that is of course open to discussion if there's a better way.

cdecker:
Very nice proposal, can't wait to see this in action. The onion
construction is mostly taken care of if you implement @rustyrussell's
offers proposal (which also uses the same construction and
parametrizes the size of the onion).

A rationale that I couldn't find, but I think Acinq are using this for
is the efficiency gain if the sender is on a slow / high latency
connection, each payment attempt requires a roundtrip to the sender
(in Eclair's case a mobile phone on a potentially flaky
connection). Outsourcing the path finding and retry logic to a better
connected node can result in far lower time-to-completion for the
payment, and the sender can save some bandwidth.

I just have the following points that I am still a bit unclear on:

 - There seem to be no limits in place for the inner onion, other than
   that it's smaller. Specifically we have an issue with the `inner x
   outer` path length. If we generate an inner onion consisting of
   nodes that are diametrically opposed in the network, we can end up
   with a route that is many times larger than the implicit 20 hop
   limit we have at the moment. The diameter of the network is
   currently about 7, so an inner onion with 20 hops (using the outer
   onion limit as a proxy here), could end up with 140 effective
   routing hops, which also acts as a force-multiplier for an attacker
   wanting to hold up as much liquidity in the network as possible.
   
 - Are trampoline hints intended to replace or augment the route
   hints? In the former case we end up in a situation where the
   destination must be known to at least one trampoline, in the latter
   we may end up duplicating a lot of information in the
   invoice. Further the duplication seems to be solely due to the fact
   that route hints don't have a feature bitset.
    - Should we just bump route hints to v2 and make them a TLV so we
      can add a feature bitset? It'd potentially save us some trouble
      in the future.
    - Do we want to have trampoline hints at all? Isn't it in the
      sender's interest to not have them, and instead adding the
      existing route hints to the last trampoline's payload? That way
      we don't make the invoices larger for everyone, while still
      maintaining their functionality.

t-bast:
> There seem to be no limits in place for the inner onion, other than that it's smaller. (...) which also acts as a force-multiplier for an attacker wanting to hold up as much liquidity in the network as possible.

This is true. However, attackers have no control on what channels will be used to relay with trampoline, so this would be a blind attack against the whole network rather than targetting specific nodes (which is still concerning).
There has been some progress recently on proposals to fight spam, so we can hope that this issue will eventually be fixed.
But in the meantime, implementations can restrict the maximum size of the trampoline onion (we can add a recommended value in the spec) just like we restricted channel capacity at first.

> Are trampoline hints intended to replace or augment the route hints?

I think we want to introduce a different, more flexible invoice routing hint.
It's probably a good opportunity, as you mention, to introduce the tlv format in these routing hints v2.
I know @rustyrussell mentioned it multiple times in the past, do you have some requirements / early designs for these new routing hints?

Let's explore the advantages and drawbacks of using the current routing hints (let me know if I'm missing important points there):

* Advantages:
  * Invoices don't get bigger than they are today
* Drawbacks:
  * It takes a lot of space in the trampoline onion
  * It encourages single-trampoline payments, which is bad for privacy (trampoline node may know both sender and recipient)
  * It offers a less flexible design space for recipient anonymity schemes

It's also important to note that to receive trampoline payments, the recipient still needs to upgrade his software to support trampoline onion decryption, so it's a good opportunity to implement a new routing hint format at the same time.

The advantages and drawbacks of using new routing hints are:

* Advantages:
  * Opportunity to add missing fields (features) and add future fields while staying backwards-compatible
  * Opportunity to remove unused fields when unnecessary, resulting in smaller invoices
  * No additional data needs to be transmitted in the trampoline onion
  * Trampoline routes must use at least two trampolines, which is much better for privacy
  * No channel information is leaked in invoices, which is better for privacy
  * More flexible design space for recipient anonymity schemes
* Drawbacks:
  * Invoices get bigger if we want them to potentially be paid by "legacy" senders (but note that in some cases we could instead issue two invoices, a legacy one and a trampoline one, and that will eventually go away when everyone support the new format)

To exemplify the argument that the design space for recipient privacy would be bigger, I've been toying with the idea of short-lived tor-like circuits between mobile wallet recipients and trampoline nodes.
Imagine we have a graph like this, where Bob wants to be paid without revealing his identity to trampoline nodes and without revealing private channel information:

```text
                            public              private
                      T1 ------------> Carol -------------+
                                                          |
                                                          v
                                                         Bob
                                                          ^
                                                          |
      public                public             private    |
T2 ------------> Mallory ------------> Dave --------------+
```

Obviously if Bob uses a mobile wallet with private channels, Carol and Dave know that he is the final recipient when forwarding payments to him (frequent IP address changes, offline most of the time, etc).
But Bob doesn't want his invoices to reveal that he's connected to Carol and Dave (e.g. to hide what lightning service provider he's using).
Bob could send anonymous onion messages to T1 and T2 to create short-lived circuits (for the duration of the invoice validity) to get them to route payments in the right direction, without revealing more than the next hop.
He then only needs to specify T1 and T2 in his routing hints, information about Carol, Dave or concrete channels is not needed.

Of course, this idea is very hand-wavy for now and has a lot of holes (and maybe route blinding works better and achieves the same results), so please don't poke at it too much yet, you will find issues, but that's not the point. I'm mentioning it because it shows how more flexible routing hints can provide a wider design space for recipients, which is important IMO.

ecdsa:
The current proposal adds a single new failure message, `trampoline_fee_expiry_insufficient`. It does not specify any new failure message if a trampoline node fails to find a route to the next trampoline in a multi trampoline route. 

The current Eclair implementation seems to return `amount_below_minimum` in all cases, regardless of the failure scenario. For example, if I create a route with two trampolines, and the second trampoline does not exist, I receive that failure message. It might be useful to distinguish between "next trampoline unreachable" and "unknown next trampoline".

ecdsa:
> trampoline doesn't need any new gossip mechanism and instead relies on the recipient doing a small amount of work to include trampoline hints in invoices

The previous proposal had a `node_update` gossip message. That message is not equivalently replaced by the `trampoline_hints` in the invoice. If the sender decides to use an intermediate trampoline not in the invoice, they need to find out the fee/cltv parameters for that intermediate trampoline. 

It is true that we may rely on trial-and-error and the `trampoline_fee_expiry_insufficient` message for that. However, I think that this induced an important change in the semantics of the fee and cltv fields in `data`:
 - in the previous proposal, the `node_update` message was used to publish fee and cltv values. By definition, these values used to be a function of time, but not of the payment sent to the trampoline.
 - in the new proposal, the fee and cltv parameters might depend on the payment. The proposal says: "The fee amount or cltv value was below that required by the trampoline node to forward to the next trampoline node." This seems to imply that the fee and cltv fields returned in `data` will be what the trampoline node would require for the current payment, but this does not guarantee success for a payment to another route. 

I think the proposal should clarify whether the values returned in the error message are relative to the current payment, or independent of the payment.

t-bast:
> The current Eclair implementation seems to return amount_below_minimum in all cases, regardless of the failure scenario

If it does that, it's a bug. It should only do that for small payments to phoenix wallets that don't have enough incoming liquidity (and thus need a new channel opened on-the-fly). For other cases it should return `unknown_next_peer`. It's definitely possible that this case isn't currently handled correctly since only single-trampoline has been implemented.

I think the errors returned should be the following:

- I need more fees/cltv to relay: `trampoline_fee_expiry_insufficient`
- There's no route in the graph to the next trampoline node: `unknown_next_peer`
- There's a route in the graph to the next trampoline node, but I don't have enough balance: `temporary_node_failure`
- There's a route in the graph to the next trampoline node, but it fails downstream: depending on the downstream failure, either directly relay that failure or replace it with a `temporary_node_failure`

Of course, trampoline nodes may also use other errors such as `amount_below_minimum` (or other Bolt 4 errors) in specific cases where that makes sense. Does that sound good?

> The previous proposal had a node_update gossip message. That message is not equivalently replaced by the trampoline_hints in the invoice. If the sender decides to use an intermediate trampoline not in the invoice, they need to find out the fee/cltv parameters for that intermediate trampoline.

I decided to remove that mechanism because it wasn't working well: it's impossible to correctly estimate fees for all possible payments. It will force clients to overpay too often, or will still need a trial-and-error when you try to minimize the overpayment.

I'm now more in favour of the trial-and-error approach, where `trampoline_fee_expiry_insufficient` will return the fees/cltv that should will work *for this specific payment*. This guarantees that the retry should work if the trampoline correctly ran his path-finding algorithm, and used a small error buffer just in case (which minimizes sender-side retries). Some trampoline nodes may choose to return a value that would work for other payments as well (and maybe overpay this one), this is an implementation choice (and is a trade-off routing nodes should consciously choose to make or not). Clients may cache this value and use it for other payments as well, which may work (or may not, in which case the trial-and-error will let them correct this).

I think this approach is the most flexible one: we should leave some room for implementations to make different choices, to ensure we have a diverse network.

ecdsa:
> Yes, I think it's better to introduce a new one, tailored for trampoline. But that is of course open to discussion if there's a better way.

In practice, wallets will be compelled to include both `trampoline_hints` and `routing_hints` in their invoices, because they do not know if the sender understands trampoline. For an invoice with a reasonably sized description and a single routing hint, this results in a 10% size increase.

I guess the data in `trampoline_hints` will most of the time be redundant with what is in `routing_hints` (the payee is going to infer the trampoline fee/cltv from the routes it knows). Are there use cases where that would not be redundant?

t-bast:
> In practice, wallets will be compelled to include both trampoline_hints and routing_hints in their invoices, because they do not know if the sender understands trampoline.

Yes, that's true. And the transition period during which both will be required will likely last months (hopefully not years though as non-trampoline wallets will have an incentive to migrate to the new routing hints, even if they don't want to support trampoline).

I think these new routing hints will solve a few issues that the current routing hints have, so people would upgrade to use them regardless of trampoline, and once that's done the size overhead will be somewhat negligible.

> Are there use cases where that would not be redundant?

They will be different for cases where current `routing_hints` include more than one hop, but in practice I don't think anyone is using this (yet).

ecdsa:
> They will be different for cases where current routing_hints include more than one hop, but in practice I don't think anyone is using this (yet).

with multiple hops, why would you assume the trampoline fee/cltv to differ from the sum of hop fees/cltv?

cdecker:
It seems to me like this proposal should be split into two: one that
introduces just the trampoline construction and forwarding logic, and
a second one that allows recipients to specify trampolines for
incoming payments. This is because in my eyes the latter is a new
requirement, that deviates from the original trampoline proposal, and
I'd like to unbundle the two.

> It's also important to note that to receive trampoline payments, the
> recipient still needs to upgrade his software to support trampoline
> onion decryption, so it's a good opportunity to implement a new
> routing hint format at the same time.

This looks like a new requirement to me, since the last trampoline
could be signaled to pay the final destination in non-trampoline mode,
thus not requiring the recipient to even understand the trampoline
protocol. It's the difference between "some node must support
trampolines" and "this specific recipient needs to support
trampolines". This is particularly important given that trampolines
are first and foremost a tool for the sender, and requiring the
recipient to play nice for the sender is an extra requirement.

> I think we want to introduce a different, more flexible invoice
> routing hint.  It's probably a good opportunity, as you mention, to
> introduce the tlv format in these routing hints v2.  I know
> @rustyrussell mentioned it multiple times in the past, do you have
> some requirements / early designs for these new routing hints?

We can totally do that as well, however I think upgrading the
route-hints to be TLVs with all the mentioned improvements is
orthogonal to trampolines, hence my objection.

> Bob could send anonymous onion messages to T1 and T2 to create
> short-lived circuits (for the duration of the invoice validity) to get
> them to route payments in the right direction, without revealing more
> than the next hop.  He then only needs to specify T1 and T2 in his
> routing hints, information about Carol, Dave or concrete channels is
> not needed.

Is this not orthogonal to the original trampoline proposal? I don't
see why we need to mix it in here, especially since with route
blinding there is already a competing proposal that achieves pretty
much the same.

> If it does that, it's a bug. It should only do that for small payments
> to phoenix wallets that don't have enough incoming liquidity (and thus
> need a new channel opened on-the-fly). For other cases it should
> return `unknown_next_peer`. It's definitely possible that this case
> isn't currently handled correctly since only single-trampoline has
> been implemented.

Reusing an existing error code seems wrong to me
here. `unknown_next_peer` implicitly assumes the sender knows what the
next peer should be, which in this case it doesn't. It's not like
we're running out of error number, so why not add a new
`unknown_next_trampoline` error?

---

Overall I think two smaller proposals with a single feature being
added are likely to make progress much quicker than bundling them
up. The larger the surface the more likely someone will object (in
this case that's me 😉)

t-bast:
> It seems to me like this proposal should be split into two: one that
introduces just the trampoline construction and forwarding logic, and
a second one that allows recipients to specify trampolines for
incoming payments.

Sounds good, I'll keep the current PR open and up-to-date for people who want
to see _the whole thing_, and will open a smaller PR in spec format for the
onion construction.

> the last trampoline could be signaled to pay the final destination in non-trampoline mode,
thus not requiring the recipient to even understand the trampoline protocol

I don't see any satisfactory way of doing that...
All the solutions I tried involve giving that last trampoline invoice information
that force you to *trust* that this trampoline node will not cheat you, whereas it
could steal money (`payment_secret` for amountless invoices for example).

That's why I want to avoid having that scenario in the spec and prefer doing E2E
trampoline. For mobile wallets that can't fallback on normal payments when the
invoice doesn't have trampoline support, then a trusted solution where you give
parts of the invoice to a trampoline node may be implemented (that's what we do
in Phoenix), but I don't think it should be in the spec because it's unsatisfactory.

If you find a satisfying way of achieving that I'm interested, but I don't think
it's possible (it will sacrifice privacy and in some cases funds safety).

> upgrading the route-hints to be TLVs with all the mentioned improvements is
orthogonal to trampolines

That's a reasonable objection, I'll make a separate PR.

> Is this not orthogonal to the original trampoline proposal? I don't
see why we need to mix it in here, especially since with route
blinding there is already a competing proposal that achieves pretty
much the same.

I mentioned this simply as an example to show that it allows a larger design
space than we previously had. And because circuit-based approaches have had
more research put into them than route blinding, I would feel better having
the possibility of studying multiple approaches in case route blinding turns
out to be insufficient or not secure enough...

> It's not like we're running out of error number, so why not add a new
unknown_next_trampoline error?

Sure, we can do that, it's more explicit that way.

ecdsa:
> It's not like we're running out of error number, so why not add a new unknown_next_trampoline error?

Like I said above, it would be useful to have distinct error numbers for `unknown_next_trampoline` and `next_trampoline_unreachable` (in case the trampoline fails to find a path, or fails to forward the payment).

In the first case, Trampoline A does not know about Trampoline B, and the sender should not ask TA to forward to TB again. In the second case, the sender will know that TA failed to forward a certain amount to TB, but it may be able to forward a different amount right now, or the same amount later.

Perhaps even more granularity is desirable. If TA fails to find a path to TB, it might return the "capacity" it believes it can send to TB in the error message.

t-bast:
I'll integrate all these requirements in the trampoline onion PR, let's discuss it there.

t-bast:
> Is the plan to adapt this proposal to add some way for the sender to indicate to the trampoline node that they should retry the payment on their behalf? Maybe it already works as-is?

Great timing, we actually just merged a PR to eclair to experiment with that mechanism: https://github.com/ACINQ/eclair/pull/2435
We are using an experimental tlv for now that the sender puts in the trampoline onion for the first trampoline node.
This tlv is currently empty (it's just a signal), but we may enrich it with data if we find useful things to stuff in there.

proposals/trampoline.md
```diff
@@ -0,0 +1,515 @@
...
+| | | outgoing_cltv_value: 1400                  | normal onion payload | | |
+| | | payment_secret: 0x1111                     | for T1               | | |
+| | | total_amount: 53000                        |                      | | |
+| | | trampoline_onion:                          |                      | | |
+| | | +---------------------------+------------+ |                      | | |
+| | | | amt_to_forward: 50000     |            | |                      | | |
+| | | | outgoing_cltv_value: 1000 | trampoline | |                      | | |
+| | | | blinded_paths:            | payload    | |                      | | |
+| | | |   <blinded_path_from_I1>  | for N1     | |                      | | |
+| | | |   <blinded_path_from_I3>  |            | |                      | | |
```
valentinewallace, line 379: @TheBlueMatt pointed out that just including the blinded paths for the trampoline hop this way currently won't work for sender-set custom TLVs or keysend preimages for async payments. Would be great to get your thoughts on this @t-bast!

t-bast: Good point! For this use-case, we'd need the recipient to support trampoline. This way we could include a trampoline onion for the recipient that would contain any sender-set data necessary. In this specific example, instead of having `EOF` right below, the trampoline onion would contain another trampoline hop for the recipient, that must be included by the trampoline node. Does that make sense?

t-bast: No, it's just like normal trampoline relay, intermediate nodes don't need to know anything about trampoline, only the final hop will receive a trampoline onion TLV in its onion payload.

t-bast: I'll need to think about it a bit more to figure out what's the best way of doing this indeed.

t-bast: Actually, thinking about this more, this shouldn't be too hard. When the recipient advertises support for trampoline in their invoice features, the sender should include a trampoline onion for them. This trampoline onion doesn't need to contain the `cltv_expiry`, it can contain only the amount that the sender intends to send, and extra TLVs (e.g. keysend). Then the intermediate trampoline node can use whatever block height when relaying that trampoline payment, and the recipient decides whether the CLTV expiry they receive is acceptable or not.

t-bast: Exactly, I'll update this PR with that scenario to ensure we're on the same page.

valentinewallace: Any status update on this, @t-bast? :)

t-bast: No, I didn't have time to work on this recently, and won't have time to look at it in the next couple of weeks, but let's discuss it again in two weeks!

t-bast: I thought about this more, and there's a trade-off I hadn't seen before, so I'd like your feedback on two potential designs. In https://github.com/lightning/bolts/commit/7589dd8e896eb2e3d206a05abb8a8c654057b693, I detailed how a sender can include blinded paths in the *trampoline* onion, which tells the trampoline node to relay the payment to those blinded paths. When done this way, the trampoline node is the *final* hop in the *trampoline* onion, so there's no way for the sender to include data for the recipient.

What follows is a bit subtle, so please take your time to check what is in the *trampoline* onion and what is in the *outer* onion in the two options below, and how it affects the payment.

### Option 1: simply add a trampoline hop for the recipient

My first idea was to slightly change the trampoline-to-blinded case from https://github.com/lightning/bolts/commit/7589dd8e896eb2e3d206a05abb8a8c654057b693 in the following way:

- the sender creates a *trampoline* packet where:
  - the last payload is for the final recipient (encrypted using its *blinded* node_id) and contains the amount sent and any additional TLV that we'd want to include (e.g. keysend)
  - the next-to-last payload contains the invoice's blinded paths
- the trampoline node decrypting the next-to-last payload:
  - detects that it contains blinded paths instead of an outgoing `node_id`, which tells it to relay to those blinded paths
  - must include in the *outer* onion of the outgoing payment the peeled trampoline packet (which contains the last payload for the final recipient)

The issue with that option is that the trampoline packet will be big, because it must contain blinded paths from the invoice. This isn't an issue between the sender and the last trampoline node, but it may be an issue for the trampoline node relaying to the blinded paths, because its *outer* onion will contain:

- a (big) trampoline onion in the last payload
- and all the `encrypted_recipient_data` necessary for standard blinded path logic (which are somewhat duplicated, because they used a lot of space in the trampoline onion, and now *also* appear in the outer onion)

The sender thus must limit how many blinded paths they include in the *trampoline* onion to ensure that the trampoline node will have enough space available to pay to those blinded paths. I'm afraid that in practice this will make it hard for the trampoline node to relay to those blinded paths because senders won't be able to include more than 1 or 2 blinded paths.

### Option 2: move blinded paths in the *outer* onion

The obvious workaround to this is that instead of putting the blinded paths in the *trampoline* onion, we can put them in the *outer* onion. This way the trampoline onion will be very small.

- the sender creates a *trampoline* packet where:
  - the last payload is for the final recipient (encrypted using its *blinded* node_id) and contains the amount sent and any additional TLV that we'd want to include (e.g. keysend)
  - the next-to-last payload does *not* contain blinded paths nor an outgoing `node_id`: it only contains the outgoing amount and cltv that the trampoline node must use
- the sender creates an *outer* packet where:
  - the last payload is for the trampoline node
  - the last payload contains the *trampoline* onion
  - the last payload contains the blinded paths
- the trampoline node decrypting the *outer* packet and its inner *trampoline* packet:
  - detects that it contains blinded paths and no outgoing `node_id`, which tells it to relay to those blinded paths
  - must include in the *outer* onion of the outgoing payment the peeled trampoline packet (which contains the last payload for the final recipient)

Since the trampoline onion can now be quite small, the trampoline node should have no issue relaying to the included blinded paths. The blinded paths only use space in the *outer* onion between the sender and the trampoline node. But the issue with this option is that the sender cannot use multiple trampoline nodes, since the blinded paths aren't in the *trampoline* onion but rather in the *outer* onion. Would that be an issue? Or is it fine because when combining trampoline and blinded paths, there is no need to use multiple trampoline hops?

If we think this is an issue, we could find a different way for the sender to encrypt some data for the recipient: but that would introduce yet another mechanism, so I'd like to avoid it to keep things simple.

Thoughts @valentinewallace @TheBlueMatt ?

valentinewallace: I think I see what you mean. To be clear, my understanding is that the trampoline packet would be too large in option 1 because even after the final trampoline hop unwraps its layer, per standard onion forwarding it essentially "pads" the trampoline onion that it forwards to be the same length as the one that it received. It's not completely obvious to me that there can't be a way around that, but I haven't looked into it. 

In any case, IMO option 2 doesn't sound unreasonable and sounds simpler. Would like to get @TheBlueMatt's thoughts on this though.

t-bast: > the trampoline packet would be too large in option 1 because even after the final trampoline hop unwraps its layer, per standard onion forwarding it essentially "pads" the trampoline onion that it forwards to be the same length as the one that it received

Yes, that's exactly it!

> It's not completely obvious to me that there can't be a way around that, but I haven't looked into it.

We cannot work around that unless we stop doing standard Sphinx, and thus lose its privacy properties...if we think we should work around it, we should find a completely different way of encrypting data only for the recipient, but I'd like to avoid that to avoid creating too many new protocols.

> In any case, IMO option 2 doesn't sound unreasonable and sounds simpler. Would like to get @TheBlueMatt's thoughts on this though.

I think it's reasonable too: having multiple trampoline hops doesn't give you any additional privacy when using blinded paths, and only has drawbacks (longer paths so more likely to fail). So it seems fine to me to use a single trampoline node when paying a trampoline recipient using blinded paths.

t-bast: I don't think that attack would work: the recipient sees that the `invoice_request` contains the trampoline feature bit and returns an invoice that contains the trampoline feature bit as `mandatory` (since the payer told them they support it). The recipient can then just fail the payment if it doesn't contain a trampoline onion.

We must make sure the trampoline node cannot create the trampoline onion themselves though, but that's easy to do by using the `payer_key` somehow (just including it again in the trampoline payload is probably enough?).

For the case where we're using trampoline to pay a blinded path where the recipient does *not* support trampoline, the trampoline node is not able to steal some sats as long as the payer pays exactly the invoice amount (which they decide in the `invoice_request`, so it should be ok?).

valentinewallace: Still thinking on this, but note that we may not be able to require trampoline feature support in invoices post-#1149, since async recipients will be creating static invoices in advance and will want to accept payment from non-trampoline-supporting senders.

t-bast: Right, but in the case where the recipient cannot come online to validate the payment, it knows that there is some trust involved anyway?

I'm still not convinced that it is a good idea, I'd rather always wait for the recipient to come online and generate a real invoice (and otherwise fail the payment), even with all the caveats it creates when waking up the payer or recipient takes some time...I'm not sure we'll ever implement the static invoice part in eclair, but if you want it for LDK I'm fine with having it in the spec, it's your call!

t-bast: > Sure, but there's nothing specific about the sender's trampoline onion? The malicious intermediary can also create a new trampoline onion.

Right, that's what was addressed in the next paragraph, where we need to reuse the `payer_key` to tie the `invoice_request` to the payment. But if you feel this is a layering violation, we'll need to come up with something else...

> If we can provide the trampoline with a list of (intro point, onion tail) that'd be much better than revealing the blinded path itself.

I'm not sure how we could do something like this. It seems to me exactly like rendezvous routing, which lacks flexibility for lightning payments because it doesn't let the trampoline node:

- set the expiry themselves (which is particularly important for async payments)
- set the amount themselves if they need to do MPP

I'd be happy to see a better proposal than my current one though if you find one, but the devil really is in the details, so it needs to be quite detailed to see if it would work or not.

t-bast: > Okay, Val pointed out that (duh) if an intermediary replaces the onion for an async payment the recipient can't claim the payment cause they won't have the preimage.

Right, I don't think the trampoline node is able to cheat if:

- we tie the trampoline onion to the sender's `payer_id` _somehow_ (an ECDH like you mention is probably the simplest way to do it without wasting additional space in the onion), which guarantees that if a trampoline onion is provided, it was created by the node who sent `invoice_request` (and thus received the invoice)
- if the recipient created an invoice with the trampoline feature bit (because the `invoice_request` advertised that feature bit as well), it rejects payments that don't contain a trampoline onion, which guarantees that the trampoline node has to include the trampoline onion
- if the recipient did *not* create the invoice themselves (which matches the async payment case when the recipient cannot be woken up to receive `invoice_request`), then we need some form of keysend anyway, which will only work if the trampoline node correctly includes the sender's trampoline onion

I can specify that in more details and add it to #836 and to this PR. I'll also need to decide between options 1 and 2 from [my previous comment](https://github.com/lightning/bolts/pull/829#discussion_r1679382517). I think we should do option 2), which restricts those payments to using a single trampoline node (see details in the comment). Are you ok with that?

remyers: > ...  I'll also need to decide between options 1 and 2 from [my previous comment](https://github.com/lightning/bolts/pull/829#discussion_r1679382517). I think we should do option 2), which restricts those payments to using a single trampoline node (see details in the comment). Are you ok with that?

I think the trade-off of a single trampoline node seems reasonable if it means that the sender can pass more blinded paths from the receiver to their trampoline node. Also if the sender pre-computes how many blinded paths they can include differently than their trampoline node we'll have some weird interop bugs. 

As I was thinking about these options, I [created some diagrams to rift on your examples](https://gist.github.com/remyers/3d31b86107dea20d338c9e4acbe1e9f3) from #836 to depict the two options. Perhaps others will find it useful.

valentinewallace: Could the sender add hops to the recipient's blinded paths to increase privacy, using `next_blinding_override`?

valentinewallace: IIRC another possibility we touched on was including trampoline hops within the blinded path. In my mind, this essentially meant that a forwarding trampoline+blinded node's TLVs would contain a `next_node_id` rather than an scid (amongst other changes). @t-bast does this option still seem worth exploring too? Happy to help btw.

proposals/trampoline.md
```diff
@@ -0,0 +1,648 @@
...
+        |                                           | |      (encrypted)      | |   |                 |                                           | expiry: 600000              |   |
+        |                                           | +-----------------------+ |   |                 |                                           | payment_secret: xxxxx       |   |
+        |                                           +---------------------------+   |                 |   HTLC(1480 msat, 600065 cltv)            | total_amount: 2500 msat     |   |
+        |                                           |             EOF           |   |                 |    +-----------------------+              | trampoline_onion:           |   |
+        |                                           +---------------------------+   |                 |    | amount_fwd: 1400 msat |              | +-------------------------+ |   |
+        |                                            HTLC(500 msat, 600112 cltv)    |                 |    | expiry: 600000        |              | | amount_fwd: 2500 msat   | |   |
```
f321x, line 578: You set `amount_fwd` == `total_amount` in the final trampoline onions for the receiver in this mpp ascii drawing. Is there any specific reason to do this? 
I cannot find details about this in the conditions for final nodes in https://github.com/lightning/bolts/blob/9938ab3d6160a3ba91f3b0e132858ab14bfe4f81/04-onion-routing.md?plain=1#L547-L553
Is it fine to set the final trampoline_onions `amount_fwd` to the outer onions `amount_fwd` as well?

t-bast: > Is it fine to set the final trampoline_onions amount_fwd to the outer onions amount_fwd as well?

That would be fine whenever you use a *single* trampoline route, but that wouldn't work if you want to split a payment across multiple trampoline routes: that's where the trampoline onion's `total_msat` field is necessary.

For example, if you want to send 250 000 sats that you split into two trampoline routes:

- sender -> T1 -> T2 -> recipient where you send 100 000 sats
- sender -> T3 -> T4 -> recipient where you send 150 000 sats

The trampoline onion's `total_msat` would be `250 000 sats` in both routes, whereas the outer onions will only believe they carry a `100 000 sats` payment for the first route and a `150 000 sats` payment for the second route.

Does that make sense?

> I cannot find details about this in the conditions for final nodes in 

I haven't duplicated every requirement because for most of the logic, the trampoline onion must be validated very similarly to the outer onion. I was hoping this case would be captured by:

```
  - MUST process the `trampoline_onion_packet` as an `onion_packet`.
  - MUST fail the HTLC if dictated by the requirements under [Failure Messages](#failure-messages).
```

But if you think it's worth explicitly detailing this requirement, I can add that!

SomberNight: Consider e.g. this paragraph:
https://github.com/lightning/bolts/blob/9938ab3d6160a3ba91f3b0e132858ab14bfe4f81/04-onion-routing.md?plain=1#L373-L377

What we mean by HTLC there becomes really fuzzy with this recursion.
When looking at the `amt_to_forward` inside the trampoline onion, first the final recipient needs to aggregate all the outer_onion level htlcs that have a matching `payment_hash+outer_payment_secret`. The htlc-value-sum of this set, should match the inner `amt_to_forward`.

Then a second-level of aggregation can be done: the sum of the inner `amt_to_forward`s matches the inner `total_amount`.

I guess the outer `total_amount` matching the inner `amt_to_forward` is invariant.

@t-bast's ascii drawings are magnificent, but we need to go even larger :D
```
                                                            HTLC(1500 msat, 600112 cltv)
                                                            +---------------------------+
                                                            | amount_fwd: 1500 msat     |
                                                            | expiry: 600112            |
                    HTLC(1560 msat, 600124 cltv)            | payment_secret: aaaaa     |
                     +-----------------------+              | total_amount: 2800 msat   |
                     | amount_fwd: 1500 msat |              | trampoline_onion:         |                                                                  HTLC1(1100 msat, 600000 cltv)
                     | expiry: 600112        |              | +-----------------------+ |                                                                 +-----------------------------+
                +--> | channel_id: 3         | ---> I1 ---> | | amount_fwd: 2500 msat | | --+                                                             | amount_fwd: 1100 msat       |
                |    |-----------------------|              | | expiry: 600000        | |   |                                                             | expiry: 600000              |
                |    |     (encrypted)       |              | | node_id: Bob          | |   |                                                             | payment_secret: xxxxx       |
                |    +-----------------------+              | +-----------------------+ |   |                     HTLC(1150 msat, 600080 cltv)            | total_amount: 2500 msat     |
                |                                           | |      (encrypted)      | |   |                      +-----------------------+              | trampoline_onion:           |
                |                                           | +-----------------------+ |   |                      | amount_fwd: 1100 msat |              | +-------------------------+ |
                |                                           +---------------------------+   |                      | expiry: 600000        |              | | amount_fwd: 2500 msat   | |
                |                                           |             EOF           |   |                 +--> | channel_id: 561       | ---> I4 ---> | | expiry: 600000          | | --+
                |                                           +---------------------------+   |                 |    |-----------------------|              | | total_amount: 5200 msat | |   |
                |                                            HTLC(800 msat, 600112 cltv)    |                 |    |     (encrypted)       |              | | payment_secret: yyyyy   | |   |
                |                                           +---------------------------+   |                 |    +-----------------------+              | +-------------------------+ |   |
                |                                           | amount_fwd: 800 msat      |   |                 |                                           | |         EOF             | |   |
                |                                           | expiry: 600112            |   |                 |                                           | +-------------------------+ |   |
                |   HTLC(820 msat, 600130 cltv)             | payment_secret: aaaaa     |   |                 |                                           +-----------------------------+   |
                |    +-----------------------+              | total_amount: 2800 msat   |   |                 |                                           |             EOF             |   |
                |    | amount_fwd: 800 msat  |              | trampoline_onion:         |   |                 |                                           +-----------------------------+   |
                |    | expiry: 600112        |              | +-----------------------+ |   |                 |                                                                             |
        +-------+--> | channel_id: 5         | ---> I2 ---> | | amount_fwd: 2500 msat | | --+---->  T1  ------+                                                                             +----+
        |       |    |-----------------------|              | | expiry: 600000        | |   |  (fee 170 msat) |                                            HTLC2(1400 msat, 600000 cltv)    |    |
        |       |    |     (encrypted)       |              | | node_id: Bob          | |   |    (delta 32)   |                                           +-----------------------------+   |    |
        |       |    +-----------------------+              | +-----------------------+ |   |                 |                                           | amount_fwd: 1400 msat       |   |    |
        |       |                                           | |      (encrypted)      | |   |                 |                                           | expiry: 600000              |   |    |
        |       |                                           | +-----------------------+ |   |                 |                                           | payment_secret: xxxxx       |   |    |
        |       |                                           +---------------------------+   |                 |   HTLC(1480 msat, 600065 cltv)            | total_amount: 2500 msat     |   |    |
        |       |                                           |             EOF           |   |                 |    +-----------------------+              | trampoline_onion:           |   |    |
        |       |                                           +---------------------------+   |                 |    | amount_fwd: 1400 msat |              | +-------------------------+ |   |    |
        |       |                                            HTLC(500 msat, 600112 cltv)    |                 |    | expiry: 600000        |              | | amount_fwd: 2500 msat   | |   |    |
        |       |                                           +---------------------------+   |                 +--> | channel_id: 1105      | ---> I5 ---> | | expiry: 600000          | | --+    |
        |       |                                           | amount_fwd: 500 msat      |   |                      |-----------------------|              | | total_amount: 5200 msat | |        |
        |       |                                           | expiry: 600112            |   |                      |     (encrypted)       |              | | payment_secret: yyyyy   | |        |
        |       |   HTLC(510 msat, 600120 cltv)             | payment_secret: aaaaa     |   |                      +-----------------------+              | +-------------------------+ |        |
        |       |    +-----------------------+              | total_amount: 2800 msat   |   |                                                             | |         EOF             | |        |
        |       |    | amount_fwd: 500 msat  |              | trampoline_onion:         |   |                                                             | +-------------------------+ |        |
        |       |    | expiry: 600112        |              | +-----------------------+ |   |                                                             +-----------------------------+        |
        |       +--> | channel_id: 7         | ---> I3 ---> | | amount_fwd: 2500 msat | | --+                                                             |             EOF             |        |
        |            |-----------------------|              | | expiry: 600000        | |                                                                 +-----------------------------+        |
        |            |     (encrypted)       |              | | node_id: Bob          | |                                                                                                        |
        |            +-----------------------+              | +-----------------------+ |                                                                                                        |
        |                                                   | |      (encrypted)      | |                                                                                                        |
        |                                                   | +-----------------------+ |                                                                                                        |
        |                                                   +---------------------------+                                                                                                        |
        |                                                   |             EOF           |                                                                                                        |
        |                                                   +---------------------------+                                                                                                        |
        |                                                                                                                                                                                        |
Alice --+                                                                                                                                                                                        +--> Bob
        |                                                                                                                                                                                        |
        |                                                                                                                                                                                        |
        |                                                   HTLC(1500 msat, 600112 cltv)                                                                                                         |
        |                                                   +---------------------------+                                                                                                        |
        |                                                   | amount_fwd: 1500 msat     |                                                                                                        |
        |                                                   | expiry: 600112            |                                                                                                        |
        |           HTLC(1560 msat, 600124 cltv)            | payment_secret: bbbbb     |                                                                                                        |
        |            +-----------------------+              | total_amount: 3000 msat   |                                                                                                        |
        |            | amount_fwd: 1500 msat |              | trampoline_onion:         |                                                                  HTLC3(1200 msat, 600000 cltv)         |
        |            | expiry: 600112        |              | +-----------------------+ |                                                                 +-----------------------------+        |
        |       +--> | channel_id: 4         | ---> J1 ---> | | amount_fwd: 2700 msat | | --+                                                             | amount_fwd: 1200 msat       |        |
        |       |    |-----------------------|              | | expiry: 600000        | |   |                                                             | expiry: 600000              |        |
        |       |    |     (encrypted)       |              | | node_id: Bob          | |   |                                                             | payment_secret: zzzzz       |        |
        |       |    +-----------------------+              | +-----------------------+ |   |                     HTLC(1250 msat, 600080 cltv)            | total_amount: 2700 msat     |        |
        |       |                                           | |      (encrypted)      | |   |                      +-----------------------+              | trampoline_onion:           |        |
        |       |                                           | +-----------------------+ |   |                      | amount_fwd: 1200 msat |              | +-------------------------+ |        |
        |       |                                           +---------------------------+   |                      | expiry: 600000        |              | | amount_fwd: 2700 msat   | |        |
        |       |                                           |             EOF           |   |                 +--> | channel_id: 563       | ---> J4 ---> | | expiry: 600000          | | --+    |
        |       |                                           +---------------------------+   |                 |    |-----------------------|              | | total_amount: 5200 msat | |   |    |
        |       |                                            HTLC(900 msat, 600112 cltv)    |                 |    |     (encrypted)       |              | | payment_secret: yyyyy   | |   |    |
        |       |                                           +---------------------------+   |                 |    +-----------------------+              | +-------------------------+ |   |    |
        |       |                                           | amount_fwd: 900 msat      |   |                 |                                           | |         EOF             | |   |    |
        |       |                                           | expiry: 600112            |   |                 |                                           | +-------------------------+ |   |    |
        |       |   HTLC(920 msat, 600130 cltv)             | payment_secret: bbbbb     |   |                 |                                           +-----------------------------+   |    |
        |       |    +-----------------------+              | total_amount: 3000 msat   |   |                 |                                           |             EOF             |   |    |
        |       |    | amount_fwd: 900 msat  |              | trampoline_onion:         |   |                 |                                           +-----------------------------+   |    |
        |       |    | expiry: 600112        |              | +-----------------------+ |   |                 |                                                                             |    |
        +-------+--> | channel_id: 6         | ---> J2 ---> | | amount_fwd: 2700 msat | | --+---->  T2  ------+                                                                             +----+
                |    |-----------------------|              | | expiry: 600000        | |   |  (fee 170 msat) |                                            HTLC4(1500 msat, 600000 cltv)    |
                |    |     (encrypted)       |              | | node_id: Bob          | |   |    (delta 32)   |                                           +-----------------------------+   |
                |    +-----------------------+              | +-----------------------+ |   |                 |                                           | amount_fwd: 1500 msat       |   |
                |                                           | |      (encrypted)      | |   |                 |                                           | expiry: 600000              |   |
                |                                           | +-----------------------+ |   |                 |                                           | payment_secret: zzzzz       |   |
                |                                           +---------------------------+   |                 |   HTLC(1580 msat, 600065 cltv)            | total_amount: 2700 msat     |   |
                |                                           |             EOF           |   |                 |    +-----------------------+              | trampoline_onion:           |   |
                |                                           +---------------------------+   |                 |    | amount_fwd: 1500 msat |              | +-------------------------+ |   |
                |                                            HTLC(600 msat, 600112 cltv)    |                 |    | expiry: 600000        |              | | amount_fwd: 2700 msat   | |   |
                |                                           +---------------------------+   |                 +--> | channel_id: 1108      | ---> J5 ---> | | expiry: 600000          | | --+
                |                                           | amount_fwd: 600 msat      |   |                      |-----------------------|              | | total_amount: 5200 msat | |
                |                                           | expiry: 600112            |   |                      |     (encrypted)       |              | | payment_secret: yyyyy   | |
                |   HTLC(610 msat, 600120 cltv)             | payment_secret: bbbbb     |   |                      +-----------------------+              | +-------------------------+ |
                |    +-----------------------+              | total_amount: 3000 msat   |   |                                                             | |         EOF             | |
                |    | amount_fwd: 600 msat  |              | trampoline_onion:         |   |                                                             | +-------------------------+ |
                |    | expiry: 600112        |              | +-----------------------+ |   |                                                             +-----------------------------+
                +--> | channel_id: 8         | ---> J3 ---> | | amount_fwd: 2700 msat | | --+                                                             |             EOF             |
                     |-----------------------|              | | expiry: 600000        | |                                                                 +-----------------------------+
                     |     (encrypted)       |              | | node_id: Bob          | |
                     +-----------------------+              | +-----------------------+ |
                                                            | |      (encrypted)      | |
                                                            | +-----------------------+ |
                                                            +---------------------------+
                                                            |             EOF           |
                                                            +---------------------------+
```

t-bast: Thanks for the awesome ASCII diagram, added a section including it in https://github.com/lightning/bolts/pull/829/commits/bc7a1a0bc97b2293e7f43dd8a06529e5fdcf7cd2

proposals/trampoline.md
```diff
@@ -0,0 +1,769 @@
...
+| Alice |              | T1 |         | T2 | (error)
++-------+              +----+         +----+
+```
+
+* T2 creates a local error and wraps it: `wrap(ss(T1,T2), wrap(tss(Alice,T2), error))`
+* T1 receives `wrap(ss(T1, M1), wrap(ss(T1,T2), wrap(tss(Alice,T2), error)))`
+* After unwrapping with the outer onion shared secrets T1 obtains: `wrap(tss(Alice,T2), error)`
+  It cannot decrypt further, so it adds its own two layers of wrapping and
+  forwards upstream `wrap(ss(Alice,T1), wrap(tss(Alice,T1), wrap(tss(Alice,T2), error)))`
+* Alice receives `wrap(ss(Alice,N1), wrap(ss(Alice,N2), wrap(ss(Alice,T1), wrap(tss(Alice,T1), wrap(tss(Alice,T2), error)))))`
```
f321x, line 241: @t-bast are you already implementing this double obfuscation scheme in Phoenix?
Right now Electrum is not able to decrypt a `wrap(tss(Alice,T2), error)` like described here, as it doesn't even stores the trampoline onions session key. So in theory if `Alice` uses Electrum and `T2` uses Phoenix, Alice shouldn't be able to decrypt the onion error as the data she sees after unwrapping all the layers encrypted with the outer onion shared secret is still encrypted.
However when testing a payment `Electrum` -> `Electrum Trampoline` -> `Acinq` -> `Phoenix` and letting `Phoenix` fail the payment `Electrum` can still decode the received onion error. This is confusing, so I was wondering if this scheme is even used in practice yet?

SomberNight: >  However when testing a payment Electrum -> Electrum Trampoline -> Acinq -> Phoenix and letting Phoenix fail the payment Electrum can still decode the received onion error. This is confusing, so I was wondering if this scheme is even used in practice yet?

To clarify, "Electrum Trampoline" is running eclair, so the payment goes: electrum -> eclair1 -> eclair2 -> phoenix.
If e.g. electrum sends an htlc with an incorrect payment_secret, phoenix fails it with `incorrect_or_unknown_payment_details`. The sender electrum then decrypts it successfully using `ss(electrum, eclair1)`. This is unexpected.

proposals/trampoline.md
```diff
@@ -0,0 +1,769 @@
...
+When used with blinded paths, there is no need for multiple trampoline hops.
+Even with a single trampoline node, the destination stays private.
+
+## Trampoline MPP
+
+Trampoline routing combines nicely with multi-part payments. When multi-part
+payment is used, we can let trampoline nodes combine all the incoming partial
+payments before forwarding. Once the totality of the payment is received, the
+trampoline node can choose the most efficient way to re-split it to reach the
+next trampoline node.
```
SomberNight, line 536: What if the recipient created an invoice with the `basic_mpp` flag unset? Perhaps they want a single HTLC because they are trying to get a JIT channel opened from an LSP who requires no-MPP for JIT.

How would the trampoline forwarder know whether they are allowed to use MPP to re-split?

To increase payment reliability, it might make sense to always allow intermediate trampoline forwaders to split. So that would mean that the invoice MPP flag should only affect the last trampoline forwarder? But that would imply giving them the knowledge that they are the last forwarder.

Or does signalling for trampoline support in an invoice simply imply supporting MPP also, for a recipient?
And then for the LSP JIT usecase, we just assume that the LSP must be the last trampoline forwarder, and that it is up to them not to re-split the payment?

(note: with the "pay-to-legacy" hack, the last trampoline forwarder gets the invoice_features, so they know whether they can do MPP)

t-bast: > Or does signalling for trampoline support in an invoice simply imply supporting MPP also, for a recipient?

Yes it does, we do assume support for MPP, because it's an important part of what makes trampoline efficient. We've discussed this a few times in the past, and there wasn't any good reason for a recipient to *not* support MPP. I should make this obvious in the spec (probably in https://github.com/lightning/bolts/pull/836 rather than here) if that's not already mentioned!

> are trying to get a JIT channel opened from an LSP who requires no-MPP for JIT.

This is a deficiency in the JIT-channel spec, which is fixed by https://github.com/lightning/blips/pull/36. There is no good reason for a JIT-channel protocol to be incompatible with MPP. With an on-the-fly-funding protocol that supports MPP, do you see a reason for a recipient to *not* want to support MPP?

PR #834, opened 2021-01-20
BOLT 1: introduce warning messages, reduce requirements to send (hard) errors

rustyrussell:
~If you really want to have a hard error, set the new fatal field.  We can still get upset if this happens during testing, and we can still log when they happen, but not break everything quite so easily!~

This follows the suggestion by @t-bast to create a new message (1) "warning".  I've implemented this for c-lightning already as a PR.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -396,7 +396,6 @@ memory.
   - if it doesn't receive a corresponding `pong`:
     - MAY terminate the network connection,
       - and MUST NOT fail the channels in this case.
-  - SHOULD NOT send `ping` messages more often than once every 30 seconds.
 
 A node sending a `pong` message:
   - SHOULD set `ignored` to 0s.
@@ -404,7 +403,6 @@ A node sending a `pong` message:
  memory.
 
 A node receiving a `ping` message:
-  - SHOULD fail the channels if it has received significantly in excess of one `ping` per 30 seconds.
   - if `num_pong_bytes` is less than 65532:
     - MUST respond by sending a `pong` message, with `byteslen` equal to `num_pong_bytes`.
   - otherwise (`num_pong_bytes` is **not** less than 65532):


rustyrussell:
I've started refactoring clightning internally so all our error routines are flagged soft or hard.  I've got a hack where soft errors send out msg number 1 (undefined, currently ignored) instead of 17, which makes it more effective with older clightning which treats all error (17) messages as "hard".

Not sure if we should endorse that as a separate "warning" message?

The other approach would be to redefine 17 (error) as a soft error and create a new message (2?) as "hard error"?

t-bast:
Interesting, eclair also defaults to fatal errors and closes channels when they happen, so old eclair will keep treating all errors as fatal regardless of the tlv flag. When the `fatal` TLV is missing, updated nodes will still want to consider these errors fatal and close the channel.

I think a separate (odd) message would be more appropriate, as it would work out of the box for non-upgraded nodes (they would simply ignore it so won't close a channel).

rustyrussell:
Yes, though we've already had to except various errors from failing the channel, since LND sends them (e.g. if you are too slow in reestablishing a channel).  Probably because LND didn't close on errors, so it didn't treat them as fatal itself.

Let me try a "warning" message instead then...

rustyrussell:
OK, now it's a separate warning message, and I went through and revised and made consistent all the errors/closing/failing language.

01-messaging.md
```diff
@@ -325,24 +327,27 @@ The fundee node:
...
+  - when sending `warning`:
+	- MAY set `channel_id` to all zero if the warning is not related to a specific channel.
+  - MAY close the connection after sending.
   - MAY send an empty `data` field.
   - when failure was caused by an invalid signature check:
     - SHOULD include the raw, hex-encoded transaction in reply to a `funding_created`, `funding_signed`, `closing_signed`, or `commitment_signed` message.
-  - when `channel_id` is 0:
-    - MUST fail all channels with the receiving node.
-    - MUST close the connection.
-  - MUST set `len` equal to the length of `data`.
```
t-bast, line 338: nit: why remove that requirement? Because it's obvious?

03-transactions.md
```diff
@@ -475,7 +475,8 @@ contribute to fees.
 
 A node:
   - if the resulting fee rate is too low:
-    - MAY fail the channel.
+    - MAY send a `warning` warn and close the connection, or send an
```
t-bast, line 478:
```suggestion
    - MAY send a `warning` and close the connection, or send an
```

t-bast (commented):
I think it's a good idea, it will be much simpler to notice issues and have users take unilateral actions without needing to contact their peer to understand what went wrong.

carlaKC:
Hi all, I started thinking about possible extensions to how the spec deals with errors towards the end of last year, and wanted to drop some thoughts in here. 

A string based error is nice for humans to read, but doesn’t give code much of a chance of elegantly reacting. If we’re going to make a change now, I think it would be worth considering the following so that we don’t have to come back again to make further updates: 
* Error codes: an enum which describes why the error is being sent
* Error-specific fields: based on the error code set, could be TLV or a fixed set of fields
* Fatal bool: similar to the idea introduced here, indicating whether we need to close the channel.

I don’t want to derail conversation here , so will make my own ML post with details if this sounds reasonable.

t-bast:
I think it could be reasonable for this new warning message (I'm not sure it makes sense for the `error` message though).

Do you have some concrete examples of what warnings could trigger an automatic action? TBH it feels like the protocols in the spec should have all cases covered to avoid needing it, but there's always a difference between theory and practice.

01-messaging.md
rustyrussell: Yeah, we no longer spell out things which are in the definition (like `len` here).  That comment predates the formal formatting of all messages.

We no longer have a "fail all channels!" error, which IMHO makes sense (we have a "warning, but unknown channel").

rustyrussell:
> Hi all, I started thinking about possible extensions to how the spec deals with errors towards the end of last year, and wanted to drop some thoughts in here.
> 
> A string based error is nice for humans to read, but doesn’t give code much of a chance of elegantly reacting. If we’re going to make a change now, I think it would be worth considering the following so that we don’t have to come back again to make further updates:
> 
>     * Error codes: an enum which describes why the error is being sent
> 
>     * Error-specific fields: based on the error code set, could be TLV or a fixed set of fields
> 
>     * Fatal bool: similar to the idea introduced here, indicating whether we need to close the channel.

In general, these errors should *never* occur.  It means one side has a bug.  Unfortunately, that happens in practice, leading to workarounds where errors are ignored.  Hence distinguishing between "Warning: you sent crap" and "Fatal: no point talking any more, I've killed the channel".

As we cannot define all possible bugs, the best you can do is log something and retry, and eventually give up (just as you would if there were no progress without any error).

There is a strong case of other kinds of errors: in the offers spec I included such things, as they can genuinely happen, and we already have onion errors.  We've also talked about soft-errors in the onion context (e.g. "I don't want to worry you, but there's a delay...")

carlaKC:
> In general, these errors should never occur. It means one side has a bug... As we cannot define all possible bugs, the best you can do is log something and retry 

Right, I think I misunderstood the original reasoning for this warning message - thanks for clarifying! Error codes make less in the infinite range of bugs that one could run into :')

> Do you have some concrete examples of what warnings could trigger an automatic action? 

@t-bast I'll make a ML post with my case for these error codes, rather than clogging up this issue more since they are related, but not directly relevant to a soft warning error.

niftynei:
> As we cannot define all possible bugs, the best you can do is log something and retry, and eventually give up (just as you would if there were no progress without any error).

I'm working on the error pathways for the dual-funding RBF right now. This disambiguation between "warning" and "error" is very useful in that context.  Particularly how 'warning/close connection' allows a retry from initial state conditions, in other words signals to abandon the failed RBF attempt (with an explanation as to why.)

Error codes are indeed useful for specific, narrowly defined interactions and recovery pathways, like payment attempts. In c-lightning we also encode them for our RPC errors, to help identify common cases. I can see a use case where we'd want error codes specifically for RBF failure cases, however these can be added ad-hoc for specific and well-understood cases in the future. 

In the general case, returning a warning to the caller with an explanation and re-starting the attempt is a *robust* solution; error coded response handling logic has a tendency to degrade robustness over time ime.

t-bast (commented):
LGTM, I finally implemented this in eclair and tested with c-lightning, and we exchange warning messages correctly.

Just needs a rebase and it's an ACK for me!

01-messaging.md
```diff
@@ -310,12 +310,14 @@ For simplicity of diagnosis, it's often useful to tell a peer that something is
...
-The 2-byte `len` field indicates the number of bytes in the immediately following field.
+1. type: 1 (`warning`)
+2. data:
+   * [`channel_id`:`channel_id`]
+   * [`u16`:`len`]
+   * [`len*byte`:`data`]
 
 #### Requirements
 
-The channel is referred to by `channel_id`, unless `channel_id` is 0 (i.e. all bytes are 0), in which case it refers to all channels.
```
TheBlueMatt, line 317: I don't see why we should get rid of this? The great thing about having `warning` messages is we can now reserve `error` messages specifically for channel-closing issues. Having a "go away, I don't want to ever talk to you again, close your channels, I hate you" message doesn't seem like a thing we should get rid of just-because.

TheBlueMatt (commented):
I'm pretty dubious of the changes to not send `error`s if a counterparty sends bogus signatures. As written, sending a warning and disconnecting seems like a great way to get into a loop constantly warning and reconnecting with a broken peer. I know there have been incompatibilities in the past where nodes have been able to get into inconsistent states due to race conditions, but those seem largely a thing of the past.

01-messaging.md
```diff
@@ -325,24 +327,27 @@ The fundee node:
...
   - upon receiving `error`:
-    - MUST fail the channel referred to by the error message, if that channel is with the sending node.
-  - if no existing channel is referred to by the message:
+    - MUST fail the channel referred to by `channel_id`, if that channel is with the sending node.
+  - upon receiving `warning`:
+    - SHOULD log the message for later diagnosis.
+	- MAY disconnect.
+	- MAY reconnect after some delay to retry.
+	- MAY attempt `shutdown` if permitted at this point.
+  - if no existing channel is referred to by `channel_id`:
```
TheBlueMatt, line 349: @ariard notes that this appears to apply to warning messages as well, implying that you should ignore and not even bother to log a warning message that is not attached to a channel. I assume that isn't the intent here.

cc https://github.com/rust-bitcoin/rust-lightning/pull/1013#discussion_r683644076

01-messaging.md
t-bast: I'm not sure it's necessary anymore, it should be preferable to either mutual close all channels when you want to "go away", or if you lost state settle them after `channel_reestablish` with `option_data_loss_protect` (which will create one error per-channel)?

I believe it was removed because all the places where the spec would recommend to close all channels were changed to only send a warning, doesn't it?

t-bast:
> I'm pretty dubious of the changes to not send errors if a counterparty sends bogus signatures. As written, sending a warning and disconnecting seems like a great way to get into a loop constantly warning and reconnecting with a broken peer. 

It's true that it will create reconnect loops.
I'd slightly prefer having that peer stuck in a loop and hopefully see them upgrade their software and then we're back to normal rather than force-closing, but in practice if that peer doesn't fix the issue we would likely end up force-closing the channels anyway, so I don't have strong feelings on that one.

01-messaging.md
TheBlueMatt: > I believe it was removed because all the places where the spec would recommend to close all channels were changed to only send a warning, doesn't it?

I don't think this should be the criteria for removing useful things in the protocol. eg I believe we currently send an `null-error` for "feature flags mismatch, we cannot connect to this peer", which seems like the right thing to do. If our counterparty upgrades to a new version which we cannot talk to at all, we should force-close and move on with our lives. `null-error` lets us do that, otherwise we're in a reconnect loop.

t-bast: > we currently send an null-error for "feature flags mismatch, we cannot connect to this peer"

IMHO this is something that will trigger many channel force-close that in practice are really unnecessary...
I believe a better behavior in that case is "send a warning, disconnect and don't initiate reconnection yourself".

We've had this kind of issue with `eclair-mobile`, which doesn't support `option_static_remotekey`, when a new lnd version made it `mandatory` by default. Users connecting their wallet to their own node couldn't connect after an upgrade. But when they noticed this and understood the issue, they simply reconfigured their node to make this feature `optional` and restarted, and everything was back to normal. With the behavior you suggest, they would have lost all their channels, and I think we would have gotten a lot of heat for it in support tickets!

In the general case though, I agree with you that we may want to keep this potential "close all channels" message (there's no strong need to remove it from the spec), let's see what others think.

TheBlueMatt: > I believe a better behavior in that case is "send a warning, disconnect and don't initiate reconnection yourself".

Eh, maybe some day we'll change it, but for now we're going to keep it because its not clear what to set all the various timeout constants to.

In any case, its just an example, and the broader point is more that if we are closing all our channels, we should be able to tell our peers that! One of the biggest annoyances with some nodes ignoring `error` messages entirely today is that nodes never learn a channel has been closed until it hits the chain, which may take quite some time depending on feerates.

rustyrussell: At one stage in c-lightning we sent this error in return for bad gossip messages.  This was, um, dumb.  So for the longest time we have ignored all-channel errors (well, treated them as a warning), and I doubt anyone else did "close all" either.

01-messaging.md
rustyrussell: I've always treated "ignore" as meaning "don't give a response", not "don't even log".

02-peer-protocol.md
```diff
@@ -496,14 +498,15 @@ A sending node:
 
 A receiving node:
   - if it hasn't received a `funding_signed` (if it is a funder) or a `funding_created` (if it is a fundee):
-    - SHOULD fail the connection
+    - SHOULD send a `warning`.
```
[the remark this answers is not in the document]
rustyrussell: Ack.

02-peer-protocol.md
```diff
@@ -496,14 +498,15 @@ A sending node:
...
   - if the `scriptpubkey` is not in one of the above forms:
-    - SHOULD fail the connection.
+    - SHOULD send a `warning`.
   - if it hasn't sent a `funding_locked` yet:
     - MAY reply to a `shutdown` message with a `shutdown`
   - once there are no outstanding updates on the peer, UNLESS it has already sent a `shutdown`:
     - MUST reply to a `shutdown` message with a `shutdown`
   - if both nodes advertised the `option_upfront_shutdown_script` feature, and the receiving node received a non-zero-length `shutdown_scriptpubkey` in `open_channel` or `accept_channel`, and that `shutdown_scriptpubkey` is not equal to `scriptpubkey`:
+    - MAY send a `warning`.
```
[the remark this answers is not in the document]
rustyrussell: Why?  I mean, you *can*, but if they screw up it seems a bit harsh.  Dropping the connection means you ignore it, and they can try again.

01-messaging.md
TheBlueMatt: We do :). And just because folks don't currently do it doesn't mean we can't in the future.

01-messaging.md
TheBlueMatt: I'm fine with that. I think "ignore" is fine wording.

02-peer-protocol.md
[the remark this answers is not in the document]
TheBlueMatt: I don't see how its "harsh"? Its a channel open - you can try again freely without reconnecting if you send an error, an error message is somewhat less annoying cause you don't have to do a whole P2P reconnect.

TheBlueMatt:
> I'd slightly prefer having that peer stuck in a loop and hopefully see them upgrade their software and then we're back to normal rather than force-closing, but in practice if that peer doesn't fix the issue we would likely end up force-closing the channels anyway, so I don't have strong feelings on that one.

I dunno, we'll probably leave the force-closes in place on our end. I don't mind it hugely being here and being optional, but the extra code to do the reconnect loop, track the reconnect and if we fail twice then force-close, or doing it on a timer, is a bunch of logic in an error path, which I despise (code that is never called is code that is probably wrong, even if its tested).

02-peer-protocol.md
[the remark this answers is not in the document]
rustyrussell: No, this is not channel open?  (That was above, and I change it as you suggested)

 This is "you tried to close to wrong scriptpubkey".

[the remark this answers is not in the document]
TheBlueMatt: Ah! I think I'm blind. I guess this still falls into the same general category as the all-0 errors - it feels like yet another case of "well, if we reconnect and then retry maybe it'll resolve itself", which requires a bunch of never-used logic to ensure we fail after a few reconnect/retry loops as its otherwise unrecoverable. I guess I'm a bit confused by "MUST fail the connection" (which means force-close, no? The updates in 00 don't mention it explicitly) and also "MAY send a warning" - we should error if we fail the channel.

t-bast (approved):

t-bast:
Needs rebase, otherwise ACK https://github.com/lightningnetwork/lightning-rfc/pull/834/commits/4f2ec0676310fbe757f05e20f791830a195ffdb9

Roasbeef:
@niftynei 

> In the general case, returning a warning to the caller with an explanation and re-starting the attempt is a robust solution; error coded response handling logic has a tendency to degrade robustness over time ime.

If the warning message has no context at all, then how is the caller meant to parse the message to restart the attempt? Totally disagree that error code response handling logic degrades over time, when was the last time you needed to redefine how you handle the HTTP 302 error code?

t-bast:
> If the warning message has no context at all, then how is the caller meant to parse the message to restart the attempt? Totally disagree that error code response handling logic degrades over time, when was the last time you needed to redefine how you handle the HTTP 302 error code?

I really don't understand what you mean. 
Node operators have context about what they're doing, they know when they're opening or closing channels.
So it's really helpful for them to get *something* when it's rejected by the other side, and a string message is actually a very flexible way of doing this.

We've already found it very useful many times in the wild compared to "the remote peer just disconnected on me, so something must be wrong with what I sent them but I have no idea what".

TheBlueMatt:
Wen rebase, lets land this :)

t-bast (approved):
ACK https://github.com/lightningnetwork/lightning-rfc/pull/834/commits/5c745fdf77de58eea4f5973991ffed5a79fc0371, let's land this!

01-messaging.md
TheBlueMatt: I think this is the last thing we need to resolve, I'm still highly unconvinced here - yes, some of the historical places all-0 channel ids have been used was nonsense, but the idea that a peer will have a channel with me and I don't want to talk to them anymore and I'm not able to tell them that seems really wrong? Implementing all the logic to stay connected to them, parse their messages for channel IDs, and then fail *those* channels seems substantially complicated, and probably not relevant in cases where you just want a peer to "go away". We can change how we communicate this if backwards compatibility is the concern, but it just seems very strange to me to not be able to communicate this at all (when we currently can).

02-peer-protocol.md
```diff
@@ -554,14 +556,15 @@ A sending node:
...
     - MUST fail the connection.
```
TheBlueMatt, line 568: Aren't we getting rid of the "fail the connection" terminology in this pr :)

02-peer-protocol.md
```diff
@@ -554,14 +556,15 @@ A sending node:
 
 A receiving node:
   - if it hasn't received a `funding_signed` (if it is a funder) or a `funding_created` (if it is a fundee):
-    - SHOULD fail the connection
+    - SHOULD send an `error` and fail the channel.
   - if the `scriptpubkey` is not in one of the above forms:
-    - SHOULD fail the connection.
+    - SHOULD send a `warning`.
```
TheBlueMatt, line 561: In general (including here and the below `warning` conversion, but basically everywhere), we should probablyld probably have something that says "and ignore the message" or so. That would imply that the sender can re-send the `shutdown` with a different script (or whatever message in most contexts), which seems to be your goal here. Maybe it could be a general comment at the top of bolt 0 or 1 whatever.

TheBlueMatt (commented):
Basically there, want to come to come conclusion on all-0 errors.

01-messaging.md
t-bast: If some implementations find it useful to keep this option, I think we should keep it.

TheBlueMatt:
Successfully used this to debug an issue in a C-Lightning unmerged PR from LDK, so I'd call that cross-implementation-more-than-tested :)

01-messaging.md
TheBlueMatt: Any update on this? It'd be nice to land this PR.

rustyrussell:
Restored all-zero errors, and trivial rebase.

t-bast (approved):
Ack https://github.com/lightning/bolts/pull/834/commits/570d3565fe5e4c765610d2169cbeafeaca23d298, thanks! Let's finally land this!

00-introduction.md
```diff
@@ -132,6 +132,20 @@ See [BOLT #11: Invoice Protocol for Lightning Payments](11-payment-encoding.md)
    * _See related: [closing transaction](#closing-transaction), [funding transaction](#funding-transaction), [penalty transaction](#penalty-transaction)_
    * _See types: [revoked commitment transaction](#revoked-commitment-transaction)_
 
+* #### *Fail the channel*:
+  * This is a forced close of the channel. Very early on (before
```
lightning-developer, line 136: Might be a contradiction to #942 ?

01-messaging.md
```diff
@@ -316,7 +316,11 @@ For simplicity of diagnosis, it's often useful to tell a peer that something is
    * [`u16`:`len`]
    * [`len*byte`:`data`]
 
-The 2-byte `len` field indicates the number of bytes in the immediately following field.
+1. type: 1 (`warning`)
+2. data:
+   * [`channel_id`:`channel_id`]
```
lightning-developer, line 321: Clarification: This assumes / includes `temporary_channel_id` during funding of the channel?

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
     - MUST use `temporary_channel_id` in lieu of `channel_id`.
 
 A sending node:
-  - when sending `error`:
-    - MUST fail the channel referred to by the error message.
   - SHOULD send `error` for protocol violations or internal errors that make channels unusable or that make further communication unusable.
   - SHOULD send `error` with the unknown `channel_id` in reply to messages of type `32`-`255` related to unknown channels.
+  - when sending `error`:
+    - MUST fail the channel(s) referred to by the error message.
```
lightning-developer, line 341: nit: referred to by the `error` ~~message~~ (pending on: #941 )

the message has only one `channel_id` field. BOLT 1 says if the  `id` is zero it refers to all channels between the peers. We ignore the situation where we have `n` channels and want to refer to `m` channels with 1 `error`?

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
...
+  - when sending `warning`:
+	- MAY set `channel_id` to all zero if the warning is not related to a specific channel.
```
lightning-developer, line 343: I guess you mean the `channel_id` in `warning` and not the `channel_id` of the channels between the peers?

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
...
+  - MAY close the connection after sending.
```
lightning-developer, line 344: Why would that not hold for `error`? I understood `error` to be the case that that a force close should trigger. I see good reasons that after publishing the commitment transaction a peer might want to close the connection.

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
...
-  - when `channel_id` is 0:
-    - MUST fail all channels with the receiving node.
-    - MUST close the connection.
-  - MUST set `len` equal to the length of `data`.
 
 The receiving node:
   - upon receiving `error`:
-    - MUST fail the channel referred to by the error message, if that channel is with the sending node.
-  - if no existing channel is referred to by the message:
+    - MUST fail the channel referred to by `channel_id`, if that channel is with the sending node.
```
lightning-developer, line 351: in the section for the sending node you wrote: 
> MUST fail the channel(s) referred to by the error message.

why is this only singular here?

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
...
+  - upon receiving `warning`:
+    - SHOULD log the message for later diagnosis.
```
lightning-developer, line 353: What does "log" mean? Keep in memory? Write to a log file? (IIRC we don't sepc logging in the bolts)

02-peer-protocol.md
```diff
@@ -484,7 +486,7 @@ A non-funding node (fundee):
   transaction after a timeout of 2016 blocks.
 
 From the point of waiting for `funding_locked` onward, either node MAY
-fail the channel if it does not receive a required response from the
+send an `error` and fail the channel if it does not receive a required response from the
```
lightning-developer, line 489: seems like a contradiction here to add that it MAY fail the channel as for what a node MUST do after sending `error` is to fail the channel (BOLT 01 line 341)

02-peer-protocol.md
lightning-developer: agree with extending the definition

02-peer-protocol.md
```diff
@@ -647,7 +650,8 @@ The sending node:
 The receiving node:
   - if the `signature` is not valid for either variant of closing transaction
   specified in [BOLT #3](03-transactions.md#closing-transaction) OR non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
-    - MUST fail the connection.
+    - MUST send a `warning` and close the connection, or send an
+      `error` and fail the channel.
```
lightning-developer, line 654: redundant "fail the channel"

02-peer-protocol.md
```diff
@@ -673,7 +677,8 @@ The receiving node:
           - MUST propose a `fee_satoshis` in the overlap between received and (about-to-be) sent `fee_range`.
   - otherwise, if `fee_satoshis` is not strictly between its last-sent `fee_satoshis`
   and its previously-received `fee_satoshis`, UNLESS it has since reconnected:
-    - SHOULD fail the connection.
+    - SHOULD send a `warning` and close the connection, or send an
+      `error` and fail the channel.
```
lightning-developer, line 681: redundant "fail the channel"

lightning-developer: ok that wording comes several times I stop pointing this out

lightning-developer (changes requested):
Concept ACK. 

I pointed out a few stylistic things and asked a few clarification questions. 

One thing that I realized towards the end there is often the phrase: 

> MUST send a `warning` and close the connection, or send an `error` and fail the channel.

I believe this is defined as things a node MUST do after sending `warning` or `error` so it seems very redundant to repeat this throughout the document.

01-messaging.md
cdecker: Correct

01-messaging.md
cdecker: That'd likely require multiple warning messages, one for each channel. Though it's hard to imagine a scenario where m < n and m > 1 channels need to close due to the same issue. The all-channel error was mostly intended to "there's something wrong but I can't identify the channel it's affecting" afaik, so "there's something wrong with one of these m channels" isn't much more useful IMHO.

01-messaging.md
```diff
@@ -331,24 +335,27 @@ The fundee node:
     - MUST use `temporary_channel_id` in lieu of `channel_id`.
 
 A sending node:
-  - when sending `error`:
-    - MUST fail the channel referred to by the error message.
   - SHOULD send `error` for protocol violations or internal errors that make channels unusable or that make further communication unusable.
   - SHOULD send `error` with the unknown `channel_id` in reply to messages of type `32`-`255` related to unknown channels.
+  - when sending `error`:
+    - MUST fail the channel(s) referred to by the error message.
+  - when sending `warning`:
```
TheBlueMatt, line 342: I thought we were restoring all-0s? Looks like its still removed here, and further down.

PR #836, opened 2021-01-25
Trampoline onion format (Feature 56/57)

t-bast:
Trampoline routing uses layered onions to trustlessly and privately offload the calculation of parts of a payment route to remote trampoline nodes.

A normal onion contains a smaller onion for the last hop of the route, and that smaller onion contains routing information about the next trampoline hop.

Intermediate trampoline nodes "fill the gap" by finding a route to the next trampoline node, and sending it the peeled trampoline onion, until that reaches the final destination.

This PR details the onion construction and requirements for supporting nodes. I advise readers to also have a look at https://github.com/lightningnetwork/lightning-rfc/pull/829 which gives a more high-level view of the different components, how they interact, and provides nice diagrams that help understand the low-level details.


04-onion-routing.md
```diff
@@ -367,9 +378,60 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+
+- If the invoice doesn't support the `trampoline_routing` feature:
+  - MUST NOT use trampoline routing to pay that invoice
+- MUST ensure that each hop in the `trampoline_onion_packet` supports `trampoline_routing`
+- MUST encrypt the `trampoline_onion_packet` with the same construction as `onion_packet`
+- MAY add trailing filler data similar to what is done in the `onion_packet`
+- MUST use a different `session_key` for the `trampoline_onion_packet` and the `onion_packet`
+- MUST include the `trampoline_onion_packet` tlv in the _last_ hop's payload of the `onion_packet`
+- MUST include the invoice's `payment_secret` in the _last_ hop's payload of the `trampoline_onion_packet`
+- MUST generate a different `payment_secret` to use in the outer onion
```
rustyrussell, line 413: Why include payment_secret at all in the outer onion?

04-onion-routing.md
```diff
@@ -367,9 +378,60 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+When processing a `trampoline_onion_packet`, a receiving node:
+
+- If it doesn't support `trampoline_routing`:
+  - MUST report a route failure to the origin node
+- Otherwise, if it supports `trampoline_routing`:
+  - MUST process the `trampoline_onion_packet` as an `onion_packet`
+  - MUST fail the HTLC if dictated by the requirements under [Failure Messages](#failure-messages)
+  - MUST compute a route to the next trampoline node
+  - MUST include the peeled `trampoline_onion_packet` in the `hop_payload` for the next trampoline node
+  - MAY aggregate the incoming multi-part payment before forwarding
```
rustyrussell, line 424: Oh, that's interesting!  Not sure we'll be implementing that for v1 ourselves though...

04-onion-routing.md
```diff
@@ -367,9 +378,60 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+  - If it uses a multi-part payment to forward to the next trampoline node:
+    - MUST generate a new `payment_secret` to use in the outer onion
```
rustyrussell, line 426: payment_secret is surely only used by the final recipient?  Nobody else can check it anyway, can they?

04-onion-routing.md
```diff
@@ -263,6 +264,15 @@ It is formatted according to the Type-Length-Value format defined in [BOLT #1](0
     2. data:
         * [`32*byte`:`payment_secret`]
         * [`tu64`:`total_msat`]
+    1. type: 10 (`outgoing_node_id`)
+    2. data:
+        * [`point`:`outgoing_node_id`]
```
rustyrussell, line 269: This addition seems orthogonal to the trampoline proposal (it's not wrong; scid is just a shortcut for spelling out this outgoing_node_id anyway)?  Not sure it's necessary though.

rustyrussell (commented):
I like it, but I'm not sure I understand the purpose of the payment_secret except inside the internal onion for the final node?

04-onion-routing.md
t-bast: It is really necessary to point to the next trampoline node. Imagine Alice chooses two trampoline nodes T1 and T2. She needs to tell T1 to forward to T2. They may not be directly connected, and even if they were, Alice doesn't necessarily have the graph information to find a matching scid. The natural way to target T2 is to include its `node_id`. Do you think it should be done differently?

04-onion-routing.md
t-bast: It's for the case where trampoline nodes aggregate incoming MPP and then re-split differently to reach the next trampoline node.
This ensures they use the normal MPP validation code and no intermediate nodes can cheat. They could just rely on the `total_amount` to verify they receive everything, but I like the fact that it works just like normal payments.

04-onion-routing.md
t-bast: Definitely something that can be done in a second step!

But since you'll be calculating routes anyway even to relay single-part HTLCs, you'll see that it won't be much additional work to aggregate all the incoming parts before relaying.

One point that was a bit hard to get right in eclair is the HTLC reconciliation after a restart (if for example the node was restarted before an HTLC was relayed out), because this breaks the mapping "1 HTLC out <-> 1 HTLC in", it's now "M HTLCs out <-> N HTLCs in".

04-onion-routing.md
t-bast: The invoice one is put in the trampoline onion to be read only by the final recipient.
But when you're constructing a new MPP set between trampoline T1 and T2, it makes sense to use a `payment_secret` for that MPP flow, and it must be generated by `T1` so it can only be in the outer onion.

04-onion-routing.md
rustyrussell: Ah, you need total_msat from payment_data.  But payment_secret here doesn't help much: the aggregating node can't know what the correct value is (and I can't see where in the spec you say it should be checked...).

An intermediate can probe by sending its own partial payment with some random payment_secret, so you can't really even say "they must all be the same...".

t-bast: It's true that in the invoice-based flow the recipient generated the `payment_secret` and can check the correctness of the value. In the trampoline flow, it's the trampoline sender who generates one, and the recipient blindly accepts and just uses it to bundle together HTLCs from the same set.

But I think it still has some value because it does ensure that HTLCs from two different nodes will not end up being bundled together (and mess up the payment). If we have the following payment (with trampoline route `Alice -> T1 -> T2 -> Bob`):

```
Alice -----> ... -----> T1 -----> I1 -----> I2 -----> T2 -----> ... -----> Bob
```

`T1` generated a `payment_secret` to send HTLCs to `T2`. `I2` is trying to interfere by generating his own (trampoline) payment to `T2`. They will have a different `payment_secret`, so they won't interfere with the ones `T1` sent.

04-onion-routing.md
ecdsa: If the receiving nodes aggregates the MPP, then I think there should be a requirement to fail the payment if the `trampoline_onion_packet` is not the same in every part of the MPP. I recently found that Eclair is not doing that.

I think this also implies that the decision to aggregate the payment should not be left with the trampoline node. If the receiver of the MPP is the trampoline node, I do not see how it could decide not to aggregate the payment.

t-bast: > I think there should be a requirement to fail the payment if the trampoline_onion_packet is not the same in every part of the MPP

Yes that would make sense to add an explicit requirement here.
For now eclair is quite lax here (which lets the final recipient decide whether the payment is valid or not), but that flexibility may not bring anything to the table, so it would make sense to reject that kind of HTLCs, I'll look into it.

> I think this also implies that the decision to aggregate the payment should not be left with the trampoline node.

I think it does, as we cannot force an implementation to do something if it's not related to security.
A trampoline node may chose to simply forward each HTLC independently without aggregating them if it doesn't have the resources to aggregate (which requires allocating resources to hodl parts until the complete payment arrives).
This is a different requirement from the case where the trampoline node is the final recipient.

ecdsa: Oh I see. That distinction was missing, I think it should be clarified:
 - if the trampoline node is the final recipient of the MPP, it must check that `trampoline_onion_packets` in each part are identical before forwarding
 - if the trampoline node is not the final recipient of the MPP, it may aggregate HTLCs before forwarding, but what does "aggregate" mean in that context? simply to wait? what if the sender is using another trampoline for another part of the payment?

t-bast: > it may aggregate HTLCs before forwarding, but what does "aggregate" mean in that context? simply to wait? what if the sender is using another trampoline for another part of the payment?

Actually the spec is quite confusing right now, my bad, I tried to keep this part concise and kinda messed it up.

If the trampoline node receives an HTLC which is a part of an MPP (because `total_amount_msat` > `htlc_amount` in the *outer* onion), it **MUST** aggregate it (ie wait to receive all the parts that he's supposed to receive) before forwarding. Otherwise we would not be able to forward the requested outgoing amount, because he could have received less than that!

An example will help clarify this. Alice sends a payment to Carol through Bob (which is a trampoline node).
Alice decides to send `10,000 sats` through Bob (and may send more sats for the same payment through another trampoline node). But Alice doesn't have a route to Bob that can carry `10,000 sats` in a single HTLC, so she splits it in two HTLCs (for example `4,000 sats` and `6,500 sats` for a total of `10,500 sats` because Bob takes a `500 sats` relay fee).

When Bob receives the first HTLC, the outer onion tells it that the `total_amount = 10,500 sats`. The inner onion tells it to relay `10,000 sat` to Carol. So Bob must wait to receive the full `10,500 sats` before forwarding the payment to Carol. That's what I call "aggregate" here.

If the sender doesn't want the trampoline node to aggregate the payments, the sender must set each HTLC's `total_amount` to equal the `htlc_amount`, and must use a different inner onion for each htlc (with a different `outgoing_amount` for each HTLC). This will likely cost more fees and be less efficient, but is supported.

ecdsa: Yes, that explanation is clear. By "the trampoline is the final recipient of the MPP", we both mean `total_amount_msat > htlc_amount` in the outer onion. That case is indeed very clear: the trampoline must aggregate. My question was about the case where the trampoline is not the recipient of the MPP. I do not think that "aggregate" makes sense in that case.

In other words, "MAY aggregate" is confusing, because it gives the feeling that the decision to aggregate is with the trampoline node.

t-bast: > My question was about the case where the trampoline is not the recipient of the MPP. I do not think that "aggregate" makes sense in that case.

You're right, in that case the trampoline even must **not** aggregate because it doesn't know how much it should even receive before forwarding. As you say, that decision is indeed instead made by the sender (not the trampoline node) depending on what `total_amount_msat` it puts in the outer onion.

04-onion-routing.md
```diff
@@ -1014,6 +1093,28 @@ the decrypted byte stream.
...
 reasonable time.
 
+1. type: NODE|24 (`temporary_trampoline_failure`)
+
+The trampoline node was unable to relay the payment to the next trampoline
+node, but may be able to handle it, or others, later.
+This error usually indicates that routes were found but failed because of
+temporary failures at intermediate hops.
+
+1. type: NODE|25 (`trampoline_fee_expiry_insufficient`)
```
lightning-developer, line 1103: Do we have a recommendation that should be used initially by the sender for CLTV and fee budgets? 

I understand from your comment in reply to @ecdsa at https://github.com/lightning/bolts/pull/829#issuecomment-756868196 that you suggest to have a trial and error mechanism with feedback from the trampoline node. I am worried if more than 1 trampoline node is involved that this might create a lot of unnecessary round trips as the sender would have to learn the fee and cltv budget for the first trampoline and then for the second one and so one. In the meantime the trampoline onion would always be cancelled in such roundtrips. 

Maybe a recommendation for the sender to take the median CLTV for paths of up to `n` hops lenghts and a fee budget for (multi)paths accordingly to start with might be helpful? I understand that especially the fee estimation might be tricky but I would prefer to start somewhere. 

I also wonder if it makes sense for trampoline nodes to replicate the `base_fee_msat` and `fee_proportional_millionths` mechanism as trampolines have a total fee budget that they will allocate to deliver the payment and earn from what they have not used. Or did you plan that if a trampoline pays too much total `base_fee_msat` but stays far below the `fee_poportional_millionths` that you would send `temporary_trampoline_failure`?

lightning-developer (changes requested):
I believe the spec should make a recommendation for CLTV and fee budgets that senders of trampoline onions should use for each trampoline to reduce the usage of `temporary_trampoline_failure`. 

Also it seems reasonable to just reply with a single `fee_budget_msat` instead of    * [`u32`:`fee_base_msat`] and [`u32`:`fee_proportional_millionths`] as those fields seem not to make sense in the context of trampolines. 

Trampolines could announce the `fee_budget_msat` and `cltv_budget`  (in case you like the new name) that they believe to be sufficient in their node announcements?

04-onion-routing.md
t-bast: > I believe the spec should make a recommendation for CLTV and fee budgets that senders of trampoline onions should use for each trampoline to reduce the usage of temporary_trampoline_failure.

The spec cannot do that, this is entirely up to node operators and implementations and we must let the market decide on good values here (which will likely change more often than the specification does).

> Also it seems reasonable to just reply with a single fee_budget_msat instead of * [u32:fee_base_msat] and [u32:fee_proportional_millionths] as those fields seem not to make sense in the context of trampolines.

I don't see why they don't make sense? Base and proportional fees do apply to trampoline nodes, and it's more consistent that way. Trampoline nodes derive values for base and proportional fees by walking the graph outwards and adding the values of individual edges.

Overall your whole comment is addressed by what will come in a second step. My first proposal for trampoline (more than 2 years ago, see #654) did contain a mechanism to gossip trampoline fees and cltv. I decided to drop this part for now, because the most common way of using trampoline doesn't really need it, so it can be added later.

The simplest path to trampoline is to have the recipient include trampoline nodes in their invoices. It's the recipient's responsibility to compute the fees that each trampoline node will need to reach them (which is easy to do by just looking at a small subset of the graph) and include those fees in the invoice. Then the sender will either:

- directly send to those trampoline nodes (if the sender is able to compute a route to them): in that case there should never be a trampoline fee / cltv failure
- pick a first trampoline node close to them: in that case they do need to guess what fees and cltv they should give that first trampoline node to allow it to reach the second trampoline node. This may require a retry, but it requires only one, which is efficient enough (the second trampoline node should never require a fee / cltv update because the recipient computed the correct values)

My goal is to work on the gossip part once this first PR is accepted, and after getting some real-world feedback on what works well and what doesn't work in practice.

lightning-developer: > The spec cannot do that, this is entirely up to node operators and implementations and we must let the market decide on good values here (which will likely change more often than the specification does).

What I meant was the idea if for example the cheapest path between two trampolines is of 3 hops the sender using these two trampoline nodes should include at least the fees that would be charged on such a path. Similarly with the `cltv delay` which MUST at least meet that of the 3 channels. While this is a perfectly reasonable recommendation I just realized that it kind of defeats the purpose of Trampolines as the sender node wants to outsource the route computation. Also trampolines might utilize unannounced channels. So yes I agree it does not make sense to add a recommendation to the spec.

> > Also it seems reasonable to just reply with a single fee_budget_msat instead of * [u32:fee_base_msat] and [u32:fee_proportional_millionths] as those fields seem not to make sense in the context of trampolines.
> 
> I don't see why they don't make sense? Base and proportional fees do apply to trampoline nodes, and it's more consistent that way. Trampoline nodes derive values for base and proportional fees by walking the graph outwards and adding the values of individual edges.

Lets say I want to pay 200k sats and I add 50 sats `base_fee` and a feerate of `1000`. This means that the trampoline node can charge a total of `250` sats. What happens if the trampoline node delivers the payment on channels charging a total of `220` sats for the fee rate and `20` sats for their `base_fee`? I think the trampoline node would have to pay a total of `240` sats and stay below the total `fee_budget`. It did however spend more on the `fee_rate`. I think in such a situation the node should process the payment and be happy it earned `10` sats. This is similar to forwarding of regular onions where the difference of outgoing HTLC and incoming HTLC can be considered a fee budget for the routing node and needs to be smaller or equal to the result of the fee formular with `fee_rate` and `base_fee`.

If the above goal to forward the payment is the case, then why requiring two separate fees? It seems the final question is only whether the trampoline was able to deliver to the next trampoline while staying within the total fee budget. Also the base fee might complicate things in case the payment amount is larger and the trampoline node has to use MPP. If there was just a single fee rate for the trampoline and the underlying channels the allocation seems much more straight forward as one would effectively have a unit cost but I guess this is more a technical detail.

> 
> Overall your whole comment is addressed by what will come in a second step. My first proposal for trampoline (more than 2 years ago, see #654) did contain a mechanism to gossip trampoline fees and cltv. I decided to drop this part for now, because the most common way of using trampoline doesn't really need it, so it can be added later.
> [...]
> My goal is to work on the gossip part once this first PR is accepted, and after getting some real-world feedback on what works well and what doesn't work in practice.

It seems to me that we will end up having a gossip for trampolines at the end anyway which is why I elaborated above on the `fee` and `cltv` budget. I understand why you want to do the upgrade in smaller steps. I will checkout out #654 and come back to the gossip questions.

thomash-acinq: Regarding the base and proportional fee, I agree with @lightning-developer that it doesn't make sense to provide both. It makes sense to have both when we don't know the amount that will be paid and want to provide a formula that works for any amount, but here we're in the context of a specific payment, we know the amount and we can compute the exact fee.

t-bast: The reason I'm currently providing both is explicitly for that: because it should be used for future payments of unknown amounts.

We don't have a gossip broadcast mechanism yet, but since we need to have a failure message, I believe it makes sense to make it compatible with future gossip. Whenever a payer receives such an error, they should store `fee-base` and `fee-proportional`, and use those values for their future payments through that trampoline node. It may be outdated by then (because we don't have yet a mechanism to receive new fees, but we will in the future), but it may still be accurate and win us a round-trip.

thomash-acinq: That would only work if the trampoline node uses the same fee structure regardless of the path used. I don't think it makes sense to advertise trampoline fees before we know what route to use, what if the best route is more expensive than the fee we've advertised? Do we refuse to route the payment? Do we increase the fees for everyone to make this case less frequent?

04-onion-routing.md
```diff
@@ -275,10 +285,10 @@ The writer:
   - For every node:
     - MUST include `amt_to_forward` and `outgoing_cltv_value`.
   - For every non-final node:
-    - MUST include `short_channel_id`
+    - MUST include `short_channel_id` or `outgoing_node_id`
```
valentinewallace, line 288: LDK's current plan is to support only scids for regular forwards and only node ids for trampoline forwards. I think this is what Acinq does as well. Maybe this should be codified in the spec, if that ends up being everyone's plan? :)

t-bast: Definitely! I'll need to do a big clean-up on this PR, now that LDK is working on it it may be the right time ;)

09-features.md
```diff
@@ -45,6 +45,7 @@ The Context column decodes as follows:
 | 46/47 | `option_scid_alias`              | Supply channel aliases for routing                        | IN       |                   | [BOLT #2][bolt02-channel-ready]   |
 | 48/49 | `option_payment_metadata` | Payment metadata in tlv record | 9 | | [BOLT #11](11-payment-encoding.md#tagged-fields)
 | 50/51 | `option_zeroconf`                | Understands zeroconf channel types                        | IN       | `option_scid_alias` | [BOLT #2][bolt02-channel-ready]   |
+| 56/57 | `trampoline_routing`              | This node supports trampoline routing                     | IN9      | `basic_mpp`               | [BOLT #4](bolt04-trampoline)                                          |
```
valentinewallace, line 48: Will need to add this feature bit for BOLT 12 invoices, and adapt the spec for the lack of payment secret in them :)

t-bast: Indeed! That should be quite easy though, since there are two "layers" of payment secrets:

- the real payment secret that comes from the invoice and is used by the sender in the trampoline onion: this will simply be replaced by usage of a blinded path
- the payment secrets generated by each trampoline node for the payment to the next trampoline node (in the _outer_ onion): those will stay, since intermediate trampoline nodes don't need to know that blinded paths are being used, except maybe for the last trampoline node before the recipient

I should also describe how blinded paths can be used on the trampoline route, which is very interesting when paying public nodes. I'll wait for the blinded path PR to be merged though before doing all that!

04-onion-routing.md
t-bast: I've clarified this in https://github.com/lightning/bolts/pull/836/commits/985cf69e25d81174545048bd70790b8a33605118

bolt04/trampoline-payment-onion-test.json
```diff
@@ -0,0 +1,117 @@
...
+        "trampoline_session_key": "0303030303030303030303030303030303030303030303030303030303030303",
+        "trampoline_onion": "0002531fe6068134503d2723133227c867ac8fa6c83c537e9a44c3c5bdbdcb1fe3371860c0749bfd613056cfc5718beecc25a2f255fc7abbea3cd75ff820e9d30807d19b30f33626452fa54bb2d822e918558ed3e6714deb3f9a2a10895e7553c6f088c9a852043530dbc9abcc486030894364b205f5de60171b451ff462664ebce23b672579bf2a444ebfe0a81875c26d2fa16d426795b9b02ccbc4bdf909c583f0c2ebe9136510645917153ecb05181ca0c1b207824578ee841804a148f4c3df7306dcea52d94222907c9187bc31c0880fc084f0d88716e195c0abe7672d15217623",
+        "hops": [
+            {
+                "alias": "Bob",
+                "node_id": "0324653eac434488002cc06bbfb7f10fe18991e35f9fe4302dbea6d2353dc0ab1c",
+                "payload": {
+                    "amt_to_forward": 100005000,
+                    "outgoing_cltv": 800250,
+                    "short_channel_id": "0x0x1105",
```
t-bast, line 38: Note that I'm using the "standard human readable format" for scids here, defined here: https://github.com/lightning/bolts/blob/master/07-routing-gossip.md#definition-of-short_channel_id

But I'm not entirely sure that this is people are using in practice and comfortable with? Let me know if that's unclear, in which case I can either use the numeric value or its hex encoding (tell me what you think is best).

09-features.md
t-bast: I have added details about the trampoline-to-blinded paths case in https://github.com/lightning/bolts/pull/836/commits/7589dd8e896eb2e3d206a05abb8a8c654057b693

Since Bolt 12 is scattered across a few PRs and pending rebases, it's a bit messy to reference the right TLV types: it will need some clean-up once Bolt 12 is merged. But the test vector should let you verify the exact encoding!

t-bast:
@arik-so @valentinewallace I have added a draft spec of trampoline payments to blinded paths in https://github.com/lightning/bolts/pull/836/commits/296ce2114a54487030389033b0c019dd32db01b6. It is probably a bit confusing because it depends on Bolt 12 types that are defined in the offers PR, and I should more explicitly spell out the requirement of which tlvs are included where, but combined with the discussions we had in [this comment](https://github.com/lightning/bolts/pull/829#discussion_r1591173093), we should be able to understand each other.

I'm particularly looking for feedback on the shared secret extension I'm using for the recipient trampoline payload to include an ECDH with the `invoice_request`'s `payer_id`, let me know what you think (see [@TheBlueMatt's comment](https://github.com/lightning/bolts/pull/829#discussion_r1692171969) for more context).

Once we have a good enough rough consensus, I'll spend some time rebasing this PR, I'll re-write the requirements in a way that is similar to what was done in #1181 and more precise that what I've currently done, and I'll finalize the test vector.

t-bast:
@arik-so I slightly changed the last commit:

- I'm now providing complete test vectors for trampoline to blinded paths (with and without a trampoline onion for the recipient)
- I changed the `invoice_request` shared secret mechanism: instead of modifying the derivation of `rho` and `mu`, I use this shared secret in the HMAC's associated data for the final hop: this is simpler and avoids changing the internals of `Sphinx`

t-bast:
There is one point that is worth discussing here: do we want to include an extension for Bolt 11 to introduce trampoline routing hints? If we want to be able to make non-blinded trampoline payments between wallets that have only private channels (e.g. Alice -> LSPA -> ... -> LSPB -> Bob), I see two possible options.

The first one is that we don't introduce any new routing hint, but if the invoice contains the (optional) trampoline feature bit, Alice _assumes_ that nodes included in Bob's routing hints support trampoline and uses them as such. In the example above, Alice would use LSPA as her trampoline node and LSPB as Bob's trampoline node, and assume that LSPB is able to route to Bob. This option is nice because it doesn't increase QR code size, but when receiving wallets provide multi-hop routing hints, Alice may get that heuristic wrong. In practice I don't think this is an issue, in the worst case Alice could make several attempts where she assumes that different nodes in the routing hints support trampoline, until one works.

The second option is that we introduce a trampoline routing hint field to Bolt 11 invoices, that are very similar to routing hints but don't include a `short_channel_id`. This will be ignored by existing nodes that don't support trampoline. The issue with that option is that it increases QR code size, because the invoice must work for non-trampoline senders as well, so it will include both the normal routing hints and the trampoline hints.

Thoughts?

04-onion-routing.md
```diff
@@ -208,12 +209,32 @@ This is formatted according to the Type-Length-Value format defined in [BOLT #1]
     1. type: 12 (`current_path_key`)
     2. data:
         * [`point`:`path_key`]
+    1. type: 14 (`outgoing_node_id`)
```
joostjager, line 212: Linking https://github.com/lightning/bolts/pull/814 again. Perhaps the intermediate step helps making this PR easier to get merged?

04-onion-routing.md
```diff
@@ -1068,6 +1178,21 @@ key, and computes the HMAC, using each hop's `um` key.
 The origin node can detect the sender of the error message by matching the
 `hmac` field with the computed HMAC.
 
+For trampoline payments, the flow is the same. Intermediate trampoline nodes
+first decrypt the downstream error using the `ammag` and `um` keys for their
+forward path. If the error comes from an intermediate node _before_ the next
+trampoline node, they may replace it with their own error for the origin node,
```
joostjager, line 1184: Checking compatibility with attributable failures. This case where the error comes from an intermediate node before the next trampoline node seems fine? The intermediate trampoline decrypts the attributable failure and creates its own with itself as the failure source.

Do trampoline nodes worry about penalties applied to them for failures on the path that they used towards the next trampoline node?

04-onion-routing.md
```diff
@@ -1068,6 +1178,21 @@ key, and computes the HMAC, using each hop's `um` key.
...
+otherwise they must encrypt on top of the next trampoline node's error.
```
joostjager, line 1185: Encrypt on top for attributable failures would I think mean that they'd just process it as usual - shifting hmacs and adding their own?

I am not sure what the intermediate nodes between the two trampolines would do though. Would that require two `attribution_data` blocks?

04-onion-routing.md
```diff
@@ -1295,6 +1420,28 @@ reasonable time.
...
 An error occurred within the blinded path.
 
+1. type: NODE|25 (`temporary_trampoline_failure`)
+
+The trampoline node was unable to relay the payment to the next trampoline
+node, but may be able to handle it, or others, later.
+This error usually indicates that routes were found but failed because of
+temporary failures at intermediate hops.
+
+1. type: NODE|26 (`trampoline_fee_or_expiry_insufficient`)
```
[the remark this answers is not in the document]
joostjager: Is it really that more additional code to use onion messages? I'd think that from a protocol perspective choosing the most efficient option in terms of liquidity (no htlc lock up for the fee probe) and fees (no guessing at the budget) is best.

Choosing a budget for the trampoline node as a sender sounds like a difficult problem also. How does that work in Phoenix?

04-onion-routing.md
joostjager: I've been reading up on trampoline more. It indeed seems that for attributable failures, you'd have two blocks of attribution data. The inner block can possibly be stored together with the failure message?

04-onion-routing.md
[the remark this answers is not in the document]
joostjager: Ah yes that works well. Then you also never have to return `trampoline_fee_or_expiry_insufficient`.

But if the goal is to not over-specify and keep it simple initially, why include in the protocol multiple trampoline routes for example with the wrapped wrapped failures etc?

04-onion-routing.md
joostjager: Easier in that things can be implemented and deployed in stages. First do pubkey routing and then continue with trampoline building on top of that.

04-onion-routing.md
joostjager: That makes sense

04-onion-routing.md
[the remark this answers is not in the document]
joostjager: It seemed to me that with a single trampoline node, you don't need to wrap errors? The trampoline node tries to complete the payment and if it fails, it returns a regular failure to the sender without the double wrapping?

04-onion-routing.md
joostjager: I just noticed the subset and that it could be an intermediate step. But not too certain that it is useful either.

The trampoline spec, does that include pubkey routing between non-trampoline routes too though?

04-onion-routing.md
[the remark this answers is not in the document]
joostjager: Ah right. I was thinking of the failure cases where the T node creates a 'outside' failure. But indeed, a final failure still needs to be wrapped back. Ok, thanks.

04-onion-routing.md
joostjager: This PR adds tlv 14 `outgoing_node_id` to the spec. Does that mean that pubkey routing can then also be used in regular non-trampoline forwards going through the node implementing this spec?

joostjager: No further comments. You are right :)

04-onion-routing.md
```diff
@@ -436,6 +458,94 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+  - MUST process the `trampoline_onion_packet` as an `onion_packet`.
+  - MUST fail the HTLC if dictated by the requirements under [Failure Messages](#failure-messages).
+  - If it is not the final node:
+    - If the incoming payment is a multi-part payment:
+      - MUST wait to receive all the payment parts before forwarding.
+    - If `encrypted_recipient_data` is included:
+      - MUST use `current_path_key` from the trampoline onion or the outer onion to decrypt it.
+      - MUST validate its content as it would for the non-trampoline case.
+      - MUST include the next `current_path_key` in the `hop_payload` for the next trampoline node.
+    - MUST compute a route to the next trampoline node.
```
joostjager, line 534: Are there any thoughts on how to protect against DoS here? Also wondering what the most effective way would be for an attacker to overload trampoline nodes. I suppose a max length route sending the htlc in circles with each hop needing to search for a hard to find path would be a component of it.

It seems to be a channel jamming like problem because there's again no info on sender or receiver to filter out / limit trampoline activitiy.

04-onion-routing.md
joostjager: How we implemented this in LDK is to just drop the attribution data if it cannot be interpreted by the sender.

thomash-acinq: I don't think you need a second block of attribution data, there is only one failure packet and there should be only one attribution data to go with it.
The process for the attribution data is the same as for the failure packet:
- If the error comes from an intermediate node that you've chosen and not from a trampoline node: You should maybe retry and if it doesn't succeed, create a failure yourself, you are the failing trampoline node.
- If the error comes from a trampoline node: You unwrap the failure packet and the attribution data until you reach the next trampoline node and then wrap it with both shared secrets (the one from the sender and the one from the previous trampoline node).

joostjager: The second case indeed, trampoline node A receiving an error from a next trampoline node B. The attribution data that goes along with that isn't interpretable by the sender, so I think there is nothing to do other than drop it? And then let trampoline node A instantiate a new attr data block for the sender, to at least get attribution for the remaining downstream nodes. Not 100% sure if we are talking about the same thing, because you mentioned 'wrap it with both shared secrets'. You can't do that with attribution data?

thomash-acinq: "wrap it with both shared secrets" it what the spec requires for the failure packet and we can do the same to the attribution data for consistency. Why do you think we can't do that? It's just xoring with a pseudo-random stream.

thomash-acinq: If it makes it easier, you can just add a virtual hop from the trampoline node to itself. The trampoline node would populate the attribution data twice, once with each shared secret.

joostjager: We can xor the attribution data for sure, but unlike the failure message, which is ultimately readable by the sender, attribution data produced by 'inner route' nodes isn't. That attribution data is useless to begin with (to the sender), and no point in trying to wrap it so that the sender can read it. Unless I am missing something?

joostjager: Ah got it now, thanks. My thought was that once these inner route nodes start processing the attribution data, it becomes unusable. But as you point out, the trampoline node just shifts that the other way again to 'recover' the attribution data as it was passed back by the next trampoline node.

04-onion-routing.md
joostjager: I am wondering what happens if a node doesn't filter for its own clients and just offers trampoline routing to anyone. Would it be advisable at all to enable this, or is this asking for trouble without having the heuristics in place too?

I wouldn't even know how to do these heuristics.

04-onion-routing.md
```diff
@@ -436,6 +458,94 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+
+- If it doesn't support `trampoline_routing`:
+  - MUST report a route failure to the origin node.
+- Otherwise, if it supports `trampoline_routing`:
+  - MUST process the `trampoline_onion_packet` as an `onion_packet`.
+  - MUST fail the HTLC if dictated by the requirements under [Failure Messages](#failure-messages).
+  - If it is not the final node:
+    - If the incoming payment is a multi-part payment:
+      - MUST wait to receive all the payment parts before forwarding.
+    - If `encrypted_recipient_data` is included:
```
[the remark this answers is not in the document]
carlaKC: lgtm!

04-onion-routing.md
```diff
@@ -1068,6 +1189,21 @@ key, and computes the HMAC, using each hop's `um` key.
 The origin node can detect the sender of the error message by matching the
 `hmac` field with the computed HMAC.
 
+For trampoline payments, the flow is the same. Intermediate trampoline nodes
+first decrypt the downstream error using the `ammag` and `um` keys for their
+forward path. If the error comes from an intermediate node _before_ the next
+trampoline node, they may replace it with their own error for the origin node,
+otherwise they must encrypt on top of the next trampoline node's error.
+
+Intermediate trampoline hops apply the obfuscation step twice: first with the
```
a-mpch: It is no clear to me that whenever the receiver is the first trampoline and something fails, should the receiver node just use the outer onion shared secret or would it use the double obfuscation?
It could be seen in different perspectives:
1. payment reaches the trampoline and then try to route it, figuring out it is itself.
2. Before trying to route it as trampoline, payment is received.

I didn't find a test case for this in eclair (don't know much of Scala tho)

t-bast: You're right, there is some subtlety here, this depends on who the error is encrypted for:

- if the failing trampoline node wants the error to be for the previous trampoline node, they use the outer onion shared secret: this is used for example when the payment is using MPP but not all parts were received in time, which is an error made by the previous trampoline node which could (probably should) retry, so we don't want the error to go all the way back to the sender
- if the failing trampoline node wants the error to be for the sender, they use the double obfuscation, which forces the previous trampoline node to relay the error upstream: this is used for example when receiving a payment that an invoice that was already paid, or that has a trampoline expiry that is too close

I don't think we can list all cases in the spec, but we probably should better explain the distinction between those two cases. Is that clear enough? Otherwise feel free to suggest the wording you'd like to see in the spec that would make this clear.

a-mpch: I think that wording makes sense. Thanks!

04-onion-routing.md
```diff
@@ -436,6 +458,100 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+    - If `encrypted_recipient_data` is included:
+      - If `current_path_key` is not set in the trampoline onion or the outer onion:
+        - MUST reject the payment.
+      - If `current_path_key` is set in both the trampoline onion and the outer onion:
+        - MUST reject the payment.
+      - Otherwise:
+        - MUST use `current_path_key` from the trampoline onion (when it is the introduction node
+          of the blinded path) or the outer onion (when it is an intermediate node) to decrypt it.
+        - MUST validate its content as it would for the non-trampoline case.
+        - MUST include the next `current_path_key` in the `hop_payload` for the next trampoline node.
```
carlaKC, line 554: The use of `current_path_key` in trampoline + blinded paths doesn't quite match our error handling instructions.

For trampoline:
* Introduction node gets `current_path_key` in trampoline onion
* Relaying node gets `current_path_key` in regular onion

Our existing error handling in bolt-02:
* Introduction node (non-receiving): should send `update_fail_htlc` because `current_path_key` is in onion.
* Relaying node should send `update_malformed_htlc` because `current_path_key` is in `update_add_htlc`.

With the current wording, we'll always return `update_fail_htlc` because we get our `current_path_key` from the onion (regular or trampoline) and never in `update_add_htlc`.  

I'm not sure whether there's a probing issue here, but we don't achieve the same "blinded tunnel / always fails at introduction" effect that we do with regular blinded payments, as the sender can now learn which blinded trampoline its payment failed at.

t-bast: Good catch, thanks for pointing it out! Here is my current thinking about what we should do:

- when failing at an intermediate trampoline node (inside the blinded path), I don't think we should send `update_fail_malformed_htlc`: if the previous node was *not* a trampoline node (because the previous trampoline node inserted a "normal" path to reach us), they are unaware that they are sitting between two blinded trampoline nodes, and we'd be giving away information by sending them `update_fail_malformed_htlc` instead of `update_fail_htlc`
- so trampoline nodes inside a blinded path should always use `update_fail_htlc` with `invalid_onion_blinding`
- BUT the first trampoline node (the introduction node) should ignore the downstream error and return its own error (instead of re-wrapping the downstream trampoline error onion): it is best effort, but if the introduction point is honest then the sender will not learn where the payment failed inside the blinded path

How does that sound?

carlaKC: Seems reasonable to me 👍

Assuming that we'd also return regular errors if we're the introduction trampoline and the final node?
(really don't see the point of a blinded trampoline to yourself, but to match the current spec)

t-bast: > Assuming that we'd also return regular errors if we're the introduction trampoline and the final node?
(really don't see the point of a blinded trampoline to yourself, but to match the current spec)

Yes, that sounds good :+1:

t-bast: Clarified in https://github.com/lightning/bolts/pull/836/commits/644b82f1b931a1e1dfca449f30eaf661822a85ef. I think the introduction node can even return whatever error message it wants (even if it is not the final node), it won't leak anything from the blinded path itself.

04-onion-routing.md
t-bast: Clarified in https://github.com/lightning/bolts/pull/836/commits/a38fdfb99ddba87bc9c14793ee001714b76bc8d8

04-onion-routing.md
```diff
@@ -436,6 +458,108 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+  - If `recipient_blinded_paths` is included:
+    - MUST forward the payment using the blinded paths provided.
+    - MAY use features included in `recipient_features`.
+  - If it is the final node:
+    - MUST reject the payment if:
+      - The outer onion's `outgoing_cltv_value` is smaller than the trampoline onion's `outgoing_cltv_value`.
+      - If this is a multi-part payment:
+        - The outer onion's `total_msat` is smaller than the trampoline onion's `amt_to_forward`.
+      - Otherwise:
+        - The outer onion's `amt_to_forward` is smaller than the trampoline onion's `amt_to_forward`.
```
SomberNight: Why are these checks conditioned on `If it is the final node`?
Shouldn't Trampoline Forwarders enforce the same checks?

t-bast: It may be a bit confusing indeed. I wanted to make sure that those requirements were explicitly there for the final node case, to ensure that recipients correctly validate values across both the outer and inner onion.

But as you noted, this must (of course) be enforced by intermediate nodes as well, otherwise they would be relaying more than what they receive, which means they would be losing funds! I'll clarify this while rebasing to fix conflicts.

t-bast:
Rebased to fix conflicts (BOLT 4 had seen quite a lot of changes with attribution data and fulfillment payload).

04-onion-routing.md
t-bast: I've clarified this in https://github.com/lightning/bolts/pull/836/commits/99f561533e226256365a76e14da7ecc2d3012881. Note that I also simplified the "if MPP" statement by assuming that we'll merge #1355 before trampoline.

04-onion-routing.md
```diff
@@ -439,6 +461,112 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+        - MUST send `update_fail_htlc` with `invalid_onion_blinding` instead.
+  - If it is not the final node:
+    - If the incoming payment is a multi-part payment:
+      - MUST wait to receive all the payment parts before forwarding.
+    - If the outer onion's `outgoing_cltv_value` is smaller than the trampoline onion's `outgoing_cltv_value`:
+      - MUST reject the payment.
+    - If the outer onion's `total_msat` is smaller than the trampoline onion's `amt_to_forward`:
+      - MUST reject the payment.
+    - If the outer onion's `amt_to_forward` is smaller than the trampoline onion's `amt_to_forward`:
+      - MUST reject the payment.
```
SomberNight: The last check is incorrect. In case of MPP, the outer onion's `amt_to_forward` can be smaller than the inner's. See https://github.com/lightning/bolts/blob/bc7a1a0bc97b2293e7f43dd8a06529e5fdcf7cd2/proposals/trampoline.md?plain=1#L612

The check should be:
```
    - If the outer onion's `total_msat` is smaller than the trampoline onion's `amt_to_forward`:
      - MUST reject the payment.
```

This is hard to get right.

(EDIT: same for the final node)

SomberNight: Oh wait, you have my suggested check on the line above.
Well, soooo, I think the (amount) checks should be:
```
assert outer_onion.total_msat >= outer_onion.amt_to_forward
assert outer_onion.total_msat >= inner_onion.amt_to_forward
```

t-bast: Right, actually removing the "if" by assuming #1355 was incorrect. I just removed one of the branches, which actually becomes obsolete afte #1355 (I've amended the last commit for that).

PR #839, opened 2021-01-29
Add funding expiry to channel open

t-bast:
Routing nodes have an incentive to use low fees when opening channels to ensure their activity is economically viable.

However, when a funding transaction takes too long to confirm, the fundee may have forgotten the channel. In that case the funder is forced to broadcast the first commit tx to get his funds back and then open a new channel, which is costly.

We can avoid this issue by having an explicit commitment to a block before which the funding tx will be confirmed. The fundee will keep the channel around for that duration, and the funder needs to ensure the funding tx confirms before that deadline (using CPFP on a change output for example). This gives clearer guarantees for both the funder and fundee.

As discussed during the spec meeting, I update this PR to a simple version where the timeout is 2016 blocks. We can later add another improvement suggested by @cdecker: if a fundee forgets about a channel, she could send a signed mutual close to the funder to help him recover his funds more efficiently. That should be added in a separate PR.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -351,6 +351,7 @@ The sender MUST set:
 The sender:
   - when creating the funding transaction:
     - SHOULD use only BIP141 (Segregated Witness) inputs.
+    - SHOULD ensure the funding transaction confirms in the next 2016 blocks.
 
 The recipient:
   - if `signature` is incorrect OR non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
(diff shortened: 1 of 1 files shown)


viaj3ro:
This is very much needed in an environment where fees can stay high for long periods of time. I just had 4 channel openings "expire" due to confirmations took more than 10 days and now have no choice but to force close and reopen.

cdecker:
> As discussed during the spec meeting, I update this PR to a simple version where the timeout is 2016 blocks. We can later add another improvement suggested by @cdecker: if a fundee forgets about a channel, she could send a signed mutual close to the funder to help him recover his funds more efficiently. That should be added in a separate PR.

Let's call that option "so long and thanks for all the sigs"-mode 😋

t-bast:
Merging as discussed during yesterday's spec meeting.

PR #847, opened 2021-02-15
Update closing_signed fee requirement

t-bast:
With anchor outputs, we can keep the commit tx feerate lower than the real on-chain feerate.

That means that when closing the channel, the closing fee will not necessarily be lower than the current commit tx fee, the existing requirement doesn't make sense to be that strict (at least for anchor outputs).

Imagine for example that the commitment feerate is kept at 10 sat/byte (what lnd currently does when using anchor outputs).
Imagine you want to do a mutual close on that channel and target a confirmation in N blocks, which results in a feerate higher than 10 sat/byte.
That's perfectly reasonable and should be allowed, otherwise you have no way of negotiating a feerate higher than 10 sat/byte (which may confirm in weeks if the mempool is quite full, like it is nowadays).

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -566,12 +566,13 @@ The funding node:
     - SHOULD send a `closing_signed` message.
 
 The sending node:
-  - MUST set `fee_satoshis` less than or equal to the
- base fee of the final commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
-  - SHOULD set the initial `fee_satoshis` according to its
- estimate of cost of inclusion in a block.
-  - MUST set `signature` to the Bitcoin signature of the close
- transaction, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+  - If the channel does not use `option_anchor_outputs`:
+    - MUST set `fee_satoshis` less than or equal to the base fee of the final
+    commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
+  - SHOULD set the initial `fee_satoshis` according to its estimate of cost of
+  inclusion in a block.
+  - MUST set `signature` to the Bitcoin signature of the close transaction,
+  as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
 The receiving node:
   - if the `signature` is not valid for either variant of closing transaction
@@ -580,9 +581,9 @@ The receiving node:
   - if `fee_satoshis` is equal to its previously sent `fee_satoshis`:
     - SHOULD sign and broadcast the final closing transaction.
     - MAY close the connection.
-  - otherwise, if `fee_satoshis` is greater than
-the base fee of the final commitment transaction as calculated in
-[BOLT #3](03-transactions.md#fee-calculation):
+  - otherwise, if `fee_satoshis` is greater than the base fee of the final
+  commitment transaction as calculated in [BOLT #3](03-transactions.md#fee-calculation)
+  and the channel does not use `option_anchor_outputs`:
     - MUST fail the connection.
   - if `fee_satoshis` is not strictly
 between its last-sent `fee_satoshis` and its previously-received
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -579,9 +579,6 @@ The funding node:
     - SHOULD send a `closing_signed` message.
 
 The sending node:
-  - If the channel does not use `option_anchor_outputs`:
-    - MUST set `fee_satoshis` less than or equal to the base fee of the final
-    commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
   - SHOULD set the initial `fee_satoshis` according to its estimate of cost of
   inclusion in a block.
   - SHOULD set `fee_range` according to the minimum and maximum fees it is
@@ -603,10 +600,6 @@ The receiving node:
     - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
     - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
     - MAY close the connection.
-  - otherwise, if `fee_satoshis` is greater than the base fee of the final
-  commitment transaction as calculated in [BOLT #3](03-transactions.md#fee-calculation)
-  and the channel does not use `option_anchor_outputs`:
-    - MUST fail the connection.
   - if the message contains a `fee_range`:
     - if there is no overlap between that and its own `fee_range`:
       - SHOULD send a `warning`
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -553,11 +553,24 @@ the other node then replies similarly, using a fee it thinks is fair.  This
 exchange continues until both agree on the same fee or when one side fails
 the channel.
 
+In the modern method, the funder sends its permissable fee range, and the
+non-funder has to pick a fee in this range. If it chooses the same value,
+negotiation is complete after two messages, otherwise the funder will reply
+with the same value (completing after three messages).
+
 1. type: 39 (`closing_signed`)
 2. data:
    * [`channel_id`:`channel_id`]
    * [`u64`:`fee_satoshis`]
    * [`signature`:`signature`]
+   * [`closing_signed_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `closing_signed_tlvs`
+2. types:
+    1. type: 1 (`fee_range`)
+    2. data:
+        * [`u64`:`min_fee_satoshis`]
+        * [`u64`:`max_fee_satoshis`]
 
 #### Requirements
 
@@ -571,6 +584,11 @@ The sending node:
     commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
   - SHOULD set the initial `fee_satoshis` according to its estimate of cost of
   inclusion in a block.
+  - SHOULD set `fee_range` according to the minimum and maximum fees it is
+  prepared to pay for a close transaction.
+  - if it is not the funder:
+    - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
+	  - SHOULD set `min_fee_satoshis` to a fairly low value
   - MUST set `signature` to the Bitcoin signature of the close transaction,
   as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
@@ -581,30 +599,53 @@ The receiving node:
   - if `fee_satoshis` is equal to its previously sent `fee_satoshis`:
     - SHOULD sign and broadcast the final closing transaction.
     - MAY close the connection.
+  - if `fee_satoshis` matches its previously sent `fee_range`:
+    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
+    - MAY close the connection.
   - otherwise, if `fee_satoshis` is greater than the base fee of the final
   commitment transaction as calculated in [BOLT #3](03-transactions.md#fee-calculation)
   and the channel does not use `option_anchor_outputs`:
     - MUST fail the connection.
-  - if `fee_satoshis` is not strictly
-between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, UNLESS it has since reconnected:
+  - if the message contains a `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
+      - SHOULD fail the connection
+    - otherwise:
+      - if it is the funder:
+        - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
+          - SHOULD fail the connection
+        - otherwise:
+          - MUST reply with the same `fee_satoshis`.
+      - otherwise (it is not the funder):
+        - if it has already sent a `closing_signed`:
+          - if `fee_satoshis` is not the same as the value it sent:
+            - SHOULD fail the connection.
+        - otherwise:
+          - MUST propose a `fee_satoshis` in the overlap between received and (about-to-be) sent `fee_range`.
+  - otherwise, if `fee_satoshis` is not strictly between its last-sent `fee_satoshis`
+  and its previously-received `fee_satoshis`, UNLESS it has since reconnected:
     - SHOULD fail the connection.
-  - if the receiver agrees with the fee:
+  - otherwise, if the receiver agrees with the fee:
     - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value.
   - otherwise:
     - MUST propose a value "strictly between" the received `fee_satoshis`
-  and its previously-sent `fee_satoshis`.
+    and its previously-sent `fee_satoshis`.
 
 #### Rationale
 
-The "strictly between" requirement ensures that forward
-progress is made, even if only by a single satoshi at a time. To avoid
-keeping state and to handle the corner case, where fees have shifted
+When `fee_range` is not provided, the "strictly between" requirement ensures
+that forward progress is made, even if only by a single satoshi at a time.
+To avoid keeping state and to handle the corner case, where fees have shifted
 between disconnection and reconnection, negotiation restarts on reconnection.
 
 Note there is limited risk if the closing transaction is
 delayed, but it will be broadcast very soon; so there is usually no
 reason to pay a premium for rapid processing.
+
+Note that the non-funder is not paying the fee, so there is no reason for it
+to have a maximum feerate. It may want a minimum feerate, however, to ensure
+that the transaction propagates. It can always use CPFP later to speed up
+confirmation if necessary, so that minimum should be low.
 
 ## Normal Operation
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -587,7 +587,7 @@ The sending node:
     - MUST fail the channel
   - if it is not the funder:
     - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
-	  - SHOULD set `min_fee_satoshis` to a fairly low value
+    - SHOULD set `min_fee_satoshis` to a fairly low value
   - MUST set `signature` to the Bitcoin signature of the close transaction,
   as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -583,6 +583,8 @@ The sending node:
   inclusion in a block.
   - SHOULD set `fee_range` according to the minimum and maximum fees it is
   prepared to pay for a close transaction.
+  - if it doesn't receive a `closing_signed` response after a reasonable amount of time:
+    - MUST fail the channel
   - if it is not the funder:
     - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
 	  - SHOULD set `min_fee_satoshis` to a fairly low value
@@ -603,16 +605,17 @@ The receiving node:
   - if the message contains a `fee_range`:
     - if there is no overlap between that and its own `fee_range`:
       - SHOULD send a `warning`
+      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
     - otherwise:
       - if it is the funder:
         - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
-          - SHOULD send a `warning`
+          - MUST fail the channel
         - otherwise:
           - MUST reply with the same `fee_satoshis`.
       - otherwise (it is not the funder):
         - if it has already sent a `closing_signed`:
           - if `fee_satoshis` is not the same as the value it sent:
-            - SHOULD send a `warning`
+            - MUST fail the channel
         - otherwise:
           - MUST propose a `fee_satoshis` in the overlap between received and (about-to-be) sent `fee_range`.
   - otherwise, if `fee_satoshis` is not strictly between its last-sent `fee_satoshis`
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -609,17 +609,17 @@ The receiving node:
     - MUST fail the connection.
   - if the message contains a `fee_range`:
     - if there is no overlap between that and its own `fee_range`:
-      - SHOULD fail the connection
+      - SHOULD send a `warning`
     - otherwise:
       - if it is the funder:
         - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
-          - SHOULD fail the connection
+          - SHOULD send a `warning`
         - otherwise:
           - MUST reply with the same `fee_satoshis`.
       - otherwise (it is not the funder):
         - if it has already sent a `closing_signed`:
           - if `fee_satoshis` is not the same as the value it sent:
-            - SHOULD fail the connection.
+            - SHOULD send a `warning`
         - otherwise:
           - MUST propose a `fee_satoshis` in the overlap between received and (about-to-be) sent `fee_range`.
   - otherwise, if `fee_satoshis` is not strictly between its last-sent `fee_satoshis`
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -554,9 +554,9 @@ exchange continues until both agree on the same fee or when one side fails
 the channel.
 
 In the modern method, the funder sends its permissable fee range, and the
-non-funder has to pick a fee in this range. If it chooses the same value,
-negotiation is complete after two messages, otherwise the funder will reply
-with the same value (completing after three messages).
+non-funder has to pick a fee in this range. If the non-funder chooses the same
+value, negotiation is complete after two messages, otherwise the funder will
+reply with the same value (completing after three messages).
 
 1. type: 39 (`closing_signed`)
 2. data:
@@ -604,7 +604,7 @@ The receiving node:
     - MAY close the connection.
   - if the message contains a `fee_range`:
     - if there is no overlap between that and its own `fee_range`:
-      - SHOULD send a `warning`
+      - SHOULD fail the connection
       - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
     - otherwise:
       - if it is the funder:


02-peer-protocol.md
```diff
@@ -555,12 +555,12 @@ The funding node:
     - SHOULD send a `closing_signed` message.
 
 The sending node:
-  - MUST set `fee_satoshis` less than or equal to the
- base fee of the final commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
-  - SHOULD set the initial `fee_satoshis` according to its
- estimate of cost of inclusion in a block.
-  - MUST set `signature` to the Bitcoin signature of the close
- transaction, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+  - SHOULD set `fee_satoshis` less than or equal to the base fee of the final
```
ariard: AFAIU the current spec the MUST sounds more like a mark point to start the closing fee negotiation. It's assuming right now that's the closing fee is more likely to be under commitment tx one (no-HTLC outputs) but in fact last `update_fee`/`commitment_signed` might have been done a while back and feerate already obsolote due to mempool spikes.

So, I would say the benefit of this change is to save in negotiations rounds (and thus reduce the risk of mempool fluctuations during the negotiation falsifying it). 

I believe you should also update the receiving part :
"the base fee of the final commitment transaction as calculated in, MUST fail the connection"

ariard (commented):
I agree on the advantage of this change but I believe we should relax receiver-side first otherwise, you will have unilateral channel closure for nothing between upgraded-more-tolerant peers and non-upgrades ones ?

02-peer-protocol.md
t-bast: > So, I would say the benefit of this change is to save in negotiations rounds (and thus reduce the risk of mempool fluctuations during the negotiation falsifying it).

Not only, right now the spec doesn't let us propose a feerate higher than the commit tx feerate, which is absurd when using anchor outputs.

> we should relax receiver-side first otherwise

That's unnecessary, this condition will only hit with anchor output channels which means the receiver have been upgraded to support anchor outputs, so let's update both at the same time for simplicity.

t-bast: I updated the receiver requirements in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/bb1012218f72a328e9c945a597ff223a8098d617

halseth:
cc

t-bast:
I added a `fee_range` tlv in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/3566bcfb567ce4d34ceed231c5d95840ca736856 as discussed during the last spec meeting.

I chose flat fees instead of feerates to be consistent with the existing fields.

viaj3ro:
Hoping to see this implemented across implementations soon. I have lot's of funds stuck indefinitely due to some nodes (even high profile nodes like LOOP) probably accidentally insisting on <2 s/b closing fees sometimes. 
Same goes the other way. Lots of funds are wasted due to peers insisting on overpaying significantly.

rustyrussell:
OK, perhaps we should split this further, to make it explicit that there's no back-and-forth if both sides support this?

1. If you both include the TLV, there's no back-and-forth, it's either two messages or three.
2. If nodes receive the tlv, it picks from that range for its reply: if it doesn't care, uses the same rate.
3. If reply has different rate (and it's acceptable), reply with using that rate.  Otherwise it's complete.

That will be a nice simplification in future...

t-bast:
> OK, perhaps we should split this further, to make it explicit that there's no back-and-forth if both sides support this?

Updated in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/9a0b1f68d6c6feefe2c459e28adee3475d02a62e

niftynei:
> OK, perhaps we should split this further, to make it explicit that there's no back-and-forth if both sides support this?
> 
> 1. If you both include the TLV, there's no back-and-forth, it's either two messages or three.
> 2. If nodes receive the tlv, it picks from that range for its reply: if it doesn't care, uses the same rate.
> 3. If reply has different rate (and it's acceptable), reply with using that rate.  Otherwise it's complete.
> 
> That will be a nice simplification in future...

Ok, so reading the spec as is this intended exchange isn't obvious.

If I understand correctly, the 3 message exchange would be as follows:
- A sends  1000 fee, range 100-5000.
- B replies with 3000 fee, range 200-6000.

If I'm reading @rustyrussell's comment correctly, A should at this point sign + broadcast AS WELL AS replying with
- A sends 3000 fee, range 100-5000

The spec, as written, only details two of these three requirements

```
  - if `fee_satoshis` matches its previously sent `fee_range`:
    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
    - MAY close the connection.
```

I'd propose updating this section to:
```
  - if `fee_satoshis` does not match the previously sent `fee_satoshis` and the `fee_satoshis` is within the previously sent `fee_range`:
    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
    - MUST reply with a matching `fee_satoshis` and their `fee_range`
```

Or something along these lines ...

t-bast:
> Ok, so reading the spec as is this intended exchange isn't obvious.

You're right, it was too implicit, I added one step in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/1684acc073d72add60d0d70008b91a9d19095133, let me know if that's better.

The main difficulty in making this clear comes from the fact that we keep specifying the legacy negotiation and the requirements must reflect that.

The flow you describe is what I expect implementations to do, so hopefully we won't have surprises when doing interop testing!

t-bast:
@niftynei did you have time to finalize this on the c-lightning side?

niftynei:
Not yet :(

rustyrussell:
OK, I'm working on implementing this now, at @niftynei request.

There are two things here:
1. The new "quick-close" negotiation method using TLVs.
2. Allowing fees to exceed the final commitment tx iff option_anchor_outputs applies to the channel.

Now, LND is doing option_zero_htlc_tx_fee instead, so not sure how that fits (we need more internal work before we can support this).  I don't know if they allow 2?

So I am going with:

1. Only send the tlvs if option_anchor_outputs is *applies to this channel*.
2. Only allow sending higher fee if option_anchor_outputs is *applies to this channel*.
3. Use normal negotiation if they don't send TLV.

Normally we don't do 2, since we don't reduce fees for option_anchor_outputs commitments (yet) anyway,
so we get away with it anyway.

Hopefully that will allow reasonable transition on any existing (experimental!) deployments.

t-bast:
> There are two things here:
>    The new "quick-close" negotiation method using TLVs.
>    Allowing fees to exceed the final commitment tx iff option_anchor_outputs applies to the channel.

Yes, at first I opened the PR only for the second point (which is a spec defect with anchor outputs), then we decided it was a good opportunity to improve closing negotiation, hence the first point. But they are really two separate things.

> Now, LND is doing option_zero_htlc_tx_fee instead, so not sure how that fits (we need more internal work before we can support this). I don't know if they allow 2?

I think this PR is fully orthogonal to what fees are applied to htlc txs, only the commit tx fee is important and it's not different between `option_zero_htlc_tx_fee` and `option_anchor_outputs`. But maybe I'm missing something here? 

> So I am going with:
> Only send the tlvs if option_anchor_outputs is applies to this channel.

Why? I think this is a change that can benefit all channels right now, I don't see a compelling reason to only apply it to anchor output channels? And if you do, does that mean you would simply ignore the TLVs if you receive them for a standard channel?

rustyrussell:
> > Now, LND is doing option_zero_htlc_tx_fee instead, so not sure how that fits (we need more internal work before we can support this). I don't know if they allow 2?
> 
> I think this PR is fully orthogonal to what fees are applied to htlc txs, only the commit tx fee is important and it's not different between `option_zero_htlc_tx_fee` and `option_anchor_outputs`. But maybe I'm missing something here?

Agreed, logic applies to either.

> > So I am going with:
> > Only send the tlvs if option_anchor_outputs is applies to this channel.
> 
> Why? I think this is a change that can benefit all channels right now, I don't see a compelling reason to only apply it to anchor output channels? And if you do, does that mean you would simply ignore the TLVs if you receive them for a standard channel?

Our RPC interface to "close" allows you to specify a "fee_negotiation_step".  I'd really like to know, up-front, whether we should allow them to specify that (or, perhaps in future a new "feerange" argument).

But thinking harder, I think we can say "always send if anchor-outputs applies (and don't let them specify fee_negotiation_step), otherwise, send *unless* they specified fee_negotiation_step, implying they wanted old-style negotiation)?

> > Normally we don't do 2, since we don't reduce fees for option_anchor_outputs commitments (yet) anyway,
> > so we get away with it anyway.
> 
> In that case you don't need to change anything when you're funder, but you still need to change something when you're fundee, because your counterparty may reduce fees for `option_anchor_outputs` commitments (but since you haven't officially activated anchor outputs yet, it's not an issue in practice).

Yes.  We will always allow lower, but not actually send lower.  Testing now...

Thanks!
Rusty.

t-bast:
> But thinking harder, I think we can say "always send if anchor-outputs applies (and don't let them specify fee_negotiation_step), otherwise, send unless they specified fee_negotiation_step, implying they wanted old-style negotiation)?

That sounds perfectly reasonable. That would work for me, and would allow me to easily test interop with eclair.

rustyrussell:
OK, I think the language needs some tweaking, in particular, there should be either 2 or 3 packets sent if both are sending the feerange tlvs?

1. Opener sends range, fee.
2. Non-opener checks that its range overlaps opener range, and if so, chooses a fee in the overlap, sends range, fee.
3. Opener is finished if sent and received fees match. Otherwise, checks fee is in its range, if so *MUST mirror* fee back.

As written, it seems like there's still a negotiation possible, since opener could reply with YA different fee?  Which we would like to get away from!

rustyrussell:
Here's a diff, which I think makes it clearer.  Even more clear might be a complete separate section, but maybe with this explanation it's OK?

```diff
diff --git a/02-peer-protocol.md b/02-peer-protocol.md
index 58c0579..89421f9 100644
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -542,6 +542,11 @@ the other node then replies similarly, using a fee it thinks is fair.  This
 exchange continues until both agree on the same fee or when one side fails
 the channel.
 
+In the modern method, the funder sends its permissable fee range, and
+the non-funder has to pick a fee in this range.  If it chooses the
+same value, negotiation is complete after two messages, otherwise the
+funder will reply with the same value (completing after three messages).
+
 1. type: 39 (`closing_signed`)
 2. data:
    * [`channel_id`:`channel_id`]
@@ -589,10 +594,16 @@ The receiving node:
   and the channel does not use `option_anchor_outputs`:
     - MUST fail the connection.
   - if the message contains a `fee_range`:
-    - if it disagrees with that `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
       - SHOULD fail the connection
     - otherwise:
-      - MUST propose a `fee_satoshis` in that range
+      - if it is the funder:
+        - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
+          - SHOULD fail the connection
+        - otherwise:
+          - MUST reply with the same `fee_satoshis`.
+      - otherwise (it is not the funder):
+        - MUST propose a `fee_satoshis` in the overlap between received and (about-to-be) sent `fee_range`.
   - otherwise, if `fee_satoshis` is not strictly between its last-sent `fee_satoshis`
   and its previously-received `fee_satoshis`, UNLESS it has since reconnected:
     - SHOULD fail the connection.
```

t-bast:
Thanks, I've applied these suggestions in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/9ee62c4cffa2c8aac1e30db3b40fd1440513ac33

I tried to be clever and not explicitly expand the funder vs fundee case when receiving the `closing_signed`, but you're right that it was too implicit. This section is now quite hard to read for newcomers, but I'm hoping that a few months after all major implementations have shipped this, we can remove the legacy negotiation from the spec which will make this section a lot simpler.

rustyrussell:
@t-bast points out that the non-funder has no skin in the game, so should set max to infinite, min to a pretty low number:

```diff
diff --git a/02-peer-protocol.md b/02-peer-protocol.md
index 5ab7572..0ad9a7b 100644
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -575,6 +575,9 @@ The sending node:
   inclusion in a block.
   - SHOULD set `fee_range` according to the minimum and maximum fees it is
   prepared to pay for a close transaction.
+  - if it is not the funder:
+    - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
+	- SHOULD set `min_fee_satoshis` to a fairly low value
   - MUST set `signature` to the Bitcoin signature of the close transaction,
   as specified in [BOLT #3](03-transactions.md#closing-transaction).
 
@@ -628,6 +631,14 @@ Note there is limited risk if the closing transaction is
 delayed, but it will be broadcast very soon; so there is usually no
 reason to pay a premium for rapid processing.
 
+Note that the non-funder is not paying the fee, so there is no reason
+for it to have a maximum feerate.  It may want a minimum feerate,
+however, since the peer can theortically use a large dependent
+transaction prevent it from using a child transaction to push the
+transaction to confirm.  However, it is perfectly reasonable for the
+funder to set quite a feerate expecting 100 blocks or more for
+confirmation, so the threshold is low.
+
 ## Normal Operation
 
 Once both nodes have exchanged `funding_locked` (and optionally [`announcement_signatures`](07-routing-gossip.md#the-announcement_signatures-message)), the channel can be used to make payments via Hashed Time Locked Contracts.
```

t-bast:
> @t-bast points out that the non-funder has no skin in the game, so should set max to infinite, min to a pretty low number:

Thanks, I applied a subset of your proposal in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/f02916485c49def7b03be75d5f0e85e0368115bb

Note that unless I missed something, the non-funder can always use CPFP (thanks to the carve-out rule) so the issue of a large dependent child transaction doesn't apply, so the minimum threshold can be lower.

02-peer-protocol.md
```diff
@@ -566,12 +579,18 @@ The funding node:
...
 The sending node:
-  - MUST set `fee_satoshis` less than or equal to the
- base fee of the final commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
-  - SHOULD set the initial `fee_satoshis` according to its
- estimate of cost of inclusion in a block.
-  - MUST set `signature` to the Bitcoin signature of the close
- transaction, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+  - If the channel does not use `option_anchor_outputs`:
+    - MUST set `fee_satoshis` less than or equal to the base fee of the final
+    commitment transaction, as calculated in [BOLT #3](03-transactions.md#fee-calculation).
```
TheBlueMatt, line 584: Our commitment transaction, or our counterparty's commitment transaction (I know this isn't new on this PR, I just realized it).

viaj3ro:
Not sure if already implemented that way anyway but I'd highly recommend the fee range for the new "quick-close" negotiation method using TLVs to be configurable via config and also manually adjustable for every channel closure via API. So far no automatic fee estimation comes anywhere close to what can be achieved manually.

t-bast:
> Not sure if already planned to be implemented that way anyway but I'd highly recommend the fee range for the new "quick-close" negotiation method using TLVs to be configurable via config and also manually adjustable for every channel closure via API.

That's what `eclair` does, you set this in the `close` API: https://github.com/ACINQ/eclair/pull/1768
From what I've seen, c-lightning implemented the same mechanism.
This is purely an implementation detail though, not something to note at the spec level.

02-peer-protocol.md
t-bast: That's a good point, it's a subtle point. We use our commitment transaction in eclair, and I think it's what makes most sense: we must create a transaction that's cheaper than the alternative we can unilaterally broadcast (even though you'd also need to factor in csv delays into the choice of which one to broadcast).

However, the receiver would probably want to receive a proposal that also pays less fees than *his* commit tx, so would it make more sense to use the min of both commit tx fees here?

TheBlueMatt: Given the recipient MUST fail/force-close if we're too high, min seems like the obvious choice. Are there any edge-cases where one party has a high dust limit, causing one side of the commitment transaction to be missing an output (skewing the fee pretty low) whereas both outputs would appear in the cooperative shutdown case? In that world you risk always having to force-close (but maybe that's what you want - at least if the side that gets to skip the counterparty's output is the initiator who pays fees).

t-bast: I don't think this edge case can meaningfully happen thanks to the channel reserve (which must be greater than the `dust_limit`), unless the channel has just been opened and the reserve requirement was never met.

I'm happy to change this to specify that it should be the min of both commitments (even though this requirement will disappear anyway since it doesn't apply to anchor outputs channels).

TheBlueMatt: Actually, should we drop this requirement for the channel initiator? I don't see a reason why the node not paying the fee should care one bit if the channel opener wants to close with a higher feerate than the existing channel one.

t-bast: That's what I did in the first version of the PR, I had dropped this requirement completely, but in order to keep existing channel behavior I was asked to drop it only for anchor outputs. I'm happy to remove it for all channel types honestly, probably worth discussing again during the next spec meeting to see if we can convince @rustyrussell ;)

TheBlueMatt: Right, its a compatibility issue with existing nodes, so not ideal, but its not a super critical one imo - a few extra force-closes instead of cooperative closes while nodes upgrade isn't the end of the world. If it resulted in force-closes of otherwise-functional channels it'd be much worse, but we're talking about closing the channel anyway.

02-peer-protocol.md
```diff
@@ -580,31 +599,54 @@ The receiving node:
...
+    and its previously-sent `fee_satoshis`.
 
 #### Rationale
 
-The "strictly between" requirement ensures that forward
-progress is made, even if only by a single satoshi at a time. To avoid
-keeping state and to handle the corner case, where fees have shifted
+When `fee_range` is not provided, the "strictly between" requirement ensures
+that forward progress is made, even if only by a single satoshi at a time.
+To avoid keeping state and to handle the corner case, where fees have shifted
```
TheBlueMatt, line 638: Also not a new issue, but its kinda unacceptable that this is in the `Rationale` section - it "feels" like it intends to be normative behavior - that you "forget" the current suggested fee ranges on reconnection, but its only mentioned in `Rationale`?

rustyrussell: It's in requirements (phew!):

```
  - if `fee_satoshis` is not strictly
between its last-sent `fee_satoshis` and its previously-received
`fee_satoshis`, UNLESS it has since reconnected:
    - SHOULD fail the connection.
```

TheBlueMatt: Ohhhh, ok, I'd missed that, thanks. Sly little thing at the end of a sentence that's easy to miss :).

TheBlueMatt (commented):
LGTM. Still implementing but if others have compat testing done I'm happy with this as-is.

02-peer-protocol.md
```diff
@@ -580,31 +596,50 @@ The receiving node:
...
-  - if `fee_satoshis` is not strictly
-between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, UNLESS it has since reconnected:
+  - if `fee_satoshis` matches its previously sent `fee_range`:
+    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
+    - MAY close the connection.
+  - if the message contains a `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
+      - SHOULD send a `warning`
```
TheBlueMatt, line 605: Wait, why a `warning`? At this point you're unable to agree and are force-closing, thus it should be an `error`.

t-bast: I find it too strict and unnecessary to force-close in that case, especially since that `fee_range` will usually be provided manually by a node operator that can adapt it depending on the response from its peer. A warning is the best mechanism we currently have for that (disconnecting doesn't make much sense because at reconnection you must send the exact same `closing_signed` which will trigger another disconnection, which doesn't give the node operator the opportunity to send a second `closing_signed` with a different set of `fee_range`).

TheBlueMatt: There's a few ways we could go with this, but I don't think simply "send a warning, but leave the channel in an unusable state" with no further guidance is an acceptable one (especially since the "manual intervention" process isn't defined here).

For example, we could say that the node that initiated the shutdown MUST either send an error or send a new closing_signed with a new fee range (and any updates needed to ensure starting over with a new fee range is allowed), plus eg that the node that did not initiate the shutdown MUST send a warning and fail the channel after N minutes if a new closing_signed isn't sent.

t-bast: You're totally right, I went with a half-measure here, we can surely improve that.

I don't think responding with a different `fee_range` that doesn't overlap with the sender's `fee_range` helps though, it would either create a loop where each side sends a different `fee_range` or we'd be reintroducing something very similar to the previous fee negotiation mechanism (which we want to get rid of!).

What do you think about this variation of your last proposal:

- if you receive a `fee_range` that you disagree with, you send a warning, and force-close after N minutes if your peer doesn't send you another `closing_signed` with a different `fee_range`
- if you sent a `closing_signed` and didn't receive a response after N minutes, force-close

NB: this N minutes counter should probably take into account whether the peer is offline or not (you may want to only count N minutes when he is online), but that's really an implementation policy, not something that belongs in the spec

TheBlueMatt: > I don't think responding with a different fee_range that doesn't overlap with the sender's fee_range helps though, it would either create a loop where each side sends a different fee_range or we'd be reintroducing something very similar to the previous fee negotiation mechanism (which we want to get rid of!).

I don't see why not? As long as the new fee range is only sent in response to user input and not in an automated way, we still have the nice "here's my fee range, if you accept, pick something in the fee range and sign and then we're done" property.

> What do you think about this variation of your last proposal:

I thought that's what I was proposing, but if I was unclear, yep, what you wrote sounds good!

t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/034486cb49d8c50fbad2c845109684a0048aeafb, let me know if that's better.

TheBlueMatt (approved):
LGTM. I'm not a huge fan of this depending on `warning` messages, as the current PR needs rebase and includes several unrelated changes that may slow it down, but I get the rationale for it.

Crypt-iQ:
Will this come with a feature bit so that upgraded nodes won't send `fee_range` to older nodes? Currently this pull reads like the recipient will fail the channel in that case instead of going with the older negotiation path? Perhaps I'm wrong and missing something.

TheBlueMatt:
> Will this come with a feature bit so that upgraded nodes won't send fee_range to older nodes? Currently this pull reads like the recipient will fail the channel in that case instead of going with the older negotiation path? Perhaps I'm wrong and missing something.

No, the current intent is that it will not. See discussion at https://github.com/lightningnetwork/lightning-rfc/pull/847#discussion_r671873899 and in the last meeting. There was some reasonable agreement that (a) there's good reason for nodes to want to send a fee higher than current channel fee, the two values mean two very different things, (b) the worst-case of a force-close isn't that bad if you were already closing, given the old node refuses to accept a higher fee that you have reason to want anyway, and (c) nodes can wait as long as they want before (potentially) sending higher fees than the channel fee if they want to be conservative.

Crypt-iQ:
> > Will this come with a feature bit so that upgraded nodes won't send fee_range to older nodes? Currently this pull reads like the recipient will fail the channel in that case instead of going with the older negotiation path? Perhaps I'm wrong and missing something.
> 
> No, the current intent is that it will not. See discussion at [#847 (comment)](https://github.com/lightningnetwork/lightning-rfc/pull/847#discussion_r671873899) and in the last meeting. There was some reasonable agreement that (a) there's good reason for nodes to want to send a fee higher than current channel fee, the two values mean two very different things, (b) the worst-case of a force-close isn't that bad if you were already closing, given the old node refuses to accept a higher fee that you have reason to want anyway, and (c) nodes can wait as long as they want before (potentially) sending higher fees than the channel fee if they want to be conservative.

Gotcha.

02-peer-protocol.md
```diff
@@ -566,12 +579,17 @@ The funding node:
...
- transaction, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+  - SHOULD set the initial `fee_satoshis` according to its estimate of cost of
+  inclusion in a block.
+  - SHOULD set `fee_range` according to the minimum and maximum fees it is
+  prepared to pay for a close transaction.
+  - if it doesn't receive a `closing_signed` response after a reasonable amount of time:
+    - MUST fail the channel
+  - if it is not the funder:
+    - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
+	  - SHOULD set `min_fee_satoshis` to a fairly low value
```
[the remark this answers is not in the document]
t-bast: Good catch, tab vs space failure. Fixed in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/aaae6bc85a2d48a6dcca6369bf7da6a6a938a0cc

t-bast:
Cross-compatibility has been tested between eclair and c-lightning (based on [this branch](https://github.com/ElementsProject/lightning/pull/4599)). Everything was green on my end!

TheBlueMatt:
> Cross-compatibility has been tested between eclair and c-lightning (based on this branch). Everything was green on my end!

I believe that was against an older version of this PR (the ~only changes are the introduction of `warning` messages here and the removal of the lower-than-current-fee limit for non-anchor channels?)? I'm happy to get this merged as-is even though cross-compatibility testing was against an older version, but also happy to contribute to cross-compatibility testing of this now.

We really want to ship a minimal implementation of this ASAP on the LDK end.

t-bast:
> I believe that was against an older version of this PR (the ~only changes are the introduction of warning messages here and the removal of the lower-than-current-fee limit for non-anchor channels?)?

https://github.com/ACINQ/eclair/pull/1768 contains a version of eclair that implements the latest version of this PR if you want to test (let me know if you need help running eclair locally). The only thing missing for now is the mechanism to automatically force-close the channel if no progress has been made for some time, I'm waiting to gather user feedback to figure out the right values here.

If you have a partial implementation in the LDK, can you point me to the right resources to run an LDK-based node locally and do some cross compatibility tests? I'm curious to play around with the LDK, it would be a good opportunity for that.

> We really want to ship a minimal implementation of this ASAP on the LDK end.

I agree, I'd also like to ship this in eclair and lightning-kmp ASAP as well, it feels like there's been rough consensus and running code for a while so hopefully we can merge this PR soon?

viaj3ro:
> The only thing missing for now is the mechanism to automatically force-close the channel if no progress has been made for some time, I'm waiting to gather user feedback to figure out the right values here.

Couldn't the mutual close attempt just return an error and leave it to the user to try again later, leave the channel open or force close?!

TheBlueMatt:
> Couldn't the mutual close attempt just return an error in this case and leave it to the user to try again later, leave the channel open or force close?!

This is outside the scope of the spec, probably worth commenting on the eclair PR linked above :).

> ACINQ/eclair#1768 contains a version of eclair that implements the latest version of this PR if you want to test (let me know if you need help running eclair locally)

I'll see if I can get to it over the weekend, but no promises. In the mean time, I've got the latest version of our patch running on this node on testnet: 0296e20fa99d2940b8b00117e65d27003f0d8f81a0c960f71a5466d1aadf5ea5ea@69.59.18.201:9735

> If you have a partial implementation in the LDK, can you point me to the right resources to run an LDK-based node locally and do some cross compatibility tests? I'm curious to play around with the LDK, it would be a good opportunity for that.

Simplest way would be to take Val's LDK sample node and run that. I pushed a branch that depends on the latest LDK udpate fee PR at https://git.bitcoin.ninja/index.cgi?p=ldk-sample;a=shortlog;h=refs/heads/2021-08-new-closing-signed which you should basically be able to `git clone ... &&  cd ldk-sample && cargo build && target/debug/ldk-tutorial-node $(cat ~/.bitcoin/.cookie)@localhost:18332 ./ldk-storage` and have a running CLI-operated LDK node.

TheBlueMatt:
Meeting followup, there was strong agreement to Just Merge This, but its noted that this does rely on `warning` messages, which is not merged yet. @t-bast indicated he'd just drop the `warning` message parts and we can merge this asap, re-adding `warning`s later.

02-peer-protocol.md
```diff
@@ -553,11 +553,24 @@ the other node then replies similarly, using a fee it thinks is fair.  This
 exchange continues until both agree on the same fee or when one side fails
 the channel.
 
+In the modern method, the funder sends its permissable fee range, and the
+non-funder has to pick a fee in this range. If it chooses the same value,
```
Roasbeef, line 557: Would be clearer if the non-funder was explicitly mentioned here instead of "it" (ambiguous context imo). 

> otherwise the funder will reply with the same value (completing after three messages) 

Also it isn't guaranteed that things complete in three messages if the funder chooses to pick a different value, right (responder wants a fee too low, maybe below the current mempool high water mark)? Since the non-funder may not like that advertised fee. 

Instead, seems this is just best effort, and if the responder doesn't immediately reply with the same value, then things fall back to the "old" method?

02-peer-protocol.md
```diff
@@ -553,11 +553,24 @@ the other node then replies similarly, using a fee it thinks is fair.  This
 exchange continues until both agree on the same fee or when one side fails
 the channel.
 
+In the modern method, the funder sends its permissable fee range, and the
```
Roasbeef, line 556: Don't we need a feature bit for this? As otherwise, an upgraded funder can send these new fields, for it only to be ignored by the non-funder (or the other way around). IMO it's also desirable to be able to _force_ this behavior in the future by having nodes advertise it as a required feature bit. 

Independent of that, it may be desirable for nodes to seek out other nodes that also understand this new behavior as it may allow them to pay lower co-op closing fees on average.

Roasbeef: Related to the comment below, and also to this feature bit idea: what if we stopped including the signature (for the new co-op close style), until one side fully "accepts" a proposal? As is, you can't actually continue using the channel once a co-op close attempt is initiated, since if it fails, or both sides D/C (or w/e) one side has a valid signature to close out the channel, so they can use that at anytime. On top of that, as is, a sort of signing oracle is exposed, since AFAIK implementations don't really clamp down on the number of iterations during negotiation, so they may just force another node to continue signing new messages which _may_ leak information. 

So if we drop the sigs (say they're all zero) until one side "accepts" (signalled by sending an actual signature), the only valid sigs ever sent across are for the final agreed fee rate. If for w/e reason, negotiation breaks down (neither side sends a sig), then its safe for both sides to use the channel. The other added benefit here is that when the "negotiation" method is used, you don't need to sign so it's more performant CPU-wise.

02-peer-protocol.md
```diff
@@ -566,12 +579,17 @@ The funding node:
...
- estimate of cost of inclusion in a block.
-  - MUST set `signature` to the Bitcoin signature of the close
- transaction, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+  - SHOULD set the initial `fee_satoshis` according to its estimate of cost of
+  inclusion in a block.
+  - SHOULD set `fee_range` according to the minimum and maximum fees it is
+  prepared to pay for a close transaction.
+  - if it doesn't receive a `closing_signed` response after a reasonable amount of time:
+    - MUST fail the channel
+  - if it is not the funder:
```
Roasbeef, line 588: Things seem undefined if the receiver finds the range sent over unacceptable. Say they want to close ASAP, and want a fee greater than the max fee proposed, what happens then? Do negotiations just break down and both sides need to reconnect and try again?

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
+    - MAY close the connection.
+  - if the message contains a `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
+      - SHOULD send a `warning`
+      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
+    - otherwise:
+      - if it is the funder:
+        - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
+          - MUST fail the channel
+        - otherwise:
```
Roasbeef, line 613: This nesting tho...

Am I the only one that finds this stuff hard to read/parse w/ all the nesting and inherited implicit context from the parent's indentation?

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
-  - otherwise, if `fee_satoshis` is greater than
-the base fee of the final commitment transaction as calculated in
-[BOLT #3](03-transactions.md#fee-calculation):
-    - MUST fail the connection.
-  - if `fee_satoshis` is not strictly
-between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, UNLESS it has since reconnected:
+  - if `fee_satoshis` matches its previously sent `fee_range`:
+    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
```
Roasbeef, line 603: So we're also changing the requirements for _existing_ nodes that don't understand this new feature?

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
+    - if there is no overlap between that and its own `fee_range`:
+      - SHOULD send a `warning`
+      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
+    - otherwise:
+      - if it is the funder:
+        - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
+          - MUST fail the channel
+        - otherwise:
+          - MUST reply with the same `fee_satoshis`.
+      - otherwise (it is not the funder):
```
Roasbeef, line 615: Should be: "if it is not" ?

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
-    - MUST fail the connection.
-  - if `fee_satoshis` is not strictly
-between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, UNLESS it has since reconnected:
+  - if `fee_satoshis` matches its previously sent `fee_range`:
+    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
+    - MAY close the connection.
+  - if the message contains a `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
```
Roasbeef, line 606: Warnings don't exit yet though right? Also this says nothing w.r.t what should be in the warning message itself, so the message would be sent w/o any context.

Roasbeef (changes requested):

02-peer-protocol.md
TheBlueMatt: > Instead, seems this is just best effort, and if the responder doesn't immediately reply with the same value, then things fall back to the "old" method?

Things never fall back to the "old" method? The funder can start over, so, yes, the text should be changed, but there's no fallback.

02-peer-protocol.md
TheBlueMatt: > IMO it's also desirable to be able to force this behavior in the future by having nodes advertise it as a required feature bit.

What does this accomplish? At least on the LDK end we already have (reasonably-deterministically-generated) an internally-stored min/max, exposing that to the counterparty just lets things go quicker.

> what if we stopped including the signature (for the new co-op close style), until one side fully "accepts" a proposal? 

because it breaks backwards compatibility.

02-peer-protocol.md
TheBlueMatt: You send a `warning` and then wait some time and then force-close. The "wait some time and then force-close" part is deliberately left to the implementation to decide how they want to do that, or, even, *if* they want to do that vs just making the user manually force-close.

02-peer-protocol.md
TheBlueMatt: Correct, see https://github.com/lightningnetwork/lightning-rfc/pull/847#issuecomment-899832302. As for warning messages being more clearly defined, that can come in a followup to the warning message pr.

02-peer-protocol.md
TheBlueMatt: If you don't understand the new feature then you won't have a "previously sent `fee_range`" to match, presumably if you're sending one you should understand it too.

02-peer-protocol.md
t-bast: This is mostly because we need to keep the previous requirements.
I'm hoping that in X months when everybody has implemented this, we can remove the old negotiation mechanism from the spec which will greatly simplify this section.

02-peer-protocol.md
t-bast: No, this is an assertion. We're in the `otherwise` of `if it is the funder`, so this is here only to help readers.

02-peer-protocol.md
t-bast: We've discussed several times why we don't use a feature bit (see previous comments on this PR).
This is quite a simple feature which will simply replace the old negotiation (which we'll completely remove from the spec in X months).

02-peer-protocol.md
t-bast:
```
- if there is no overlap between that and its own `fee_range`:
      - SHOULD send a `warning`
      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
```

@Roasbeef the case you describe is a spec violation. The non-funder either chooses a fee inside the advertised `fee_range`, or doesn't answer (sending a warning once that has been added to the spec) potentially leading to a force-close instead if no progress is made.

We only fall back to the old negotiation when the response doesn't contain a `fee_range` (indicating that the receiver didn't support this new negotiation scheme).

02-peer-protocol.md
t-bast: Done in https://github.com/lightningnetwork/lightning-rfc/pull/847/commits/92d2af04069ad5f3668bc3605353243a463c6169

t-bast:
I've removed the reference to warning messages, let's wait for @Roasbeef to ack before merging as he left some comments?

02-peer-protocol.md
t-bast: The flow from the node operator's point of view in that case is:

- use the API to close a channel, with an explicit `fee_range` that the node operator think makes sense
- if the peer accepts that, great!
- otherwise the node operator receives a warning, and can choose to either retry with a different `fee_range`, do nothing or force-close

In the future we may want to automate that retry, but based on our users' feedback, they would really prefer having a human in the loop to decide whether they should adapt their `fee_range` or not. Many have been disgruntled by the feerates chosen by estimators, so they're not interested in an automated retry here.

t-bast:
> Simplest way would be to take Val's LDK sample node and run that. I pushed a branch that depends on the latest LDK udpate fee PR at https://git.bitcoin.ninja/index.cgi?p=ldk-sample;a=shortlog;h=refs/heads/2021-08-new-closing-signed which you should basically be able to git clone ... && cd ldk-sample && cargo build && target/debug/ldk-tutorial-node $(cat ~/.bitcoin/.cookie)@localhost:18332 ./ldk-storage and have a running CLI-operated LDK node.

@TheBlueMatt I'm getting the following error, any idea what I might be doing wrong (or if I'm missing steps in the setup process)?

```
$ cargo build

Updating git repository `https://git.bitcoin.ninja/rust-lightning`
error: failed to get `lightning` as a dependency of package `ldk-tutorial-node v0.1.0 (/home/bast/Documents/lightning/ldk-sample)`

Caused by:
  failed to load source for dependency `lightning`

Caused by:
  Unable to update https://git.bitcoin.ninja/rust-lightning?branch=2021-07-new-closing-fee#9425aadb

Caused by:
  object not found - no match for id (9425aadbb574a1f4b08e3e2929676b45c99b3c69); class=Odb (9); code=NotFound (-3)
```

TheBlueMatt:
> @TheBlueMatt I'm getting the following error, any idea what I might be doing wrong (or if I'm missing steps in the setup process)?

Huh, I seem to have screwed up my git branch, whatever, the PR for this is merged now, so you should be able to just `git clone https://github.com/lightningdevkit/ldk-sample && cd ldk-sample....`

TheBlueMatt (approved):

TheBlueMatt:
> let's wait for @Roasbeef to ack before merging as he left some comments?

I've reached out to roasbeef to ask for a timeline, but haven't yet gotten a response. This PR is now many months old with cross-implementation testing complete and multiple ACKs from different implementations. Ideally we should wait, but lets not wait too long.

t-bast:
> the PR for this is merged now

Are you sure? It looks like it's merged to `rust-lightning`, but I'm not sure it's included in the latest `ldk-sample`...
I've tried it and the ldk node gladly accepts whatever closing fee eclair sends (without any negotiation regardless of the proposed fee), but doesn't include a `fee_range` in the `closing_signed` it sends to eclair.

TheBlueMatt:
> Are you sure? It looks like it's merged to rust-lightning, but I'm not sure it's included in the latest ldk-sample...
I've tried it and the ldk node gladly accepts whatever closing fee eclair sends (without any negotiation regardless of the proposed fee), but doesn't include a fee_range in the closing_signed it sends to eclair.

You may need to run `cargo update` in `ldk-sample`, it by default points to git rust-lightning, but it won't magically update on its own unless you tell it to, IIRC.

02-peer-protocol.md
Roasbeef: > This is quite a simple feature which will simply replace the old negotiation (which we'll completely remove from the spec in X months).

How can it be removed without having a new feature bit? If we have a new bit, we'd simply all switch it over to _required_ which effectively allow us to remove the old logic for these clients.

TheBlueMatt: > Nodes in the wild can seek out peers that support this feature so they don't need to implement the "old" negotiation protocol. Nodes that already have the old logic may also want to instead

Do you plan on doing this? If not, then I think no one will and its probably not worth worrying about.

> What's the tangible cost of adding a feature bit here?

Complexity and more stuff to check in a chunk of code that already has a ton of tests.

> How can it be removed without having a new feature bit? If we have a new bit, we'd simply all switch it over to required which effectively allow us to remove the old logic for these clients.

You haven't explained why this is different with or without a feature bit.

t-bast:
Merging as per yesterday's spec meeting.

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
+      - SHOULD fail the connection
+      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
+    - otherwise:
+      - if it is the funder:
+        - if `fee_satoshis` is not in the overlap between the sent and received `fee_range`:
+          - MUST fail the channel
+        - otherwise:
+          - MUST reply with the same `fee_satoshis`.
```
Crypt-iQ, line 614: Sorry to gravedig, but what is meant by overlap here? Is it the union of the two ranges or the intersection? If it's the intersection of the two ranges, then the statement a bit above `if fee_satoshis matches its previously sent fee_range:` could make this "overlap" statement redundant by changing the SHOULD to a MUST.

02-peer-protocol.md
```diff
@@ -580,31 +598,51 @@ The receiving node:
...
-between its last-sent `fee_satoshis` and its previously-received
-`fee_satoshis`, UNLESS it has since reconnected:
+  - if `fee_satoshis` matches its previously sent `fee_range`:
+    - SHOULD use `fee_satoshis` to sign and broadcast the final closing transaction
+    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value if it is different from its previously sent `fee_satoshis`
+    - MAY close the connection.
+  - if the message contains a `fee_range`:
+    - if there is no overlap between that and its own `fee_range`:
+      - SHOULD fail the connection
+      - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
```
Crypt-iQ, line 608: Does this imply that you are allowed to send multiple `closing_signed` messages without "waiting your turn" if you send bad fee ranges? If so, this interacts with musig2+taproot channels

t-bast: Yes, this is not a turn-based protocol, the initiator may send another `closing_signed` with a different fee range.

Crypt-iQ: Just to clarify, legacy coop close _is_ turn-based, but fee_range isn't? 

What is the rationale to make this not turn-based? Shouldn't the recipient of a fee_range ensure that the sent fee_range has overlap, or else don't send one?

Crypt-iQ: I haven't posted my set of comments there yet but you don't need to use nonces in shutdown -- your point still remains the same as you'd be using up the latest nonces you have. since there's no acknowledgement in `closing_signed`, you either need:
- turn-based coop closing always
- some acknowledgement message
- deterministic nonce points only for the coop closing domain (send these in shutdown, don't use for regular commitments)
- pre-share a bunch of nonces in shutdown (impossible to guess how many nonces is enough without setting a limit on the number of `closing_signed` sent without receiving a response)

Crypt-iQ: Is the mismatched fee range only possible in this flow (A is funder):
```
A                 B
   --cl_sig-->  (100, 200) fee=150
   <-cl_sig---  (300, 400) fee=350
   <-cl_sig---  (150, 350) fee=175
   --cl_sig-->  (100, 200) fee=175
```

Or is it possible for the fundee B to be on the receiving end of the non-turn-based behavior where A is sending multiple `closing_signed` without a break (I don't think so given my interpretation of the spec)?

I am trying to draft an update to the taproot channels proposal that eliminates signatures in some cases like if the sent fee_range has no overlap (and only if the taproot feature bit was negotiated).

Crypt-iQ: I see now that you've answered my question in that A (funder) is the one that sends multiple closing_signed if B sends a `warning`

t-bast: Yes, sorry for the late reply. There are two possible flows for `fee_range` closing. The first is the happy flow:

* A sends `closing_signed`
* B responds with `closing_signed` inside the `fee_range` proposed by A (B is not allowed to answer with a fee outside of A's `fee_range`)
* A responds with `closing_signed` with the same fee as the previous message

The second possible flow is the unhappy flow where B isn't satisfied with A's `fee_range`:

* A sends `closing_signed`
* B doesn't like the `fee_range`, sends a `warning`
* A sends another `closing_signed`
* B still doesn't like it, sends another `warning`
* ...
* A sends another `closing_signed`
* B likes that `fee_range`, responds with `closing_signed`
* A responds with `closing_signed` with the same fee as the previous message

We can make that strictly turn-based by having Bob send a new message (e.g. `reject_closing_signed`) to express his disagreement with the `fee_range`, in which case we could have nonces in both messages.

t-bast: I don't mind doing a coop close v3 for taproot, with a more strictly specified protocol if it's easier than tacking it on to the existing one!

Crypt-iQ: Right I dont mind either, I can draft some coop changes to add onto @Roasbeef taproot proposal

PR #849, opened 2021-02-18
Clarify nonce size for onion cipher

matheusd:
Some ChaCha20 implementations API's support both 64- and 96-bit nonces, while
others only support a single one.

Functionally, both nonce sizes are equivalent for LN usage, since the
nonce is always zeroed. However, while evaluating spec compliance of
ChaCha20 libraries, the fact that some do not support the 8 byte nonce
variant prompted a closer investigation about the nonce requirement.

Since RFC8439 is the one linked to in the current BOLT0004 spec and that
RFC only specifies the 96-bit nonce variant, that requirement is made
more explicit by this commit.


t-bast (commented):
Good catch, ACK, we are indeed using 96-bit nonces.

cdecker:
Oh nice one, we are using libsodium which takes 64bits, hence the implicit assumption in my writeup.

ACK

cfromknecht (approved):
LGTM 👍

PR #852, opened 2021-03-05
Clarify Bolt 3 htlc tx output test vectors

t-bast:
It was sometimes unclear whether we indexed by the output or the htlc id.
This is a follow-up from discussions made in #539.


t-bast:
@rustyrussell @jkczyz you may be interested in these test vector updates

rustyrussell:
Tested-Ack: 2f0cce1a12e889a544e48d7e39cd90d80cf1efb9

t-bast:
@jkczyz can you test this for the rust-lightning side? If these test vectors work for you as well, we'll merge this.

jkczyz:
@t-bast It's unclear to me what has changed in these test vectors. A cursory look seems to suggest only labels have changed. Am I missing something?

Also, I think the PR description and commit message aren't very clear. 

> It was sometimes unclear where we indexed by the output or the htlc id.

Which one is it then and what is being indexed?

t-bast:
@jkczyz yes only the labels have changed, but you may (or may not) be using them in your tests.

It was previously unclear in some places whether the number that was in the comment for htlcs was the `htlc_id` or the index of the htlc output in the commit tx. We also clearly specify now which are `htlc-success` and which are `htlc-timeout` txs and have a more coherent grouping of signatures and txs.

It's mostly a cosmetic change, but it should make these test vectors easier to understand for newcomers (at least that's the goal!).

t-bast:
Rebased. @rustyrussell confirmed that c-lightning validates these test vectors, @jkczyz can you have a look at the latest state of the PR? I think it's worth merging, it makes the test vectors clearer for future readers.

jkczyz:
ACK 2a015741ebe4792b760e53604d6bbf4bac7d8789

Sorry about the delay! RL's tests don't use any output indexes directly. Rather they use RL's state machine logic to build a commitment transaction from the included HTLCs. So the change itself shouldn't affect us.

Took a pass through and it seems easier to understand now. Though regardless of the change, the vectors could use better formatting, IMHO. Not sure if anyone is parsing these with code, but they don't seem to optimized for the reader. It's also not clear to me why `local_htlc_signature` lines start with `#` comments. But this outside the scope of the PR.

t-bast:
Thanks @jkczyz, I agree that formatting to get something more easily machine-readable would be great.
We've talked about it for years, but the cost of change overran the benefits so no-one made a PR :slightly_smiling_face: 

@rustyrussell shall we merge this now that we have 3 implementations acking?

PR #862, opened 2021-04-20
Deterministic points

rustyrussell:
Built on top of @niftynei 's https://github.com/lightningnetwork/lightning-rfc/pull/851/commits, this alters that protocol to use @LLFourn's deterministic points.

This is useful, because splice wants to change the points, and now it's as easy as updating the generation.


rustyrussell:
After careful consideration, I am withdrawing this.  The issue is that it is no longer possible to have separate secret keys for each channel: a compromise of one channel's keys would compromise the others.  That isn't a security requirement right now, but it's definitely a valid security model: you can hold all your own channels' keys and contain the damage that can be done.  c-lightning early on operated in this manner, for example.

PR #866, opened 2021-04-26
proposal: Added Funding Timeout Recovery proposal (KEYOLO variant)

cdecker:
This is the counter-proposal to #854, using the more flexible, but
also potentially more dangerous direct private key exchange, rather
than sending over a `sighash_none` signature for a close transaction.

It was proposed due to the observation that, while not yet required by
todays spec, all implementations decided to use hardened derivation
for the channel `funding_privkey`, making this a safe option.

Please continue discussing tradeoffs on #854 so we can keep them in
one coherent place, wording recommendations on this PR are still
welcome.

The two proposals are mutually exclusive, so we need to hammer out the
differences in the PRs :-)

Closes #854

### DIFF
--- a/proposals/001-funding-timeout-recovery.md
+++ b/proposals/001-funding-timeout-recovery.md
@@ -1,0 +1,177 @@
+# Funding Timeout Recovery (KEYOLO)
+
+> "So long, and thanks for all the keys" 
+>     -- Hitchhikers guide to Lightning
+
+
+## Problem description
+
+Due to circumstances it may take considerable time for a funding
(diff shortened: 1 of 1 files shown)


niftynei:
```
<BlueMatt> didnt we discuss this last time or two times ago and decide against the keyolo variant (after having been weakly in favor the week before)?
<BlueMatt> i had thought there was rough consensus against keyolo
```

Closing this as per June 7, 2020 IRC meeting in #lightning-dev (libera)

PR #867, opened 2021-04-27
Feature 106/107: option_simplified_update.

rustyrussell:
This feature turns on a simplification of the current update scheme:
1. Turns are taken, with all updates settling before switching sides.
2. update_fee is always a separate commitment by itself.

This is a fairly easy addition for existing implementations, but
a much-reduced burden for new implementations.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -670,9 +670,13 @@ If one side sends an update when it is not its turn, the other side
 can either ignore it (in favor of its own update), or reply with a
 `yield` message to hand the turn over.
 
-#### Swapping turns: `yield`
+#### Swapping turns: `yield` and `update_noop`
 
-1. type: 138 (`yield`)
+1. type: 129 (`update_noop`) (`option_simplified_update`)
+2. data:
+   * [`channel_id`:`channel_id`]
+
+1. type: 138 (`yield`) (`option_simplified_update`)
 2. data:
    * [`channel_id`:`channel_id`]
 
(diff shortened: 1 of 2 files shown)


02-peer-protocol.md
```diff
@@ -644,6 +645,118 @@ transactions may be out of sync indefinitely. This is not concerning:
...
+  - MUST NOT send `yield` unless `option_simplified_update` is negotiated.
+  - MUST track whose turn it is, starting with the peer with the lesser SEC1-encoded node_id.
+  - MUST give up its turn when:
+    - sending `revoke_and_ack`
+    - sending a `yield`
+  - MUST accept its turn when:
+    - receiving `revoke_and_ack`
+    - receiving a `yield`
+  - During this node's turn:
+    - if it receives an update message or `commitment_signed`:
```
instagibbs, line 695: the remote node sends a `commitment_signed` during the local node's turn, after the local node has sent a `commitment_signed` message though? diagram, step (5)...

02-peer-protocol.md
```diff
@@ -644,6 +645,118 @@ transactions may be out of sync indefinitely. This is not concerning:
...
+      - if it has sent its own update or `commitment_signed`:
+        - MUST ignore the message
```
instagibbs, line 697: the remote node sends a commitment_signed during the local node's turn, after the local node has sent a commitment_signed message though? diagram, step (5)...

02-peer-protocol.md
```diff
@@ -1119,6 +1232,8 @@ given in [BOLT #3](03-transactions.md#fee-calculation).
 The node _responsible_ for paying the Bitcoin fee:
   - SHOULD send `update_fee` to ensure the current fee rate is sufficient (by a
       significant margin) for timely processing of the commitment transaction.
+  - if `option_simplified_update` is negotiated:
+    - MUST NOT mix `update_fee` with other updates in the same commitment.
```
instagibbs, line 1236: does this mean any other updates, including other `update_fee` messages in same turn, or just `update_fee` vs all other update types?

rustyrussell: Good q! I think it means the former, but should be more explicit!

rustyrussell:
Someone should update this and reintroduce this for the modern world.  

In particular, on top of this we can start rejecting HTLCs we don't want to include, meaning we don't have to tell the peer in advance what the limits of the channel are (minimum htlcs, etc) in advance: we can just safely reject things we don't like and not rely on our peer never sending them to us.

PR #851, opened 2021-03-02
interactive-tx: Add dual-funding flow, using the interactive tx protocol (feature 28/29)

niftynei:
This commit adds the interactive transaction construction protocol, as
well as the first practical example of using it, v2 of channel
establishment.

Note that for v2 we also update the channel_id, which now uses the hash
of the revocation_basepoints. We move away from using the funding
transaction id, as the introduction of RBF makes it such that a single
channel may have many funding transaction id's over the course of
its lifetime.

The easiest way to understand the gist of this PR is this video, which explains the prior protocol and overviews this one.

https://youtu.be/i_GxmNZjwhk

PR #878, opened 2021-06-07
option_will_fund: liquidity ads

niftynei:
Draft spec for adding 'liquidity ads' to the dual-funding flow. Allows peers to signal fees that they'll charge for committing liquidity into a dual-funded open, which are then locked until the lease expires.

Currently a work in progress.

Builds on #851.


07-routing-gossip.md
```diff
@@ -312,6 +324,19 @@ The origin node:
   - MUST set `features` according to [BOLT #9](09-features.md#assigned-features-flags)
   - SHOULD set `flen` to the minimum length required to hold the `features`
   bits it sets.
+ - if supports `option_will_fund`:
+  - MUST include `option_will_fund` tlv type
+  - MUST set `funding_fee_base_sat` to the base fee (in satoshi) it will charge
+    for any reciprocated funding.
+  - MUST set `funding_fee_proportional_basis` to the amount (in
+    thousandths of a satoshi) it will charge per contributed satoshi.
```
Kixunil, line 332: Why not millionths?

07-routing-gossip.md
```diff
@@ -359,6 +384,12 @@ to be ordered in ascending order, unknown ones can be safely ignored.
 Additional fields beyond `addresses` may also be added in the future—with
 optional padding within `addresses`, if they require certain alignment.
 
+If a node signals `option_will_fund`, they are signaling that they
+will provide funding to a node at the stated terms. They also commit
+to a feerate they will charge for transmitting funds over the channel
```
Kixunil, line 389: Nit: "to a *maximum* feerate"

proposals/010-will-fund-for-food.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Overview
+
+This proposal adds a TLV to the `node_announcement`,
+which permits a node to advertise a rate for which they will contribute
+funds to an open channel request. Due to the nature of the protocol, these
+opens are only possible on the v2 channel protocol.
+
+Any advertised liquidity lease is for a duration of 4032 blocks, or
+approximately 28 days.
```
Kixunil, line 17: Seems like it'd be better to keep this configurable. Is there any reason not to?

07-routing-gossip.md
rustyrussell: Because this fits better in u16?  As these msgs are gossiped all over the network, bytes matter.  Though basis points would be 10,000th, so perhaps rename to "funding_fee_proportional_thousanths" explicitly?  (Allowing up to 65% seems nice; 6.5% is probably a bit low).

proposals/010-will-fund-for-food.md
rustyrussell: Because the more variables you have the harder it is to computationally compare rates.  It's actually pretty easy to add a parameter for this later, if it proves popular, but I think we need to see what it looks like in practice (should the fee guaranty be separately configurable, for example).

02-peer-protocol.md
```diff
@@ -968,6 +997,42 @@ additions.
...
 The accepting node:
     - MAY respond with a `funding_satoshis` value of zero.
+    - if the `option_will_fund` tlv was sent in `open_channel2`:
+        - if they decline to contribute funds or to be paid for their
+          contributions:
+            - SHOULD omit the `will_fund` tlv
+        - if they decide to accept the offer:
+            - MUST include a `will_fund` tlv
+            - MUST set `funding_satoshis` to a value greater than 0msat
+            - MAY provide a `funding_satoshis` of any amount.
```
rustyrussell, line 1007: Does this imply WUMBO?

02-peer-protocol.md
```diff
@@ -968,6 +997,42 @@ additions.
...
+            - MUST set `funding_fee_base_sat` to the base fee
+              (in satoshi) it will charge for the `funding_satoshis`
+            - MUST set `funding_fee_proportional_basis` to the amount
+              (in thousandths of satoshi) it will charge per `funding_satoshi`
+            - MUST set `funding_weight` to the weight they
+              will contribute to this channel, to fund the request.
```
rustyrussell, line 1013: Not *quite*, this is the weight that the initiating node will pay for (I suspect this language predates that change)

02-peer-protocol.md
```diff
@@ -985,6 +1050,41 @@ basepoint for the peer's revocation basepoint.  This allows the peer to
...
+If the initiating node sent `request_funds` and the accepting node replied
+with `will_fund` this channel is considered 'leased'.  The lease ends
+4032 blocks after the `open_channel2` `blockheight`, i.e.
+`lease_end` is defined as the `open_channel2`.`blockheight` + 4032.
+
+During the lease, the initiator should consider limiting
+the rate funds are routed through the channel as this provides the opener
+the opportunity to move the funds elsewhere (and renders the
+lease fairly useless). A sensible policy here depends on the motivation of the
+opener in acquiring the leased funds.
```
rustyrussell, line 1086: This causes a weird case, where Alice requests 1M sats, Bob has 1M+dust sats available.  If Bob puts that dust into the channel, Alice pays a bit extra for it, which is weird.  Is this worth fixing?  Maybe not?  The only fix I can think of is to use min(`accept_channel2`.`funding_satoshis`, `open_channel2.request_funds.requested_sats`) * `funding_fee_proportional_basis` instead of `accept_channel2`.`funding_satoshis` * `funding_fee_proportional_basis` but that's also a bit messy?

02-peer-protocol.md
```diff
@@ -968,6 +997,42 @@ additions.
...
+
+The receiving node:
+  - SHOULD fail the negotiation if:
+    - they sent `request_funds` tlv and:
+      - the `funding_satoshi` is less than required
+      - the `funding_fee_proportional_basis` is too high
+      - the `funding_fee_base_sat` is too high
+      - the `funding_weight` is too high
+      - the `channel_fee_proportional_basis_max` is too high
+      - the `channel_fee_base_max_msat` is too high
```
rustyrussell, line 1033: Maybe - SHOULD accept a `funding_satoshi` in excess of `requested_sats` ?

02-peer-protocol.md
```diff
@@ -1862,6 +1962,49 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+The node _not responsible_ for initiating the channel:
+  - MUST NOT send `update_blockheight`.
+
+A receiving node:
+  - if the `update_blockheight` is less than the last received `blockheight`:
+    - SHOULD fail the channel.
+  - if the sender is not the initiator:
+    - MUST fail the channel.
+  - if `blockheight` is more than 504 blocks behind the current blockheight:
+    - SHOULD fail the channel
```
rustyrussell, line 1996: Not quite right, this needs a separate section (hmm, may need an intro which defines lease nomenclature?  Lessor, end of lease, etc?)

A node which offered a lease on a channel, which is still being enforced:
- if the last received `blockheight` is more than 504 blocks behind the current blockheight:
  - SHOULD fail the channel.

With rationale:

The lessor's lease is only shortened when a new blockheight is committed, otherwise it's always 4032 blocks in the future. If the peer doesn't update the blockheight, at some point it has to start the closing process to retrieve its funds.  504 blocks seems a reasonable (but not compulsory!) answer here.

02-peer-protocol.md
```diff
@@ -2095,6 +2238,49 @@ fall-behind detection.  An implementation can offer both, however, and
...
+	`funding_fee_base_sat` +
+	`accept_channel2`.`funding_satoshis` * `funding_fee_proportional_basis` / 1000 +
+	`funding_weight` * `funding_feerate_perkw` / 1000
+
+E.g.
+An node requests 100.0000sats at a feerate of 2500perkw. They
+are contributing 50.0000sats. Their weight contribution to the
+funding transaction will be 720.
+
+The accepter charges a base funding fee of 233sats with a
```
rustyrussell, line 2259: Importantly, the accepter offers funding_satoshis 100sats

02-peer-protocol.md
```diff
@@ -2095,6 +2238,49 @@ fall-behind detection.  An implementation can offer both, however, and
...
+
+The lease fee is calculated as:
+
+	`funding_fee_base_sat` +
+	`accept_channel2`.`funding_satoshis` * `funding_fee_proportional_basis` / 1000 +
+	`funding_weight` * `funding_feerate_perkw` / 1000
+
+E.g.
+An node requests 100.0000sats at a feerate of 2500perkw. They
+are contributing 50.0000sats. Their weight contribution to the
```
rustyrussell, line 2256: 4 decimal point is particularly weird? but the decimals here are really confusing if you're from Europe and read this ias 100,000 sats!

02-peer-protocol.md
```diff
@@ -2095,6 +2238,49 @@ fall-behind detection.  An implementation can offer both, however, and
...
+The miner fee for the initiator will be 720 * 2500 / 1000 or 1800sats.
+
+Minimum funds that the initiator must contribute to the channel open
+transaction:
+
+	open_channel2.funding_satoshis: 50.0000sats
+	lease fee:			 3.3333sats
+	miner fee:			   1800sats
+
+	total required contribution:	53.5133sats
```
rustyrussell, line 2281: Rounded.... down?

07-routing-gossip.md
```diff
@@ -262,6 +262,18 @@ nodes not associated with an already known channel are ignored.
...
+
+1. `tlv_stream`: `node_ann_tlvs`
+2. types:
+   1. type: 1 (`option_will_fund`)
+   2. data:
+       * [`u16`:`funding_fee_proportional_basis`]
+       * [`u32`:`funding_fee_base_sat`]
+       * [`u16`:`funding_weight_max`]
+       * [`u16`:`channel_fee_proportional_basis_max`]
+       * [`tu32`:`channel_fee_base_max_msat`]
```
rustyrussell, line 275: Use subtype `lease_rates` directly? Yes, I know this creates a dep between the two BOLTs, but that's OK, and if it's not we fix our own tooling :)

rustyrussell: Oh, and it means we need to put signature field first in BOLT 2, since tu32 must be at end of TLV.

07-routing-gossip.md
rustyrussell: ... names don't match fields, particularly missing _max?

07-routing-gossip.md
```diff
@@ -359,6 +384,12 @@ to be ordered in ascending order, unknown ones can be safely ignored.
 Additional fields beyond `addresses` may also be added in the future—with
 optional padding within `addresses`, if they require certain alignment.
 
+If a node signals `option_will_fund`, they are signaling that they
+will provide funding to a node at the stated terms. They also commit
+to a feerate they will charge for transmitting funds over the channel
+for the duration of the funding lease.
+
+
 ### Security Considerations for Node Aliases
```
rustyrussell, line 393: Need a new requirement in `channel_update` that we don't send out feerates that violate any active lease.

02-peer-protocol.md
```diff
@@ -968,6 +997,42 @@ additions.
...
+            - MUST set `channel_fee_base_max_msat` to the base fee
+              (in millisatoshi) it will charge for any HTLC on this channel
+              during the funding period.
+            - MUST set `channel_fee_proportional_basis_max` to the amount
+              (in thousandths of a satoshi) it will charge per transferred
+              satoshi during the funding period.
+	    - MUST set `signature` to the ECDSA signature of
+	      SHA256("option_will_fund" || `funding_pubkey`|| `blockheight` ||
+	      `channel_fee_base_max_msat` || `channel_fee_proportional_basis_max`)
+              using the node_id key.
```
rustyrussell, line 1023: Let's make this `blockheight` + 4032?  That way if we ever make lease duration configurable, it doesn't matter, since this commits to the final block.

03-transactions.md
```diff
@@ -132,6 +154,19 @@ The output is spent by an input with `nSequence` field set to `1` and witness:
 
 Otherwise, this output is a simple P2WPKH to `remotepubkey`.
 
+##### Leased channel (`option_will_fund`)
+
+In a 'leased' channel, the `to_remote` output that pays the `accepter` node
+is modified so that it is equal to the greater of the
+1 or the `lease_end` - `blockheight`.
```
rustyrussell, line 161: Well, yeah.  If we define lease nomenclature it's easier to change this to say *why*:

`If a lease applies to the channel, the `to_remote` output of the `initiator` ensures the lessor won't get their funds until after lease expiry.`

03-transactions.md
```diff
@@ -120,6 +120,28 @@ If a revoked commitment transaction is published, the other party can spend this
 
     <revocation_sig> 1
 
+##### Leased channel (`option_will_fund`)
+In a leased channel, the `to_local` output that pays the `accepter` node
+is modified so that its CSV is equal to the greater of the
+`to_self_delay` or the `lease_end` - `blockheight`.
```
rustyrussell, line 126: See comment below on explaining why, not what?

rustyrussell (commented):
Wording nitpicks and terminology to make things clearer (esp definitions of lessor and active lease or whatever), and naming fixes.  Only two actual spec changes:

1. Request to use the same subtype in both init tlv and announce.
2. Maybe have them only pay for what they requested?
3. Make signature commit to *end* not *start* of lease.

07-routing-gossip.md
Kixunil: Good point, my thinking was to be consistent and sufficiently granular. I would like 10000ths more than 1000ths.

02-peer-protocol.md
Kixunil: Min sounds reasonable to me.

02-peer-protocol.md
Kixunil: As a European, I got used to interpreting `.` as a decimal point in programming-related texts. Can't say about others but I'd be surprised if there are many people who didn't do similarly. Personally, I get sometimes confused in Europe actually. :D

02-peer-protocol.md
niftynei: Not intentionally; I mean to signal that you MAY provide less (or more) than `requested_sats`.

Will update to `MAY send less`...

02-peer-protocol.md
niftynei: Good point, will update.

02-peer-protocol.md
niftynei: Yes this is sensible. :+1:

02-peer-protocol.md
niftynei: Right, this makes sense.

02-peer-protocol.md
```diff
@@ -1862,6 +1962,49 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+
+#### Requirements
+
+The node _responsible_ for initiating the channel:
+  - MUST send `update_blockheight` to ensure the CSV lock mandated by
+    `option_will_fund` will expire at the end of the funding lease.
+  - MUST send `update_blockheight` when their local blockheight updates
+
+The node _not responsible_ for initiating the channel:
+  - MUST NOT send `update_blockheight`.
```
niftynei, line 1988: note to self: if last received blockheight is > XXX, should fail the channel.

07-routing-gossip.md
```diff
@@ -312,6 +324,19 @@ The origin node:
...
+    for any reciprocated funding.
+  - MUST set `funding_fee_proportional_basis` to the amount (in
+    thousandths of a satoshi) it will charge per contributed satoshi.
+  - MUST set `funding_weight_max` to the maximum weight that will be charged
+    for any lease.
+  - MUST set `channel_fee_base_max_msat` to the maximum base fee
+    (in millisatoshi) it will charge for any HTLC during the funding period.
+  - MUST set `channel_fee_proportional_basis_max` to the max amount (in
+    thousandths of a satoshi) it will charge per transferred satoshi during
+    the funding period.
```
Kixunil, line 339: I'm thinking would it make sense to advertise a minimum required contribution to the channel capacity from the taker?

07-routing-gossip.md
niftynei: I updated this to say 'thousandths' -- generally it's advisable to expose these to end users as ppm though -- using parts-per-thousandths is a shortcut to reduce the size of the node_announcement TLV

07-routing-gossip.md
niftynei: What's the rationale here?

Including this would add another 4-8 bytes to the node_announcement, which we're attempting to keep as small as possible.

Kixunil: Without it one could open a channel with e.g. 100 sats and request 10M of liquidity. This is not a problem if the node charges for liquidity sufficiently but perhaps the node would be willing to charge less if there's liquidity in the opposite direction.

Maybe we could represent it as a percentage of taker contribution which would fit into one byte.

Alternatively, maybe instead of putting it into channel announcement there could be a way to get details by connecting to the node directly and thus not causing load on the rest of the network?

niftynei: >  Alternatively, maybe instead of putting it into channel announcement there could be a way to get details by connecting to the node directly and thus not causing load on the rest of the network?

￼the `init` message would fit this description; adding it to the init TLV is a superb idea. if you want to take a stab at drafting a PR for it, i can help get it added to clightning :)

07-routing-gossip.md
Kixunil: @niftynei do you mean that when exposed to user as ppm some rounding will be implemented or a more precise value will be negotiated when actually opening the channel?

07-routing-gossip.md
Kixunil: Glad that you like the idea! Unfortunately I'm both incredibly overwhelmed by other stuff and not familiar with C-lightning codebase. I'm mainly interested in the protocol because I believe it may solve some high-level problems I observe as a node operator.

proposals/010-will-fund-for-food.md
```diff
@@ -0,0 +1,89 @@
...
+for any node looking to accept inbound payments.
+
+## Overview
+
+This proposal adds a TLV to the `node_announcement`,
+which permits a node to advertise a rate for which they will contribute
+funds to an open channel request. Due to the nature of the protocol, these
+opens are only possible on the v2 channel protocol.
+
+Any advertised liquidity lease is for a duration of 4032 blocks, or
```
joostjager, line 16: Is this fixed lease duration that is also used in Lightning Pool really the best model? If I want to keep the channel after 28 days, but the lessor thinks differently because the funding fee has been collected already anyway, what can I do? I'd need to hit the chain again to restore my connection to the network.

An alternative could be to stay closer to the actual cost of keeping the channel open. For example: the lessee periodically pays interest calculated over the lessor's channel balance. These payments could be integrated as a channel update message with the channel being failed automatically if the contract is not followed.

niftynei: The rationale for presenting channel leases as fixed time block entities is to create a standard contract of liquidity leases that makes it easy to compare offers. In other words, it makes it easier to evaluate what certain liquidity offers are worth across the network.

> what can I do? I'd need to hit the chain again to restore my connection to the network.

I think what you're proposing here is a lease extension negotiation protocol -- this something that we can add on top of this existing initial protocol, and perhaps should be something to consider for splices (though that's also an on-chain operation). Which is to say, I think this is a really important point to make and that negotiations for "extending a channel lease" would be very welcome!

> the lessee periodically pays interest calculated over the lessor's channel balance.

Yes, as currently proposed, the period of the payment is 4032 blocks; a renewed interest payment would need to be made at the 4033'd block to keep the channel open. This deserves further investigation, but shouldn't be a blocker for shipping this initial lease proposal imo.

joostjager: I think it could work like that indeed. After each 4032 blocks, the lessor will demand an extension fee based on their average channel balance over the past period.

This could be a protocol that is totally independent from dual funding. Because the counterparty of a single-funder channel that is spent down will probably also consider closing unless they are compensated somehow.

03-transactions.md
```diff
@@ -120,6 +120,28 @@ If a revoked commitment transaction is published, the other party can spend this
...
+
+    OP_IF
+        # Penalty transaction
+        <revocationpubkey>
+    OP_ELSE
+        MAX(`to_self_delay`, `lease_end` - `blockheight`)
```
ariard: Do you need to encumber `leasor` (iiuc the counterparty offering the lease?)'s HTLC transactions outputs with OP_MAX too ?

Otherwise I think as a leasor I can send all my balance (minus `max_htlc_value_in_flight_msat`) in an offered HTLC output, of which I'm never going to give you the preimage. At expiration broadcast commitment + HTLC-timeout, wait for the CSV delay, and withdraw the balance to a single-owned output ?

proposals/010-will-fund-for-food.md
```diff
@@ -0,0 +1,89 @@
...
+
+A new message, `update_blockheight` is added to the Normal Operation of
+a channel's lifetime. As with `update_feerate`, only the channel opener
+may send this message.
+
+The CSV value for the commitment transaction outputs to the accepter
+are modified. They now incorporate the funding lease and time since
+the channel has been opened. As blocks are published, the CSV
+for the accepter's outputs is decremented. Once the funding lease
+has passed, the CSV value outputs return to the 'normal', non-leased values.
```
ariard, line 71: I think you should introduce a liveliness duty on the leavee ? I.e if you're not online every X, the leasor is allowed to close the channel, otherwise as the lease enforcement is relying on relative timelocks they're never going to be triggered and expire at the expected height of the end lease ?

02-peer-protocol.md
```diff
@@ -1862,6 +1962,49 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+
+A receiving node:
+  - if the `update_blockheight` is less than the last received `blockheight`:
```
ariard: "if the `lease_end - blockheight` is less than the receiver announced `to_self_delay` MUST fail the channel?" to avoid justice delays downgrads ?

ariard (commented):
Really cool proposal :) Still thinking more about timelocks/channel policies interactions.

proposals/010-will-fund-for-food.md
niftynei: > This could be a protocol that is totally independent from dual funding.

Totally, though I think we might be able to reuse the CSV locks presented here?

joostjager: The CSV locks indeed look useful to keep the lessor honest. Although there are still other ways to play unfair. One thing I thought about is for the lessor to lease out a channel and then push out all balance and stop routing. That way they collect the lease fee without having to commit capital for the duration of the channel. The CSV lock probably doesn't hurt that much anymore at that point. This obviously doesn't work if the lessee is not a routing node.

Maybe it is worth to already think through (not implement yet) what that lease extension protocol could look like. Only on the message level. With the goal to identify anything that must already be taken into account for `openchannel2`.

joostjager: Would it be possible to have the lessor and lessee switch roles during the lifetime of a channel? Initially A charges B for the liquidity, but over time, A gets low on liquidity and B now wants to charge A to keep the same channel open. Maybe they even mutually charge each other based on their average local channel balances?

niftynei: > and then push out all balance and stop routing

You could push out up to 99% of the channel balance, the reserve would still be locked in. Assuming the channel fee is <1% of the total channel balance (which is .. likely?) then you'd never be able to push out more than you'd originally locked in. Maybe this is fine?

It's recommended that lessors implement routing policies for the capital that they've leased -- e.g. if you're leasing the capacity with the aim of accepting payments, you may reconsider any routed payment that comes in via that channel.

02-peer-protocol.md
niftynei: Hmm I think I'm missing something about your point here. Anywhere there's a to_self_delay, the csv is max(to_self_delay, lease_end - blockheight).

proposals/010-will-fund-for-food.md
niftynei: Yes, iiuc this is encapsulated in the suggestion to close the channel if they haven't updated the `blockheight` in 1008 blocks.

02-peer-protocol.md
ariard: Right, you're using OP_MAX directly in the script? So should be good with concern of justice delay downgrades, but still you can save few bytes of `witnessScript` by checking this during message reception or transaction construction ?

niftynei:
The max is done prior to the construction of the script, the actual on
chain script is unchanged.

proposals/010-will-fund-for-food.md
niftynei: >  Initially A charges B for the liquidity, but over time, A gets low on liquidity and B now wants to charge A to keep the same channel open. Maybe they even mutually charge each other based on their average local channel balances?

This is interesting. Here's a few thoughts.

First, having a balance on your side of the channel represents an opportunity to earn routing fees. By charging your peer for the maintenance of a balance on your side of the channel, you're effectively charging them for inbound capacity. A fee for capital maintenance would drive up routing fees, as you're essentially trading off future cash from your channel partner for routing income. Plus there's the potential that routing a payment would then tip the scales in their favor such that you'd then be paying that routing income back over to them (given a long enough time scale).

Assuming zero routing takes place and over a long enough time frame, a policy which charges a fee on the un-capitalized side of the channel would mean that the non-dominant side of the channel would slowly leak funds to the larger side, resulting in an increasingly unbalanced channel. The only way to avoid this would be by maintaining 50-50 balanced channels. So basically, either you allocate a full 50% of a channel's capacity or you slowly lose balance in that channel.
 
One interesting side effect of this is that it'd incentivize a "mean reversion" on every channel -- e.g. there's an incentive to keep your channel balances at 50/50 to avoid losing fees to your peer. This may not be possible on every channel and would give a strong deadline for channel closure, as opposed to now, where most closures seem to be predicated around either a need to reallocate funds, a total usage of all channel funds, or whimsy.

02-peer-protocol.md
niftynei: The max is done prior to the construction of the script, the actual onchain script is unchanged.

proposals/010-will-fund-for-food.md
joostjager: > Assuming the channel fee is <1% of the total channel balance (which is .. likely?) then you'd never be able to push out more than you'd originally locked in

I am thinking of dual-funding where the leasor funds (almost) the complete channel. Then the leasor pushes out 99% and frees up their capital to repeat the same trick with the next lessee, and the next. You sell 100 channels and collect channel fees with the capital that you could normally only use to sell a single channel.

Kixunil: I'm not sure it matters. Let's look at various cases:

* The user has no prior channels - circularly routing is impossible - the lessor can't do that
* The user (U) already has a channel to another, higher-quality node (A) than the quality of lessor (L). L -> U -> A -> L causes improvement in quality aside from A becoming SPOF (likely the reason why U wanted another channel). If U charges sufficient fees, it can cover the cost of opening third channel. Note that if B -> L -> U is possible then B -> L -> A -> U is also possible even though a bit worse.
* A has same strategy as L and will also attempt to push out - will end up in balanced situation while U gets free money or one node routes and the other gives up, still U getting money. This scenario may be even incentive not to attempt it in the first place.
* A is lower quality than L and doesn't employ this strategy. The question is: is this a contradiction? A node that automatically destroys your ability to receive over its channel should probably be considered worse. Again, if the fee is sufficient it may pay for another channel.

07-routing-gossip.md
niftynei: The channel fee is expressed as "channel_fee_proportional_millionths" and is broadcast via the "channel_update" gossip message. The channel lease commits to a "channel_fee_max_proportional_thousandths". Let's call these "ppm" and "ppt" for short. 1 ppt is equal to 1,000ppm. Practically, setting a 'channel_fee_max_proportional_thousandth" to 1 would be a commitment to never set the "channel_fee_max_proportional_millionth" larger than 1000.

07-routing-gossip.md
```diff
@@ -262,6 +262,14 @@ nodes not associated with an already known channel are ignored.
    * [`32*byte`:`alias`]
    * [`u16`:`addrlen`]
    * [`addrlen*byte`:`addresses`]
+   * [`node_ann_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `node_ann_tlvs`
+2. types:
+   1. type: 1 (`option_will_fund`)
+   2. data:
+       * [`lease_rates`:`lease_rates`]
```
t-bast, line 271: How would you set `funding_weight` here?
It feels like a very dynamic value that would depend on `requested_sats` and your utxo set at channel open time, it's really hard to choose a value that would fit all requests...

02-peer-protocol.md
```diff
@@ -968,6 +997,45 @@ additions.
...
+	      `channel_fee_max_base_msat` || `channel_fee_max_proportional_thousandths`)
+              using the node_id key.
+
+The receiving node:
+  - SHOULD fail the negotiation if:
+    - they sent `request_funds` tlv and:
+      - the `funding_satoshi` is less than required
+      - the `lease_fee_basis` is too high
+      - the `lease_fee_base_sat` is too high
+      - the `funding_weight` is too high
```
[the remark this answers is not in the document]
Kixunil: > let the opener include in `request_funds` the maximum `funding_weight` they are prepared to sponsor

BIP78 has a similar feature and it recommends sponsoring one input. Sounds like same recommendation could apply here.

03-transactions.md
niftynei: Yes, this a hole in the protocol at the moment. We chose not to update the HTLC contracts, as that was a much larger change for the first draft. iiuc it also impacts the lockup for failed payments across the entire payment chain, which seems suboptimal as well.

You can limit your exposure to this today by limiting the max htlc inflight for this channel.

03-transactions.md
```diff
@@ -120,6 +120,32 @@ If a revoked commitment transaction is published, the other party can spend this
 
     <revocation_sig> 1
 
+##### Leased channel (`option_will_fund`)
+
+If a `lease` applies to the channel, the `to_local` output of the `accepter`
+ensures the `leasor` funds are not spendable until the lease expires.
+
+In a leased channel, the `to_local` output that pays the `accepter` node
+is modified so that its CSV is equal to the greater of the
```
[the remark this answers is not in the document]
niftynei: We used CSV since that was already introduced to the scripts for `option_anchor_outputs`, and therefore required less surgery to the onchain scripts. :angel: 

I could further make the argument that if your peer isn't actively updating the CSV lock, then it's a good signal to go ahead and shut the channel with them but that's not exactly a water-tight argument for keeping it as a CSV ...

07-routing-gossip.md
niftynei: The idea is that you pick a value that you want the opening peer to pay you for -- they're not going to pay you for more than that, input aggregation you should do on your own dime ;)

c-lightning defaults to charging for the weight of two P2WPKH inputs and a P2WPKH output.

02-peer-protocol.md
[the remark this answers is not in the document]
niftynei: > Why don't we let the opener include in request_funds the maximum funding_weight they are prepared to sponsor

In the c-lightning implementation, the way we handle mismatches between the accepter's response of their current rates vs the currently advertised rates it that, when you're offering to buy a channel lease from someone based on their advertised rates, you pass in the rates that you're expecting the peer to charge you. 

If the returned rates in `accept_channel2` don't exactly match what you expected, we fail the open and return the new rates to the opener's caller. They can then decide to retry using the new rates or find a lease from a different node.

03-transactions.md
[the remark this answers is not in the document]
niftynei: > Since the lease expires at an absolute block, why do you use a CSV and not a CLTV (which would remove the need for the update_blockheight mechanism)?

I really like this suggestion!! the `update_blockheight` hack feels very hacky imo haha. But I *do* like how cleanly it reuses the CSV lock that anchors introduced. hmm

02-peer-protocol.md
```diff
@@ -884,6 +885,10 @@ This message initiates the v2 channel establishment workflow.
    2. data:
        * [`u16`:`shutdown_len`]
        * [`shutdown_len*byte`:`shutdown_scriptpubkey`]
+   1. type: 3 (`request_funds`)
```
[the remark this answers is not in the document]
niftynei: true but you can also ignore it and be fine?

02-peer-protocol.md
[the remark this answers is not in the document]
niftynei: the other way to handle this of course would be for us to expose a hook that allows plugins to be written to more finely modify acceptance behavior if the accepter's rates have changed, of course.

02-peer-protocol.md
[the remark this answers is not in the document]
t-bast: True, it's just my OCD of seeing an unnecessary gap between our tlv field types :laughing:

03-transactions.md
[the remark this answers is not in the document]
t-bast: trade-offs

proposals/010-will-fund-for-food.md
```diff
@@ -0,0 +1,89 @@
...
+## Liquidity Advertisements
+
+## Problem Description
+
+Sourcing liquidity on the lightning network is a common problem
+for any node looking to accept inbound payments.
+
+## Overview
+
+This proposal adds a TLV to the `node_announcement`,
```
joostjager, line 11: Is it worth splitting this part off into its own PR? Extensible gossip probably has uses beyond liquidity ads.

renepickhardt:
Concept ACK 

I am however uncertain about how the choices for `lease_rates` (and other parameters) will influence the emerging market / routing game. We have seen that the choice to add the `base_fee` to the `routing_fees` [led to some nasty side effects](https://lists.linuxfoundation.org/pipermail/lightning-dev/2021-August/003174.html). I am currently studying [the price of anarchy and the game between routing nodes and sending nodes of payments](https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-May/003590.html). As pointed out I am not sure yet how the dominant strategies and nash equilibrium will look like but routing fees going to zero and routing nodes earning via liquidity ads may not be unlikely.  Thus I fear that some choices for the `lease_rates` can lead to higher / lower prices of anarchy if the dominant strategies are being applied by the participants of the routing game. As we saw in the `base_fee` case that changing the rules of the game (or in our terms the protocol) can be tricky at a later point I would love to understand the implications of the proposed rules of this game. Obviously we would like to create a market / game where the price of anarchy if nodes use dominant strategies is small / low. Unfortunately I cannot say how much time it will take to understand this properly.

morehouse (commented):
Cool stuff.

Overall I think we could simplify the proposal using CLTV as @t-bast suggested.

And in a future change we could address the HTLC problem in a fairly straightforward way (see comment below), if needed.  For now throttling the leased channel seems the simplest way to avoid getting cheated.  Maybe that will be good enough in practice...

03-transactions.md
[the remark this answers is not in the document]
t-bast: Agreed, I plan on prototyping the CLTV version soon(ish). I'm not 100% sure it will work, as we're already using the `nLockTime` of commitment txs to encode data, I'll report back when it's done.

03-transactions.md
```diff
@@ -120,6 +120,32 @@ If a revoked commitment transaction is published, the other party can spend this
...
+
+The output is spent by an input with `nSequence` field set to
+MAX(`to_self_delay`, `lease_end` - `blockheight`)
+(which can only be valid after that duration has passed) and witness:
+
+    <local_delayedsig> <>
+
 #### `to_remote` Output
 
 If `option_anchor_outputs` applies to the commitment transaction, the `to_remote` output is encumbered by a one block csv lock.
```
morehouse, line 151: We'll also need to add a lease term timelock to the `to_remote` output of the lessee.  Otherwise the lessor could goad the lessee into force-closing (e.g., by letting an HTLC time out) and get their funds back immediately.

03-transactions.md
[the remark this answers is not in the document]
morehouse: FYI - It looks like LND [already](https://github.com/lightningnetwork/lnd/blob/master/input/script_utils.go) uses CLTV leases for Pool.  They also have CLTVs added to HTLC-success and HTLC-timeout transactions, which partially plugs the HTLC hole @ariard pointed out.

I think this is a much simpler approach.

03-transactions.md
morehouse: Note that LND [already](https://github.com/lightningnetwork/lnd/blob/master/input/script_utils.go) adjusts `to_self` in the HTLC-timeout and HTLC-success transactions for Pool.  This partially plugs the hole, but the lessor can still evade the timelock by routing an HTLC to itself and then goading the lessee into force closing.

The solution I described above should block this case as well, but I'm thinking just having HTLC-success and HTLC-timeout lease locks may be enough for the first protocol.  It would be nice to avoid the more complex HTLC surgery.

03-transactions.md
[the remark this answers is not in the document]
t-bast: > Agreed, I plan on prototyping the CLTV version soon(ish). I'm not 100% sure it will work, as we're already using the nLockTime of commitment txs to encode data, I'll report back when it's done.

I don't know what I was smoking when I wrote that message, using CLTV of course works, we don't care about the `nLockTime` of the commitment tx, the `nLockTime` that matters is the one of the transaction that spends the leasor's main output.

02-peer-protocol.md
```diff
@@ -968,6 +997,45 @@ additions.
...
+            - MUST set `funding_weight` to the weight they will charge
+              the peer for.
+            - MUST set `channel_fee_max_base_msat` to the base fee
+              (in millisatoshi) it will charge for any HTLC on this channel
+              during the funding period.
+            - MUST set `channel_fee_max_proportional_thousandths` to the amount
+              (in thousandths of a satoshi) it will charge per transferred
+              satoshi during the funding period.
+	    - MUST set `signature` to the ECDSA signature of
+	      SHA256("option_will_fund" || `funding_pubkey`|| `blockheight` + 4032 ||
```
t-bast, line 1021: How are those encoded? How many bytes should the `blockheight` part use?

02-peer-protocol.md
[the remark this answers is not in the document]
niftynei: We'll just make the number line longer??

07-routing-gossip.md
```diff
@@ -262,6 +262,14 @@ nodes not associated with an already known channel are ignored.
    * [`32*byte`:`alias`]
    * [`u16`:`addrlen`]
    * [`addrlen*byte`:`addresses`]
+   * [`node_ann_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `node_ann_tlvs`
+2. types:
+   1. type: 1 (`option_will_fund`)
```
joostjager, line 269: You are claiming NUMBER ONE here? 😅

03-transactions.md
niftynei: > We could use the multisig + secondary transaction trick we already use with HTLC-success and HTLC-timeout to avoid cltv_expiry lockup across the payment chain. But yes this would complicate the proposal a bit.

I think I'm going to propose doing exactly this (creating second stage txs for the leasor in their remote's commitment tx HTLC outputs), as it's a complete patch for this hole. It is more surgery, but I think worth it in terms of eradicating a potential hole in the protocol. 

I'll also add HTLC locks on the to-self HTLC second stage txs, as you've suggested (and LND does!)

02-peer-protocol.md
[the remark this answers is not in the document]
niftynei: `type 2` is now used by `require_confimed_inputs`, marking this as resolved :)

t-bast:
As discussed offline, I think we should add the `request_funds` and `will_fund` TLV fields to `tx_init_rbf` and `tx_ack_rbf`. This is especially useful when combining liquidity ads with splicing, because we will heavily use `tx_init_rbf` to replace an unconfirmed splice with a new one that potentially represents a different set of user actions, and it would be beneficial to be able to include a liquidity request in such a replacement.

02-peer-protocol.md
```diff
@@ -1247,10 +1283,43 @@ The accepting node:
...
+      contributions:
+      - SHOULD omit the `will_fund` tlv
+    - if they decide to accept the offer:
+      - MUST include a `will_fund` tlv
+      - MUST set `funding_satoshis` to a value greater than 0msat
+      - MAY send `funding_satoshis` less than `requested_sats`
+      - MUST set `lease_fee_base_sat` to the base fee
+        (in satoshi) it will charge for the `funding_satoshis`
+      - MUST set `lease_fee_basis` to the amount
+        (in thousandths of satoshi) it will charge per `funding_satoshi`
```
t-bast, line 1297: I thought this was basis points, thus ten thousandths of satoshi? Same for `channel_fee_basis_max` below. The formula used in the examples below uses 1/10_000.

02-peer-protocol.md
```diff
@@ -1235,6 +1260,17 @@ acceptance of the new channel.
...
+   2. data:
+       * [`signature`:`signature`]
+       * [`lease_rates`:`lease_rates`]
+
+1. subtype: `lease_rates`
+2. data:
+       * [`u16`:`funding_weight`]
+       * [`u16`:`lease_fee_basis`]
+       * [`u16`:`channel_fee_basis_max`]
+       * [`u32`:`lease_fee_base_sat`]
```
t-bast, line 1273: We can keep channel base fees in a way where we can omit it thanks to truncated ints:

```suggestion
       * [`u32`:`lease_fee_base_sat`]
       * [`tu32`:`channel_fee_base_msat_max`]
```

02-peer-protocol.md
```diff
@@ -1263,6 +1332,49 @@ Instead, the channel reserve is fixed at 1% of the total channel balance
 rounded down to the nearest whole satoshi or the `dust_limit_satoshis`,
 whichever is greater.
 
+If the opener requested a `option_will_fund` peer to contribute funds to
```
t-bast, line 1335: `option_will_fund` doesn't exist, this doesn't need a feature bit since it instead relies on a TLV in `node_announcement`.

02-peer-protocol.md
```diff
@@ -1263,6 +1332,49 @@ Instead, the channel reserve is fixed at 1% of the total channel balance
...
+total lease fee at a rate of `funding_weight` times the
+`funding_feerate_perkw` established in `open_channel2`.
+
+The total lease fee is added to the accepter node's output in the
+commitment transaction. It is equal to the `lease_fee_base_sat` +
+min(`accept_channel2`.`funding_satoshis`,`open_channel2`.`requested_sats`)
+* `lease_fee_basis` / 10_000 + `funding_feerate_perkw` *
+`funding_weight` / 1000, rounded down to the nearest satoshi. See [Appendix A: The lease fee](#appendix-a-the-lease-fee).
+
+The `channel_fee_max_base_msat` is an implicit limit set
```
t-bast, line 1354: Remove this section once we restore `channel_fee_max_base_msat` in `lease_rates`.

02-peer-protocol.md
```diff
@@ -2283,6 +2395,12 @@ be trimmed at the updated feerate, this could overflow the configured
 `max_dust_htlc_exposure_msat`. Whether to close the channel preemptively
 or not is left as a matter of node policy.
 
+The lessor's lease is only shortened when a new `blockheight` is committed,
```
t-bast, line 2398: This paragraph is obsolete, it only made sense when using CSVs?

03-transactions.md
```diff
@@ -223,6 +271,45 @@ If a revoked commitment transaction is published, the remote node can spend this
 
 The sending node can use the HTLC-timeout transaction to timeout the HTLC once the HTLC is expired, as shown below. This is the only way that the local node can timeout the HTLC, and this branch requires `<remotehtlcsig>`, which ensures that the local node cannot prematurely timeout the HTLC since the HTLC-timeout transaction has `cltv_expiry` as its specified `locktime`. The local node must also wait `to_self_delay` before accessing these funds, allowing for the remote node to claim these funds if the transaction has been revoked.
 
+### Leasee Offered HTLC Outputs (`option_will_fund`)
```
t-bast, line 274: As discussed on the mailing list, I think we should remove this, because we have no way of actually exchanging signatures for those transactions unless we add a round-trip, which isn't desirable. The leasee should adapt their `max_accepted_htlc` and `max_htlc_value_in_flight_msat` if they're afraid that the seller will cheat.

t-bast:
After thinking about this more, I have more feedback :innocent:. As mentioned in my [mailing list post](https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-December/004227.html), I think we should make the lease enforcement through CLTV optional. We should also take the duration into account in the lease rates. Here is what I think the types should be:

- `lease_rate`:
  - [`u16`:`lease_duration`] -> in blocks (up to a year-ish, I don't think we should use a u32 and allow longer values?)
  - [`u16`:`lease_fee_basis`]
  - [`u32`:`lease_fee_base_sat`]
- `lease_rates`: included in `node_announcement` and `init`, default rates that the node will use, but there's no guarantee that this is what will be offered to you when you actually do a `request_funds`:
  - [`u16`:`funding_weight`]
  - [`u8`:`rates_count`]
  - [`rates_count*8*byte`:`rates`] -> we offer multiple rates depending on the duration (e.g. 1 week = 0,5% with 50 sat flat, 1 month = 1% with 75 sat flat, 1 year = 5% with 2000 sat flat)
  - [`u16`:`max_channel_fee_basis`]
  - [`tu32`:`max_channel_fee_base_msat`]
- `request_funds`:
  - [`u32`:`requested_sats`]
  - [`u16`:`lease_duration`] -> must exactly match one of the `lease_duration`s proposed by the seller
  - [`u32`:`lease_expiry`]
- `will_fund`:
  - [`64*byte`:`signature`] -> signature of the `lease_witness`
  - [`byte`:`enforce_strict_lease`] -> boolean that tells whether we add the lease CLTV to the seller's main outputs or not
  - [`u16`:`lease_fee_basis`]
  - [`u32`:`lease_fee_base_sat`]
  - [`u16`:`max_channel_fee_basis`]
  - [`tu32`:`max_channel_fee_base_msat`]
- `lease_witness`:
  - [`u16`:`funding_script_size`]
  - [`funding_script_size`:`funding_script`] -> to be compatible with taproot channel, we should commit to the script of the funding output instead of just the `funding_pubkey`?
  - [`u16`:`lease_duration`]
  - [`u32`:`lease_expiry`]
  - [`u16`:`max_channel_fee_basis`]
  - [`tu32`:`max_channel_fee_base_msat`]

ariard:
@t-bast why do you ask niftynei opinion when asking to close her issues, but not mine when closing my own issues like #1043 ?

ariard:
Okay, just as a matter of courtesy better to ask people before to force-close an issue, wait for weeks for a status update and if no answer, then close it. More or less what folk do on the core side with the “shill, kill or merge” tradition.

PR #881, opened 2021-06-28
Peer backup storage (feature 36/37)

t-bast:
Nodes can offer to altruistically store small amounts of data on behalf of their channel peers.
It complements `option_data_loss_protect` and can let nodes that lost data fully recover their channels without any on-chain transaction.

The contents of the `peer_backup` is left unspecified to offer wide flexibility for implementations. It can for example contain an encrypted version of your internal channel state, or a mix of data from several channels (if you want to shard your backups across multiple nodes to maximize your chances of recovery).

The version we currently use for Phoenix only needs to send a `peer_backup` in messages that change the channel state, but maybe we want to extend this optional TLV and include it in more messages? Do you see use-cases where that would be useful?

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -25,6 +25,7 @@ operation, and closing.
       * [Completing the Transition to the Updated State: `revoke_and_ack`](#completing-the-transition-to-the-updated-state-revoke_and_ack)
       * [Updating Fees: `update_fee`](#updating-fees-update_fee)
     * [Message Retransmission: `channel_reestablish` message](#message-retransmission)
+    * [Peer Backup Storage](#peer-backup-storage)
   * [Authors](#authors)
 
 # Channel
@@ -379,6 +380,12 @@ This message introduces the `channel_id` to identify the channel. It's derived f
     * [`channel_id`:`channel_id`]
     * [`signature`:`signature`]
 
+1. `tlv_stream`: `funding_signed_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
+
 #### Requirements
 
 Both peers:
@@ -480,6 +487,12 @@ along with the `scriptpubkey` it wants to be paid to.
    * [`u16`:`len`]
    * [`len*byte`:`scriptpubkey`]
 
+1. `tlv_stream`: `shutdown_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
+
 #### Requirements
 
 A sending node:
@@ -558,6 +571,12 @@ the channel.
    * [`channel_id`:`channel_id`]
    * [`u64`:`fee_satoshis`]
    * [`signature`:`signature`]
+
+1. `tlv_stream`: `closing_signed_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
 
 #### Requirements
 
@@ -1002,6 +1021,12 @@ sign the resulting transaction (as defined in [BOLT #3](03-transactions.md)), an
    * [`u16`:`num_htlcs`]
    * [`num_htlcs*signature`:`htlc_signature`]
 
+1. `tlv_stream`: `commitment_signed_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
+
 #### Requirements
 
 A sending node:
@@ -1065,6 +1090,12 @@ The description of key derivation is in [BOLT #3](03-transactions.md#key-derivat
    * [`channel_id`:`channel_id`]
    * [`32*byte`:`per_commitment_secret`]
    * [`point`:`next_per_commitment_point`]
+
+1. `tlv_stream`: `revoke_and_ack_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
 
 #### Requirements
 
@@ -1179,6 +1210,12 @@ messages are), they are independent of requirements here.
    * [`u64`:`next_revocation_number`]
    * [`32*byte`:`your_last_per_commitment_secret`]
    * [`point`:`my_current_per_commitment_point`]
+
+1. `tlv_stream`: `channel_reestablish_tlvs`
+2. types:
+    1. type: 1 (`peer_backup`)
+    2. data:
+        * [`...*byte`:`backup_data`]
 
 `next_commitment_number`: A commitment number is a 48-bit
 incrementing counter for each commitment transaction; counters
@@ -1382,6 +1419,31 @@ however), but the disclosure of previous secret still allows
 fall-behind detection.  An implementation can offer both, however, and
 fall back to the `option_data_loss_protect` behavior if
 `option_static_remotekey` is not negotiated.
+
+## Peer Backup Storage
+
+Nodes that advertise the `peer_backup_storage` feature offer storing arbitrary
+data for their peers. The data stored must fit in a lightning message, so it is
+inherently limited (less than 65535 bytes per channel).
+
+Nodes can verify that storage providers correctly store their backup data at
+each reconnection, by comparing the contents of the `peer_backup` of the
+`channel_reestablish` message to the last one they sent.
+
+A node connected to a storage provider:
+  - when it sends a message that completes an update of the channel state
+    (`funding_signed`, `commitment_signed`, `revoke_and_ack`, `shutdown` or
+    `closing_signed`):
+    - MAY include an optional `peer_backup` TLV field
+  - when it receives `channel_reestablish` with an outdated or missing
+    `peer_backup`:
+    - SHOULD fail the channel
+
+A storage provider:
+  - when it receives a `peer_backup`:
+    - MUST store this backup data
+  - when it sends `channel_reestablish`:
+    - MUST include the last `peer_backup` it received for that channel
 
 # Authors
 
--- a/09-features.md
+++ b/09-features.md
@@ -40,7 +40,8 @@ The Context column decodes as follows:
 | 18/19 | `option_support_large_channel`   | Can create large channels                                 | IN       |                   | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
 | 20/21 | `option_anchor_outputs`          | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)         |
 | 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC transactions   | IN       |                   | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-sighash-single-harmful]|
-| 26/27 | `option_shutdown_anysegwit`         | Future segwit versions allowed in `shutdown`              | IN       |                   | [BOLT #2][bolt02-shutdown]   |
+| 26/27 | `option_shutdown_anysegwit`         | Future segwit versions allowed in `shutdown`           | IN       |                   | [BOLT #2][bolt02-shutdown]   |
+| 36/37 | `peer_backup_storage`               | Allow storing encrypted peer backup data               | IN       |                   | [BOLT #2][bolt02-peer-backup]   |
 
 ## Requirements
 
@@ -84,6 +85,7 @@ This work is licensed under a [Creative Commons Attribution 4.0 International Li
 [bolt02-open]: 02-peer-protocol.md#the-open_channel-message
 [bolt03-htlc-tx]: 03-transactions.md#htlc-timeout-and-htlc-success-transactions
 [bolt02-shutdown]: 02-peer-protocol.md#closing-initiation-shutdown
+[bolt02-peer-backup]: 02-peer-protocol.md#peer-backup-storage
 [bolt04]: 04-onion-routing.md
 [bolt07-sync]: 07-routing-gossip.md#initial-sync
 [bolt07-query]: 07-routing-gossip.md#query-messages


SomberNight:
Thank you for writing this up!
In general I like this proposal very much, and we would definitely want to implement this (or whatever it evolves into) into Electrum.

-----

On first read I had assumed that the backup blobs are per-peer but then realised they are actually per-channel.

One thing I would like to note is the usecase where, after restoring from seed, wallet software might be able to find some of their channel peers but not the channels themselves. i.e. Alice's wallet might know they have some channels with Bob but not know any of the channel ids (funding outpoints). In such a case Alice can establish a transport with Bob, and then wait for Bob to send one or more `channel_reestablish` messages. As there is no way for Alice to "request" the backup blobs (and she does not even know what the channel ids are), she must rely on Bob sending the `channel_reestablish` messages (regardless of who initated the transport).

-----

One thing I am not clear is the (a)symmetry of this feature. I think this needs to be cleared up.
The PR text talks about "storage provider"s, and suggests an asymmetric relationship where e.g. a forwarding note offers storing backups for a light client. However, negotiating a feature such as `peer_backup_storage` is inherently symmetric with current BOLT-09. So how exactly does this look like?

Let's say there is a light client who wants to store backups with their peer (Carol); and let's say the light client does not want to store backups for the other party (or does it need to? it could I guess). Whould the light client set `peer_backup_storage` in e.g. the `init` message? Would Carol set it? How would Carol know that they are the "service provider" in this relationship?
Or is it that they both must store blobs for the other one, and then the relationship is symmetric?

I am also wondering if the feature can even be used in a symmetric way.
If two counterparties of a channel, Alice and Bob, want to store backups with each other, and Alice loses state, then she will not be able to send the last backup sent by Bob. So assuming the backup blobs are used to in a Phoenix-like usecase (resume channel operation after restoring from seed); if the feature was used in a symmetric way (Alice storing blobs at Bob, and Bob storing blobs at Alice), then if either party loses state, the channel would get closed by the other party, I guess(?).

t-bast:
Thanks @SomberNight this is very good feedback. I've started with the minimal approach we use in Phoenix, which is simpler than the general case because Phoenix is always fundee and initiates reconnections, but this PR is a good opportunity to offer a backup mechanism that supports more scenarios.

I think it would make sense to have two distinct types of backups, which would address the scenario described in your first paragraph:

- channel backups (one per channel, as described in https://github.com/lightningnetwork/lightning-rfc/pull/881/commits/d87341663244da63833f3d37a4018674447ae7c3), which may be updated at every commitment change
- a node backup which can contain data about channels and peers

We could store in the node backup the list of our peers and the channel IDs we have with each peer (encrypted of course). That node backup could be sent to all of our peers: this way we only need to remember and reconnect to one of them to discover the others and the list of our channels.

We would need a new message to ask our peers to store an updated version of our `node_backup_data`. We would send that message for example whenever we've opened a new channel, to add the new channel ID to our backup. Our storage provider peers would send us back the latest version of that `node_backup_data` in `init`. At a high-level, does that sound like it would address your first comments?

Regarding the asymmetry of this feature, I think it is both desired and representative of real-world scenarios.
Let's detail for example the scenario where a mobile wallet Alice connects to a storage provider Bob.
Alice would *not* turn on the `peer_backup_storage` feature, because Alice doesn't offer to store people's data.
Bob would turn on the `peer_backup_storage` feature, which tells Alice that she can send her backups.

You're completely right to highlight that the case where both peers store each other's backups doesn't work: if one of them loses data, that will force the other to close channels. I don't think this feature is meant to be used between routing nodes running on servers. This type of node should implement DB redundancy internally to ensure they never lose data (instead of relying on their peers for that). I think this feature should only be used for mobile wallets (or more generally light clients) connecting to "service providers". But your comment made me realize that if two service providers connect to each other they will both have the `peer_backup_storage` feature activated but should not send backups to each other. I'm wondering if this should simply be covered by implementations (only mobile wallet implementations would actually send their backup data, server implementations would never send them nor expect peers to store their data) or if we should use distinct feature bits for "I can store your data" and "I'd like you to store my data" (but with the somewhat complex restriction that you cannot have two peers turn on both at the same time otherwise they'd run in the issue you mention if one of them loses data). What do you think?

SomberNight:
> We would need a new message to ask our peers to store an updated version of our `node_backup_data`. We would send that message for example whenever we've opened a new channel, to add the new channel ID to our backup. Our storage provider peers would send us back the latest version of that `node_backup_data` in `init`. At a high-level, does that sound like it would address your first comments?

Yes, that sounds good. I think it might not be strictly needed, as a workaround could be the light client always waiting for the remote sending `channel_reestablish` first; and this kind of data (list of your peers and list of channel ids) could be stored as part of the per-channel blob. However, that is a bit hackish and I think it would be worth it to introduce this new message for per-peer backups; IMHO it would even simplify implementations overall. Also, having both per-channel and per-peer backups could enable more use cases in the future. For example, not tying some backups to a specific channel could allow a market where nodes could pay peers they don't have a channel with to store data for them.

> if two service providers connect to each other they will both have the `peer_backup_storage` feature activated but should not send backups to each other. I'm wondering if this should simply be covered by implementations (only mobile wallet implementations would actually send their backup data, server implementations would never send them nor expect peers to store their data) or if we should use distinct feature bits for "I can store your data" and "I'd like you to store my data" (but with the somewhat complex restriction that you cannot have two peers turn on both at the same time otherwise they'd run in the issue you mention if one of them loses data). What do you think?

I think one feature bit is sufficient. It could behave as follows, given a transport between Alice and Bob:
- if Alice has the feature set, Bob does not: Bob can send backups to be stored at Alice, Alice must not
- if Alice and Bob both have the feature set, neither of them is allowed to send backups to be stored by the other
So the feature bit means "I am willing to store your backups", but if both parties set it then neither can utilise it.
So light clients would never set the feature, "service providers" could always set it (if they implement it); and if two service providers peer, they would recognise it.

One thing to note here is that light clients might want to "require" this feature but there isn't really a way to communicate that fact. Maybe that is fine as they could simply not peer with nodes that do not set it.

-----

Another point is about deleting backups. If Alice ("light client") and Bob ("service provider") have a single channel, and Alice stores backups with Bob, and Bob e.g. force-closes the channel, would Bob now delete Alice's backup? I think Bob could keep the backups for some time or maybe even indefinitely (on a best effort basis) - as even then the service seems hard to DOS: opening channels is probably costly enough even if they end up being closed. I suggest the spec include a recommendation about what to do if there no longer are any open channels with a peer but the peer has stored backups with us.

-----

One higher level point re use cases: I am wondering what the assumptions are when using this for "resumable channels". e.g. restoring from seed and discovering your existing channels, keeping them open and keep using them. Phoenix does this of course; but in that case the clients probably already have some trust in the ACINQ node... What do you think would be the implications of using it with arbitrary nodes? I think it might only work well if there is at least *some* trust. Unfortunately I don't think the game theory works out well: if the channel funder is the light client (user), there is virtually no risk for the remote for sending an old blob to the client on reestablish. They only risk missing out on future forwarding fees... their potential gain compared to that is huge :/

t-bast:
> I think one feature bit is sufficient. 

Yes, I think so as well, adding a second one wouldn't be a very clean "fix".

> One thing to note here is that light clients might want to "require" this feature but there isn't really a way to communicate that fact. Maybe that is fine as they could simply not peer with nodes that do not set it.

Exactly, I think that instead of requiring it, light clients would just scan `node_announcement`s and connect only to nodes that advertised the feature (if they need peers that can store their backups).

> I suggest the spec include a recommendation about what to do if there no longer are any open channels with a peer but the peer has stored backups with us.

This is a good point, the spec should mention that storage providers should (best effort) keep the latest backups stored even after closing the channel. That's what eclair currently does, as it doesn't cost much space (and as you mention, it's not subject to DoS since the size of backups is limited by the message size and there is an economic cost in the channel creation).

I'll add a section mentioning that.

> What do you think would be the implications of using it with arbitrary nodes? I think it might only work well if there is at least some trust. Unfortunately I don't think the game theory works out well: if the channel funder is the light client (user), there is virtually no risk for the remote for sending an old blob to the client on reestablish. They only risk missing out on future forwarding fees... 

I agree with you, this backup mechanism isn't completely bullet-proof. You can check that your peer behaves correctly at each `channel_reestablish` and `init`, but if they misbehave they won't be directly economically penalized (there's no penalty fee that you can automatically claim). As you mention, the only drawback for the cheating storage provider is that you won't use their services, and will probably tell your friends to avoid them as well. Is that a good enough incentive? I don't know, but I'm not sure we can do better...

I'll work on adding a commit to include the `node_backup_data` and the new associated message.

t-bast:
@SomberNight I add the `node_backup` and several other requirements and explanations in https://github.com/lightningnetwork/lightning-rfc/pull/881/commits/18bd2af1e4ac91db10b0ba1c89634dfec17178ec, let me know if that's better.

01-messaging.md
```diff
@@ -443,6 +449,60 @@ every message maximally).
 Finally, the usage of periodic `ping` messages serves to promote frequent key
 rotations as specified within [BOLT #8](08-transport.md).
 
+## Node Backup Storage
+
+Nodes that advertise the `peer_backup_storage` feature offer storing arbitrary
+data for their peers. The data stored must not exceed 32000 bytes, which lets
+it fit in existing lightning messages (e.g. a `commitment_signed` message with
+483 htlc signatures).
```
t-bast, line 457: Can you please double-check my calculations here? I *think* the biggest message that should be taken into account is `commitment_signed` with 483 HTLCs, with takes a bit more than 31000 bytes, so 32000 bytes is good (it leaves a bit more than 2000 bytes for future HTLC fields we'll add).

rustyrussell:
I like this!  But I think we should downgrade it to strictly "best effort", at least for peer_backup.

1. If they give you an old state, don't get upset.  This can happen anyway, since there's no ack.
2. Require only best effort, not require them to perform a db commit on every new update.
3. Perhaps we should have an "append" flag to reduce data traffic?  (The total size must be less than 32000 bytes still?).

The way I would use this feature is to store the commitment number in every peer_backup, and update at least three peers on every send of revoke_and_ack.  *That* is the critical piece of information so we don't accidentally go backwards on restore.

Then have a rule on restart that insists on a threshold of init msgs before sending reestablish messages.

I also think it would be informative (perhaps in a new doc) to indicate the format implementations used: a TLV seems to fit pretty well here, IMHO, but details of encryption matter too.

rustyrussell:
Features clash with #759  BTW!

t-bast:
> I like this! But I think we should downgrade it to strictly "best effort", at least for peer_backup.
> If they give you an old state, don't get upset. This can happen anyway, since there's no ack.

You're right, I made it best effort in https://github.com/lightningnetwork/lightning-rfc/pull/881/commits/3129979beb48ab983e6028d6a8e094d7984ddbdd

> Perhaps we should have an "append" flag to reduce data traffic?

It's worth discussing, if we think it makes sense we should include it now in the backup tlv (to avoid adding a new tlv later for it).
I'm afraid it's useless with encrypted backups, so it only makes sense for unencrypted data (array of `(channel_id, commitment_number)`), but I personally would only send encrypted data to avoid leaking anything (just in case).

> I also think it would be informative (perhaps in a new doc) to indicate the format implementations used: a TLV seems to fit pretty well here, IMHO, but details of encryption matter too.

I agree, bLIPs are likely a good place for that.

> Features clash with #759 BTW!

Fixed in https://github.com/lightningnetwork/lightning-rfc/pull/881/commits/42e68e0d85a31a7132f8f8a6aededa48b3ae553d

---

I realized that my calculation for the maximum size for backups is completely wrong...
I considered a `commitment_signed` message with 483 htlc signatures, but it can actually contain two times that since there can be 483 incoming htlcs **and** 483 outgoing htlcs :facepalm: (even though nodes may configure a lower `max_accepted_htlcs` to ensure there is some space available for backup data).

When that happens, there isn't much space available in the `commitment_signed` message for backup data (especially if we want to leave some space for future additional tlv fields). Maybe that's ok as you will soon follow up with a `revoke_and_ack` where you'll have enough space to send your backup data...

Should we still define an arbitrary maximum size for backup data (e.g. 32768 bytes, powers of two are nice) and add a requirement that the sender must not include the backup if that would cause the message to be bigger than the 65535 bytes message size limit? Or should we completely drop the maximum size for backup data and let it be bounded by the message size limit (65535 bytes)?

btcontract:
@t-bast 

> We could store in the node backup the list of our peers and the channel IDs we have with each peer (encrypted of course). That node backup could be sent to all of our peers: this way we only need to remember and reconnect to one of them to discover the others and the list of our channels.

It seems to me this is actually the most important part at least for as long as we need to store revoked commit history for each channel. That history can't be attached to channel messages TLV and restoring and then keep using channels without it is obviously dangerous, they better be closed ASAP.

OTOH an encrypted list of peers is small and wallet user won't need to remember which peers they had a channels with on wallet recovery, which is very useful.

rustyrussell:
Don't add TLVs everywhere, simply use a separate message?  That increases the limit (to its natural amount), as well.

SomberNight:
> Don't add TLVs everywhere, simply use a separate message?

The idea is that having TLVs as part of `commitment_signed`, `revoke_and_ack`, etc, allows atomically updating the channel backups with the peer, which is not possible with a separate message.
This allows using the channel backups to store our own latest state with the remote, making it possible to e.g. keep using a channel after restoring from seed -- although this requires either some trust in the peer or cross-checking against some other source regarding what the latest commitment numbers are. I expect this is how Phoenix works when restoring from seed (and the ACINQ node is trusted not to replay an old state).

t-bast:
> The idea is that having TLVs as part of commitment_signed, revoke_and_ack, etc, allows atomically updating the channel backups with the peer, which is not possible with a separate message.

Exactly, that's the argument for doing it in existing messages instead of defining new ones.

However we simply cannot guarantee how much space will be available in existing messages (`commitment_signed` really is an issue when there are many pending HTLCs), so the whole scheme could fall apart, whereas having dedicated new messages would ensure we can always transmit backups.

We have to choose our trade-off here:

- if we reuse existing messages, we must limit the number of pending htlcs to a value below 483 and backups cannot exceed whatever space would be available in a worst-case `commitment_signed` message. Note that we cannot really evaluate what the worst-case size of a `commitment_signed` message is because we may need to add new tlvs to the extension tlv stream in the future...
- if we use separate messages, we're not atomic, but we can have strict requirements on backup sizes

t-bast:
I rebased and squashed this PR. I also moved to using separate messages instead of putting the backups in the tlv extension of existing messages, because @pm47 made a really good comment that trivially addresses the atomicity issue: if we send the updated backups *before* the message that updates the channel state (e.g. `commit_sig`) we can provide atomicity without needing to piggy-back on existing messages.

02-peer-protocol.md
```diff
@@ -1515,6 +1522,51 @@ fall-behind detection.  An implementation can offer both, however, and
...
+    - when it receives `channel_reestablish` with an outdated or missing `channel_backup`:
+      - MAY send a warning
+
+A node with `provide_peer_backup_storage` activated:
+  - when it receives `update_channel_backup`:
+    - if the `channel_backup` exceeds 32000 bytes:
+      - SHOULD send a warning
+      - MUST NOT store this backup data
+    - otherwise:
+      - SHOULD store this backup data
```
SomberNight, line 1555: why is this not a "MUST"?

Also, in BOLT-01 you write
> However, nodes should not expect
their peers to always have their latest backup available (note that there is no
acknowledgement to ensure a backup has been correctly stored).

and below you write
> The `update_channel_backup` should be sent before the message that updates the
channel state (e.g. `commit_sig`) to ensure that the backup is stored with that
latest update, which can be retransmitted on reconnection. If the backup is
sent after the channel update, there is a risk that the channel update is
committed by the remote peer but not the corresponding backup, which may lead
to data loss.

It might be worth specifying that the backup data should be persisted to disk in a blocking manner, before processing subsequent messages.
In particular, imagine a weird architecture where you have a thread that is responsible for persisting backup data blobs to disk and you keep queueing items onto this thread. As such a thread could lag behind the channel state machine (revack, commit_sig) by multiple messages, and we would lose all guarantees about how recent the latest stored backup is compared to the channel state -- ideally (it should be +/-1)

SomberNight (commented):
Looks good! Just one comment inline.

> I also moved to using separate messages instead of putting the backups in the tlv extension of existing messages, because @pm47 made a really good comment that trivially addresses the atomicity issue: if we send the updated backups before the message that updates the channel state (e.g. commit_sig) we can provide atomicity without needing to piggy-back on existing messages.

Ah, such a simple solution! Hard to unsee now :) I suppose it means ~halving the available max space as both the current and the next state should be bundled into a single blob, but that's fine.

02-peer-protocol.md
t-bast: > why is this not a "MUST"?

This was made a `SHOULD` based on one of rusty's earlier comment that since this feature is free, you cannot expect your peer to provide a perfect SLA, it's a best effort on their part. In practice I agree with you that for many cases you will really expect your peer to perfectly store your backups (because you're indirectly paying them by connecting to them and creating payment traffic through them), so I don't know whether we should make it a `MUST` or not.

> It might be worth specifying that the backup data should be persisted to disk in a blocking manner, before processing subsequent messages.

Good idea, I'll add something like that.

t-bast:
> I suppose it means ~halving the available max space as both the current and the next state should be bundled into a single blob, but that's fine.

I don't think it does! It depends on what you put in the backup exactly, but for example for `commit_sig` and `revoke_and_ack` you only need to put the next backup: if it contains your sig and the details of the revocation, you can discard the previous one, the backup contains what you need to re-transmit `commit_sig` / `revoke_and_ack` on reconnection if necessary. We'd need to double-check for each state transition, but I believe this should work exactly as if it were atomically sent with the channel message.

02-peer-protocol.md
t-bast: I have slightly extended the requirement in https://github.com/lightning/bolts/pull/881/commits/cac404d2eb7c43b6615c5a0cf798738a57868421, if you have a better wording suggestion don't hesitate to share it!

SomberNight: > This was made a SHOULD based on one of rusty's earlier comment that since this feature is free, you cannot expect your peer to provide a perfect SLA, it's a best effort on their part. 

Ah I see, rusty said:
> Require only best effort, not require them to perform a db commit on every new update.

So is it a DOS concern that a channel-peer can send many `channel_backup` messages, forcing us to do lots of db commits? If so, we could maybe relax the requirement to say that next db commit that includes channel-state-changes, must also include the latest received  `channel_backup` blob. i.e. the blob does not have to be persisted as soon as received, but it must be persisted together with the next channel-state-change.

01-messaging.md
```diff
@@ -473,6 +480,53 @@ every message maximally).
...
+    - when it receives `init` with an outdated or missing `node_backup`:
+      - MAY send a warning
+
+A node with `provide_peer_backup_storage` activated:
+  - when it receives `update_node_backup`:
+    - if the `node_backup` exceeds 32000 bytes:
+      - SHOULD send a warning
+      - MUST NOT store this backup data
+    - otherwise:
+      - SHOULD store this backup data
```
SomberNight, line 524: what if the sender does not have a channel with us? maybe discard the blob and send a warning?

02-peer-protocol.md
t-bast: > the blob does not have to be persisted as soon as received, but it must be persisted together (at the latest) with the next channel-state-change.

Do you have a specific sentence in mind to to express that requirement? Otherwise I'll try something.

01-messaging.md
t-bast: It should be up to the implementation to decide, probably need a `MAY ignore backup data from peers with whom they have no channels` or something similar

SomberNight: Sounds good.

t-bast: Added and rebased in https://github.com/lightning/bolts/pull/881/commits/daa853ce357203ddebb0bc988ac54b9af60fb140

02-peer-protocol.md
t-bast: I like `channel state change`, let's see what other people think. I included this and rebased in https://github.com/lightning/bolts/pull/881/commits/daa853ce357203ddebb0bc988ac54b9af60fb140

PR #887, opened 2021-07-12
Make payment secret mandatory and update Bolt 11 test vectors

t-bast:
Update Bolt 11 test vectors to always include a payment secret.
We want to make it mandatory in invoices which would make the existing test vectors invalid.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -159,7 +159,7 @@ Currently defined tagged fields are:
 
 A writer:
   - MUST include exactly one `p` field.
-  - If the `payment_secret` feature is set, MUST include exactly one `s` field.
+  - MUST include exactly one `s` field.
   - MUST set `payment_hash` to the SHA2 256-bit hash of the `payment_preimage`
   that will be given in return for payment.
   - MUST include either exactly one `d` or exactly one `h` field.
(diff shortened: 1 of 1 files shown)


11-payment-encoding.md
```diff
@@ -355,16 +355,21 @@ https://github.com/rustyrussell/lightning-payencode
 
 # Examples
 
-NB: all the following examples are signed with `priv_key`=`e126f68f7eafcc8b74f54d269fe206be715000f94dac067d1c04a8ca3b2db734`.  Also, the first 9 examples are legacy: modern invoices have an `s` field. 
+NB: all the following examples are signed with `priv_key`=`e126f68f7eafcc8b74f54d269fe206be715000f94dac067d1c04a8ca3b2db734`.
+Also, the first 9 examples are legacy: modern invoices have an `s` field.
```
rustyrussell, line 359: Remove this sentence?

rustyrussell (commented):
Seems OK, but the checksums and signatures in the breakdowns are incorrect.

Let me write a tool to generate these, and I can create another diff...

t-bast:
> Seems OK, but the checksums and signatures in the breakdowns are incorrect.

Right, good catch. Don't worry, I'll fix these today!

rustyrussell:
Already done, might as well take my changes!

Also, validated the vectors, thanks!

t-bast:
Great, I had done it in parallel and I just checked your commits, we find the same results so we should be good to go ;)

rustyrussell:
Ack https://github.com/lightningnetwork/lightning-rfc/pull/887/commits/42bd71d49c68075f54a31558af67af8c9892a08f

rustyrussell (approved):
Ack ec1d4dc6651953b2778343e44e3fc9d7195ed3b8

TheBlueMatt (approved):
Didn't check the new vectors, but text changes look good.

Roasbeef (approved):
LGTM 🐳

PR #889, opened 2021-07-19
Replace "than" with "as"

OrfeasLitos:


OrfeasLitos:
Replace "than" with "as"

t-bast (approved):
Sure, this works too. I believe the previous wording was also correct, but why not.

PR #891, opened 2021-07-27
websocket address type: allow transport over RFC6455

rustyrussell:
~And allow transparent upgrade~.

Allows you to advertize an alternate port.  This allows a wider range of software to interact with the lightning network.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -307,15 +307,15 @@ The origin node:
   - MUST place address descriptors in ascending order.
   - SHOULD NOT place any zero-typed address descriptors anywhere.
   - SHOULD use placement only for aligning fields that follow `addresses`.
-  - MUST NOT create a `type 1`, `type 2` or `type 5` address descriptor with `port` equal
+  - MUST NOT create a `type 1`, `type 2` or `type 6` address descriptor with `port` equal
   to 0.
   - SHOULD ensure `ipv4_addr` AND `ipv6_addr` are routable addresses.
   - MUST set `features` according to [BOLT #9](09-features.md#assigned-features-flags)
   - SHOULD set `flen` to the minimum length required to hold the `features`
   bits it sets.
-  - MUST NOT add a `type 5` address unless there is also at least one address of different type.
-  - if it adds a type 5 address:
-    - MUST allow unencrypted RFC6455<sup>[3](#reference-3)</sup> as a transport when a connection is made to at least one of the other addresses, with the type 5 `port` substituted for that address's `port` 
+  - MUST NOT add a `type 6` address unless there is also at least one address of different type.
+  - if it adds a type 6 address:
+    - MUST allow unencrypted RFC6455<sup>[3](#reference-3)</sup> as a transport when a connection is made to at least one of the other addresses, with the type 6 `port` substituted for that address's `port` 
 	- SHOULD allow this on ALL of the other addresses.
 
 The receiving node:
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -285,6 +285,7 @@ The following `address descriptor` types are defined:
          onion service addresses; Encodes:
          `[32:32_byte_ed25519_pubkey] || [2:checksum] || [1:version]`, where
          `checksum = sha3(".onion checksum" | pubkey || version)[:2]`.
+   * `5`: WebSocket port; data = `[2:port]` (length 2)
 
 ### Requirements
 
@@ -306,12 +307,16 @@ The origin node:
   - MUST place address descriptors in ascending order.
   - SHOULD NOT place any zero-typed address descriptors anywhere.
   - SHOULD use placement only for aligning fields that follow `addresses`.
-  - MUST NOT create a `type 1` OR `type 2` address descriptor with `port` equal
+  - MUST NOT create a `type 1`, `type 2` or `type 5` address descriptor with `port` equal
   to 0.
   - SHOULD ensure `ipv4_addr` AND `ipv6_addr` are routable addresses.
   - MUST set `features` according to [BOLT #9](09-features.md#assigned-features-flags)
   - SHOULD set `flen` to the minimum length required to hold the `features`
   bits it sets.
+  - MUST NOT add a `type 5` address unless there is also at least one address of different type.
+  - if it adds a type 5 address:
+    - MUST allow unencrypted RFC6455<sup>[3](#reference-3)</sup> as a transport when a connection is made to at least one of the other addresses, with the type 5 `port` substituted for that address's `port` 
+	- SHOULD allow this on ALL of the other addresses.
 
 The receiving node:
   - if `node_id` is NOT a valid compressed public key:
@@ -358,6 +363,12 @@ New address types may be added in the future; as address descriptors have
 to be ordered in ascending order, unknown ones can be safely ignored.
 Additional fields beyond `addresses` may also be added in the future—with
 optional padding within `addresses`, if they require certain alignment.
+
+Websockets generally are run on adjacent ports (or even overloaded on
+the same port) as existing "raw" transports, so including just the
+port is a compromise which avoids replacating all the addresses.  It's
+ideal if all addresses support this, but it's not a hard requirement:
+at least one must.
 
 ### Security Considerations for Node Aliases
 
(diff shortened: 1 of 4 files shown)


TheBlueMatt:
Is this better off done via a bridge instead? Last I heard there weren’t any decent websocket implementations that weren’t huge, which (maybe incorrectly) led me to understand that it was reasonably complicated. In either case, it seems if the only options are fresh code or huge dependency, it’s probably a bad idea for us to recommend users integrate it instead of simply using one of the websocket bridge services.

kristapsk:
@TheBlueMatt has a valid point here, we used https://github.com/novnc/websockify at my previous workplace.

Roasbeef:
> Last I heard there weren’t any decent websocket implementations that weren’t huge, which (maybe incorrectly) led me to understand that it was reasonably complicated. 

Go has its own implementation that's packaged as part of the standard library, so we effectively get this for free. The advantage of simply letting a node advertise that it supports doing the brontide handshake over WS is that you can avoid needing to use a proxy all together.

08-transport.md
```diff
@@ -253,7 +256,9 @@ and 16 bytes for the `poly1305` tag.
     * The raw bytes of the remote party's ephemeral public key (`re`) are to be
       deserialized into a point on the curve using affine coordinates as encoded
       by the key's serialized composed format.
-3. If `v` is an unrecognized handshake version, then the responder MUST
+3. If `v` is an unrecognized handshake version, and the responder supports
```
Roasbeef, line 259: So the version here was intended to be used to swap out for new _crypto_ in the future. 

When I saw the title of this PR, I thought this was instead adding a new address type (ws) to the node announcement message's set of addresses, and leaving everything else unmodified. This would let nodes advertise they support ws (maybe on a different interface/port) and keep everything else in place.

Roasbeef: IMO the address route also means that the message encryption layer doesn't need to be aware of what's being used to transport it. Instead nodes may optionally (if the include the new addr type) start to listen on a new port, expecting a WS handshake as the first step.

08-transport.md
```diff
@@ -402,6 +407,43 @@ construction, and 16 bytes for a final authenticating tag.
...
+- MAY attempt to initiate an unencrypted WebSocket as specified in RFC6455<sup>[4](#reference-4)</sup>:
+  - MUST send at least 50 bytes before awaiting a response.
+  - MUST abort the connection attempt if WebSocket upgrade fails.
+  - MUST begin the [Handshake Exchange](#handshake-exchange) as initiator 
+    as soon as upgrade succeeds.
+
+The responder:
+- if it supports `option_websocket`:
+  - SHOULD set `option_websocket` in its node announcements
+  - MUST attempt WebSocket upgrade if the Act 1 handshake it receives
```
Roasbeef, line 438: See comment above re separation of the layers. When nodes connect over Tor today, `brontide` implementations don't need to do anything different, and we can inherit that behavior here by defining the web sockets feature as a new _address_ type.

Roasbeef: Also by having it as a new address type, nodes can advertise another addr here, which might actually just be a bridge they're using.

Roasbeef (dismissed):

TheBlueMatt:
I feel like this would be a good candidate for a bLIP - I don't think we will reasonably consider this something that "most nodes should implement", but instead a totally-optional cool thing for those for whom its easy.

rustyrussell:
> Is this better off done via a bridge instead? Last I heard there weren’t any decent websocket implementations that weren’t huge, which (maybe incorrectly) led me to understand that it was reasonably complicated. In either case, it seems if the only options are fresh code or huge dependency, it’s probably a bad idea for us to recommend users integrate it instead of simply using one of the websocket bridge services.

Yes, I was looking for implementations and had the same sinking feeling.  But it's actually just modern software which sucks ass.  This is a minimal proxy implementation which I made from reading the RFC; it's 350 well-commented lines:

https://github.com/ElementsProject/lightning/pull/4685/files#diff-5397f54245a40c2eb2457fe205631701dba00e4ea957793f8545674758b38603

08-transport.md
rustyrussell: Yeah, this is a bit awkward since "this port supports ws" is orthogonal to the address format, so I just made it a 2-byte port number, implying it's an alternate to the other address types which have a port.

And yeah, that's exactly what happens: you get a HTTP incoming, upgrades to ws, then you transport lightning over the top (we ignore the ws framing and treat it as a stream, but that's trivial).

Roasbeef: > Yeah, this is a bit awkward since "this port supports ws" is orthogonal to the address format, so I just made it a 2-byte port number, implying it's an alternate to the other address types which have a port.

Yeah so what I'm suggesting is to effectively just treat it as a new address type, since that lets nodes handle it fully indepndently (they can bridge or make it native). 

The difference here is (also from looking at your current impl), is that the noise layer doesn't need to "peak" into the handshake state to understand it. Instead we get a clean separation, just like it doesn't need to know anything about Tor when that's used, as in the context of the language runtime, it likely just gets some "connection-like" object to use once that transport handshake has been completed.

07-routing-gossip.md
```diff
@@ -285,6 +285,7 @@ The following `address descriptor` types are defined:
          onion service addresses; Encodes:
          `[32:32_byte_ed25519_pubkey] || [2:checksum] || [1:version]`, where
          `checksum = sha3(".onion checksum" | pubkey || version)[:2]`.
+   * `5`: WebSocket port; data = `[2:port]` (length 2)
```
Roasbeef, line 288: Doesn't this also need an interface/IP? Or it's meant to be a sort of modifier?

07-routing-gossip.md
```diff
@@ -306,12 +307,16 @@ The origin node:
...
   - SHOULD NOT place any zero-typed address descriptors anywhere.
   - SHOULD use placement only for aligning fields that follow `addresses`.
-  - MUST NOT create a `type 1` OR `type 2` address descriptor with `port` equal
+  - MUST NOT create a `type 1`, `type 2` or `type 5` address descriptor with `port` equal
   to 0.
   - SHOULD ensure `ipv4_addr` AND `ipv6_addr` are routable addresses.
   - MUST set `features` according to [BOLT #9](09-features.md#assigned-features-flags)
   - SHOULD set `flen` to the minimum length required to hold the `features`
   bits it sets.
+  - MUST NOT add a `type 5` address unless there is also at least one address of different type.
```
Roasbeef, line 316: What's the rationale here? To ensure a node is able to serve the "greater" network w/ a normal TCP (or w/e) transport?

Roasbeef (commented):
Dig the new direction, just a few comments about the new addr type and behavior related to it

07-routing-gossip.md
rustyrussell: The alternative is to duplicate all the entries, so this is simply a modifier.

07-routing-gossip.md
rustyrussell: Yeah, there's already a js library which talks WS to this: https://github.com/rustyrussell/bolt12/

It's kinda cute...

m-schmoock:
@rustyrussell @TheBlueMatt 
Do we need to limit the number of allowed `type 6` similar to https://github.com/lightning/bolts/pull/911/commits/3e88d0344d7dd210a7346648e0573b4805af33b0 as currently being discussed in https://github.com/lightning/bolts/pull/911 ?

07-routing-gossip.md
```diff
@@ -285,6 +285,7 @@ The following `address descriptor` types are defined:
          onion service addresses; Encodes:
          `[32:32_byte_ed25519_pubkey] || [2:checksum] || [1:version]`, where
          `checksum = sha3(".onion checksum" | pubkey || version)[:2]`.
+   * `6`: WebSocket port; data = `[2:port]` (length 2)
```
TheBlueMatt, line 288: I'd rather we allow specification of an IP as well here. Though maybe that interacts poorly with the hostname gossip thing? There may likely end up being nodes that would love to allow websocket connections, but sit behind cloudflare/nginx/whatever when doing so, whereas native LN protocol obviously doesn't proxy through a web proxy. Thus, they'd want to be able to specify a different host for websockets. I suppose for cloudflare-proxied nodes (dear god people stop using cloudflare) you'd need a domain name, not an IP, and have to send the HTTP Host header.

Maybe this should be hostname-only? Or maybe two different gossips, one for hostnames and one for ips?

rustyrussell: You basically iterate through all the other addresses, trying this port.  It's not ideal, but it seemed easier than some kind of specification which indicated *which* other descriptors it applied to...

TheBlueMatt: Yea, totally fair, I think my broader point here is that we really should consider `wss` a required part of this, because supporting websockets so browser clients can connect but then not support `wss` so browser clients can't connect in a secure context seems like it almost completely neuters the value here, sadly. For `wss` you need a hostname, ultimately, and I think we'd probably want to support a different hostname than the lightning node itself, as the `wss` part will probably be nginx or some other proxy, hence...I think we should just shove a hostname here :(

rustyrussell:
Plan is to update to DNS-record style (TLS!) instead....

07-routing-gossip.md
TonyGiorgio: +1 for wanting `wss`, we've created a websocket to tcp proxy server for connecting in a browser, but we can't do `ws` unfortunately.

PR #902, opened 2021-08-29
BOLT 7: remove 2^32 restriction from channel_update as well.

rustyrussell:
This was missed in #877 (84213f45c05a39dd0812604a6a11d9d3c1207425).

As reported in https://github.com/ElementsProject/lightning/issues/4746


t-bast (approved):
Good catch, thanks!

PR #910, opened 2021-09-12
Explicitly allow funding_locked early, and support alias scids.

rustyrussell:
This lets you add your brand new channel to routehints, and also
means you can use a routehinted channel even if you (later?) have a
real channel.

This supports both trusted and untrusted zero-conf channels: in the
trusted case you can use it immediately like any other channel,
and for the untrusted case you simply use any push_msat they gave you
for outgoing payments, but fail incoming.

Closes: #895


02-peer-protocol.md
```diff
@@ -409,25 +409,54 @@ this channel will continue to use `option_static_remotekey` or `option_anchor_ou
...
 
-The sender MUST:
-  - NOT send `funding_locked` unless outpoint of given by `funding_txid` and
+The sender:
+  - MUST NOT send `funding_locked` unless outpoint of given by `funding_txid` and
    `funding_output_index` in the `funding_created` message pays exactly `funding_satoshis` to the scriptpubkey specified in [BOLT #3](03-transactions.md#funding-transaction-output).
-  - wait until the funding transaction has reached `minimum_depth` before
-  sending this message.
-  - set `next_per_commitment_point` to the per-commitment point to be used
+  - MUST set `next_per_commitment_point` to the per-commitment point to be used
```
TheBlueMatt, line 435: So this implies we always send the thing we sent in our last `revoke_and_ack` (if there was a previous one)?. Just making sure we're all on the same page.

TheBlueMatt (commented):
This seems to address all my questions/feedback I had while implementing, outside of the questions around feature bit in the previous PR - nodes that *want* to be able to use an alias for their incoming payments may still want to connect to nodes that implement sending aliases. That said, I dunno if anyone has any plans on implementing such peer-seeking behavior.

02-peer-protocol.md
```diff
@@ -409,25 +409,61 @@ this channel will continue to use `option_static_remotekey` or `option_anchor_ou
...
+  - if it sets `alias`:
+    - if the `announce_channel` bit was set in both `open_channel` and `accept_channel`:
+      - SHOULD initially set `alias` to value not related to the real `short_channel_id`.
+      - SHOULD re-transmit `funding_locked` with the real `short_channel_id` once the channel is confirmed.
+    - otherwise:
+      - MUST set `alias` to a value not related to the real `short_channel_id`.
+    - MUST NOT send the same `alias` for multiple peers.
+    - MUST always recognize the `alias` as a `short_channel_id` for incoming HTLCs to this channel.
+    - if `option_scid_alias_only` was negotiated:
+      - MUST NOT allow incoming HTLCs to this channel using the real `short_channel_id`
```
Crypt-iQ, line 451: If `option_scid_alias_only` is negotiated and incoming HTLCs of the real `short_channel_id`, does this break the spontaneous nature of key-send? The payer will need to know one of the alias `short_channel_id` which will be different from the publicly announced "real" one in the `channel_announcement` / `channel_update` messages.  Additionally, regular routing may require a payer to know aliases of ALL `option_scid_alias_only`-negotiating hops along the way.

02-peer-protocol.md
Crypt-iQ: I think this means send the per-commitment-point for the first commitment transaction (hence the #1)? Rather than the last one

TheBlueMatt: Yes, this changed in the second (fixup) commit that was pushed after this comment was made :).

TheBlueMatt:
@rustyrussell you'd mentioned previously this may make sense/need to be a channel type given the persistent scid alias info - I think I agree (but don't feel strongly at all really) - do you plan to update this PR?

02-peer-protocol.md
Crypt-iQ: It also makes sense in the zero-conf case before the scid is known

pm47:
> Shouldn't a node just advertise its ability to _understand_ `scid` aliases? Something like `option_scid_alias`.

I have explored this approach in this branch https://github.com/pm47/lightning-rfc/tree/zeroconf-as-alias-pm. In summary:

- make zero-conf and scid alias independent
    Fundee may signal that it is willing to trust the funder by setting minimum_depth to zero, but that's unenforceable anyway.
- make scid alias and channel announcement independent
    There is no reason to advertize the "real" scid in funding_locked, because the real scid is already provided in announcement_signatures
- have nodes only advertise their ability to understand scid alias (feature option_scid_alias)
    If this feature is negotiated (both peers support it), then they must only use aliases (but that's unenforceable).
- accept that some combinations don't make sense, like "zero-conf without alias", or "alias and public channels"


>> I don't understand the point of option_scid_alias_only.

> No, this is critical to protect against probing.

My point is if both parties support `alias`es, they must only use them. Why create a corner case where parties understand `alias`es but keep using real `scid`s? Plus this is unenforceable anyway.

Answering https://github.com/lightningnetwork/lightning-rfc/pull/923#issuecomment-941290456:

>> make zero-conf and scid alias independent Fundee may signal that it is willing to trust the funder by setting minimum_depth to zero, but that's unenforceable anyway.

> Why? You need an SCID alias in order for this to work, presumably?

Not for e.g. Rusty's "spend push_msat" use case which is valid imo.

>> have nodes only advertise their ability to understand scid alias (feature option_scid_alias)

> Did you see my comment on 970? Should this be a channel type instead?

Nope, I missed it (970??). I guess it could be a channel type (@t-bast wdyt?) but this is a separate concern from the semantics of this option.

>> "alias and public channels"

> One question on this - do we want to support this with an assumption that in the future we'll have public key aliases as well? Like, for non-routing payments that are just to the final node over a public channel, we may want to support that lost hop being "private" by having an alias, and if we support it here the only requirement is a public key alias which can come separately.

That's a very good point: in fact, if we were to implement `scid` `alias` in Phoenix we would treat any _channel_ alias as a _node_ alias, because the general case is that a new channel may be created for a given payment (plus: only providing one routing hint reduces the size of the payment request). Also, putting Phoenix aside, given all (I think?) implementations do non-strict forwarding, there is no real difference between _channel_ and _node_ alias. This leads me to YA another alternative to this PR: have the `alias` be provided in the `init` tlv instead of `funding_locked`. Problem: we don't want to assign an `alias` to every incoming connection before there is a channel. Also, the `alias` could only rotate at reconnection, but that's fine for mobiles, which disconnect/reconnect all the time.

Answering https://github.com/lightningnetwork/lightning-rfc/pull/923#issuecomment-941242029:

> How can a zero-conf channel be public from the start if it does not have a short_channel_id assigned and only an alias?

It's public in the sense that both peers have set the `announce_channel` bit in `open`/`accept`. This is a separate concern, unrelated to zero conf: we are in the same situation when a funding tx has more than `minimum_depth` and less than 6 confirmations.

>> That would be a spec violation though: if both peers support option_scid_alias, they must send an alias.

> What happens if an alias is not received?

I guess it's a channel failure, or a warning, depending on how much you care.

rustyrussell:
> I don't understand the point of `option_scid_alias_only`. Shouldn't a node just advertise its ability to _understand_ `scid` aliases? Something like `option_scid_alias`.
> 
> Then if you want your peer to _only use_ aliases, just request private channels? The only `scid` that will be used will be those provided in routing hints. Otherwise things gets confusing, as this comment thread shows: https://github.com/lightningnetwork/lightning-rfc/pull/910/files#r722406596.

Yes, text fixed.

The intent was that everyone should be setting this option in future, and thus respecting the alias.  We can certainly rename it to `option_scid_alias`, which makes more sense?

rustyrussell:
> > Shouldn't a node just advertise its ability to _understand_ `scid` aliases? Something like `option_scid_alias`.
> 
> I have explored this approach in this branch https://github.com/pm47/lightning-rfc/tree/zeroconf-as-alias-pm. In summary:
> 
>     * make zero-conf and scid alias independent
>       Fundee may signal that it is willing to trust the funder by setting minimum_depth to zero, but that's unenforceable anyway.
> 
>     * make scid alias and channel announcement independent
>       There is no reason to advertize the "real" scid in funding_locked, because the real scid is already provided in announcement_signatures

True, and you won't be able to because it probably is still in mempool.

>     * have nodes only advertise their ability to understand scid alias (feature option_scid_alias)
>       If this feature is negotiated (both peers support it), then they must only use aliases (but that's unenforceable).

Yep, done.

>     * accept that some combinations don't make sense, like "zero-conf without alias", or "alias and public channels"

Zero-conf without alias would work for direct payments, but they're not generally useful.  But aliases on public channels are good for routehints before funding tx is 6 deep.

> > > I don't understand the point of option_scid_alias_only.
> 
> > No, this is critical to protect against probing.
> 
> My point is if both parties support `alias`es, they must only use them. Why create a corner case where parties understand `alias`es but keep using real `scid`s? Plus this is unenforceable anyway.

Yep.  Spec now says this means you gotta ignore requests for "real" scid on unannouncable channels.

> Answering [#923 (comment)](https://github.com/lightningnetwork/lightning-rfc/pull/923#issuecomment-941290456):
> 
> > > make zero-conf and scid alias independent Fundee may signal that it is willing to trust the funder by setting minimum_depth to zero, but that's unenforceable anyway.
> 
> > Why? You need an SCID alias in order for this to work, presumably?
> 
> Not for e.g. Rusty's "spend push_msat" use case which is valid imo.

Or trust.  But just supply a damn scid, it's easy :)

> > > have nodes only advertise their ability to understand scid alias (feature option_scid_alias)
> 
> > Did you see my comment on 970? Should this be a channel type instead?
> 
> Nope, I missed it (970??). I guess it could be a channel type (@t-bast wdyt?) but this is a separate concern from the semantics of this option.
> 
> > > "alias and public channels"
> 
> > One question on this - do we want to support this with an assumption that in the future we'll have public key aliases as well? Like, for non-routing payments that are just to the final node over a public channel, we may want to support that lost hop being "private" by having an alias, and if we support it here the only requirement is a public key alias which can come separately.
> 
> That's a very good point: in fact, if we were to implement `scid` `alias` in Phoenix we would treat any _channel_ alias as a _node_ alias, because the general case is that a new channel may be created for a given payment (plus: only providing one routing hint reduces the size of the payment request). Also, putting Phoenix aside, given all (I think?) implementations do non-strict forwarding, there is no real difference between _channel_ and _node_ alias. This leads me to YA another alternative to this PR: have the `alias` be provided in the `init` tlv instead of `funding_locked`. Problem: we don't want to assign an `alias` to every incoming connection before there is a channel. Also, the `alias` could only rotate at reconnection, but that's fine for mobiles, which disconnect/reconnect all the time.

SCID inside an onion is a compact representation of the next node id.  Always has been.  It's only a direct channel reference for channel_update.

> Answering [#923 (comment)](https://github.com/lightningnetwork/lightning-rfc/pull/923#issuecomment-941242029):
> 
> > How can a zero-conf channel be public from the start if it does not have a short_channel_id assigned and only an alias?
> 
> It's public in the sense that both peers have set the `announce_channel` bit in `open`/`accept`. This is a separate concern, unrelated to zero conf: we are in the same situation when a funding tx has more than `minimum_depth` and less than 6 confirmations.
> 
> > > That would be a spec violation though: if both peers support option_scid_alias, they must send an alias.
> 
> > What happens if an alias is not received?
> 
> I guess it's a channel failure, or a warning, depending on how much you care.

I will close the channel.  Don't do that :)

rustyrussell:
Yerch... you need to use an alias, but whose, and which one?

I think the general rule here is "update_channel should use your own alias for unannounced channels".  That works whether the update_channel is for an error response, as well as this case.

@TheBlueMatt wanted to allow multiple aliases, but this kills that idea, and I think it should be removed.

(Aside: I am looking at switching this to channel types, too: this makes far more sense when we try to upgrade, though it's a little more work in the short term.  I'll try implementing it today...)

rustyrussell:
You can't use 0, because we can have more than one channel.  You can't use the "real" scid since it might not exist yet (zeroconf).  You could send a channel_update every time the you get a new alias, I guess that's friendly (you call this channel 7x7x7?  OK, here's the channel_update for "7x7x7").

I'm actually not sure what happens when we get a bad fee on a private channel, whether we send a channel_update or not.  I'm pretty sure we don't in c-lightning.  We should def. not leak the real scid in that case!

TheBlueMatt:
> Rotating can be probed without pubkey rotation which doesn't work without telling your peer what the association is (otherwise they can send via the new pubkey on the old channel scid, etc).

Indeed, that's exactly what I said, gotta have pubkey aliases :)

> And we decided not to do that, going for blinded paths instead.

"we" did? I think I'd love to go that route too, but last I heard work had largely stalled and there was some significant disagreement about trampoline and friends and how to build it appropriately. I'd love to learn that work is not stalled, however!

rustyrussell:
Well, the onion message enctlv format is the same as the HTLC one, so hopefully progress in one leads to progress in the other?

TheBlueMatt:
> Well, the onion message enctlv format is the same as the HTLC one, so hopefully progress in one leads to progress in the other?

True! I guess what you're telling me is the issues that folks had are resolved, and I shouldn't worry and its definitely moving forward? If that's definitely the case than I'm totally happy to drop the alias-rotation stuff, I just figured it had ~no implementation cost and was an easy future win, given murkyness of trampoline's future.

Crypt-iQ:
> > The implementation cost is low, but you have to remember every rotation that you sent out, which makes this unappealing to me.
> 
> Why? "seems unappealing to me" isn't generally an argument :). The sender (who has to store them) is the one who decides when to send another, it costs ~nothing to store a bunch of 48-bit integers, and it could enable substantially more privacy if we add pubkey aliases. If we're not moving towards pubkey aliases (because we're definitely gonna do trampoline), then it doesn't get us anything, of course, but that's doubly so a question for you - I know there was some hesitation from the lnd end towards that.

I agree it's not an argument - I don't understand the point of rotation. What is the risk in using the same alias SCID over and over?

Crypt-iQ:
> > I agree it's not an argument - I don't understand the point of rotation. What is the risk in using the same alias SCID over and over?
> 
> Its not ideal for privacy - if you're generating an invoice with a single route hint to a large public node/LSP, ideally each invoice would be unique and a sender wouldn't be able to identify the recipient across each invoice. If we add node id aliases (with rotation) then all we'd need is SCID aliases (with rotation). Alternatively, we could accomplish this with blinded paths in the invoice, providing the entire route hints as randomized blobs.

"generating an invoice .. to a large public node/LSP" - need clarification - does this mean you're giving the LSP the invoice?

Crypt-iQ:
> > "generating an invoice .. to a large public node/LSP" - need clarification - does this mean you're giving the LSP the invoice?
> 
> Sorry, that was confusing, I mean "a private node that just has a single or two channels to a large public node/LSP" wants to generate an invoice with a single/two route-hints.

I think I understand. In some cases, you won't know who you're paying but you'll have a node id. With node id rotation and without SCID rotation, you have a privacy leak. So you need to rotate node id & scid ?

benthecarman:
Is alias the same format as short channel id? It is not clear what the format is

Crypt-iQ:
> Is alias the same format as short channel id? It is not clear what the format is

its a Short channel id

rustyrussell:
OK, rebased and I added the channel_type, but it's only for the case where you *don't* want the real scid used (i.e. this is the property that persists for this channel).

I believe that when you send a private channel_update, you should use the *peers* scid alias.  Reasons are simple: they committed to remember it, and you don't actually ever send this to anyone anyway (they just use the info to populate route hints).

If Alice calls the channel A, and Bob calls it B, and Alice uses B in a routehint for Carol, if Carol gets the fee wrong it's *Bob* who will be the one to respond with a channel_update, and he will use B in that naturally (it should specify that nodes use the scid used by the sender in this case: this is no longer clear once we have aliases!).

Crypt-iQ:
> I believe that when you send a private channel_update, you should use the peers scid alias. Reasons are simple: they committed to remember it, and you don't actually ever send this to anyone anyway (they just use the info to populate route hints).

Exactly. One thing I find a little weird is that issuing an invoice only provides a 1-way pass for forwarding over the channel. To actually use the channel as a 3rd-party you'd need both scids. Maybe this is something that can be improved upon instead of receiving 2 invoices.

02-peer-protocol.md
```diff
@@ -1344,7 +1396,8 @@ A node:
   sent and received:
     - MUST retransmit `funding_locked`.
   - otherwise:
-    - MUST NOT retransmit `funding_locked`.
+    - MUST NOT retransmit `funding_locked`, but MAY send `funding_locked` with
```
TheBlueMatt, line 1399: I think this may be a backwards incompatible change unless we require the feature bit here - at least we used to check the `funding_locked` `next_per_commitment_point` by checking it against the channel's current commitment id, which will fail if a `funding_locked` is sent after the channel is in normal operation.

TheBlueMatt (commented):
OK, sorry, finally getting back to this.

TheBlueMatt:
Needs rebase.

02-peer-protocol.md
```diff
@@ -460,25 +466,61 @@ is negotiated.
...
 This message indicates that the funding transaction has reached the `minimum_depth` asked for in `accept_channel`. Once both nodes have sent this, the channel enters normal operating mode.
 
+Note that `minimum_depth` can be 0 if the accepter trusts the opener,
+in which case this message is a misnomer.  The opener is free to send this
+message at any time (since it presumably trusts itself).
+
 1. type: 36 (`funding_locked`)
 2. data:
     * [`channel_id`:`channel_id`]
     * [`point`:`next_per_commitment_point`]
```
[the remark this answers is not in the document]
cdecker: Shall we just call it `channel_ready`, since that's the semantics were trying to convey? Fwiw I'm also happy to just keep the current name and remove the misnomer comment (specs are authoritative, and not intended to convey all the details and rationale since they tend to bloat and obfuscate the important Infos).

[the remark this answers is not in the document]
TheBlueMatt: Oh, I meant next_per_commitment_point specifically, not `funding_locked`. I'm happy to see `funding_locked` renamed `channel_ready`, but it'd be nice to see `next_per_commitment_point` renamed as well given it too is a misnomer.

rustyrussell:
OK, new commits:

1. Add a new feature for zeroconf, and a new zeroconf channel_type.  You're only supposed to accept this if you trust the other side, and you must also set minimum_depth to 0 in that case.
2. Renamed some fields.
3. Rebased onto master

02-peer-protocol.md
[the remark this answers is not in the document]
rustyrussell: It's actually the *second* per-commitment point though.  (We already have first_per_commitment_point in `open_channel`).  But renaming is easy; will check that the `next_per_commitment_point` requirements elsewhere apply properly...

Crypt-iQ:
Given that there's now the channel type, is there any reason not to just *ignore* the min_depth field?

TheBlueMatt:
> Given that there's now the channel type, is there any reason not to just ignore the min_depth field?

Well, we have to set it anyway, so we might as well define it to be zero.

02-peer-protocol.md
```diff
@@ -239,6 +243,8 @@ The sending node:
     - MUST set it to a defined type representing the type it wants.
     - MUST use the smallest bitmap possible to represent the channel type.
     - SHOULD NOT set it to a type containing a feature which was not negotiated.
+    - if `announce_channel` is `true` (not `0`):
+      - MUST NOT send `channel_type` with the `option_scid_alias` bit set.
```
TheBlueMatt, line 247: This needs some kind of feature gate - prior to this PR sending a channel_ready after the channel has advanced state` is not allowed (and, at least, we would barf on it).

TheBlueMatt (commented):
Quick glance, will do more later.

TheBlueMatt:
Should the `channel_update` you send for private channels to tell your peer where to relay use an SCID alias if you've received one from your peer? It seems kinda nice to, but its certainly not all that important either way.

TheBlueMatt:
Yea, following up on the above, because of the way we serialize objects and support backwards and forwards compat, LDK will probably never send `option_zeroconf` features, so we definitely want to leave `minimum_depth` in place - let it control whether a channel is zero conf, the feature bit is just a "yo, zero conf or bust" signal, nothing more.

Crypt-iQ:
There is a channel type that negotiates whether or not zero-conf should be used, minimum_depth seems kinda pointless

TheBlueMatt:
The channel type as proposed here does not "negotiate whether or not zero-conf should be used", that's exactly what `minimum_depth` is for. The channel_type as proposed here allows the initiator to say "zero conf or bust, buddy". The initiat-ee is allowed to turn on zero conf whether the feature is set or not.

Roasbeef:
> My current read of the spec here, and my current implementation, never sends a zero conf feature flag at all, and can still happily do zero conf.

I think this only works one way? 

Assume implementation `A` always uses the channel type, and implementation `B` only ever sends `min_depth`:
  * If `A` is the funder, and they sent the channel type, `B` sends `min_depth` to zero (in `accept_channel`), zero conf all around, gg. 
  * If `B` is the funder, and they _don't_ send the channel type, then:
      * If `A` is an older node, they'll reject that `min_depth` value (as it's zero), or they'll fall back to 6 conf. 
      * If `A` is updated, then they won't send funding locked early (assuming they always want the chan type to signal), and no zero conf happens. 
          * For `B` this silent fallback isn't detectable until the funding is complete, at which point they need to double spend the input if they want to back out of the channel. 

One thing I learned from our prior interop issue when the `channel_type` field was originally introduced is: if there's a channel type, it should _always_ be sent in the appropriate context. This simplifies code, and also makes intent explicit all the way around, without needing to fall back to implicit negotiation which at times can be ambiguous. 

> and c-lightning, and maybe us too later, has stated intent to allow this ever after the channel has been funded).

So this would take the form of an adhoc "zero conf nao" API that can retroactively be applied to channels where `min_depth` zero wasn't sent? If you're applying that to a non-explicit zero conf channel, then I don't think that gets you anything, as both sides need to send funding locked for the channel to be useable in the first place. Single sided zero conf is indistinguishable from a normal wait-for-confs-channel.

Roasbeef:
> Given that there's now the channel type, is there any reason not to just ignore the min_depth field?

IMO it should always be set, this way it acts as a sort of double opt in: funder sets the channel_type, responder sents min_depth: </insert-buff-dude-handshake.png>.

TheBlueMatt:
> I think this only works one way?

I don't see why that would be the case. The implementation I have works just fine without it :).

> If A is an older node, they'll reject that min_depth value (as it's zero), or they'll fall back to 6 conf.

B doesn't send a min_depth, so there's nothing to reject. Can you restate your issue here there's some major confusion.

> One thing I learned from our prior interop issue when the channel_type field was originally introduced is: if there's a channel type, it should always be sent in the appropriate context. This simplifies code, and also makes intent explicit all the way around, without needing to fall back to implicit negotiation which at times can be ambiguous.

There's a specific concern there from our end - we write the channel type to disk, and when loading, if we see a channel type that we don't support, we refuse to read the data. Further, we always guarantee that *previous* versions of LDK can read data written by *new* versions, at least for a version or two. Our users rely on this pretty heavily. Thus, if we start sending the new channel type immediately, we'll needlessly break backwards compat for no reason.

In this specific case, we *already* have the relevant field - `minimum_depth`, it can be set to 0 and nodes can look at that, there's no reason it *has* to be a channel type, unless you really want the "funder demands it and channel fails immediately" logic, which, fine, y'all want that, I don't care, I'll implement it as long as it doesn't break other things, on the fundee side.

> IMO it should always be set, this way it acts as a sort of double opt in: funder sets the channel_type, responder sents min_depth: </insert-buff-dude-handshake.png>.

Again, that is not possible for several use-cases that have been described, c-lightning (and probably us) will eventually implement the ability to make a channel 0-conf *after* the funding has been negotiated, its a great API for users.

Roasbeef:
> B doesn't send a min_depth, so there's nothing to reject. Can you restate your issue here there's some major confusion.

Ah yeah you're right, got the scenario mixed up half way, but then that case just results in `B` never being able to open a zero conf channel? As they'll never set the channel type, so the remote party doesn't know if `min_depth` of zero is requested/warranted. If `A` doesn't have an existing trust relationship with `B`, and `B` sends the channel type, `A` can just reject it up front, and both sides immediately know what's going on. 

> I don't see why that would be the case. The implementation I have works just fine without it :).

I don't see how it can given, that we now require the `channel_type` to [_always_ be echoed back in the accept message](https://github.com/lightning/bolts/pull/906), or maybe I'm misunderstanding? This was one of the things we had to fix in `lnd`: we didn't always echo back which caused the interop break. If you send `channel_type` and the other node doesn't  echo it back, that's a spec violation. 

> There's a specific concern there from our end - we write the channel type to disk, and when loading, if we see a channel type that we don't support, we refuse to read the data.

Why isn't this handled by the even/odd semantics? If you get an open channel with a required channel type you don't understand, you can halt there. Unless y'all always write every incoming message to disk before validating it? 

> Further, we always guarantee that previous versions of LDK can read data written by new versions, at least for a version or two.

Something something extra layer of internal TLV with odd types? :p 

> In this specific case, we already have the relevant field - minimum_depth, it can be set to 0 and nodes can look at that, there's no reason it has to be a channel type, unless you really want the "funder demands it and channel fails immediately" logic, which, fine, y'all want that, I don't care, I'll implement it as long as it doesn't break other things, on the fundee side.

Agree we already have that field, what I'm proposing is a double opt in: so initiator sets the channel type, responder sets that type (as they already _have_ to), _and_ they set that `min_depth` to zero. Atm internally we use a zero min depth as a signal of a non user specified value, so we'll need to change that to be something else. 

> Again, that is not possible for several use-cases that have been described, c-lightning (and probably us) will eventually implement the ability to make a channel 0-conf after the funding has been negotiated, its a great API for users.

By after do you mean once both open+accept have been exchanged? I don't see how that's possible, as if one side sends early but the other doesn't, you can't use the channel. So there's an assumption of an out-of-band double opt-in to upgrade the channel _after_ open has been sent but not accept? I'd be curious to hear more about the concrete use case, all the users we've talked to so far always know up front what they expect, as that trust relationship exists before the channel has been funded (Alice is using Bob's wallet and added services on top of it). Unclear to me how two mutually distrusting sides end up getting close enough to decide to do a zero conf after the fact (unless they do jitsi over onion messages mid way and find some common interests)?

TheBlueMatt:
Maybe an impromptu call would be better than going back and forth here. Happy to set up some time or feel free to call.

> Ah yeah you're right, got the scenario mixed up half way, but then that case just results in B never being able to open a zero conf channel? As they'll never set the channel type, so the remote party doesn't know if min_depth of zero is requested/warranted. If A doesn't have an existing trust relationship with B, and B sends the channel type, A can just reject it up front, and both sides immediately know what's going on.

No? The remote party knows whether it has a trust relationship or not before they receive the open_channel, they have to no matter whether the feature bit is set or not. Based on that trust relationship they can set the conf threshold. Note that you can always send the funding_locked early, no matter what you set it to, so you can still do it after the fact and just ignore the channel type no matter what.

> I don't see how it can given, that we now require the channel_type to https://github.com/lightning/bolts/pull/906, or maybe I'm misunderstanding?

Again, the channel type, as phrased, is a statement by the initiator that 0conf is *required*, even if it is unset, the initiatee can always declare 0conf and it can go through if both parties `funding_locked` immediately.

> Why isn't this handled by the even/odd semantics? If you get an open channel with a required channel type you don't understand, you can halt there. Unless y'all always write every incoming message to disk before validating it?

Channel types are not allowed to have an odd bit, ever.

> Agree we already have that field, what I'm proposing is a double opt in: so initiator sets the channel type, responder sets that type (as they already have to), and they set that min_depth to zero. Atm internally we use a zero min depth as a signal of a non user specified value, so we'll need to change that to be something else.

Yep, we can't do that, see the desire to 0conf after init.

>      By after do you mean once both open+accept have been exchanged? I don't see how that's possible, as if one side sends early but the other doesn't, you can't use the channel. So there's an assumption of an out-of-band double opt-in to upgrade the channel after open has been sent but not accept?

In c-lightning, and probably LDK, in many cases we'll send the funder funding_locked immediately whether we set 0conf or not. so it doesn't matter.

TheBlueMatt:
> I don't see how it can given, that we now require the channel_type to https://github.com/lightning/bolts/pull/906, or maybe I'm misunderstanding?

Note that PR's text is *only* about `open_channel`, not `accept_channel`.

Roasbeef:
> Note that PR's text is only about open_channel, not accept_channel.

Hmm, you're right. We adopted that behavior after an implementation (I think it was CL) failed funding attempts as we didn't echo it back. Seems we still need to tighten up that text to reflect what impls do in the wild (the bit was added after the initial merge so we tried to make things work w/ the older and newer impls).

02-peer-protocol.md
TheBlueMatt: Maybe we just only allow this with the SCID alias channel type bit? There's not much reason to do it otherwise.

Roasbeef:
Spec meeting update: some rough consensus to allow both the explicit zero conf chan type feature bit and also just the normal min_depth=0 vector. We'll make sure to do interop testing to make sure both path ways work. I also need to catch up on the latest text (funding_locked rename, etc).

Crypt-iQ:
@TheBlueMatt will LDK echo the channel_type in accept_channel if lnd sends an open_channel with the bit set? And to clarify from yesterday, the use case was LDK will send the open_channel w/o the bit and lnd will reply with min_depth=0 also w/o the bit? It would be nice to document what each implementation will support

TheBlueMatt:
> will LDK echo the channel_type in accept_channel if lnd sends an open_channel with the bit set?

Yes, we will accept the 0conf channel type if the initiator requests it, depending on local configuration (obviously).

> And to clarify from yesterday, the use case was LDK will send the open_channel w/o the bit and lnd will reply with min_depth=0 also w/o the bit?

Yes, the bitset in the channel type is absolutely required to match in any case. LDK intends to never set the bit, and instead rely on whatever out-of-band negotiation had to occur to open the channel in the first place to negotiate the 0conf min_depth. We'll honor any 0conf request from the fundee (in the form of the min_depth parameter, with a user config knob to turn this behavior off), and allow the user to specify 0conf on any inbound channel whether it has the feature bit or not.

02-peer-protocol.md
```diff
@@ -461,31 +471,66 @@ use `option_static_remotekey`, `option_anchor_outputs` or
...
+    1. type: 1 (`short_channel_id`)
+    2. data:
+        * [`short_channel_id`:`alias`]
 
 #### Requirements
 
-The sender MUST:
-  - NOT send `funding_locked` unless outpoint of given by `funding_txid` and
+The sender:
+  - MUST NOT send `channel_ready` unless outpoint of given by `funding_txid` and
```
dunxen, line 496:
```suggestion
  - MUST NOT send `channel_ready` unless outpoint given by `funding_txid` and
```

Think either this `of` needs to be removed or a noun is missing.

Crypt-iQ:
> Yes, the bitset in the channel type is absolutely required to match in any case.

I think that either c-lightning or eclair was (dunno if still happens) would send an accept_channel w/ channel_type set even though lnd sent an open_channel without a channel_type. So either those implementations are wrong, I'm mis-remembering, or the spec allows this and the channel type doesn't have to match in this case.

> has stated intent to allow this ever after the channel has been funded

How would this be done?

TheBlueMatt:
> I'm mis-remembering, or the spec allows this and the channel type doesn't have to match in this case.

See #960

> How would this be done?

You just...send funding_locked? If you're the funder (and you aren't gonna double-spend to RBF) you should probably always send funding_locked immediately. The fundee can then pick whenever to lock the funding, irrespective of what they put in the `accept_channel`.

02-peer-protocol.md
```diff
@@ -461,31 +471,66 @@ use `option_static_remotekey`, `option_anchor_outputs` or
...
-  sending this message.
-  - set `next_per_commitment_point` to the per-commitment point to be used
-  for the following commitment transaction, derived as specified in
+  - if it is not the node opening the channel:
+    - SHOULD wait until the funding transaction has reached `minimum_depth` before
+      sending this message.
+  - MUST set `second_per_commitment_point` to the per-commitment point to be used
+  for commitment transaction #1, derived as specified in
   [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
+  - if `option_scid_alias` was negotiated:
```
Crypt-iQ, line 504: Is this the channel_type or the feature bit?

02-peer-protocol.md
```diff
@@ -207,12 +207,16 @@ definitions they reuse even feature bits, but they are not an
...
-The currently defined types are:
+The currently defined basic types are:
   - no features (no bits set)
   - `option_static_remotekey` (bit 12)
   - `option_anchor_outputs` and `option_static_remotekey` (bits 20 and 12)
   - `option_anchors_zero_fee_htlc_tx` and `option_static_remotekey` (bits 22 and 12)
 
+Each basic type has the following variations allowed:
+  - `option_scid_alias` (bit 46)
+  - `option_zeroconf` (bit 48)
```
Crypt-iQ, line 218: Does the `option_zeroconf` bit depend on `option_scid_alias`? I feel like it should since you need an alias anyways, and with `option_scid_alias`, you MUST set an alias.

Crypt-iQ: I guess then you couldn't have a public zero-conf channel, but I still think the spec should say MUST set an alias if zero-conf

02-peer-protocol.md
TheBlueMatt: Agree it should be clarified but it means the feature bit. Providing an SCID alias for a channel that does not have the scid_alias feature set is an important supported behavior.

02-peer-protocol.md
TheBlueMatt: You can totally have a public zero conf channel! Obviously it can't be announced until the channel confirms, but you can use it privately until then, and you can't have the SCID_alias feature on a public channel.

02-peer-protocol.md
```diff
@@ -461,31 +471,66 @@ use `option_static_remotekey`, `option_anchor_outputs` or
...
+1. type: 36 (`channel_ready`)
 2. data:
     * [`channel_id`:`channel_id`]
-    * [`point`:`next_per_commitment_point`]
+    * [`point`:`second_per_commitment_point`]
+    * [`channel_ready_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `channel_ready_tlvs`
+2. types:
+    1. type: 1 (`short_channel_id`)
```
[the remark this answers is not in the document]
TheBlueMatt: No, the feature bit doesn't indicate anything about the presence or lack of SCID alias field. You're also welcome to send an SCID alias without the feature bit, and the peer can interpret that or ignore it as they see fit. This greatly simplifies the logic in all clients (except apparently lnd, which has a different approach to features).

Note further that the scid_alias feature is somewhat of a misnomer, it's really "scid alias *only*" (in LDK we're calling it "scid privacy" instead), not "scid alias".

TheBlueMatt:
Should we rename the scid_alias feature to scid_privacy? That seems a bit more clear as to what it is (and this is what we ended up calling it in LDK).

02-peer-protocol.md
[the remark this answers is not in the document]
Crypt-iQ: If the spec allows sending a SCID alias without the feature bit (i.e. multiple funding_locked), then what was the reasoning behind this comment https://github.com/lightning/bolts/pull/910/files#r807436433

02-peer-protocol.md
Crypt-iQ: Shouldn't this include private zero-conf channels?

TheBlueMatt: Its already implied cause you're not allowed to set a public channel as `scid_alias`.

02-peer-protocol.md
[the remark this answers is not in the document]
TheBlueMatt: I believe that comment is on the wrong line.

02-peer-protocol.md
TheBlueMatt: Yea, that seems reasonable, the gating here should be on the feature bit, not the channel type.

TheBlueMatt:
One note about renaming `scid_alias` to `scid_privacy`, see note above about gating sending multiple `funding_locked` messages on the feature bit - meaning the feature bit implies "ability to receive multiple funding_locked/channel_ready messages" while the channel type means something completely different. Maybe they need different bits but not a lot of reason to not re-use it, sitll means the rename would be confusing.

02-peer-protocol.md
Crypt-iQ: The clause you're referencing only requires that if the option_scid_alias channel_type was negotiated rather than the feature bit?

02-peer-protocol.md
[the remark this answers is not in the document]
Crypt-iQ: So the comment I linked was about sending multiple funding_locked, which as I understand it should be gated on the feature bit. But you're saying here that you can send anything in the TLV (as usual) including this scid alias regardless of the feature bit. If I understand you correctly here, then I agree

02-peer-protocol.md
TheBlueMatt: Yea, the current text of this PR, however, only adds the only-use-alias-when-forwarding requirement for the "`channel_type` has `option_scid_alias` set", which is correct, I believe. There is nothing implied by the feature bit here.

Crypt-iQ: For private zero-conf channels, this line seems to be ignored (the zero-conf and option-scid-alias channel types seem to be exclusive)

pm47 (commented):
Totally dropped the ball 😬, but I'm back to it now!

07-routing-gossip.md
```diff
@@ -445,10 +445,12 @@ or `node_id_2` otherwise.
 ### Requirements
 
 The origin node:
-  - MUST NOT send a created `channel_update` before `funding_locked` has been received.
+  - MUST NOT send a created `channel_update` before `channel_ready` has been received.
   - MAY create a `channel_update` to communicate the channel parameters to the
```
Crypt-iQ, line 449: I think once the `announcement_signatures` has been exchanged, each side should send the `channel_update` with the real SCID to each other since prior to this point, they might have been using alias SCIDs in the `channel_update`. Unless this is already implied since both sides will broadcast (to each other, the network) the "real" `channel_update`?

09-features.md
```diff
@@ -42,7 +42,9 @@ The Context column decodes as follows:
 | 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC transactions   | IN       | `option_static_remotekey` | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-sighash-single-harmful]|
 | 26/27 | `option_shutdown_anysegwit`         | Future segwit versions allowed in `shutdown`              | IN       |                   | [BOLT #2][bolt02-shutdown]   |
 | 44/45 | `option_channel_type`            | Node supports the `channel_type` field in open/accept     | IN       |                   | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
+| 46/47 | `option_scid_alias`              | Supply channel aliases for routing                        | IN       |                   | [BOLT #2][bolt02-channel-ready]   |
 | 48/49 | `option_payment_metadata` | Payment metadata in tlv record | 9 | | [BOLT #11](11-payment-encoding.md#tagged-fields)
+| 50/51 | `option_zeroconf`                | Understands zeroconf channel types                        | IN       | `option_scid_alias` | [BOLT #2][bolt02-channel-ready]   |
```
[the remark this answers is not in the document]
Crypt-iQ: If there is no dependency link, how do you enforce aliases are used for zero-conf channels? Is this not the point of option-scid-alias feature bit?

02-peer-protocol.md
TheBlueMatt: I'm honestly entirely unsure what we're discussing here at this point - the above comments all seem to be on the same page - you can't set scid_alias on a public channel, and for private channels you can set it allowing the limiting of relay to the alias.

09-features.md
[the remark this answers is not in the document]
TheBlueMatt: The dependency here only applies to the init and node_announcement contexts. There is no such dependency in the channel type contexts. Seems to me like it makes perfect sense to have an scid-alias dependency in those contexts.

07-routing-gossip.md
```diff
@@ -445,10 +445,12 @@ or `node_id_2` otherwise.
...
   channel peer, even though the channel has not yet been announced (i.e. the
   `announce_channel` bit was not set).
+    - MUST set the `short_channel_id` to either an `alias` it has
+	  received from the peer, or the real channel `short_channel_id`.
```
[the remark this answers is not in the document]
TheBlueMatt: > That's really confusing I think. In your example below: if B is the alias set by Bob for Bob->Alice, then B is what Bob should use in the channel_update for Bob->Alice. Otherwise the same alias points to one direction (in the routing hint) and the opposite direction (in a payment error)

Its confusing in that you have to generate channel updates differently based on the context, but you already likely have to do that to handle the different cases for onions now anyway - using real scids for public channels and refusing to do so for private ones.

> Curious how you have implemented it @TheBlueMatt? Did you use the sender or receiver alias for channel_update?

The receiver's alias makes sense here - the receiver isn't otherwise required to understand the sender's alias at all, and moreso its an optional TLV so we have to assume they don't know how to interpret it at all.

09-features.md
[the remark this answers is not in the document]
pm47: Assuming that we rename `option_scid_alias` to `option_scid_privacy` (as you suggested and I agree with you).

The two features introduced by this PR: `option_scid_privacy` and `option_zeroconf`, are very different and independent from each other. We can have one without the other. It just happens that they both depend on supporting scid alias, but there is no explicit feature for supporting aliases, just like e.g. `option_payment_metadata` depends on supporting tlvs.

> The dependency here only applies to the init and node_announcement contexts. There is no such dependency in the channel type contexts

Dependencies as layed out in BOLT 9 are not context dependent though.

07-routing-gossip.md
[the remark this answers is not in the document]
TheBlueMatt: Right, but the receiver doesn't have to use the alias in their invoices at all, if they don't want to (without the scid_privacy channel type set). This is, again, consistent with the SCID alias being an optional TLV, with no feature bit assigned to describing its presence.

09-features.md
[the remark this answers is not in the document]
cdecker: ~How can we ever announce a channel, if setting the `option_scid_alias` literally forces us to set the `announce_channel` flag in the `open_channel` message, and that flag is impossible to renegotiate once the negotiation is done? I think this literally makes public/announced channels and zeroconf mutually exclusive.~

Oh I see, it can still be negotiated as a blank feature, not as part of the `channel_type` at all.

https://github.com/lightning/bolts/pull/910/files#diff-ed04ca2c673fd6aabde69389511fa9ee60cb44d6b2ef6c88b549ffaa753d6afeR246-R247

07-routing-gossip.md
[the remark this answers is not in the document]
pm47: So, we are optimizing for the case where both the sender and the receiver understand aliases, but neither of `option_scid_privacy` or `option_zeroconf` are enabled, and the alias isn't used for anything except in the `channel_update`, and in a way that creates the confusion I layed out above? I don't know why we would want that.

In my opinion we should instead:
- if `option_scid_privacy` or `option_zeroconf` is enabled: have each side always use their own alias for their own `channel_update`
- otherwise: use the real `scid` for `channel_update`s.

This too would be consistent with the alias being an optional TLV: in the latter case the sender is always free to set the alias, and the receiver may or may not use it in their invoice.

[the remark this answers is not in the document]
Crypt-iQ: For public zero-conf channels, you'd also want to exchange the real SCID in updates

02-peer-protocol.md
```diff
@@ -207,12 +207,16 @@ definitions they reuse even feature bits, but they are not an
...
 affect the channel operation).
 
-The currently defined types are:
+The currently defined basic types are:
   - no features (no bits set)
   - `option_static_remotekey` (bit 12)
   - `option_anchor_outputs` and `option_static_remotekey` (bits 20 and 12)
   - `option_anchors_zero_fee_htlc_tx` and `option_static_remotekey` (bits 22 and 12)
 
+Each basic type has the following variations allowed:
```
Crypt-iQ, line 216: are these mutually exclusive

07-routing-gossip.md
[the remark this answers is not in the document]
TheBlueMatt: > So, we are optimizing for the case 

I wouldn't say we're "optimizing" for any case here - this is the only way it can work. Otherwise what value do you put in the channel_update - you sent a SCID alias, you don't know if the peer supports it or remembered it at all, you cannot, there is no feature to indicate the peer must support it, there is no ACK of the scid alias you sent. They may have sent their channel_ready before you, so you can't use that as an ACK that they stored your SCID alias either.

> if option_scid_privacy or option_zeroconf is enabled:

Its not crazy to want to use SCID aliases even if the privacy extension is not negotiated on the channel, eg for backwards compat many LDK clients will do this today.

02-peer-protocol.md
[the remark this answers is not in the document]
TheBlueMatt: Ping on this one.

02-peer-protocol.md
TheBlueMatt: I assume you're saying we should be explicit in the text to say that they are not mutually exclusive? You obviously want to support both.

07-routing-gossip.md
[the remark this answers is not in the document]
pm47: >  what value do you put in the channel_update [...], you don't know if the peer supports it or remembered it at all

If `option_scid_privacy` or `option_zeroconf` is negotiated, you use the scid alias in the `channel_update` because you know that the peer supports it and they must remember it. Otherwise, you use the real scid just like before this PR, and the peer still has the option to use your alias in their invoice just like you wish.

> there is no ACK of the scid alias you sent

No ACK is required if the alias is used only when `option_scid_privacy` and/or `option_zeroconf` have been negotiated. You only have this ACK problem if you don't know whether the peer supports it, which is why we should simply use the real scid in that case, instead of sometimes using the sender's alias, sometimes the receiver's alias for the same `channel_update`.

> Its not crazy to want to use SCID aliases even if the privacy extension is not negotiated on the channel

I agree, and again, what I propose do not prevent that.

> For public zero-conf channels, you'd also want to exchange the real SCID in updates

Indeed: we would use the scid alias until the funding tx is confirmed, and the real scid after that.

cdecker:
One thing that I've been wondering while implementing this is that we pretty much are deciding whether we allow a zeroconf channel to be opened to us at connection time by either signaling or not signaling support in the `init` message. If we signal there seems to be no good way of backing out again, since the funder can chose to add it to its `channel_type` or not. If the funder decides to add it the only option we have if we don't want to allow it is to reject the `open_channel`, or is there something I'm overlooking?

This is a bit unwieldy since it doesn't allow us to decide at channel open time, because at that time we're already locked into what we signalled before, so I'm pretty sure I'm misreading this, but I can't find where.

pm47:
> One thing that I've been wondering while implementing this is that we pretty much are deciding whether we allow a zeroconf channel to be opened to us at connection time by either signaling or not signaling support in the init message. If we signal there seems to be no good way of backing out again, since the funder can chose to add it to its channel_type or not. If the funder decides to add it the only option we have if we don't want to allow it is to reject the open_channel, or is there something I'm overlooking?

I understand it the same way as you. The way it makes sense to me is in the context of only signalling the support for a set of peers that you trust (can you do that with cln?) and with which you would always agree to do a zero-conf channel. But now that I think of it, even for trusted peers you may agree to go zero-conf only when fees are low, or up to a certain funding amount. 🤔

07-routing-gossip.md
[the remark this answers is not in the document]
TheBlueMatt: > If option_scid_privacy or option_zeroconf is negotiated, you use the scid alias in the channel_update because you know that the peer supports it and they must remember it.

Or we could not add any branching based on feature flags, no complicated logic to figure out which one to use and just keep it as-is :). I'm not really sure why you're advocating this - yea, its a bit awkward from a theoretical perspective, but the way its written now leaves the code much, much simpler.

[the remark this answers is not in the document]
pm47: There will be branching anyway: either based on feature flags (I'd argue they are built for that purpose), or based on context (depending on what the `channel_update` is built for). Anyway I think both positions are pretty clear let's see what others think 🤓.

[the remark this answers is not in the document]
TheBlueMatt: Oh, apologies, using only "option_scid_privacy or option_zeroconf" doesn't work - you're totally allowed to start using a channel at 0conf even without `option_zeroconf`. This is in fact an important use-case because I believe only lnd is gonna bother using `option_zeroconf`, others can/will simply send a channel_ready without any confirmations.

TheBlueMatt:
Also, re: negotiation, note that Rusty's originally proposed implementation, which the spec very carefully supports here, totally allows using a channel without any confirmations and without the zero_conf channel type. It was added because the lnd folks demanded a way to negotiate it up front (after out-of-band negotiation) instead of only relying on out-of-band negotiation. Its totally reasonable for an implementation to never request the zero_conf channel type and simply send a channel_ready once the user indicates that out-of-band negotiation has completed and the channel is ready to go without confirmations.

cdecker:
Very interesting, so it seems we agree on the interpretation. However it makes negotiation relatively inflexible.

One thing that came to mind, to avoid the restart semantics, would be to weaken the requirement for `accept_channel` to have to return the exact `channel_type` and allowing it to blank variants, such as `option_scid_alias` and `option_zeroconf` to signal whether they're happy with the negotiated paramenters. This would allow a more dynamic way of determining whether we want zeroconf or not (in CLN as a hook deferring the decision to a plugin for example).

Validation would then just mean that the funder checks that the `channel_type` is a subset of the offered `channel_type` bits and act accordingly. WDYT?

TheBlueMatt:
This seems like an unrelated conversation - redoing the whole channel type negotiation flow is something we could maybe consider, but its already (I believe) implemented in every BOLT-participating lightning node implementation today, so I'm not sure we need to address that here.

cdecker:
> This seems like an unrelated conversation - redoing the whole channel type negotiation flow is something we could maybe consider, but its already (I believe) implemented in every BOLT-participating lightning node implementation today, so I'm not sure we need to address that here.

Absolutely, not suggesting we change it here, it's better suited for a new PR, with associated discussion. Just wanted to float the idea, since we have some context here :-)

07-routing-gossip.md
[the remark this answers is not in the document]
TheBlueMatt: > which allows each side to tell whether the other side understand aliases

Not currently, you'd be adding an assumption that if a peer sent you an alias they also stored the alias you sent them, which is a new assumption that doesn't seem like it needs to make?

> It is still strictly simpler than using the receiver's scid alias, which requires the same branching + makes the scid context-dependent for the channel_update.

This doesn't match up with our implementation experience at all. On the sending side we just do `chan.get_short_channel_id().or(chan.latest_inbound_scid_alias())` which is nice and clean. On the receiving side, we use the same map we use for outbound relay - we already (obviously) have to have some listing of SCIDs that includes real and outbound-aliases in one place, we currently do not anywhere have a listing of real and inbound-alias SCIDs anywhere. Your proposed change would mandate we store a new hashmap for SCID aliases for the inbound side.

Crypt-iQ:
> The channel type as proposed here does not "negotiate whether or not zero-conf should be used", that's exactly what `minimum_depth` is for. The channel_type as proposed here allows the initiator to say "zero conf or bust, buddy". The initiat-ee is allowed to turn on zero conf whether the feature is set or not (and c-lightning, and maybe us too later, has stated intent to allow this ever after the channel has been funded).
> 
> My current read of the spec here, and my current implementation, never sends a zero conf feature flag at all, and can still happily do zero conf.

Does the rust lightning impl use the scid-alias feature bit? I ask since aliases are necessary to use zero-conf regardless which path is used to enable it

02-peer-protocol.md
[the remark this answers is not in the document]
Roasbeef: Yeah this should be **50** unless we're missing something here. I think maybe we had a different set of feature bits but then were re-org'd as all the pending PR titles were updated?

cdecker:
ACK https://github.com/lightning/bolts/pull/910/commits/34e9cd99dba8d2f562bc63500bacae4b50223a98

ellemouton:
@1440000bytes , yes it is. But this is why if you are on the receiver side you should only accept an htlc on such a channel if you trust the peer you are doing this with (in the pre-six-conf phase). You would not do this with someone that you think will double spend. So this will mostly be between a user and their LSP.

PR #912, opened 2021-09-21
Add payment metadata to payment request (feature 48)

joostjager:
This PR adds a new field to bolt11 invoices for payment metadata. Senders are supposed to carry over this data to a tlv field to provide context for stateless invoice applications.

More background on stateless invoices: https://lists.linuxfoundation.org/pipermail/lightning-dev/2021-September/003236.html

_Please attach comments to the most appropriate line of the diff, so that discussions remain threaded and can eventually be resolved. We all know what PRs look like with too many top level comments._

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -263,6 +263,9 @@ It is formatted according to the Type-Length-Value format defined in [BOLT #1](0
     2. data:
         * [`32*byte`:`payment_secret`]
         * [`tu64`:`total_msat`]
+    1. type: 16 (`payment_metadata`)
+    2. data:
+        * [`...*byte`:`payment_metadata`]
 
 ### Requirements
 
@@ -280,6 +283,9 @@ The writer:
       - MUST include `payment_data`
       - MUST set `payment_secret` to the one provided
       - MUST set `total_msat` to the total amount it will send
+    - if the recipient provided `payment_metadata`:
+      - MUST include `payment_metadata` with every HTLC
+      - MUST not apply any limits to the size of payment_metadata except the limits implied by the fixed onion size
 
 The reader:
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
@@ -301,6 +307,9 @@ Note that `amt_to_forward` is the amount for this HTLC only: a
 ultimate sender that the rest of the payment will follow in succeeding
 HTLCs; we call these outstanding HTLCs which have the same preimage,
 an "HTLC set".
+
+`payment_metadata` is to be included in every payment part, so that
+invalid payment details can be detected as early as possible.
 
 #### Requirements
 
@@ -948,9 +957,9 @@ handling by the processing node.
    * [`u32`:`height`]
 
 The `payment_hash` is unknown to the final node, the `payment_secret` doesn't
-match the `payment_hash`, the amount for that `payment_hash` is incorrect or
+match the `payment_hash`, the amount for that `payment_hash` is incorrect,
 the CLTV expiry of the htlc is too close to the current block height for safe
-handling.
+handling or `payment_metadata` isn't present while it should be.
 
 The `htlc_msat` parameter is superfluous, but left in for backwards
 compatibility. The value of `htlc_msat` always matches the amount specified in
--- a/09-features.md
+++ b/09-features.md
@@ -42,6 +42,7 @@ The Context column decodes as follows:
 | 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC transactions   | IN       |                   | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-sighash-single-harmful]|
 | 26/27 | `option_shutdown_anysegwit`         | Future segwit versions allowed in `shutdown`              | IN       |                   | [BOLT #2][bolt02-shutdown]   |
 | 44/45 | `option_channel_type`            | Node supports the `channel_type` field in open/accept     | IN       |                   | [BOLT #2](02-peer-protocol.md#the-open_channel-message) |
+| 48/49 | `option_payment_metadata` | Payment metadata in tlv record | 9 | | [BOLT #11](11-payment-encoding.md#tagged-fields)
 
 ## Definitions
 
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -140,6 +140,8 @@ Currently defined tagged fields are:
 * `p` (1): `data_length` 52. 256-bit SHA256 payment_hash. Preimage of this provides proof of payment.
 * `s` (16): `data_length` 52. This 256-bit secret prevents forwarding nodes from probing the payment recipient.
 * `d` (13): `data_length` variable. Short description of purpose of payment (UTF-8), e.g. '1 cup of coffee' or 'ナンセンス 1杯'
+* `m` (27): `data_length` variable. Additional metadata to attach to the
+  payment. Note that the size of this field is limited by the maximum hop payload size. Long metadata fields reduce the maximum route length.
 * `n` (19): `data_length` 53. 33-byte public key of the payee node
 * `h` (23): `data_length` 52. 256-bit description of purpose of payment (SHA256). This is used to commit to an associated description that is over 639 bytes, but the transport mechanism for the description in that case is transport specific and not defined here.
 * `x` (6): `data_length` variable. `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
@@ -216,7 +218,8 @@ A reader:
     - MUST use that as [`payment_secret`](04-onion-routing.md#tlv_payload-payload-format)
   - if the `c` field (`min_final_cltv_expiry`) is not provided:
     - MUST use an expiry delta of at least 18 when making the payment
-
+  - if an `m` field is provided:
+    - MUST use that as [`payment_metadata`](04-onion-routing.md#tlv_payload-payload-format)
 ### Rationale
 
 The type-and-length format allows future extensions to be backward
@@ -234,6 +237,9 @@ complex orders. Thus, the `h` field allows a summary: though the method
 by which the description is served is as-yet unspecified and will
 probably be transport dependent. The `h` format could change in the future,
 by changing the length, so readers ignore it if it's not 256 bits.
+
+The `m` field allows metadata to be attached to the payment. This supports
+applications where the recipient doesn't keep any context for the payment.
 
 The `n` field can be used to explicitly specify the destination node ID,
 instead of requiring signature recovery.
@@ -703,6 +709,33 @@ Breakdown:
   * `qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqq`
 * `z599y53s3ujmcfjp5xrdap68qxymkqphwsexhmhr8wdz5usdzkzrse33chw6dlp3jhuhge9ley7j2ayx36kawe7kmgg8sv5ugdyusdcq`: signature
 * `zn8z9x`: Bech32 checksum
+
+> ### Please send 0.01 BTC with payment metadata 0x01fafaf0
+> lnbc10m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdp9wpshjmt9de6zqmt9w3skgct5vysxjmnnd9jx2mq8q8a04uqsp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q2gqqqqqqsgq7hf8he7ecf7n4ffphs6awl9t6676rrclv9ckg3d3ncn7fct63p6s365duk5wrk202cfy3aj5xnnp5gs3vrdvruverwwq7yzhkf5a3xqpd05wjc
+
+Breakdown:
+
+* `lnbc`: prefix, Lightning on Bitcoin mainnet
+* `10m`: amount (10 milli-bitcoin)
+* `1`: Bech32 separator
+* `pvjluez`: timestamp (1496314658)
+* `p`: payment hash
+  * `p5`: `data_length` (`p` = 1, `5` = 20; 1 * 32 + 20 == 52)
+  * `qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypq`: payment hash 0001020304050607080900010203040506070809000102030405060708090102
+* `d`: short description
+  * `p9`: `data_length` (`p` = 1, `9` = 5; 1 * 32 + 5 == 37)
+  * `wpshjmt9de6zqmt9w3skgct5vysxjmnnd9jx2`: 'payment metadata inside'
+* `m`: metadata
+  * `q8`: `data_length` (`q` = 0, `8` = 7; 0 * 32 + 7 == 7)
+  * `q8a04uq`: 0x01fafaf0
+* `s`: payment secret
+  * `p5`: `data_length` (`p` = 1, `5` = 20; 1 * 32 + 20 == 52)
+  * `zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs`: 0x1111111111111111111111111111111111111111111111111111111111111111
+* `9`: features
+  * `q2`: `data_length` (`q` = 0, `2` = 10; 0 * 32 + 10 == 10)
+  * `gqqqqqqsgq`: [b01000000000000000000000000000000000100000100000000] = 8 + 14 + 48
+* `7hf8he7ecf7n4ffphs6awl9t6676rrclv9ckg3d3ncn7fct63p6s365duk5wrk202cfy3aj5xnnp5gs3vrdvruverwwq7yzhkf5a3xqp`: signature
+* `d05wjc`: Bech32 checksum
 
 # Examples of Invalid Invoices
 


t-bast:
Interestingly, this will come for free with route blinding: https://github.com/lightningnetwork/lightning-rfc/pull/765
We introduce a `recipient_secret` in a payload that is encrypted for the recipient-only, and lets it avoid having to store data to validate a blinded route. That field can hold arbitrary data, so it does also enable your use-case.

11-payment-encoding.md
```diff
@@ -140,6 +140,8 @@ Currently defined tagged fields are:
 * `p` (1): `data_length` 52. 256-bit SHA256 payment_hash. Preimage of this provides proof of payment.
 * `s` (16): `data_length` 52. This 256-bit secret prevents forwarding nodes from probing the payment recipient.
 * `d` (13): `data_length` variable. Short description of purpose of payment (UTF-8), e.g. '1 cup of coffee' or 'ナンセンス 1杯'
+* `m` (14): `data_length` variable. Additional metadata to attach to the
```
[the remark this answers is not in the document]
t-bast: Agreed, I believe this needs a feature bit and it should be set as mandatory in the invoice to ensure payers know how to behave.

11-payment-encoding.md
```diff
@@ -235,6 +239,9 @@ by which the description is served is as-yet unspecified and will
 probably be transport dependent. The `h` format could change in the future,
 by changing the length, so readers ignore it if it's not 256 bits.
 
+The `m` field allows metadata to be attached to the payment. This supports
```
t-bast, line 242: Even if we don't set a hard rule in the spec on the maximum size of this field, implementations will likely set a threshold to ensure there is enough space in the onion for other fields and a long enough route, so we should at least mention it here.

11-payment-encoding.md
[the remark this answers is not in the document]
joostjager: Ok, added.

11-payment-encoding.md
joostjager: Added

09-features.md
```diff
@@ -41,6 +41,7 @@ The Context column decodes as follows:
 | 20/21 | `option_anchor_outputs`          | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)         |
 | 22/23 | `option_anchors_zero_fee_htlc_tx` | Anchor commitment type with zero fee HTLC transactions   | IN       |                   | [BOLT #3][bolt03-htlc-tx], [lightning-dev][ml-sighash-single-harmful]|
 | 26/27 | `option_shutdown_anysegwit`         | Future segwit versions allowed in `shutdown`              | IN       |                   | [BOLT #2][bolt02-shutdown]   |
+| 48    | `option_payment_metadata` | Payment metadata in tlv record | 9 | | [BOLT #11](11-payment-encoding.md#tagged-fields)
```
joostjager, line 44: Grabbed feature bit 48 for this.

04-onion-routing.md
```diff
@@ -263,6 +263,9 @@ It is formatted according to the Type-Length-Value format defined in [BOLT #1](0
     2. data:
         * [`32*byte`:`payment_secret`]
         * [`tu64`:`total_msat`]
+    1. type: 10 (`payment_metadata`)
+    2. data:
+        * [`len*byte`:`payment_metadata`]
```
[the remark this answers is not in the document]
joostjager: Haha, yes I wasn't sure either. Copied this from another part of the spec where a var array was specified.

rustyrussell:
So, @joostjager pointed out that you could just use "beer" and mix with an internal secret; the metadata doesn't need to be a  UUID.

In general, however, this is not robust: you would not know if you have issued two such invoices simultaneously (if you did it serially, it's OK, since presumably you remember successfully-paid invoices, and could check for clashes).

And, if it serves as a UUID, it also serves as a payment secret?

09-features.md
TheBlueMatt: This should have an optional bit too, no? And the category is missing.

TheBlueMatt (commented):
Agreed, lets get payer-side support for this sooner rather than later.

09-features.md
joostjager: My thought was that if an invoice specifies metadata, you have to include it in the onion as a sender. Therefore a required invoice feature. We can add an optional bit, but is there a use case for that? A use case where including the metadata is somehow better for the sender, but it will also work without?

t-bast (commented):
I have a draft PR ready on the eclair side if you want to do cross-compatibility tests: https://github.com/ACINQ/eclair/pull/2063

11-payment-encoding.md
```diff
@@ -140,6 +140,8 @@ Currently defined tagged fields are:
 * `p` (1): `data_length` 52. 256-bit SHA256 payment_hash. Preimage of this provides proof of payment.
 * `s` (16): `data_length` 52. This 256-bit secret prevents forwarding nodes from probing the payment recipient.
 * `d` (13): `data_length` variable. Short description of purpose of payment (UTF-8), e.g. '1 cup of coffee' or 'ナンセンス 1杯'
+* `m` (14): `data_length` variable. Additional metadata to attach to the
+  payment. Note that implementations may apply limits to the size of this field because the size of the hop payload itself is limited as well.
```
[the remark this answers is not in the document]
joostjager: Can decode it alright, but when I encode myself I get `lnbc10m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdp9wpshjmt9de6zqmt9w3skgct5vysxjmnnd9jx2wq8q8a04uqnp4q0n326hr8v9zprg8gsvezcch06gfaqqhde2aj730yg0durunfhv66sp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q2gqqqqqqsgq7h4ea65u4zwf4j6jjwru7wqcvdtes8z54pphp08as630aljqvlxpyxrde8ddfyy8fggcctlzkqwvhggd6lkg00406urur6yc2p8djwgp6tyhkv`.

One difference seems to be the order of the fields. Is there a requirement for order?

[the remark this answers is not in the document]
joostjager: Btw, do you have an easy way to get that bolt11 string decomposition that the other test cases have?

09-features.md
TheBlueMatt: Yea, I don't see why we should forbid using metadata as an optional thing. We could debate whether we should set a feature bit for that vs just have the field in the invoice, but why would we want to forbid such a use?

11-payment-encoding.md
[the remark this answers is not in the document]
joostjager: Okay, there are some problems here. `m` isn't 14. Will get back with a proper vector.

09-features.md
joostjager: My general answer would be because every variation has a development and maintenance cost.

Maybe it is worth thinking a little harder about the actual use case. An invoice with optional metadata. New implementations will send the metadata, but older ones will not. The receiver needs that metadata, but is also able to handle a payment without. How can this be used?

04-onion-routing.md
```diff
@@ -263,6 +263,9 @@ It is formatted according to the Type-Length-Value format defined in [BOLT #1](0
     2. data:
         * [`32*byte`:`payment_secret`]
         * [`tu64`:`total_msat`]
+    1. type: 10 (`payment_metadata`)
```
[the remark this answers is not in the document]
joostjager: Updated

11-payment-encoding.md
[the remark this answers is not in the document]
joostjager: Okay, I have added a test vector with decomposition. Tag `m` is now number 27 as is should be.

11-payment-encoding.md
```diff
@@ -235,6 +238,9 @@ by which the description is served is as-yet unspecified and will
 probably be transport dependent. The `h` format could change in the future,
 by changing the length, so readers ignore it if it's not 256 bits.
 
+The `m` field allows metadata to be attached to the payment. This supports
+applications where the recipient doesn't keep any context for the payment.
```
cdecker, line 242: This is more a rationale for how this field could be used, but i can see other use cases as well (split PoS and recipient, ...). I'd personally prefer having these separate from the prescriptive text.

09-features.md
t-bast: I have a different argument for allowing the optional bit number.

If we don't allow the optional bit to be set, we cannot monitor whether payers have upgraded or not, so we will never have enough data to know whether we can turn it on with the mandatory bit. We will be forced to just decide at some point to make it mandatory in a specific release, and realize afterwards that it is actually breaking a lot of payments.

The way I plan on using it in eclair is that in the initial release, the optional bit will be set and a small, dummy metadata field will be included in every invoice. This way receivers can track what percentage of their payers actually support this option (by checking whether they included the metadata in the payment or not), and when it reaches 100% (or a good enough threshold) they can decide to make it mandatory, knowing that it will not affect their existing customers.

Does that sound reasonable?

11-payment-encoding.md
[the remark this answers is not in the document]
t-bast: LGTM, I added your test vector to https://github.com/ACINQ/eclair/pull/2063 and it works fine.
Good catch on the `m (14)` instead of `m (27)`, I somehow missed that.

11-payment-encoding.md
```diff
@@ -216,7 +218,8 @@ A reader:
     - MUST use that as [`payment_secret`](04-onion-routing.md#tlv_payload-payload-format)
   - if the `c` field (`min_final_cltv_expiry`) is not provided:
     - MUST use an expiry delta of at least 18 when making the payment
-
+  - if an `m` field is provided:
+    - MUST use that as [`payment_metadata`](04-onion-routing.md#tlv_payload-payload-format)
```
t-bast, line 222: nit: please add an empty line between this and `### Rationale`

t-bast (commented):
ACK https://github.com/lightning/bolts/pull/912/commits/41962a73cc4d95e69428bb69e8ba37054b856057

11-payment-encoding.md
t-bast: But this part is already inside the rationale section, isn't it? So it should be ok? Or do you mean that you don't want something as concrete here, and would rather have a separate doc (a spark?) explain how the `payment_metadata` field can be used for invoices not stored in a local DB?

09-features.md
TheBlueMatt: > My general answer would be because every variation has a development and maintenance cost.

Hmm, but I'm not even sure what the variation here is? As far as I'm aware, all implementations today have some logic to parse feature bits and reject things if there's unknown even bits, and ignore odd bits. This bit being odd should be +0-0 in terms of LoC, at least it would be in LDK.

I guess I don't understand what your view is on optional use (and maybe the spec needs clarification) - is it okay to include a metadata field and no feature bit (and should the sender, then, include the metadata if they understand it?)?

> I have a different argument for allowing the optional bit number.

It seems like this is doable with or without the feature bit? You can still track the number of senders that included the new data.

joostjager: >The way I plan on using it in eclair is that in the initial release, the optional bit will be set and a small, dummy metadata field will be included in every invoice. This way receivers can track what percentage of their payers actually support this option

Interesting use case for optional. I hadn't thought about this before, but indeed, it is a critical that a large majority has upgraded before switching over to stateless invoices.

t-bast: > It seems like this is doable with or without the feature bit? You can still track the number of senders that included the new data.

True, but it requires you to include a `payment_metadata` in the invoice without activating the feature bit, which is a bit less clean in my opinion, if we have a feature bit for this, we should use it.

joostjager: Agreed. We would need to implement optimistic metadata sending without feature bit, not the cleanest logic.

lightning-developer (changes requested):
> ACK [41962a7](https://github.com/lightning/bolts/commit/41962a73cc4d95e69428bb69e8ba37054b856057)

Maybe I misunderstood the discussion following your ACK between @joostjager @t-bast and @TheBlueMatt but wasn't there an argument that we should add the optional feature bit so that we can track adoption in wallet software?

t-bast:
@lightning-developer this is being discussed here: https://github.com/lightning/bolts/pull/912#discussion_r717313582

09-features.md
joostjager: Added optional bit

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/912/commits/1f041b110f5faac879e95827673b36a6e083cea7

11-payment-encoding.md
lightning-developer, line 222: What happens if the payer does not use this as [`payment_metadata`](04-onion-routing.md#tlv_payload-payload-format) in the `tlv_payload`? 

In [BOLT 04 we have the onion failure message](https://github.com/joostjager/lightning-rfc/blob/custom-invoice-data/04-onion-routing.md#failure-messages)  1. type: PERM|15 (`incorrect_or_unknown_payment_details`)


https://github.com/joostjager/lightning-rfc/blob/custom-invoice-data/04-onion-routing.md#L953 currently reads: 

> The `payment_hash` is unknown to the final node, the `payment_secret` doesn't
> match the `payment_hash`, the amount for that `payment_hash` is incorrect or
> the CLTV expiry of the htlc is too close to the current block height for safe
> handling.

Maybe we should also include that this error is sent if the `payment_metadata` was not included?

lightning-developer (changes requested):
According to the rational in BOLT11  the payer MUST use [`payment_metadata` in the `tlv_payload`](04-onion-routing.md#tlv_payload-payload-format). Maybe we should also send `incorrect_or_unknown_payment_details` if the `payment_metadata` field was not included in the `tlv_payload`?

11-payment-encoding.md
```diff
@@ -140,6 +140,8 @@ Currently defined tagged fields are:
 * `p` (1): `data_length` 52. 256-bit SHA256 payment_hash. Preimage of this provides proof of payment.
 * `s` (16): `data_length` 52. This 256-bit secret prevents forwarding nodes from probing the payment recipient.
 * `d` (13): `data_length` variable. Short description of purpose of payment (UTF-8), e.g. '1 cup of coffee' or 'ナンセンス 1杯'
+* `m` (27): `data_length` variable. Additional metadata to attach to the
+  payment. Note that implementations may apply limits to the size of this field because the size of the hop payload itself is limited as well.
```
TheBlueMatt, line 144: This isn't clear, and really hard to implement to - if an implementation hits the limit does it fail to send the payment or does it truncate the result? Are implementations expected to simply add a hard-coded limit and fail invoices over a threshold? If so what is that threshold supposed to be (and, critically, below which number should an invoice-creator target keeping this value to ensure their invoices will be parsed?).

joostjager: What I implemented is:
* Fail payment when limit is exceeded, no truncation.
* No hard-coded limit. It tries to find a route that can deliver the full metadata and stay below the 1300 bytes max size. This means that the more metadata there is, the shorter the max route length. This may be a concern, because the receiver can influence pathfinding indirectly.
* Receivers need to be aware that long metadata strings may decrease the success chance for the payment.

11-payment-encoding.md
joostjager: Or we can use `invalid_onion_payload`?

lightning-developer: I am a bit confused right now. Which of the two error codes protects the privacy of the recipient better?

I guess if an intermediate malicious node interrupts routing and sends onions to various candidate recipients without including `payment_metadata` those wrong candidates would regularly send `incorrect_or_unknown_payment_details` back to the intermediate node. However following your suggestion the real recipient would send back `invalid_onion_payload` as they expected the `payment_metadata` field. Thus my feeling is that `incorrect_or_unknown_payment_details` will be more secure as it prevents such an attack. Following #627 I am under the impression that `invalid_onion_payload` is intended to be used if something goes wrong for a routing node when decoding the onion. But maybe I missed something and @t-bast can clarify?

t-bast: > Thus my feeling is that incorrect_or_unknown_payment_details will be more secure as it prevents such an attack.

I agree. If you make `payment_metadata` mandatory and if it wasn't included in the onion you receive, I would send back `incorrect_or_unknown_payment_details`. You mandated the payment to include a `payment_metadata` and it didn't, so it's really an incorrect payment (and gives the best anonymity set as it's the error used whenever something is invalid with the payment at the recipient's end).

joostjager: Updated PR to return `incorrect_or_unknown_payment_details`.

11-payment-encoding.md
t-bast: > If so what is that threshold supposed to be (and, critically, below which number should an invoice-creator target keeping this value to ensure their invoices will be parsed?)

That's a hard question. AFAICT we have only two options:

1. The spec provides a number here (e.g. 64 bytes). It's almost the same as making this field fixed-size, which is the mistake we made with `payment_secret` and don't want to make again. But the encoding is variable-size, so it's better than `payment_secret` and probably easier to update later (if we want to make the size bigger)?
2. The spec doesn't provide a number: every implementation will use a different limit (or every payment will use a different limit with Joost's implementation), which makes it really hard for invoice creators to ensure their invoices can be paid.

I don't think there is a clear winner. Even if invoice creators keep their `payment_metadata` small (below a spec-defined limit), there is always a risk that payers need many hops and are unable to include it in the onion for the (long) paths they find.

joostjager: Limiting `payment_metadata` to something like 64 bytes does sort of defeat the purpose of stateless invoices, unless it is deemed to be sufficient to just include basic invoice data in there for reconstruction.

If you want to put in a delivery address for example, it wouldn't really work.

I would be in favor of letting services decide on the routability/size tradeoff themselves. For custom records, it works in a similar way. The more data is included, the shorter the max route length.

11-payment-encoding.md
lightning-developer: Dear @joostjager I am very pleased that you found my review & suggestion useful. Yet I am a bit surprised that you did not accept the patch that I provided 5 days ago with my suggestion and for your convenience directly as a pull request to your branch at https://github.com/joostjager/lightning-rfc/pull/1. I hope the reason is that you have simply overseen my pull request and patch. As you did not credit my contribution in your recent commit may I ask you kindly to revert the most recent addition and accept the provided patch instead? Thank you very much!

11-payment-encoding.md
cdecker: Also not quite clear on what the right thing to do is here. Any fixed value we put in here is both limiting for some applications and signaling safety to others ("64 bytes will definitely work" 🤔). It's probably best to mention the risk of not fitting in an onion and defer the decision to the developers using it.

Given the tiny network diameter I personally don't see it increasing the failure rate considerably (6 hops << 20 hops -> ~14*65 bytes of room for random signalling in the onion)

TheBlueMatt: I agree its not clear what to do here. I imagine 64 bytes (plus the 32 byte payment secret) should suffice for most use-cases, but if we don't want to hard-code it then I believe the spec should clearly state that nodes MUST NOT apply a limit except as required by routing failures due to too-large-onion.

Relatedly, and maybe invalidating the above, is there a concern of someone detecting the payment distance in a repeated payment by varying the length here and detecting when the payment fails? Maybe we can limit to, like, 128 bytes?

joostjager: Discussed in the spec meeting: allow metadata to use all remaining space in the onion and state that implementations shouldn't limit the size to a lower number.

joostjager: Added

11-payment-encoding.md
joostjager: @lightning-developer, No I didn't see your patch on my personal repo.

Generally speaking and also in this particular case, lots of people contribute ideas. I think it will get complicated to list all of them as co-authors. They may not all have contributed text literally, but an idea often deserves credit too.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/912/commits/b5db023b06c0f5dc2691e37fa5615a0218857e2d, this is implemented by https://github.com/ACINQ/eclair/pull/2063 (I removed the hard-coded limit) if you want to do cross-compatibility tests.

04-onion-routing.md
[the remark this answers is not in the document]
joostjager: Actually, it looks like AMP is using 14 already. Shall we move to 16 then?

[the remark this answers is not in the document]
t-bast: What about AMP updating to another field? It's not documented that they use 14, not even in #658 AFAICT...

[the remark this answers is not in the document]
joostjager: Haha, that is not a q for me to answer :popcorn: 

To be consequent, I think we should use 10 then?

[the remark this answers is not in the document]
t-bast: > To be consequent, I think we should use 10 then?

Not quite, route blinding has documented that it uses 10 and 12 (for 6+ months) and already has two implementations' ACKs and cross-compatibility successfully tested (even before `payment_metadata`, so it does take precedence). AMP has no spec and only one implementation...

Your call, you can either update to 16 here or ask lnd folks. In all cases using 14 for AMP without specifying it anywhere is reckless, it just creates a big risk that another feature will ship across multiple implementations and use 14 for something else which will be a huge pain to update. Until specified, AMP should use a TLV in the high range to avoid running into this kind of compatibility problems.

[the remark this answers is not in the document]
joostjager: Let's update to 16 so that we are unblocked.

Tagging @Roasbeef here for comment on AMP.

04-onion-routing.md
```diff
@@ -280,6 +283,9 @@ The writer:
       - MUST include `payment_data`
       - MUST set `payment_secret` to the one provided
       - MUST set `total_msat` to the total amount it will send
+    - if the recipient provided `payment_metadata`:
+      - MUST include `payment_metadata`
+      - MUST not apply an arbitrary limit to the size of `payment_metadata`.
```
TheBlueMatt, line 288: I mean I'd say "128 bytes" isn't "arbitrary" :). Maybe rephrase as "MUST not apply any limits to the size of `payment_metadata` except the limits implied by the fixed onion size" or so?

04-onion-routing.md
```diff
@@ -948,9 +954,9 @@ handling by the processing node.
    * [`u32`:`height`]
 
 The `payment_hash` is unknown to the final node, the `payment_secret` doesn't
-match the `payment_hash`, the amount for that `payment_hash` is incorrect or
+match the `payment_hash`, the amount for that `payment_hash` is incorrect,
 the CLTV expiry of the htlc is too close to the current block height for safe
-handling.
+handling or `payment_metadata` isn't present while it should.
```
TheBlueMatt, line 959: nit: while it should *be*

TheBlueMatt (commented):
LGTM

04-onion-routing.md
joostjager: Updated.

joostjager:
Completed interop test Eclair->LND: https://github.com/ACINQ/eclair/pull/2063#issuecomment-988597577

11-payment-encoding.md
```diff
@@ -704,6 +710,36 @@ Breakdown:
...
+* `p`: payment hash
+  * `p5`: `data_length` (`p` = 1, `5` = 20; 1 * 32 + 20 == 52)
+  * `qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypq`: payment hash 0001020304050607080900010203040506070809000102030405060708090102
+* `d`: short description
+  * `p9`: `data_length` (`p` = 1, `9` = 5; 1 * 32 + 5 == 37)
+  * `wpshjmt9de6zqmt9w3skgct5vysxjmnnd9jx2`: 'payment metadata inside'
+* `m`: metadata
+  * `q8`: `data_length` (`q` = 0, `8` = 7; 0 * 32 + 7 == 7)
+  * `q8a04uq`: 0x01fafaf0
+* `n`: node id
```
TheBlueMatt, line 731: Can you add a version that doesn't require an explicit node id? Seems this is also the only test that uses an explicit node id (which, btw, can we just deprecate that? no one uses it, right?)

t-bast: I was surprised by this as well, eclair doesn't use the explicit `node_id` either.
It should be ignored so it's fine, but if we plan on deprecating it some day (not sure if that will happen though), it's probably better to avoid including it in a test vector.

joostjager: :+1:, test vector updated

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/912/commits/42dc90bd5b331aa9276748d9c99179adc2d6752c, I verified the updated test vector with eclair.

TheBlueMatt (changes requested):
This needs to specify that the metadata is duplicated across each MPP part (I assume its duplicated? I guess we could go with 'only have to specify it on one path' if we want to avoid breaking longer paths more than we need?).

TheBlueMatt:
Other than the above, ACK. Feel free to land with a comment noting duplication for MPP.

joostjager:
I also thought about the duplication. What I did in `lnd` is to just store the metadata for each htlc that comes in and leave it up to an application to process it. But we should indeed give guidance to senders. It seems there are three options:

1. Attach to only one htlc
2. Attach to all htlcs
3. Attach to at least one htlc

Also do we want to specify that the receiver should fail an MPP when inconsistent metadata is received? So two htlcs with different metadata or too many or not enough htlcs with metadata (in case we go for option 1 or 2).

TheBlueMatt:
My (weak) preference would be to attach to all HTLCs (what if someone wants to authenticate individual HTLCs like we do for payment secrets?) and that a receiver MAY fail the payment if they aren't all the same/don't match what they expect (but leaving that unspecified seems fine too - a receiver MAY always fail a payment for any reason).

t-bast:
I also think we should attach to all payments, especially if at some point in the future we want to deprecate `payment_secret` in favor of `payment_metadata`.

joostjager:
Good points - will update the pr to instruct the sender to send metadata with all htlcs.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/912/commits/2ab3a9f0221e601ba58ebb8bcaee55158b21bf80

TheBlueMatt (approved):

rustyrussell:
Ack https://github.com/lightning/bolts/commit/2ab3a9f0221e601ba58ebb8bcaee55158b21bf80

joostjager:
Carried out interop test lnd->eclair too, that works too.

vincenzopalazzo (commented):
ack https://github.com/ACINQ/lightning-kmp/commit/39e8a459ef8efac6d85a3714c4479756fa801839

with latency, sorry :)

joostjager:
@t-bast I vaguely remember that you planned to add marker metadata to all invoices generated in your systems, to be able to gauge the uptake of the payment metadata feature. Did you do that in the end, and if yes, can you share how it is going?

t-bast:
This was added in eclair and phoenix, but it doesn't give us a very easy way of measuring support. Only the phoenix recipient knows whether the sender included the payment metadata and this information isn't shared with our node. Our node doesn't issue invoices so we don't have any data about that.

This would be more interesting to measure at big hubs that acts as receiver nodes: Wallet of Satoshi, Muun and the likes. Those nodes are generating a lot of invoices and receiving payments and can better measure what percentage of senders do include the metadata.

PR #923, opened 2021-10-12
Feature `option_scid_alias` separate from zero-conf

pm47:
This is an alternative to https://github.com/lightningnetwork/lightning-rfc/pull/910, it is very close with the following changes:

- make zero-conf and `scid` `alias` independent
Fundee may signal that it is willing to trust the funder by setting `minimum_depth` to zero, but that's unenforceable anyway.

- make `scid` `alias` and channel announcement independent
There is no reason to advertize the "real" `scid` in `funding_locked`, because the real `scid` is already provided in `announcement_signatures`

- have nodes only advertise their ability to understand `scid` `alias` (feature `option_scid_alias`)
If this feature is negotiated (both peers support it), then they must only use `alias`es (but that's unenforceable).

- accept that some combinations don't make sense, like "zero-conf without alias", or "alias and public channels"

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -409,10 +409,23 @@ this channel will continue to use `option_static_remotekey` or `option_anchor_ou
 
 This message indicates that the funding transaction has reached the `minimum_depth` asked for in `accept_channel`. Once both nodes have sent this, the channel enters normal operating mode.
 
+Nodes which have funded the channel, or trust their peers to have
+done, or wish to use `push_msat` from other side's opening (without
+accepting incoming HTLCs) can simply send their `funding_locked` right
+away. As soon as both parties have sent their `funding_locked`, the channel
+can be used even if the funding transaction hasn't yet been confirmed.
+
 1. type: 36 (`funding_locked`)
 2. data:
     * [`channel_id`:`channel_id`]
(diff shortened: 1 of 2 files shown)


Crypt-iQ:
> accept that some combinations don't make sense, like "zero-conf without alias", or "alias and public channels"

Some thoughts:
- Is it expected that zero-conf channels are always private even after being confirmed for some amount of time? I could see users wanting it public after some time. However, this amount of time would need to be signalled by using the `minimum_depth` field (e.g. non-zero) before sending `AnnouncementSignatures`?
- If A sends B a `funding_locked` early with an alias, and B responds early without an alias, the zero-conf channel can still be used but A will not be able to issue invoices. I think this "degenerate" case should be marked out explicitly.

Crypt-iQ:
Ah I see `announcement_signatures` are only after 6 confirmations regardless of depth, so scratch my initial comment there. How can a zero-conf channel be public from the start if it does not have a short_channel_id assigned and only an alias? I was specifically referencing this `"zero-conf without alias", or "alias and public channels"` which seems to imply zero-conf cannot be public since it has an alias?

> That would be a spec violation though: if both peers support option_scid_alias, they must send an alias.
What happens if an alias is not received?

PR #932, opened 2021-10-27
Clarify and relax requirements on `channel_reestablish`

pm47:
The peer that is using an outdated commitment must not fail the channel itself, rather it should request the other peer to fail the channel.

Also, I think we could relax the requirements on the "up-to-date" side of the channel: if we detect that remote is using an outdated state, we could also stay idle and give remote a chance to fix its data? Maybe the node operator just accidentally started their node using a wrong directory. Current behavior is a bit trigger happy. Of course if remote sends an `error`, then we force-close.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1381,7 +1381,7 @@ A node:
     `your_last_per_commitment_secret` is correct for that
     `next_revocation_number` minus 1:
       - MUST NOT broadcast its commitment transaction.
-      - SHOULD fail the channel.
+      - SHOULD send an `error` to request the peer to fail the channel.
     - otherwise:
       - if `your_last_per_commitment_secret` does not match the expected values:
         - SHOULD fail the channel.
@@ -1390,7 +1390,7 @@ A node:
     `your_last_per_commitment_secret` is correct for that
     `next_revocation_number` minus 1:
       - MUST NOT broadcast its commitment transaction.
-      - SHOULD fail the channel.
+      - SHOULD send an `error` to request the peer to fail the channel.
       - SHOULD store `my_current_per_commitment_point` to retrieve funds
         should the sending node broadcast its commitment transaction on-chain.
     - otherwise (`your_last_per_commitment_secret` or `my_current_per_commitment_point`


rustyrussell:
There's nothing wrong with this clarification.  A node is supposed to fail the channel on either send or receive of an error; we don't define "fail the channel" but logically "send errors from now on and broadcast the latest commit tx": the spec already says when you must not broadcast (because it's not the latest).

If you didn't want to fail the channel, don't send an error, send a warning.

(C-lightning doesn't fail on error receipt, we'd have to fix that if we wanted to start sending a warning in these cases!).

02-peer-protocol.md
```diff
@@ -1381,7 +1381,7 @@ A node:
     `your_last_per_commitment_secret` is correct for that
     `next_revocation_number` minus 1:
       - MUST NOT broadcast its commitment transaction.
```
lightning-developer, line 1383: as @rustyrussell mentioned and according to BOLT 1 upon sending an `error` the node SHOULD fail the channel  and failing usually means broadcasting the commitment transaction. 

This situation seems a little bit paradox because obviously in this case the node does not want to / MUST NOT broadcast its commitment transaction and thus can't really fail the channel. 

I wonder if this could also leave the node in a problematic situation indicating  the peer that they may attempt to cheat? But I guess it can't be worse than the already existing situation.

lightning-developer (commented):
I wonder if we can either make the recommendation who fails the channel when sending or receiving an error more precise or relax them in BOLT1.

One idea that comes to my mind is that we could define that failing a channel always starts with sending an `error` depending on the state / situation we have to define how the channel is to be failed / how the `error` is being handled. 

Also @rustyrussell wrote: 

> If you didn't want to fail the channel, don't send an error, send a warning.

edit: 

~~Do we have warnings? I could not find a definition for warning messages in BOLT 1.~~

Never mind! I discovered #834 which is a delight because after my review yesterday I was about to open an issue suggesting that we could properly define what we mean with `fail the channel`.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/932/commits/ad3ed9d9fac470a403c35f077d7bdd51f0712253

vincenzopalazzo (approved):
ack https://github.com/lightning/bolts/pull/932/commits/ad3ed9d9fac470a403c35f077d7bdd51f0712253

lightning-developer:
As requested by @t-bast I have cherry-picked from his branch and updated #942. Thus I believe this PR can be closed and #942 is ready to be merged.

t-bast:
Closing in favor of #942 (which has already been merged)

PR #935, opened 2021-11-10
Add table of contents to `README.md`

Bosch-0:
Added table of contents to `README.md`

Lots of blanks entries so opening as a draft for now so people can fill them in.


vincenzopalazzo (commented):
What is the reason to remove the link to the main page, and add it on the Readme directly?

Bosch-0:
Visually it's easier to find the content you are looking for rather than having to go into the introduction to see the table of contents which requires extra unnecessary effort.

Issue #937, opened 2021-11-23
BOLT01 example of `init` message incorrect?


instagibbs:
> 0x00100000000001012a030104: the extension contains two odd TLV records (with types 0x01 and 0x03)

IIUC, `0x01` is `networks`, a known even feature which requires a multiple of 32-byte as a data payload("...*chain_hash"), and this example has single byte lengths for each type. This should cause it to be invalid?

t-bast:
You're right, these test vectors were made before we added the `networks` tlv field to the `init` message.
My bad, I should have made this more future-proof by using tlvs in a much higher range, I'll fix that.
Thanks for the report!

instagibbs:
I could also submit a patch if it's not too much trouble

t-bast:
Great, that would be nice, I'll let you update this one, thanks!

PR #938, opened 2021-11-23
BOLT01: Switch  extension records to unknown fields in test vector

instagibbs:
Should resolve https://github.com/lightning/bolts/issues/937 for quite a long while @t-bast

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -918,7 +918,7 @@ message. The base `init` message (without extensions) for these examples is
 The following `init` messages are valid:
 
 - `0x001000000000`: no extension provided
-- `0x00100000000001012a030104`: the extension contains two _odd_ TLV records (with types `0x01` and `0x03`)
+- `0x00100000000009012a0b0104`: the extension contains two unknown _odd_ TLV records (with types `0x09` and `0x0b`)
 
 The following `init` messages are invalid:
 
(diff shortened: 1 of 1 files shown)


01-messaging.md
```diff
@@ -918,7 +918,7 @@ message. The base `init` message (without extensions) for these examples is
 The following `init` messages are valid:
 
 - `0x001000000000`: no extension provided
-- `0x00100000000001012a030104`: the extension contains two _odd_ TLV records (with types `0x01` and `0x03`)
+- `0x00100000000009012a0b0104`: the extension contains two unknown _odd_ TLV records (with types `0x09` and `0x0b`)
```
t-bast, line 921: Could you use even higher values, so that we're pretty sure we won't have to update it again in 6 months? For example `201` and `203`?

Could you also update the tests in the section below:

- `0x00100000000002012a` should use tlv type `202` instead of `2` otherwise it's likely it will be wrong soon
- `0x001000000000010101010102` should probably use tlv type `201` instead of `1`, right now it's invalid mostly because tlv type 1 should be 32 bytes, not necessarily because the type is duplicated

Thanks!

instagibbs: gave it a shot, agreed

t-bast (approved):
Thanks! ACK https://github.com/lightning/bolts/pull/938/commits/d852387fb2f6506e12e81e9a78ca78ba8d95dd6a

PR #942, opened 2021-12-05
Removed requirement to broadcast an outdated commitment transaction

lightning-developer:
Following the discussion at #932 and looking at the proposal of #834 I thought it might be good to clarify in BOLT 5 when a node MUST NOT (potentially SHOULD NOT) broadcast its own commitment transaction when failing the channel. 

If a node has to fail a channel but knows that its latest commitment transaction is outdated it should not be required to send it but rather wait for the peer to unilaterally close the channel. 

The proposed solution is not so clean because it might produce a deadlock in which two peers assume they have outdated state and send `error` back and forth without actually force closing. Maybe in such a scenario we could create a protocol that mutually closes with split balance? 

Also replaced the word use with broadcast as it seems more accurate.

### DIFF
--- a/05-onchain.md
+++ b/05-onchain.md
@@ -143,15 +143,14 @@ A node:
       sufficient fee:
         - SHOULD use this fee to perform a *mutual close*.
       - otherwise:
-      	- if the node knows or assumes its channel state is outdated it
-      	  -  MUST NOT broadcast its *last commitment transaction*. 
-      	  -  SHOULD send an `error`.
+      	- if the node knows or assumes its channel state is outdated:
+      	  - MUST NOT broadcast its *last commitment transaction*.
         - otherwise:
           - MUST broadcast the *last commitment transaction*, for which it has a
-        signature, to perform a *unilateral close*.
-          - MUST spend any `to_local_anchor` output, providing sufficient fees as incentive to include the commitment transaction in a block
-        Special care must be taken when spending to a third-party, because this re-introduces the vulnerability that was
-        addressed by adding the CSV delay to the non-anchor outputs.
+          signature, to perform a *unilateral close*.
+          - MUST spend any `to_local_anchor` output, providing sufficient fees as incentive to include the commitment transaction in a block.
+          Special care must be taken when spending to a third-party, because this re-introduces the vulnerability that was
+          addressed by adding the CSV delay to the non-anchor outputs.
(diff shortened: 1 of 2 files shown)


t-bast (commented):
ACK https://github.com/lightning/bolts/pull/942/commits/fdd791afdb736e1a32058c74ebb9cb259e1ba2ec

05-onchain.md
```diff
@@ -143,12 +143,16 @@ A node:
       sufficient fee:
         - SHOULD use this fee to perform a *mutual close*.
       - otherwise:
-        - MUST use the *last commitment transaction*, for which it has a
+      	- if the node knows or assumes its channel state is outdated it
+      	  -  MUST NOT broadcast its *last commitment transaction*. 
+      	  -  SHOULD send an `error`.
```
Roasbeef, line 148: Seems related to #932. I think this is a weird place to add this requirement as BOLT 2 is where we define the _p2p_ interaction, while BOLT 5 so far has been reserved primarily for _on chain_ interaction. In other words, this seems to sort of "breach" layers. 

This was discussed during the latest dev call, with some inclination to sort of merge this and #932, remove this line (as it's in #932) and keep the rest of the diff.

Roasbeef (changes requested):

rustyrussell:
This is already stated explicitly in reestablish, but probably worth mentioning here too, indeeed.

```
 A receiving node:
  - if `option_static_remotekey` or `option_anchors` applies to the commitment
    transaction:
    - if `next_revocation_number` is greater than expected above, AND
    `your_last_per_commitment_secret` is correct for that
    `next_revocation_number` minus 1:
      - MUST NOT broadcast its commitment transaction.
      - SHOULD fail the channel.
```

05-onchain.md
t-bast: @lightning-developer I've prepared a commit on top of your branch that does this: https://github.com/t-bast/bolts/commit/ae2d1e9f7128adb19ca2e9d6043505467eb19af3

Can you cherry-pick it and update this PR? We can then include it and simply close #932.

lightning-developer: Done as requested.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/942/commits/b60e2b24b0abdbe13c8a0ef0946c9531c992379f

TheBlueMatt (approved):

vincenzopalazzo (approved):
ack https://github.com/lightning/bolts/pull/942/commits/b60e2b24b0abdbe13c8a0ef0946c9531c992379f

PR #962, opened 2022-02-22
BOLT 4: Remove legacy format, make var_onion_optin compulsory.

rustyrussell:
My measurements a few weeks ago reveal that only 5 nodes do not advertize this feature, of over 17000.  I have a patch to remove support from c-lightning, too.


04-onion-routing.md
```diff
@@ -164,91 +161,29 @@ It is 1300 bytes long and has the following structure:
...
-     values or that the origin node has an obsolete `cltv_expiry_delta` value.
-     The hop MUST be consistent in responding to an unexpected
-     `outgoing_cltv_value`, whether it is the final node or not, to avoid
-     leaking its position in the route.
-
-   * `padding`: This field is for future use and also for ensuring that future non-0-`realm`
-     `hop_data`s won't change the overall `hop_payloads` size.
-
-When forwarding HTLCs, nodes MUST construct the outgoing HTLC as specified
-within `hop_data` above; otherwise, deviation from the specified HTLC
```
t-bast, line 243: Wouldn't it be useful to keep this information?

t-bast (commented):
The future is here! Concept ACK, glad to remove this!

664tub

09-features.md
```diff
@@ -61,6 +61,9 @@ The origin node:
   * MUST NOT set feature bits in fields not specified by the table above.
   * MUST set all transitive feature dependencies.
 
+The origin node:
```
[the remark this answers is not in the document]
rustyrussell: Hmm, I wanted to start a separate list of compulsory features. Will reformat to make that clearer...

04-onion-routing.md
rustyrussell: I can rework it. The Old format mixed requirements with discussion: I moved the requirements, but should probably put the discussion somewhere...

t-bast (commented):
LGTM, just a typo to fix and should be ready to go.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/962/commits/7a6291d4b11b2733a90b40741ab3104c258bb042, I'll prepare a branch to remove the legacy code in eclair as well.

t-bast:
Actually while updating eclair, I realized that we should also update the test vectors:

- remove https://github.com/lightning/bolts/blob/master/bolt04/onion-test-v0.json
- udpate https://github.com/lightning/bolts/blob/master/bolt04/onion-test-multi-frame.json to only contain variable-size payloads

That does mean a bit more work to do (sorry!).

TheBlueMatt (approved):

rustyrussell:
Hmm, I'm out of time to update the multiframe test, though you're right :(

rustyrussell:
Ping @cdecker ?

Roasbeef:
Concept ACK, we'll likely start with just making our existing feature bit required, as we have some existing tests that still assume this is relevant.

t-bast:
I'm looking at our node's statistics, and a non-negligible portion (~30%) of the payments we relay are still using the legacy format.
I'm afraid most nodes advertise the feature, but without actually using it by default when sending payments.
Can you check relay statistics on your nodes to see if it matches what I'm seeing?
Are all implementations always generating onion with the new format (and since which version)?

Roasbeef:
> Are all implementations always generating onion with the new format (and since which version)?

`lnd` has been generating with the new format as of version v0.8.0 (released October 2019). In our path finding code, we'll use the old/new depending on which feature bits the node advertises (so we support mixing old/new along a route).

t-bast:
> In our path finding code, we'll use the old/new depending on which feature bits the node advertises (so we support mixing old/new along a route).

In that case that's probably not the culprit since more than 99% of nodes (including ours) advertise `var_onion_optin`.
I'm suspecting a popular wallet (or a few of them) is still using the legacy encoding for their payments...

vincenzopalazzo:
on c-lightning IRC we have a crash report due to the old onion format too!

>my node seems to be crashing due to invalid onion messages
Unexpected message WIRE_OBS2_ONION_MESSAGE:
using master at commit 7abc491f4c214f2bcf9e8f2b9066ca05dee0342b

t-bast:
> on c-lightning IRC we have a crash report due to the old onion format too!

I find c-lightning should definitely not be crashing, even with the legacy encoding disabled it should simply fail the adds with a `required_node_feature_missing` or something similar, shouldn't it?

Roasbeef:
Some ppl on Twitter hypothesized that maybe a lot of the traffic is actually probing. I know in the past, some systems/applications would use an API for lnd that accept a raw route for probing. I dug into things on our end, and realized that unless a user is setting some custom TLV record in the route (also not using mpp/amp), we'll _default_ to using the legacy payload (there's a bool that defaults to false). So this might be attributing to that 30% number mentioned above...

We have a new major release coming up, so this could be an opportunity to change our API to default to the modern payload (can't thing of any downside really).

Roasbeef:
We've released version v0.14.3 of lnd which removes the ability for users to specify the _legacy_ payload option using certain API calls. However if a node in the path needs the legacy format, then the normal path finding code will handle populating the correct payload.

t-bast:
It looks like lnd's release really helped, we're now only seeing ~5% of the payments we relay use the legacy format. Hopefully in a few months that will drop below 1% and we can deprecate it.

dunxen:
> Hopefully in a few months that will drop below 1% and we can deprecate it.

Any good news here? :)

t-bast:
We're still seeing payments use the legacy onion format...
Over the last 3 days, 2,8% of relayed payments used it.
Over the last week, 1,6% of relayed payments used it.
I'd still wait a bit before releasing a version that rejects those payments.

t-bast:
good news everyone

Payments using the legacy encoding have fallen below 0,2% of the total payments relayed on our node in the last week :tada: 
Let me know if other nodes see similar data, this is a good indication that we may be able to finally remove this soon!

t-bast (commented):
LGTM, I think we can now safely merge this.

This comment isn't addressed yet though: https://github.com/lightning/bolts/pull/962#issuecomment-1050733772

I have created a commit [here](https://github.com/t-bast/bolts/commit/69ea87c831eb23d0a2f328606e5960d89d5a4604) to update the test vectors and use valid tlv streams, feel-free to cherry-pick it (or provide a different set of test vectors).

Roasbeef (approved):
LGTM ❄️

rustyrussell:
> Payments using the legacy encoding have fallen below 0,2% of the total payments relayed on our node in the last week tada Let me know if other nodes see similar data, this is a good indication that we may be able to finally remove this soon!

Confirmed: my node has seen *no* legacy since 17 June (over 12000 htlcs fwd since then), and the one before that was 2 July.

He's dead, Jim!

rustyrussell:
Squashed into a single commit, and included the test vector fixups.  Will test them today...

rustyrussell:
Ack test vectors!

04-onion-routing.md
```diff
@@ -267,11 +202,34 @@ It is formatted according to the Type-Length-Value format defined in [BOLT #1](0
...
+
+For non-final nodes, this includes the origin node's computed _fee_ for the
+receiving peer, calculated according to the receiving peer's advertised fee
+schema (as described in [BOLT #7](07-routing-gossip.md#htlc-fees)).
+
+`outgoing_cltv_value` is the CLTV value that the _outgoing_ HTLC
+carrying the packet should have.  Inclusion of this field allows a hop
+to both authenticate the information specified by the origin node, and
+the parameters of the HTLC forwarded, and ensure the origin node is
+using the current `cltv_expiry_delta` value/
```
t-bast, line 220: nit:

```suggestion
using the current `cltv_expiry_delta` value.
```

bolt04/onion-test.json
```diff
@@ -0,0 +1,37 @@
+{
+  "comment": "A testcase for a variable length hop_payload. The third payload is 256 bytes long.",
```
t-bast, line 2: Looks like I forgot to update this comment, the third payload isn't 256 bytes long!

t-bast (commented):
A couple of nits, otherwise LGTM, let's kill this thing!

bolt04/onion-test.json
rustyrussell: Yep, fixed to "275 bytes long".

ViktorTigerstrom: Apologies for leaving this comment only after the PR has been merged, but I believe the "comment" is still incorrect as it's the fifth payload that is 275 bytes long, no?
There's also the second payload that's 83 bytes long, so there's now two payloads that have a different size than the rest of the payloads, so perhaps it'd be worth mentioning both if we want to explicitly mention the payloads that have a different size. If you'd like me to, I can of course open a PR editing the "comment" :).

PR #963, opened 2022-02-22
Fix typo and aspell list.

rustyrussell:
Since Travis died, we don't get CI to check these any more :(


t-bast (approved):
Damn, why did Travis die? We should setup github CI instead someday then

vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/963/commits/a97b5e15c2c2b89d8482e4f5f221567c37fa360a

PR #988, opened 2022-05-11
Make generated pubkeys slightly more grep-able

instagibbs:


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/988/commits/e6279a5da52fe17b314fe9b0126c95adb26a2a68

t-bast (approved):
ACK, this was the only place where we use that underscore.

PR #991, opened 2022-05-17
Replace "logical OR" with "bitwise OR"

morehouse:
This makes more sense given the context of combining bitmaps.


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/991/commits/dd9738c8b44702bf98819bfe117750a497aa7bd7

I like this change, because from my dictionary the bitwise is on top of logical OR :)

PR #996, opened 2022-05-31
BOLT #7: Make channel_update always have htlc_maximum_msat.

rustyrussell:
I think everyone sets this.  Vastly simplifies parsing.

Indeed, every single of the 149964 channel updates seen by my node has his set.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -407,7 +407,7 @@ of *relaying* payments, not *sending* payments. When making a payment
     * [`chain_hash`:`chain_hash`]
     * [`short_channel_id`:`short_channel_id`]
     * [`u32`:`timestamp`]
-    * [`byte`:`message_flags`]
+    * [`byte`:`must_be_one`]
     * [`byte`:`channel_flags`]
     * [`u16`:`cltv_expiry_delta`]
     * [`u64`:`htlc_minimum_msat`]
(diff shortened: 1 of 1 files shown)


cdecker:
ACK https://github.com/lightning/bolts/pull/996/commits/466317a300cded545b5f1b0fb6e2191644469a01

Roasbeef (approved):
LGTM 🏂

Would be good to run some basic analysis just to check what % of channels _aren't_ setting it.

vincenzopalazzo (approved):
Concept ack for the "metrics"

07-routing-gossip.md
```diff
@@ -432,12 +432,6 @@ fields in the `channel_update` message:
 | ------------- | ------------------------- | -------------------------------- |
 | 0             | `option_channel_htlc_max` | `htlc_maximum_msat`              |
 
-Note that the `htlc_maximum_msat` field is static in the current
```
t-bast, line 435: Why not also remove lines 428 to 433 entirely as well? We'll never want to use those flags in the future, we'll use tlvs instead, so just say that this whole `message_flags` field must be hard-coded to 1 (we could even go further and rename it to show that it's just an obsolete constant)?

t-bast (commented):
Concept ACK, I think we can remove even more stuff

07-routing-gossip.md
rustyrussell: Done.

t-bast (dismissed):
ACK https://github.com/lightning/bolts/pull/996/commits/f99b5494c8d52ac08d006179ca1c3e46cae86880

t-bast:
As noted in https://github.com/lightningdevkit/rust-lightning/issues/1515 and discussed during the dev summit, it would also make sense to verify that every implementation sets this value to `max_htlc_value_in_flight_msat` instead of using the channel capacity.

t-bast:
Retracting my ACK, I made a proposal at #999 that would benefit from message flags (but I did make `htlc_maximum_msat` mandatory, which was the goal of this PR).

vincenzopalazzo (approved):

arik-so:
I have just downloaded a fresh batch of 134k channel updates from an LND peer, 364 of which don't have an HTLC max set, which is 2.7%.

Roasbeef:
Replaced by https://github.com/lightning/bolts/pull/999

PR #999, opened 2022-06-10
Bolt7: add channel update flag for scid alias

t-bast:
Since #910, nodes are allowed to use aliases instead of real scids. It would be quite helpful to be explicit about it and set a flag in `channel_update` when an alias is used instead of a real scid for a few reasons:

- the receiving peer knows that this `channel_update` is meant to be kept private and can internally store it separately from "real" channel updates without requiring a lookup into an internal mapping table
- the sender knows whether they're using an alias or not, so it's trivial to set that bit

We also make the `htlc_maximum_msat` field mandatory: every node on the network currently sets it, so we can simplify message parsing. This makes this PR an alternative to #996

@cdecker @TheBlueMatt @crypt-iQ is that message flag helpful for your implementation's router logic?

DAMN I could have gotten PR number 1000, I'm always off-by-one (unless there was a PR number 0)

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -413,7 +413,7 @@ of *relaying* payments, not *sending* payments. When making a payment
     * [`u64`:`htlc_minimum_msat`]
     * [`u32`:`fee_base_msat`]
     * [`u32`:`fee_proportional_millionths`]
-    * [`u64`:`htlc_maximum_msat`] (option_channel_htlc_max)
+    * [`u64`:`htlc_maximum_msat`]
 
 The `channel_flags` bitfield is used to indicate the direction of the channel: it
 identifies the node that this update originated from and signals various options
@@ -425,18 +425,12 @@ individual bits:
 | 0             | `direction` | Direction this update refers to. |
 | 1             | `disable`   | Disable the channel.             |
 
-The `message_flags` bitfield is used to indicate the presence of optional
-fields in the `channel_update` message:
+The `message_flags` bitfield is used to provide additional details about the message:
 
-| Bit Position  | Name                      | Field                            |
-| ------------- | ------------------------- | -------------------------------- |
-| 0             | `option_channel_htlc_max` | `htlc_maximum_msat`              |
-
-Note that the `htlc_maximum_msat` field is static in the current
-protocol over the life of the channel: it is *not* designed to be
-indicative of real-time channel capacity in each direction, which
-would be both a massive data leak and uselessly spam the network (it
-takes an average of 30 seconds for gossip to propagate each hop).
+| Bit Position  | Name           |
+| ------------- | ---------------|
+| 0             | `must_be_one`  |
+| 1             | `dont_forward` |
 
 The `node_id` for the signature verification is taken from the corresponding
 `channel_announcement`: `node_id_1` if the least-significant bit of flags is 0
@@ -450,7 +444,8 @@ The origin node:
   channel peer, even though the channel has not yet been announced (i.e. the
   `announce_channel` bit was not set).
     - MUST set the `short_channel_id` to either an `alias` it has
-	  received from the peer, or the real channel `short_channel_id`.
+    received from the peer, or the real channel `short_channel_id`.
+    - MUST set `dont_forward` to 1 in `message_flags`
     - MUST NOT forward such a `channel_update` to other peers, for privacy
     reasons.
     - Note: such a `channel_update`, one not preceded by a
@@ -464,15 +459,8 @@ The origin node:
     - MUST set the `direction` bit of `channel_flags` to 0.
   - otherwise:
     - MUST set the `direction` bit of `channel_flags` to 1.
-  - if the `htlc_maximum_msat` field is present:
-    - MUST set the `option_channel_htlc_max` bit of `message_flags` to 1.
-    - MUST set `htlc_maximum_msat` to the maximum value it will send through this channel for a single HTLC.
-      - MUST set this to less than or equal to the channel capacity.
-      - MUST set this to less than or equal to `max_htlc_value_in_flight_msat`
-    it received from the peer.
-  - otherwise:
-    - MUST set the `option_channel_htlc_max` bit of `message_flags` to 0.
-  - MUST set bits in `channel_flags` and `message_flags `that are not assigned a meaning to 0.
+  - MUST set `must_be_one` in `message_flags` to 1.
+  - MUST set bits in `channel_flags` and `message_flags` that are not assigned a meaning to 0.
   - MAY create and send a `channel_update` with the `disable` bit set to 1, to
   signal a channel's temporary unavailability (e.g. due to a loss of
   connectivity) OR permanent unavailability (e.g. prior to an on-chain
@@ -525,14 +513,11 @@ The receiving node:
     - otherwise:
       - SHOULD queue the message for rebroadcasting.
       - MAY choose NOT to for messages longer than the minimum expected length.
-  - if the `option_channel_htlc_max` bit of `message_flags` is 0:
-    - MUST consider `htlc_maximum_msat` not to be present.
+  - if `htlc_maximum_msat` is greater than channel capacity:
+    - MAY blacklist this `node_id`
+    - SHOULD ignore this channel during route considerations.
   - otherwise:
-    - if `htlc_maximum_msat` is not present or greater than channel capacity:
-      - MAY blacklist this `node_id`
-      - SHOULD ignore this channel during route considerations.
-    - otherwise:
-      - SHOULD consider the `htlc_maximum_msat` when routing.
+    - SHOULD consider the `htlc_maximum_msat` when routing.
 
 ### Rationale
 
@@ -553,12 +538,6 @@ message can rebroadcast it just by changing the `s` component of signature with 
 This should however not result in the blacklist of the `node_id` from where
 the message originated.
 
-The explicit `option_channel_htlc_max` flag to indicate the presence
-of `htlc_maximum_msat` (rather than having `htlc_maximum_msat` implied
-by the message length) allows us to extend the `channel_update`
-with different fields in future.  Since channels are limited to 2^32-1
-millisatoshis in Bitcoin, the `htlc_maximum_msat` has the same restriction.
-
 The recommendation against redundant `channel_update`s minimizes spamming the network,
 however it is sometimes inevitable.  For example, a channel with a
 peer which is unreachable will eventually cause a `channel_update` to
@@ -571,6 +550,10 @@ When a node creates a new `channel_update` to change its channel parameters,
 it will take some time to propagate through the network and payers may use
 older parameters. It is recommended to keep accepting older parameters for
 at least 10 minutes to improve payment latency and reliability.
+
+The `must_be_one` field in `message_flags` was previously used to indicate
+the presence of the `htlc_maximum_msat` field. This field must now always
+be present, so `must_be_one` is a constant value, and ignored by receivers.
 
 ## Query Messages
 


Crypt-iQ:
I don't think this helps the lnd implementation any since we know which SCID aliases we are using for a given channel

t-bast:
> What does this mean? You should never "wait" for a channel_announcement - if you receive a channel_update first you should drop it on the floor and ignore it (unless its your own channel/direct peer).

Sorry, this was poorly phrased. We validate `channel_announcement`s asynchronously (since this requires calls to `bitcoind`), so we can receive a valid `channel_update` while the `channel_announcement` is being validated, in which case we don't want to ignore that `channel_update`, we want to stash it to validate it once the `channel_announcement` has been validated. Does that make more sense?

Mostly what we'd like is to be able to properly type `channel_update` at decoding time. Right now when we receive a `channel_update` from the wire, we need to look at several internal maps to choose what processing flow we should apply to the `channel_update` and where to store it.

Being able to properly type `channel_update`s would ensure that it's impossible to have a `channel_update` for a private channel using aliases end up being sent to the network because of a bug in internal maps / processing flow. We can do that proper typing later in the flow, once we've identified what kind of updates it is, but we figured that it would be much better to be explicit directly in the message since it's trivial to do and would simplify the flow for some implementations (even though I think every implementation could benefit from it, because it should remove the need for a lookup in an internal map).

It's trivial for senders to set this bit, it remove some ambiguity in how the message is interpreted, so it feels like a useful change?

t-bast:
> I don't think this helps the lnd implementation any since we know which SCID aliases we are using for a given channel

But as a receiver, when you receive a `channel_update` on a connection, you'd have to do a lookup on a mapping table to know that this update is for a local private channel, right?

07-routing-gossip.md
```diff
@@ -450,7 +445,11 @@ The origin node:
   channel peer, even though the channel has not yet been announced (i.e. the
   `announce_channel` bit was not set).
     - MUST set the `short_channel_id` to either an `alias` it has
-	  received from the peer, or the real channel `short_channel_id`.
+    received from the peer, or the real channel `short_channel_id`.
+    - if it sets `short_channel_id` to an `alias`:
+      - MUST set `scid_alias` to 1 in `message_flags`
+    - otherwise:
+      - MUST set `scid_alias` to 0 in `message_flags`
```
Crypt-iQ, line 452: What about older nodes?

Crypt-iQ:
> > I don't think this helps the lnd implementation any since we know which SCID aliases we are using for a given channel
> 
> But as a receiver, when you receive a `channel_update` on a connection, you'd have to do a lookup on a mapping table to know that this update is for a local private channel, right?

Yes, we have to lookup that we know of the alias, but I think this patch doesn't get rid of that except if the update designates that the SCID is not an alias -- If I receive an update saying this is for an alias, I still need to check my maps to check that the alias exists?

07-routing-gossip.md
t-bast: They already set it to `0`, otherwise they wouldn't be spec-compliant, because of this requirement (that has been here forever): `MUST set bits in channel_flags and message_flags that are not assigned a meaning to 0.`

Crypt-iQ: Ah ok, missed that they were bits

Crypt-iQ:
Lnd uses a range of [16_000_000, 16_250_000] and if we receive a value in that range, we know it's an alias so it serves the same purpose as this SCID bit

t-bast:
> Lnd uses a range of [16_000_000, 16_250_000] and if we receive a value in that range, we know it's an alias so it serves the same purpose as this SCID bit

What?!? But you can't expect other implementations to use exactly that range (and we expect our node to be able to handle more than 250k channels with wallets, so we'd need a bigger range than that)! A specified bit that everybody agrees on is widely better than a heuristic that will work only for lnd...

Crypt-iQ:
> > Lnd uses a range of [16_000_000, 16_250_000] and if we receive a value in that range, we know it's an alias so it serves the same purpose as this SCID bit
> 
> What?!? But you can't expect other implementations to use exactly that range (and we expect our node to be able to handle more than 250k channels with wallets, so we'd need a bigger range than that)! A specified bit that everybody agrees on is widely better than a heuristic that will work only for lnd...

It's like 2^58 possible SCIDs, that's just the block range. Since the counter-party only sends _my_ aliases during gossip, this is fine - lnd stores counter-party aliases in a separate table and JIT inserts the alias in the gossip message if necessary? Rust-lightning was able to handle our SCID range just fine

t-bast:
> It's like 2^58 possible SCIDs, that's just the block range.

Ok, I didn't understand that this was the range of the `block_height` part of the scid (for aliases we're not treating them as block_height/tx_index/output_index triplets at all in `eclair`, just as a 64-bit value).

You're also completely right that we only care about the range we use for our own aliases, because the only `channel_update`s you'll receive that use an alias use your local alias, that's a good point. But you may have a backwards-compat' issue if someday you change the range you're using (and you update your node after it has generated a local alias that doesn't fit in the new range), that's why I think a bit in the message would be cleaner.

t-bast:
> Right, I guess my reaction here is more that its not strongly-typed as a result, you've taken one of many different cases (public-our-channel, public-not-our-channel, nonpublic-our-channel, nonpublic-alias-our-channel) and flagged it, but you haven't flagged the other cases, which have similar handling issues/concerns. If we want to strongly-type this we should flag more than just this type.

It's true, it isn't perfect, but it's already better than nothing, because it lets you immediately identify `channel_update`s for which you'll never need to validate a `channel_announcement` (the channel is either private or public with 0-conf) and which you shouldn't hand to your staggered rebroadcast component. For such a trivial change on the sender side, I personally think it's worth it, but if others disagree we'll work without it!

Crypt-iQ:
> > Lnd uses a range of [16_000_000, 16_250_000] and if we receive a value in that range, we know it's an alias so it serves the same purpose as this SCID bit
> 
> Bleh, why not use realistic values? Especially post-taproot-in-use its a nice privacy win to do so.

Less chance of me messing something up in a tricky part of the codebase

> for which you'll never need to validate a channel_announcement (the channel is either private or public with 0-conf

is it not necessary to validate the announcement for public 0-conf after 6 confs?

t-bast:
Thinking about this more, I'm not sure we need more flags. Correct me if I'm missing something, but I believe the only thing we want to know is: "is this channel update supposed to stay private between me and my peer?". There are two cases where that is true:

- unannounced channels with `scid_alias`
- unannounced channels without `scid_alias`

The `uses_alias` field covers the first case. The second case would need a new flag, but if we want to be able to actually rely on it, we'd need all implementations to support this. Given that all implementations are currently working on shipping `scid_alias`, does it make sense to require another change? Since we'll be waiting for our direct peers to update, we simply need them to update to support `scid_alias` and at that point require that private channels use `scid_alias`, which makes the second case obsolete?

t-bast:
The issue is that an additional `isPrivate` flag cannot be backwards-compatible with existing nodes (who will set it to `0` even for private channels), while a `uses_alias` flag can be fully backwards-compatible since existing nodes don't use aliases on mainnet yet and currently set that bit to `0`.

> I'm maybe a bit dubious that we should add a flag for something that people rely on meaning X, when the protocol allows it to mean Y, even though it usually means X. That seems like a great way to mis-implement it.

Are you talking about the `uses_alias` flag here? The meaning of that one should be completely straight-forward, it just adds more information about the `short_channel_id` included in that `channel_update`. Or am I missing a different way people may understand this?

I think I'd rather add only this flag to the spec, wait for 6 months until enough nodes on the network advertise support for `option_scid_alias` then make it mandatory on `eclair`. At that point we'll know for sure that:

- a `channel_update` that sets `uses_alias` must never be broadcast to the network and is only useful for invoice generation
- a `channel_update` that does *not* set `uses_alias` is for a public channel and must be broadcast to the network

WDYT?

t-bast:
> I mean this isn't entirely true, scid aliases have been deployed in LDK for some time now, but, sure, its not super broad and those users upgrade quickly currently.

I didn't know it had been deployed so soon after the spec PR was merged, it indeed means we'll need those early adopters to upgrade quickly.

> Still, I think the broader point is that this flag is really a hint to the recipient to skip the processing of it as if it were about a public channel, which doesn't really need backwards compatibility?
> Sure, but you can do the same whether the bit(s) are defined as uses-alias as you can do if the bits are defined more generally.

My point is that with this exact flag + making `scid_alias` mandatory on your node, this hint becomes more than just a hint: if it's not set, you're 100% sure this is a public `channel_update` meant for rebroadcast. I believe other uses of `message_flags` would only be hints, which means implementations will still need to handle the case where the condition is true but the hint isn't set, which means you can't simplify your codebase to rely entirely on those flags.

t-bast:
> Seems like the meeting conclusion here was to either use a flag to mean "not public channel update" or have two flags as I'd proposed. Is that what you took away form the conversation?

That's how I understand it as well, I forgot to update the PR, thanks for the nudge.
I changed it in https://github.com/lightning/bolts/pull/999/commits/9f1a704790e055098ae5dbebfaf964c66d02c667, let me know what you think.

We can bikeshed the name of the flag, I hesitated between `dont_forward`, `private` and `no_broadcast`, which one do you think works best?

07-routing-gossip.md
```diff
@@ -565,6 +544,11 @@ the channel when the peer reestablishes contact.  Because gossip
 messages are batched and replace previous ones, the result may be a
 single seemingly-redundant update.
 
+The `must_be_one` field field in `message_flags` was previously used to
```
niftynei, line 547:
```suggestion
The `must_be_one` field in `message_flags` was previously used to
```

niftynei: drive by change.

t-bast: Good catch, thanks! Rebased and fixed in https://github.com/lightning/bolts/pull/999/commits/0d10541b73d97ed5f56b02564830048506e060ab

vincenzopalazzo (approved):

rustyrussell:
Ack, this seems good.  Closes 996...

Roasbeef (approved):
LGTM 🍬

t-bast:
This is implemented in https://github.com/ACINQ/eclair/pull/2362, let me know if you want me to do cross-compat tests between eclair and another implementation!

harding:
After the merge of this PR, the spec no longer describes what `htlc_maximum_msat` is for.  E.g. none of these is informative,

```
$ git grep htlc_maximum_msat
07-routing-gossip.md:    * [`u64`:`htlc_maximum_msat`]
07-routing-gossip.md:  - if `htlc_maximum_msat` is greater than channel capacity:
07-routing-gossip.md:    - SHOULD consider the `htlc_maximum_msat` when routing.
07-routing-gossip.md:the presence of the `htlc_maximum_msat` field. This field must now always
```

In particular, I think it would be helpful to restore the following two lines to somewhere in the spec:

```
-    - MUST set `htlc_maximum_msat` to the maximum value it will send through this channel for a single HTLC.
-      - MUST set this to less than or equal to the channel capacity.
```

It looks to me like the PR that this PR superseded did that: https://github.com/lightning/bolts/pull/996

FWIW, I noticed this when trying to confirm my understanding of `htlc_maximum_msat` after reading the recent mailing list discussion about using it as a flow control device: https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-September/003686.html

Alternatively, maybe the term should be added to the glossary in BOLT0

t-bast:
Thanks @harding, you're right this is my mistake! I'll correct that with a follow-up PR.

t-bast:
@harding done in #1029, thanks for the report!

PR #1001, opened 2022-06-15
Add grace period for older channel parameters

t-bast:
When a node creates a new `channel_update` to change its channel parameters, it will take some time to propagate through the network and payers may use older parameters. It is recommended to keep accepting older parameters for a while to improve payment latency and reliability. Eclair and Core Lightning already implement this.


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1001/commits/ce99310fe84d339a49b1148b4f8eb8f2afed0fc8

t-bast:
Notes from the meeting: the spec should mention what a reasonable number is.

07-routing-gossip.md
```diff
@@ -565,6 +567,11 @@ the channel when the peer reestablishes contact.  Because gossip
 messages are batched and replace previous ones, the result may be a
 single seemingly-redundant update.
 
+When a node creates a new `channel_update` to change its channel parameters,
+it will take some time to propagate through the network and payers may use
+older parameters. It is recommended to keep accepting older parameters for a
+while to improve payment latency and reliability.
```
renepickhardt, line 573: Maybe we should define a number here? It seems 10 minutes are reasonable?

t-bast: Done.

vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1001/commits/2d008dd09f9aef03cd179c901df15458a5adfe64

Roasbeef:
Makes sense, I think some nodes also already do this. FWIW, lnd doesn't do this _yet_, but iiuc the main motivation here is better compatibility with route blinding (routes can be used for a longer period of time).

t-bast:
While it also helps route blinding, it's really useful today as well to make up for delays in receiving `channel_update`s or a sender missing our latest `channel_update`. This is why cln and eclair have already implemented this a while ago regardless of route blinding.

cdecker:
As rationale for the 10 minutes: the current network diameter is about 7, if every node has a staggered broadcast interval of 60 seconds, and we assume worst-case (i.e., they all synchronize their clocks, thus from receiving to forwarding we get exactly 60 seconds + 1/2 RTT of the link that it is forwarded over), we get a worst-case delay from emitting the `channel_update` to everybody receiving it of 7 minutes + 3.5 RTT. For some RTT < 0.85 minutes we are therefore sure the upadte has propagated throughout the entire network.

Offline nodes may of course come back up and not see the `channel_update`, however they have to handle outdated routing information anyway, so no additional failure surface is introduced.

ProofOfKeags:
> the current network diameter is about 7

How do we know what the gossip networks diameter is? Isn't the p2p linkage much more dynamic than the channel graph? Or is that network diameter the measure of the channel graph?

cdecker:
> > the current network diameter is about 7
> 
> How do we know what the gossip networks diameter is? Isn't the p2p linkage much more dynamic than the channel graph? Or is that network diameter the measure of the channel graph?

Good catch, I implicitly assumed the gossip network to be representative, which isn't quite true, but can serve as an upper bound. Any real world end-to-end latency would be lower than this estimate

PR #1004, opened 2022-06-23
gossip: delay considering a channel edge deleted for 12-blocks

niftynei:
Allows us time to propagate a splice update (new channel_announcement)

Suggested-By: @t-bast
Co-Authored: @ddustin

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -224,7 +224,7 @@ The receiving node:
       - otherwise:
         - SHOULD store this `channel_announcement`.
   - once its funding output has been spent OR reorganized out:
-    - SHOULD forget a channel.
+    - SHOULD forget a channel after a 12-block delay.
 
 ### Rationale
 
@@ -248,6 +248,10 @@ New channel features are possible in the future: backwards compatible (or
 optional) features will have _odd_ feature bits, while incompatible features
 will have _even_ feature bits
 (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide)).
+
+A delay of 12-blocks is used when forgetting a channel on funding output spend
+as to permit a new `channel_announcement` to propagate which indicates this
+channel was spliced.
 
 ## The `node_announcement` Message
 


07-routing-gossip.md
```diff
@@ -249,6 +249,10 @@ optional) features will have _odd_ feature bits, while incompatible features
 will have _even_ feature bits
 (["It's OK to be odd!"](00-introduction.md#glossary-and-terminology-guide)).
 
+A delay of 12-blocks is used when forgetting a channel on funding output spend
```
tnull, line 252: I have a hard time parsing "channel on funding output spend". What is exactly meant by that?

Also, nit: s/12-blocks/12 blocks/

t-bast:
I'd like to spend more time exploring the design space for splicing before acking this, sorry if it takes some time before I can comment!

Roasbeef:
> What's the worry with Just Doing This? It seems good to get moving on this a ways before it's needed.

IMO in practice the signal will be pretty unreliable due to strictly relying on timely gossip message propagation. There also isn't a good way to track the actual observance of this, since it's more or less just a guideline. 

I wrote more about this here, along with a suggestion for a primary chain signal, with the gossip signal (w/e that looks like) being secondary: https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-June/003616.html

t-bast:
That's what I had in mind, my thinking was that the splicing tx would be somehow identifiable as a splicing tx so that we don't need this 12-blocks delay at all. But the issue is that we'd like implementations to deploy code to identify this signal _before_ we actually deploy splicing...

We can also choose to do both: delay considering a channel edge deleted for 12 blocks, which implementations can do today and will help splicing when it happens, even though it's not a perfect solution, and then implement a signal in the splicing tx to improve reliability.

niftynei:
Here's a thought (also posted to ML): if you're in possession of a gossip message which indicates that it's a splice *and* you've seen a channel btw the same two peers close, you SHOULD prioritize spreading that gossip message to every peer you're connected to/have a channel with?

This is nice because it uses the on-chain signaling that @Roasbeef proposed, but doesn't require new protocol design, just changing the priority/scope of which you gossip a subset of messages.

This is nicely spam compatible too imo. Splice gossip has a natural rate-limit/real world cost in that triggering the close onchain costs sats; you can't generate a lot of urgent gossip messages w/o sending money to the L1 miners.

07-routing-gossip.md
niftynei: See L226: https://github.com/lightning/bolts/pull/1004/files#diff-35876d484bc2647647dd95e16e18f0aa6c2b098b5f70834ec7fea33cdc522835R226

cdecker:
Agreed with @TheBlueMatt: the gossip is intended as a best effort system, and striving for completeness or timeliness is likely a far deeper Rabbit Hole than we're willing to tackle (now or potentially ever).

As for the actual number, 12 blocks seems like plenty: 6 blocks to make the New version announceable, plus 6 blocks for propagation (see #1001 for a back of the envelope calculation).

Do we also want onchain signaling? I'm not sure we want to make transactions identifiable beyond what we already have, but it might not be a huge loss here (we're gossiping about these outpoints anyway) and a belts and suspenders variant might be sensible.

This is all going to change when we detach gossip from onchain outpoints anyway, so maybe better not to overthink it here?

t-bast (approved):
ACK

t-bast:
Shall we merge this now? The previous spec meeting had acks from eclair, cln and ldk and @Roasbeef wasn't opposed to it?

vincenzopalazzo (approved):

Roasbeef (approved):
LGTM 🍫

PR #1006, opened 2022-06-28
bolt7 typo: "newest" channel_update, not "oldest".

fiatjaf:


07-routing-gossip.md
```diff
@@ -987,7 +987,7 @@ A node:
 #### Requirements
 
 A node:
-  - if a channel's oldest `channel_update`s `timestamp` is older than two weeks
+  - if a channel's newest `channel_update`s `timestamp` is older than two weeks
```
t-bast, line 990: nit: we don't use "newest" anywhere in the repo, but it seems like we use "latest" for that:
```suggestion
  - if a channel's latest `channel_update`s `timestamp` is older than two weeks
```

t-bast (commented):
Good catch, thanks!

rustyrussell:
Um, no, this was wrong (or, at least, a change):

From our code:

```
		/* BOLT #7:
		 *   - if a channel's latest `channel_update`s `timestamp` is
		 *     older than two weeks (1209600 seconds):
		 *    - MAY prune the channel.
		 */
		/* This is a fancy way of saying "both ends must refresh!" */
		if (!is_halfchan_defined(&chan->half[0])
		    || chan->half[0].bcast.timestamp < highwater
		    || !is_halfchan_defined(&chan->half[1])
		    || chan->half[1].bcast.timestamp < highwater) {
			status_debug(
			    "Pruning channel %s from network view (ages %"PRIu64" and %"PRIu64"s)",
			    type_to_string(tmpctx, struct short_channel_id,
```

i.e. one side can't keep a channel alive unilaterally, *both* sides must send every 2 weeks.

t-bast:
I see what you mean, and our code does this as well, but I thought the confusion fixed here was a different one, let me try to explain. Let's say we have a channel Alice <-> Bob.

Alice creates a first channel update, `channel_update_ab_1` then another one, `channel_update_ab_2`. The "oldest" channel update could be understood as being `channel_update_ab_1`, whereas we only want to consider the latest `channel_update_ab_2`. That's the confusion I thought we were fixing here.

I think we should clarify it further by saying that we consider the latest channel update in each direction, and if the oldest of those two is older than two weeks, ...

I'll open a PR.

PR #863, opened 2021-04-20
Splice draft

rustyrussell:
Based on https://github.com/lightningnetwork/lightning-rfc/pull/862

We use the interactive tx construction protocol to make a splice:
1. Initiator pays for input and output, sets fees.
2. You can do one per minute, but you have to increase feerate by >= 25% each time
3. We freeze the channel during construction (but not while waiting for confs).
4. A simple channel_update tells people not to notice the channel close, since we're splicing.

PR #1009, opened 2022-06-30
Add funding outpoint sigs to tx_signatures message

ddustin:
Dependent on PR https://github.com/lightning/bolts/pull/851


t-bast:
I'm not sure it makes sense to have this PR outside of the splicing one...at the spec level it's usually better to have one big long-lived PR that contains everything a feature needs, instead of small atomic PRs like you'd do for code.

ddustin:
Ah good point. The change is needed for both dual funding and splicing cause they share the interactive tx protocol.

Since dual funding is still in draft we are in a weird PR dependency situation =/.

Talking it over with @niftynei IIRC putting it here seemed like the ideal compromise but there may be a better way to refactor things.

t-bast:
> The change is needed for both dual funding and splicing cause they share the interactive tx protocol.

Can you explain why dual funding would need that tlv? I haven't felt the need for it while implementing dual funding in eclair.

ddustin:
> > The change is needed for both dual funding and splicing cause they share the interactive tx protocol.
> 
> Can you explain why dual funding would need that tlv? I haven't felt the need for it while implementing dual funding in eclair.

Ya totally -- dual funding doesn't need the tlv. There could be a cleaner way to do this 🤷.

The problem is that dual funding introduces `tx_signatures` and we need to modify `tx_signatures` for splicing.

The ways to do it I see are:
A) Put the change into the dual funding PR -> https://github.com/lightning/bolts/pull/851
B) Rebase the splice PR off the dual funding PR and add it there -> https://github.com/lightning/bolts/pull/863
C) Make a separate PR that's rebased off dual funding PR and add it there -> https://github.com/lightning/bolts/pull/1009

None seem ideal to me but C seemed like the simplest. B is probably the most correct but in my experience large interdependent PRs lead to chaotic problems. So far the splicing PR has been able to avoid being merge dependent on dual funding.

Once the dual funding PR gets merged in the dependency problem goes away on its own.

t-bast:
Got it, this is just a dependency issue for the spec PRs, not a real feature dependency.

I've got a better proposal to resolve this: we verify dual funding interop between eclair and cln, merge the dual funding PR :sunglasses:, and then those tlvs are on `master` and it's much easier to work on new proposals that leverage the `interactive-tx` protocol.

We should be ready for interop testing soon, so we'll have acks from eclair and cln. It would be really great it someone from ldk and lnd could grant a concept ACK on dual funding, if that happens I think we should be able to merge it.

ddustin:
> Got it, this is just a dependency issue for the spec PRs.
> 
> I've got a better proposal to resolve this: we verify dual funding interop between eclair and cln, merge the dual funding PR 😎, and then those tlvs are on `master` and it's much easier to work on new proposals that leverage the `interactive-tx` protocol.
> 
> We should be ready for interop testing soon, so we'll have acks from eclair and cln. It would be really great it someone from ldk and lnd could grant a concept ACK on dual funding, if that happens I think we should be able to merge it.

Yessssss love this plan!

optout21:
This is based on #851, which has just been merged, so this could be rebased and reviewed. Alternatively, I suggest to included it in splicing PR #863, as this change is needed by splicing.

t-bast:
I think this should be directly included in the splice PR, so that everything splice-related is in the same place. The quiescence PR shouldn't take too long to land, and at that point the splice PR won't have any dependency anymore.

PR #1014, opened 2022-08-01
Clarify Bolt 7 latest channel update pruning

t-bast:
This is a follow-up on #1006, which actually created more confusion.
This is somewhat hard to explain both clearly and concisely, let me know if I should make it less concise and more clear.
See https://github.com/lightning/bolts/pull/1006#issuecomment-1200785218 for more context.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -994,8 +994,8 @@ A node:
 #### Requirements
 
 A node:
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM 🐿

vincenzopalazzo (approved):

Issue #1013, opened 2022-07-26
bolt2: fee_range clarification


Crypt-iQ:
I am a little confused about the fee_range requirements for coop close.

I asked some clarifying questions of @t-bast in https://github.com/lightning/bolts/pull/847#discussion_r893605295:

"Yes, fee_range allows the funder to update its fee_range if the fundee **doesn't respond** or sent a warning."

I didn't realize until now that the funder may update its fee_range even if the fundee did not send a warning. I think this could lead to the following issue:

A's PoV
```
A                      B

fr=[100, 200], fs=150
----closing_signed---->

fr=[200, 300], fs=275
----closing_signed----> (A doesn't hear back soon enough and updates the range)

fr=[150, 250], fs=175
<---closing_signed----- (spec says A should fail here since 175 is not in the overlap)
```

B's PoV
```
A                      B

fr=[100, 200], fs=150
----closing_signed---->

fr=[150, 250], fs=175
<---closing_signed----- (B sends acceptable message back)

fr=[200, 300], fs=275
----closing_signed----> (spec says B should fail since 175 wasn't received)
```

* A fails since there is an overlap between A's range [200, 300] and B's range [150, 250] of [200, 250], but 175 isn't in the overlap:

    - if the message contains a fee_range:
        - if there is no overlap between that and its own fee_range:
            - SHOULD send a warning
            - MUST fail the channel if it doesn't receive a satisfying fee_range after a reasonable amount of time
        - otherwise:
            - **if it is the funder:**
                - **if fee_satoshis is not in the overlap between the sent and received fee_range:**
                    - **MUST fail the channel**

* B fails since there is an overlap between A's range [200, 300] and B's range [150, 250] of [200, 250], but 175 should have been received from A since B has already sent closing_signed.

    - if the message contains a fee_range:
        - if there is no overlap between that and its own fee_range:
            - SHOULD send a warning
            - MUST fail the channel if it doesn't receive a satisfying fee_range after a reasonable amount of time
        - otherwise:
            - **otherwise (it is not the funder):**
                - **if it has already sent a closing_signed:**
                    - **if fee_satoshis is not the same as the value it sent:**
                        - **MUST fail the channel**

t-bast:
[...]

PR #1016, opened 2022-08-10
option_closing_rejected: turn-based fee_range coop close

Crypt-iQ:
Adds a feature bit. It's optional, but does simplify some of the spec logic and I think it's nice to have. The naming can change. It works with simple-taproot-channels to allow sending musig2 nonces in a turn-based way.

Fixes https://github.com/lightning/bolts/issues/1013

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -679,6 +679,10 @@ non-funder has to pick a fee in this range. If the non-funder chooses the same
 value, negotiation is complete after two messages, otherwise the funder will
 reply with the same value (completing after three messages).
 
+If `option_closing_rejected` is negotiated, then the fee_range flow is similar to
+the case where it isn't negotiated, except that `closing_rejected` messages are
+sent by the fundee in case of mismatch.
+
 1. type: 39 (`closing_signed`)
 2. data:
    * [`channel_id`:`channel_id`]
@@ -706,7 +710,12 @@ The sending node:
   prepared to pay for a close transaction.
   - if it doesn't receive a `closing_signed` response after a reasonable amount of time:
     - MUST fail the channel
-  - if it is not the funder:
+  - if it is the funder:
+    - if `option_closing_rejected` has been negotiated:
+      - if a `closing_signed` has already been sent:
+        - MUST wait until receipt of either a corresponding `closing_signed` or
+          `closing_rejected` before sending this `closing_signed`.
+  - otherwise (it is not the funder):
     - SHOULD set `max_fee_satoshis` to at least the `max_fee_satoshis` received
     - SHOULD set `min_fee_satoshis` to a fairly low value
   - MUST set `signature` to the Bitcoin signature of the close transaction,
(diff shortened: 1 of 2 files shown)


Roasbeef:
Commentary from the latest spec meeting:
  * This splits off a bit into paths of adding more information to the warning message (error code, context information, etc), and also a possible greater re-org for co-op close in general. 
  * Instead of the new message, we can add more info to warnings, then also add a new bit to help ppl recognize/enforce that 
      * In addition to an error code, we might also want to add: suggested value, the field itself, maybe the first 16 bytes of message (or hash or something like that). 
  * Also a chance to make the "negotiation" part of co-op closing more explicit

t-bast:
Concept ACK. I would add a `feerate_range` tlv in the `closing_rejected` message, it provides more information to the initiator even though it may be obsolete by the time the initiator acts on it (and of course, this will also later contain a tlv with musig2 nonces).

I think it would be confusing to use a `warning` or an `error` for that. This is an explicit step of the protocol that will require protocol-related handling (e.g. extract musig2 nonces), so it should use a message tailored for that protocol. It's easy to understand/debug a state machine by only looking at what messages are exchanged, and using a `warning` for this would force introspection of the message (adds one layer). For example we could have very confusing message flows such as:

- A sends `closing_signed` to B
- B sends `warning`s to A for a completely unrelated issues
- B send `closing_signed` to accept A's `closing_signed`

If we had a shortage of available message types I would understand the need to create that additional layer of indirection, but since we don't have such a shortage adding such indirection is creating unnecessary complexity?

Crypt-iQ:
@t-bast

> if there is no overlap between that and its own `fee_range`:
>      - SHOULD send a warning

Should the above be changed so that if you are sending `closing_rejected`, you don't send a `warning`?

Additionally, how do we handle the requirement of "MUST fail the channel"? It seems like that'd require an Error message?

t-bast:
> Should the above be changed so that if you are sending closing_rejected, you don't send a warning?

Yes, I believe we should replace the use of warning with this new `closing_rejected` message now that we have it (it lets us provide more structured data to the sender).

Of course, if others disagree with my previous comment and think we should extend warning/error instead of introducing a new message, this point would be moot.

02-peer-protocol.md
```diff
@@ -776,6 +787,31 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
 the closing transaction will likely never reach miners.
 
+### Closing Rejected: `closing_rejected`
+
+If the fundee disagrees on the `fee_range` the funder proposed, they may send a
+`closing_rejected` to tell them to try again.
+
+1. type: 40 (`closing_rejected`)
+2. data:
+   * [`channel_id`:`channel_id`]
```
Roasbeef, line 797: Comment from spec meeting: should this also include a fee range in this message?

Crypt-iQ: I thought we were going with the whole warnings approach for telling what parameters you didn't like about _any_ message?

PR #1020, opened 2022-08-30
bolt3: Clarify order of operations for HTLC amount calculation

dunxen:
The current wording is ambiguous, although the order can be inferred if one considers the result of the fee calculations as an operand to the subtraction.

Anyway, probably better to be explicit.


t-bast (approved):
Thanks!

PR #1026, opened 2022-09-21
zzz-extension-bolt: dynamic commitments

Crypt-iQ:
This commit introduces three messages `dyn_begin_propose`, `dyn_propose`, and `dyn_propose_reply` which allow both channel participants to change their static channel parameters. Initially, only changing the dust limit and the CSV delay are supported.

I made a separate document since it's easier to see the changeset, but I could also cram it into bolt 2.

NOTE: the design here is separate from what was laid out in the mailing list post (https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-March/003531.html) since the commitment format isn't being upgraded here. Only static channel parameters are being updated, so perhaps a different name could be used.

### DIFF
--- a/zzz-dynamic-commit.md
+++ b/zzz-dynamic-commit.md
@@ -1,0 +1,217 @@
+# Extension Bolt ZZZ: Dynamic Commitments
+
+Authors:
+  * Olaoluwa Osuntokun <roasbeef@lightning.engineering>
+  * Eugene Siegel <eugene@lightning.engineering>
+
+Created: 2022-06-24
+
+# Table of Contents
+
+- [Introduction](#introduction)
+
+- [Specification](#specification)
+  * [Proposal Messages](#proposal-messages)
+    + [`dyn_begin_propose`](#-dyn_begin_propose-)
+    + [`dyn_propose`](#-dyn_propose-)
+    + [`dyn_propose_reply`](#-dyn_propose_reply-)
+  * [Reestablish](#reestablish)
+    + [`channel_reestablish`](#-channel_reestablish-)
+
+# Introduction
+
+Dynamic commitments are a way to propose changing some static channel parameters that
+are initially set in the `open_channel` and `accept_channel` messages. This document
+outlines how this can be done on-the-fly with new messages.
+
+# Specification
+
+## Proposal Messages
+
+Three new messages are introduced that let each side propose what they want to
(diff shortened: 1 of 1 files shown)


zzz-dynamic-commit.md
```diff
@@ -0,0 +1,217 @@
...
+outlines how this can be done on-the-fly with new messages.
+
+# Specification
+
+## Proposal Messages
+
+Three new messages are introduced that let each side propose what they want to
+change about the channel.
+
+### `dyn_begin_propose`
```
carlaKC, line 34: Would it make sense to re-use #869 here? At a glance it can probably be quite trivially expanded to include some of the functionality being used here: 
1. Add a `reject` field to the message to `stfu` message
2. (optionally) Add a `message_type` tlv that indicates why you're pausing the channel (eg, in this example `message_type=115/dyn_propose`)

Crypt-iQ: I made it this way because 869 allows HTLC's on the commit tx, and I didn't want to have to think about a new dust limit trimming HTLCs -- but a general flow like 869 could be useful here and probably other future flows

Crypt-iQ:
> Can you share the rationale for wanting to change the dust limit? 

I thought it made sense to change, but now I don't know. Maybe the node operator wants to have a smaller commit tx or something.

> While changing the dust limit based on fee rates sounds nice, it would re-add all of the complexity of dust exposure management that we were able to get rid of with anchor-0-htlc-fee.

I think this exists independently of anchor-0fee since both have the dust exposure issue? In our code, we have a method that calculates the exposure on the channel and in the context of this flow, if the new dust limit makes that method error, we'd reject the proposal

zzz-dynamic-commit.md
carlaKC: > but we need to figure out the warning message bit where it contains a rejected tlv which indicates the fields and why they're rejected (high/low).

I can take a stab at this, looked into it a while back. I think @rustyrussell's idea of bad field / bad value / suggested value will fit pretty well for this use case. Simple enough to repeat that field if we want to point out all the fields we don't like (although I agree that in practise the first one we encounter and don't like should be good enough)

Crypt-iQ: cool, feel free to request a review

PR #1027, opened 2022-09-21
Splicing into a musig2 funding output

Crypt-iQ:
Based on https://github.com/lightning/bolts/pull/863 and very much a work in progress. Only the last commit is relevant. Some of the splice messages are modified to allow upgrading to a simple-taproot-channel during the splice. It is convenient to combine the flows since both splicing and on-the-fly commitment format upgrades need an intermediate transaction (the splice transaction). It is possible, but not required to add or remove funds when upgrading to musig2 funding outputs.

I am not entirely sure how a disconnect is handled at various stages during a splice, so I'm not sure if anything else needs to be handled during reestablish or if any of my assumptions are wrong.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -506,6 +506,10 @@ a new one.  For simplicity, splicing takes place once a channel is
 Operation returns to normal once negotiation is done (while waiting
 for the splice transaction(s) to confirm).
 
+The splice is finally terminated when both sides send
+`splice_complete` to indicate that one of the splice transactions
+reached acceptable depth.
+
 ### The `splice` Message
 
 1. type: 74 (`splice`)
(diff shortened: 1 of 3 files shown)


02-peer-protocol.md
```diff
@@ -518,12 +518,51 @@ reached acceptable depth.
...
+    2. data:
+        * [`66*byte`: `nonces`]
+    1. type: 4 (`tx_verification_musig2_pubnonce`)
+    2. data:
+        * [`66*byte`: `nonces`]
+    1. type: 6 (`tx_signing_musig2_pubnonce`)
+    2. data:
+        * [`66*byte`: `nonces`]
+
+1. type: 78 (`splice_rejected`)
```
ddustin, line 563: We've been leaning on tx_abort for canceling in other places. What are the pros of adding a new message type for this?

PR #1032, opened 2022-10-11
Allow nodes to overshoot final htlc amount and expiry

t-bast:
When nodes receive HTLCs, they verify that the contents of those HTLCs match the intructions that the sender provided in the onion. It is important to ensure that intermediate nodes and final nodes have similar requirements, otherwise a malicious intermediate node could easily probe whether the next node is the final recipient or not.

Unfortunately, the requirements for intermediate nodes were more lenient than the requirements for final nodes. Intermediate nodes allowed overpaying and increasing the CLTV expiry, whereas final nodes required a perfect equality between the HTLC values and the onion values.

This provided a trivial way of probing: when relaying an HTLC, nodes could relay 1 msat more than what the onion instructed (or increase the outgoing expiry by 1). If the next node was an intermediate node, they would accept this HTLC, but if the next node was the recipient, they would reject it.

We update those requirements to fix this probing attack vector. We also rename the `min_final_cltv_expiry` field, which is actually an expiry _delta_.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -259,6 +259,7 @@ The reader:
     - MUST return an error if:
       - incoming `amount_msat` < `amt_to_forward`.
       - incoming `cltv_expiry` < `outgoing_cltv_value`.
+      - incoming `cltv_expiry` < `current_block_height` + `min_final_cltv_expiry_delta`.
 
 Additional requirements are specified [below](#basic-multi-part-payments).
 
@@ -397,7 +398,7 @@ the final node with the following values:
 * `payment_secret`: set to the payment secret specified by the recipient (e.g.
   `payment_secret` from a [BOLT #11](11-payment-encoding.md) payment invoice)
 * `outgoing_cltv_value`: set to the final expiry specified by the recipient (e.g.
-  `min_final_cltv_expiry` from a [BOLT #11](11-payment-encoding.md) payment invoice)
+  `min_final_cltv_expiry_delta` from a [BOLT #11](11-payment-encoding.md) payment invoice)
 * `amt_to_forward`: set to the final amount specified by the recipient (e.g. `amount`
   from a [BOLT #11](11-payment-encoding.md) payment invoice)
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -250,15 +250,15 @@ The reader:
   - if it is not the final node:
     - MUST return an error if:
       - `short_channel_id` is not present,
-       - it cannot forward the HTLC to the peer indicated by the channel `short_channel_id`.
-       - incoming `amount_msat` - `fee` < `amt_to_forward` (where `fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#htlc-fees))
-       - `cltv_expiry` - `cltv_expiry_delta` < `outgoing_cltv_value`
+      - it cannot forward the HTLC to the peer indicated by the channel `short_channel_id`.
+      - incoming `amount_msat` - `fee` < `amt_to_forward` (where `fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#htlc-fees))
+      - `cltv_expiry` - `cltv_expiry_delta` < `outgoing_cltv_value`
   - if it is the final node:
     - MUST treat `total_msat` as if it were equal to `amt_to_forward` if it
       is not present.
     - MUST return an error if:
-       - incoming `amount_msat` != `amt_to_forward`.
-       - incoming `cltv_expiry` != `cltv_expiry_delta`.
+      - incoming `amount_msat` < `amt_to_forward`.
+      - incoming `cltv_expiry` < `outgoing_cltv_value`.
 
 Additional requirements are specified [below](#basic-multi-part-payments).
 
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -877,9 +877,9 @@ received.
 
 The critical settings here are the `cltv_expiry_delta` in
 [BOLT #7](07-routing-gossip.md#the-channel_update-message) and the
-related `min_final_cltv_expiry` in [BOLT #11](11-payment-encoding.md#tagged-fields).
+related `min_final_cltv_expiry_delta` in [BOLT #11](11-payment-encoding.md#tagged-fields).
 `cltv_expiry_delta` is the minimum difference in HTLC CLTV timeouts, in
-the forwarding case (B). `min_final_cltv_expiry` is the minimum difference
+the forwarding case (B). `min_final_cltv_expiry_delta` is the minimum difference
 between HTLC CLTV timeout and the current block height, for the
 terminal case (C).
 
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -1065,7 +1065,7 @@ channel:
 3. C: 30 blocks
 4. D: 40 blocks
 
-C also uses a `min_final_cltv_expiry` of 9 (the default) when requesting
+C also uses a `min_final_cltv_expiry_delta` of 9 (the default) when requesting
 payments.
 
 Also, each node has a set fee scheme that it uses for each of its
@@ -1089,7 +1089,7 @@ The network will see eight `channel_update` messages:
 
 **B->C.** If B were to send 4,999,999 millisatoshi directly to C, it would
 neither charge itself a fee nor add its own `cltv_expiry_delta`, so it would
-use C's requested `min_final_cltv_expiry` of 9. Presumably it would also add a
+use C's requested `min_final_cltv_expiry_delta` of 9. Presumably it would also add a
 _shadow route_ to give an extra CLTV of 42. Additionally, it could add extra
 CLTV deltas at other hops, as these values represent a minimum, but chooses not
 to do so here, for the sake of simplicity:
@@ -1109,7 +1109,7 @@ per [HTLC Fees](#htlc-fees):
         200 + ( 4999999 * 2000 / 1000000 ) = 10199
 
 Similarly, it would need to add B->C's `channel_update` `cltv_expiry_delta` (20), C's
-requested `min_final_cltv_expiry` (9), and the cost for the _shadow route_ (42).
+requested `min_final_cltv_expiry_delta` (9), and the cost for the _shadow route_ (42).
 Thus, A->B's `update_add_htlc` message would be:
 
    * `amount_msat`: 5010198
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -145,7 +145,7 @@ Currently defined tagged fields are:
 * `n` (19): `data_length` 53. 33-byte public key of the payee node
 * `h` (23): `data_length` 52. 256-bit description of purpose of payment (SHA256). This is used to commit to an associated description that is over 639 bytes, but the transport mechanism for the description in that case is transport specific and not defined here.
 * `x` (6): `data_length` variable. `expiry` time in seconds (big-endian). Default is 3600 (1 hour) if not specified.
-* `c` (24): `data_length` variable. `min_final_cltv_expiry` to use for the last HTLC in the route. Default is 18 if not specified.
+* `c` (24): `data_length` variable. `min_final_cltv_expiry_delta` to use for the last HTLC in the route. Default is 18 if not specified.
 * `f` (9): `data_length` variable, depending on version. Fallback on-chain address: for Bitcoin, this starts with a 5-bit `version` and contains a witness program or P2PKH or P2SH address.
 * `r` (3): `data_length` variable. One or more entries containing extra routing information for a private route; there may be more than one `r` field
    * `pubkey` (264 bits)
@@ -175,7 +175,7 @@ A writer:
   - MAY include one `x` field.
     - if `x` is included:
       - SHOULD use the minimum `data_length` possible.
-  - MUST include one `c` field (`min_final_cltv_expiry`).
+  - MUST include one `c` field (`min_final_cltv_expiry_delta`).
     - MUST set `c` to the minimum `cltv_expiry` it will accept for the last
     HTLC in the route.
     - SHOULD use the minimum `data_length` possible.
@@ -216,7 +216,7 @@ A reader:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
   - if there is a valid `s` field:
     - MUST use that as [`payment_secret`](04-onion-routing.md#tlv_payload-payload-format)
-  - if the `c` field (`min_final_cltv_expiry`) is not provided:
+  - if the `c` field (`min_final_cltv_expiry_delta`) is not provided:
     - MUST use an expiry delta of at least 18 when making the payment
   - if an `m` field is provided:
     - MUST use that as [`payment_metadata`](04-onion-routing.md#tlv_payload-payload-format)
@@ -618,9 +618,9 @@ Breakdown:
 * `x`: expiry time
   * `qy`: `data_length` (`q` = 0, `y` = 2; 0 * 32 + 4 == 4)
   * `jw5q`: 604800 seconds (`j` = 18, `w` = 14, `5` = 20, `q` = 0; 18 * 32^3 + 14 * 32^2 + 20 * 32 + 0 == 604800)
-* `c`: `min_final_cltv_expiry`
+* `c`: `min_final_cltv_expiry_delta`
   * `qp`: `data_length` (`q` = 0, `p` = 1; 0 * 32 + 1 == 1)
-  * `2`: min_final_cltv_expiry = 10
+  * `2`: min_final_cltv_expiry_delta = 10
 * `r`: tagged field: route information
   * `zj`: `data_length` (`z` = 2, `j` = 18; 2 * 32 + 18 == 82)
   * `q0gxwkzc8w6323m55m4jyxcjwmy7stt9hwkwe2qxmy8zpsgg7jcuwz87fcqqeuqqqyqqqqlgqqqqn3qq9q`:


04-onion-routing.md
```diff
@@ -292,7 +292,7 @@ The writer:
     - otherwise:
       - MUST set `total_msat` to the amount it wishes to pay.
     - MUST ensure that the total `amount_msat` of the HTLC set which arrives at the payee
-      is equal to `total_msat`.
+      is equal to or greater than `total_msat`.
```
joostjager, line 296: I don't think this change is necessary anymore when make the change to allow `htlc_amt > amt_to_fwd` in the later commit?

t-bast: I think we have to keep it, don't we? Because if we overpay one HTLC in the set, that means that the total HTLC set will overpay the `total_amount`?

t-bast: > I don't think so, because you only count amt_to_forward towards total_msat. The overpayment is already absorbed at the htlc level.

I see what you mean, that depends on what you use to compute your "HTLC set amount". Currently the spec does it using `amount_msat`, which is the value from `update_add_htlc`: in that case, we do need to explicitly allow over-paying.

What you suggest is that we should define the "HTLC set amount" to be the sum of the onion's `amt_to_fwd`, in which case we don't necessarily need to allow over-paying.

That would be a bit awkward for the scenario highlighted in #1031 though. Let's imagine that you retry a partial payment of `500 msat` and are forced to pay `510 msat` to satisfy `min_htlc` constraints in the route. With your proposal, the sender would be forced to put `500 msat` in the onion `amt_to_fwd`, which means giving the additional fee to intermediate nodes instead of the final node (and feels a bit weird, doesn't it?). Whereas if we allow over-paying the `total_amount_msat`, the sender can decide whether they want to give that extra fee to intermediate nodes or to the final node.

I understand your point of view, but my current feeling is to go for the extra flexibility of allowing the `total_amount_msat` over-payment. Let's see what others prefer, I can change that if that's the consensus.

joostjager: >Let's imagine that you retry a partial payment of 500 msat and are forced to pay 510 msat to satisfy min_htlc constraints in the route. With your proposal, the sender would be forced to put 500 msat in the onion amt_to_fwd, which means giving the additional fee to intermediate nodes instead of the final node.

I don't think you are giving it to the intermediate node straight away? The instruction is still to forward 510 sat to the final node. Only if the intermediate node would first try to pay lower amounts, they could steal some of those sats.

Also I think this only applies to the final shard. All the other shards would just put `510 msat` in the onion `amt_to_fwd`? It satisfies `min_htlc` and makes sure that a maximum of sats is delivered to the final destination. The overshoot can be compensated in the final shard to end up with the exact pre-determined `total_msat`.

What remains then is a sender being forced to overpay on the final shard because of a `min_htlc` constraint and the possibility of a routing node trying to steal the difference. It wouldn't increase the total cost of the payment though, because all prior shards were already 'optimal' in that they don't allow any fee stealing. The difference is just in where that final overpayment ends up IF the routing node decides to probe.

Even if we fix this for the final hop, it is still possible to shift the fee distribution to the intermediate hops.

If senders would always assume that overpayment of `total_msat` is acceptable, there wouldn't be a way anymore to reject such payments. You can fail the htlc set if it overpays, but the sender wouldn't know what's wrong. As mentioned before, overpayment isn't always desirable and can complicate accounting.

joostjager: I thought the probing attack was actually fixed by allowing nodes to forward less (if the amount was inflated because of `min_htlc`), just as is allowed currently for intermediate nodes.

Can you otherwise give a very concrete example, to make sure we're really talking about the same thing?

joostjager: How can you not allow nodes to do that? They can forward whatever they want.

Getting back to an example to clarify things: I think the following table lists the possibilities to deliver 5 sat to a node through a channel that has a min_htlc routing policy of 10 sat.

| amt_to_fwd A | htlc_amt A->B | amt_to_fwd B | current behavior routing node | current behavior final node | relaxed htlc_amt check final node |
| ------------ | ------------- | ------------ | ----------------------------- | --------------------------- | --------------------------------- |
| 10           | 9             | 10           | REJECT                        | REJECT                      | REJECT                            |
| 10           | 10            | 10           | ACCEPT                        | ACCEPT                      | ACCEPT                            |
| 10           | 11            | 10           | ACCEPT                        | REJECT                      | ACCEPT                            |
| 10           | 9             | 5            | ACCEPT                        | REJECT                      | ACCEPT                            |
| 10           | 10            | 5            | ACCEPT                        | REJECT                      | ACCEPT                            |
| 10           | 11            | 5            | ACCEPT                        | REJECT                      | ACCEPT                            |

To me it seems that the relaxed `htlc_amt` check alone fixes the probing issue because it makes the behavior of final nodes and routing nodes the same.

04-onion-routing.md
```diff
@@ -250,15 +250,16 @@ The reader:
...
+      - it cannot forward the HTLC to the peer indicated by the channel `short_channel_id`.
+      - incoming `amount_msat` - `fee` < `amt_to_forward` (where `fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#htlc-fees))
+      - `cltv_expiry` - `cltv_expiry_delta` < `outgoing_cltv_value`
   - if it is the final node:
     - MUST treat `total_msat` as if it were equal to `amt_to_forward` if it
       is not present.
     - MUST return an error if:
-       - incoming `amount_msat` != `amt_to_forward`.
-       - incoming `cltv_expiry` != `cltv_expiry_delta`.
+      - incoming `amount_msat` < `amt_to_forward`.
```
halseth, line 260: If we relax this field we must also start counting `payment_amt = SUM(htlcs.amt_to_forward)` at the final node. If we don't, a second-to-last node can overpay the htlc amount and mess with the tracking of MPP shards (that would lead to a premature settle of the HTLC set).

This would become even worse with the relaxation of the `total_msat` check, as it would get very easy to cause such a premature settle for a malicious node.

I think the ultimate solution would be what is mentioned in https://github.com/lightning/bolts/pull/1032/files#r993085640, adding a`required_incoming_htlc_amt` field. This would restrict all sorts of funny business routing nodes can do to the HTLC amt they send to the next hop, and would require them to always forward exactly `amt_to_fwd`. To me that sounds like it would solve all the probing issues also, as there would not be possible for a routing node to under- nor overpay, regardless of the next hop is the final or not.

04-onion-routing.md
t-bast: > To me it seems that the relaxed htlc_amt check alone fixes the probing issue because it makes the behavior of final nodes and routing nodes the same.

I think that's what you're missing, this is incorrect and the three last rows in your table are prone to probing. The onion's `amt_to_fwd` _must be the same for the next-to-last node and the last node_ to prevent probing. The underlying reason is that _intermediate nodes will never let you underpay_ (because that wouldn't satisfy their fee requirement), while underpaying the recipient becomes possible if its onion's `amt_to_fwd` is smaller than the next-to-last node's. So intermediate nodes can always try to underpay, and if that's accepted, it means the next node is the recipient. Does that make sense? Or should I detail a complete example (which is a bit tedious, since we need to have two hops in the table to correctly explain this).

04-onion-routing.md
t-bast: I think you're over-complicating this. From the recipient's point of view, they should be happy as soon as they receive **at least** `total_amount_msat`: they've been paid what they wanted and should release the preimage. From the sender's point of view, they should be happy as soon as they receive the preimage while paying **at most** `total_amount_msat`.

If intermediate nodes overpay an HTLC, neither the sender not the recipient are negatively impacted. The sender may even get the preimage for less than what they intended to pay: this is great, why would they prevent that?

04-onion-routing.md
joostjager: Ah I get it now. This is all about setting `amt_to_fwd` (Matt mentioned this too in https://github.com/lightning/bolts/pull/1032#discussion_r992442477). Sender's must make sure that the whole path is 'tight' so that no underpayment can happen anywhere. Because of the relaxation of `amt_to_fwd`, the 1 msat overpayment will be possible for both routing and intermediate nodes which fixes the privacy issue.

The next problem arises when a sender needs to overpay because of `min_htlc` and overshoots `total_msat`. They could decide not to adhere to the final amt_to_forward setting requirement above, but would sacrifice privacy. Therefore the second change to allow overshooting `total_msat` instead.

One thing to note is that there may be an additional cost for senders to keeping the path tight. If there is an intermediate node with a high `min_htlc`, the sender may not want to carry over that high amount to all channels further downstream. For those channels a proportional fee may be charged, in which case it would be cheaper for the sender to reduce amt_to_fwd right after the `min_htlc` has been passed. If the sender decides to do so, it becomes possible again to identify a node as being non-final.

For inbound fees though, I am still not sure how to handle them with these changes. An htlc comes in, but I can only decide that not enough inbound fees were paid until the mpp set is complete. All htlcs need to be held until then, and fresh channel updates need to be returned for all of them. It also looks like a layering violation, as fees pertain to the htlc level only.

joostjager: Actually, for inbound fees there may not be a problem. In think the sender can safely set the final `amt_to_fwd` to a lower value than the second-last `amt_to_fwd` _if_ the difference is equal to the inbound fee that is required. If a routing node would try to underpay (and identify the final node), the htlc would still fail because the final node requires `htlc_amt = amt_to_fwd + inbound_fee`.

joostjager: One more remark here: if a node recently lowered its (regular outbound) fee and the sender isn't aware of that yet, a window opens up momentarily where probing could reveal some information. Maybe something to keep in mind for payments with high privacy requirements? I think that to fully seal a the route, that `required_incoming_htlc_amt` field is needed still.

t-bast: > Sender's must make sure that the whole path is 'tight' so that no underpayment can happen anywhere. 

Exactly. That is most likely the default behavior of senders, since they build the onion based on the latest channel updates they received and have no reason to overpay intermediate fees. That's why I believe the simple changes in this PR are a good enough mitigation of that probing vector.

> if a node recently lowered its (regular outbound) fee and the sender isn't aware of that yet, a window opens up momentarily where probing could reveal some information

Yes, that's a good point. If an intermediate node underpays the next node, here is what can happen:

- the payment is rejected: the attacker doesn't learn anything
- the payment is accepted: the attacker learns that the next node is *not* the final node. By looking at recent channel updates from the next node, the attacker may learn with a high probability who the subsequent node is.

I don't know if we should be concerned or not. We really don't want intermediate nodes to be able to learn that information. The only way I see to fully fix this would be to add a `required_incoming_htlc_amt` field as discussed in other comments. However, the conditions for this probing attack won't appear often in practice, as the payment needs to match a decrease in fees for one channel in the route. Also if we think something like [fee ratecards](https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-September/003685.html) is the future, this will even decrease the probability that this probing attack succeeds.

> One thing to note is that there may be an additional cost for senders to keeping the path tight. If there is an intermediate node with a high min_htlc, the sender may not want to carry over that high amount to all channels further downstream. For those channels a proportional fee may be charged, in which case it would be cheaper for the sender to reduce amt_to_fwd right after the min_htlc has been passed. If the sender decides to do so, it becomes possible again to identify a node as being non-final.

I agree in theory, but I'm not sure it actually matters in practice. You would need to be using a route with a very high `min_htlc` in the middle **and** very high proportional fees after that node for this issue to cost you meaningful fees.

Even if those conditions do happen, the prober only learns that the next node is **not** the final node, which is much less worrying that learning who the final node is, so I think this is an acceptable probing vector (but maybe I'm wrong!).

joostjager: I agree that most of this may be theoretical. On the other hand though, if we would let the per-hop payload hmac also cover the required incoming htlc amount, all of these cases would be fixed for "free".

t-bast: > if we would let the per-hop payload hmac also cover the required incoming htlc amount

That would be an alternative for adding a `required_incoming_htlc_amt` field, right? But that removes all possibility of giving some leeway for intermediate nodes to decide to forward more to pay an inbound fee for example.

joostjager: Yes, a zero bytes alternative indeed for `required_incoming_htlc_amt` indeed. Why would an intermediate need leeway to pay for example an inbound fee? I'd think this is all decided upfront by the sender.

The feature bit is required indeed, but does the whole path need to support this? I thought the sender can calculate the hmac for each hop and include the required incoming amt only if the feature bit is present.

thomash-acinq:
The objections to this change seem to only consider the privacy aspect: preventing a malicious node from learning more information than it should, assuming all other nodes follow the spec. However for this to work, we need all other nodes to follow the spec, and for that we need to make sure that incentives are aligned. Currently, the spec asks us to refuse free money to protect the privacy of others. Any economically rational actor should take the free money as there is no downside for them, the downside is only for the privacy of the others.
This changes makes it so that economically rational actors would now follow the spec, which is in itself a sufficient reason to accept it. In addition, it also improves privacy, but only because now everyone is incentivized to actually follow the spec.

04-onion-routing.md
halseth: Starting to magically include more data that's not be present in the HMAC sounds like something that would make similar changes harder later. 

What's bad about just adding an extra (optional) field `require_incoming_amount` and as more nodes start checking this, probing gets harder and harder. I still thinks this is the cleanest way of ensuring intermediate nodes can't do anything other than what the sender tells them to, which sounds like it would fix most of these probing vectors (maybe even some we haven't thought of).

Should also be totally backwards compatible IIUC.

I agree with t-bast that this would give intermediate nodes less leeway to overpay (for some reason), but as Joost is saying, shouldn't senders really be in charge of deciding that?

04-onion-routing.md
halseth: I agree with the theory, but are implementations properly doing this in practice? I guess it should be handled anyway, since there are never any guarantees all shards will be settled.

04-onion-routing.md
```diff
@@ -292,7 +293,7 @@ The writer:
     - otherwise:
       - MUST set `total_msat` to the amount it wishes to pay.
     - MUST ensure that the total `amount_msat` of the HTLC set which arrives at the payee
-      is equal to `total_msat`.
+      is equal to or greater than `total_msat`.
```
halseth, line 296: I think it is worth considering the consequences of doing this without a feature bit in the invoice to signal the recipient supports such overpaying. Although possibly rare, the sender cannot know if he can overpay `total_msat` and it will be accepted.

Is it overkill to signal in the invoice whether we support such overpayment?

04-onion-routing.md
t-bast: That's at least how eclair handles it ;)

04-onion-routing.md
t-bast: That's a good point. I'm not sure it's worth "wasting" a feature bit, but maybe a simple flag in the invoice would be enough? I'll look into it.

04-onion-routing.md
joostjager: >as an intermediate node or a final node, if I receive more money than expected, I will take it

Another way to look at this is that failing could lead to more money on the longer term. By failing you're teaching your predecessor that there is no point in trying to steal anything and next time they might just pay the instructed (higher) amount?

04-onion-routing.md
joostjager: Perhaps the two changes in this PR can be decoupled? The relaxation of `amt_to_fwd` seems fairly straight-forward, where as changing the mpp logic probably needs further discussion.

04-onion-routing.md
```diff
@@ -259,6 +259,7 @@ The reader:
     - MUST return an error if:
       - incoming `amount_msat` < `amt_to_forward`.
       - incoming `cltv_expiry` < `outgoing_cltv_value`.
+      - incoming `cltv_expiry` < `current_block_height` + `min_final_cltv_expiry_delta`.
```
joostjager, line 262: With this added, is the other condition `incoming cltv_expiry` < `outgoing_cltv_value` still useful?

04-onion-routing.md
```diff
@@ -250,15 +250,15 @@ The reader:
...
+      - incoming `amount_msat` - `fee` < `amt_to_forward` (where `fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#htlc-fees))
+      - `cltv_expiry` - `cltv_expiry_delta` < `outgoing_cltv_value`
   - if it is the final node:
     - MUST treat `total_msat` as if it were equal to `amt_to_forward` if it
       is not present.
     - MUST return an error if:
-       - incoming `amount_msat` != `amt_to_forward`.
-       - incoming `cltv_expiry` != `cltv_expiry_delta`.
+      - incoming `amount_msat` < `amt_to_forward`.
+      - incoming `cltv_expiry` < `outgoing_cltv_value`.
```
joostjager, line 261: Should this be `>` for the final node? For routing nodes a higher expiry is accepted too?

04-onion-routing.md
t-bast: They are actually already somewhat decoupled since I'm just building on top of Matt's PR.
I can rebase my changes directly on top of `master` though if needed, let's wait for other people to chime in first.

04-onion-routing.md
joostjager: Nvm, this are the error conditions.

t-bast: Can you detail? I think the current inequation is correct, you want to return an error if you receive an HTLC that expires before what is specified in the onion, don't you? But you are completely fine with accepting an HTLC that expires after what what the sender intended.

An intermediate node can decide to use a lower `cltv_expiry_delta` than what the sender used when relaying (even honestly, if they recently lowered it), so intermediate nodes could receive an HTLC with a higher `cltv_expiry` than what the sender wanted, but that's a net benefit for them so they shouldn't reject that (and with the current requirements, they won't, because when computing the next `cltv_expiry` they will find something greater than what they're instructed to relay by their `outgoing_cltv_value`).

Does that make sense or should I detail that with some concrete sample values?

04-onion-routing.md
t-bast: I think we need both: this one is for the case where the sender doesn't use any "dummy hop" to help hide the recipient, but ideally senders should use an `outgoing_cltv_value` that is much higher than `current_block_height` + `min_final_cltv_expiry_delta` to make it look like there are other hops afterwards.

04-onion-routing.md
joostjager: No it's clear, I thought I was reading the acceptance conditions.

04-onion-routing.md
joostjager: Ah yes, that makes sense. I think in lnd, we just add three blocks in case blocks are found while the htlc is in flight.

t-bast:
Rebased on `master`.

halseth (approved):
ACK 391ed8448d39329f351bc940a411fcc612d4b6d7

joostjager:
I am not so sure about this change anymore, https://github.com/lightning/bolts/pull/1031#issuecomment-1290938756

t-bast:
> I am not so sure about this change anymore, https://github.com/lightning/bolts/pull/1031#issuecomment-1290938756

As was pointed out in response to your comment, I believe this isn't a valid reason to reject the PR as we already support over-payment.

We can add a recipient feature to require strict payment amounts (at the expense of privacy) in future work, but that shouldn't have any impact on acceptance of this PR.

joostjager:
>We can add a recipient feature to require strict payment amounts (at the expense of privacy) in future work, but that shouldn't have any impact on acceptance of this PR.

I am wondering if we can have strict payment amounts without sacrificing privacy.

I think we can have that if we'd go the other previously discussed way and instead of relaxing the final amt_to_fwd requirement (this pr), we'd tighten acceptance for routing nodes via a new onion field or inclusion of the amount in the hmac'ed data.

The downside of that is, and we also discussed this before, is that an economically rational routing node would ignore the accept instruction in the onion, maximize profit without caring about privacy, bringing us back to the current situation.

So yes, maybe the outcome is that there is no other option than relaxing indeed. Unless anyone has a new creative idea?

t-bast:
I think that strict payments should not be the default as they're not what an economically rational actor would want. But there's room for making this an opt-in feature, signaled in the invoice, most likely at the expense of privacy, and maybe this is a trade-off that some nodes are ok with?

Anyway implementing strict payments is a separate proposal, and as you say hopefully by the time we really commit to implementing it we'll have found ways of doing it that doesn't sacrifice too much privacy.

joostjager (approved):

rustyrussell:
Yes.  Consistency FTW.

Ack.

Roasbeef (approved):
LGTM 💎

PR #1036, opened 2022-10-27
Add option to sign commitments at various feerates (FEAT 32/33)

t-bast:
When using `option_anchors`, nodes must keep a pool of utxos available to set the fees of commitment and HTLC transactions at broadcast time. This is a complex task that requires sophisticated utxo management to efficiently protect against malicious peers (even more so when HTLC txs pay 0 fees by default).

The `option_alternative_feerates` feature reduces that complexity, by making it possible for nodes to have versions of HTLC transactions at various feerates, which may remove the need to add external inputs in most cases and keep them for exceptional occasions, at the expense of slightly more latency, bandwidth and storage usage.

Disclaimer: this is a very low tech and unsatisfying mitigation for the issues discussed in #845, but it's trivial to implement and may provide good guarantees against attackers in practice. I'm merely chasing concept ACKs at that point and need to prototype it to fill potential gaps in the requirements. It can be especially useful for mobile wallets that don't want to manage on-chain fee-bumping reserves.

TODO:

- investigate alternatives, such as signing multiple commit txs with different amounts for the anchor output: you can then use this output to fund HTLC txs from this channel (it lets the reserve come from the channel funds inside of an external wallet). Think about the dependency with the channel reserve.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -24,6 +24,7 @@ operation, and closing.
       * [Committing Updates So Far: `commitment_signed`](#committing-updates-so-far-commitment_signed)
       * [Completing the Transition to the Updated State: `revoke_and_ack`](#completing-the-transition-to-the-updated-state-revoke_and_ack)
       * [Updating Fees: `update_fee`](#updating-fees-update_fee)
+      * [Signing alternative feerates](#signing-alternative-feerates)
     * [Message Retransmission: `channel_reestablish` message](#message-retransmission)
   * [Authors](#authors)
 
@@ -783,20 +784,25 @@ Once both nodes have exchanged `channel_ready` (and optionally [`announcement_si
 Changes are sent in batches: one or more `update_` messages are sent before a
 `commitment_signed` message, as in the following diagram:
 
-        +-------+                               +-------+
-        |       |--(1)---- update_add_htlc ---->|       |
-        |       |--(2)---- update_add_htlc ---->|       |
-        |       |<-(3)---- update_add_htlc -----|       |
-        |       |                               |       |
-        |       |--(4)--- commitment_signed --->|       |
-        |   A   |<-(5)---- revoke_and_ack ------|   B   |
-        |       |                               |       |
-        |       |<-(6)--- commitment_signed ----|       |
-        |       |--(7)---- revoke_and_ack ----->|       |
-        |       |                               |       |
-        |       |--(8)--- commitment_signed --->|       |
-        |       |<-(9)---- revoke_and_ack ------|       |
-        +-------+                               +-------+
+        +-------+                                           +-------+
+        |       |--(1)----       update_add_htlc        ---->|       |
+        |       |--(2)----       update_add_htlc        ---->|       |
+        |       |<-(3)----       update_add_htlc        -----|       |
+        |       |                                            |       |
+        |       |--(4)---       commitment_signed        --->|       |
+        |       |--(4a)--- alternative_commitment_signed --->|       |
+        |       |--(4b)--- alternative_commitment_signed --->|       |
+        |   A   |<-(5)----       revoke_and_ack        ------|   B   |
+        |       |                                            |       |
+        |       |<-(6)---       commitment_signed        ----|       |
+        |       |<-(6a)--- alternative_commitment_signed ----|       |
+        |       |<-(6b)--- alternative_commitment_signed ----|       |
+        |       |<-(6c)--- alternative_commitment_signed ----|       |
+        |       |--(7)----       revoke_and_ack        ----->|       |
+        |       |                                            |       |
+        |       |--(8)---       commitment_signed        --->|       |
+        |       |<-(9)----       revoke_and_ack        ------|       |
+        +-------+                                            +-------+
 
 Counter-intuitively, these updates apply to the *other node's*
 commitment transaction; the node only adds those updates to its own
@@ -1183,21 +1189,30 @@ sign the resulting transaction (as defined in [BOLT #3](03-transactions.md)), an
    * [`u16`:`num_htlcs`]
    * [`num_htlcs*signature`:`htlc_signature`]
 
+1. `tlv_stream`: `commitment_signed_tlvs`
+2. types:
+    1. type: 1 (`alternative_feerates`)
+    2. data:
+        * [`...*u32`:`feerates_per_kw`]
+
 #### Requirements
 
 A sending node:
   - MUST NOT send a `commitment_signed` message that does not include any
-updates.
-  - MAY send a `commitment_signed` message that only
-alters the fee.
-  - MAY send a `commitment_signed` message that doesn't
-change the commitment transaction aside from the new revocation number
-(due to dust, identical HTLC replacement, or insignificant or multiple
-fee changes).
+    updates.
+  - MAY send a `commitment_signed` message that only alters the fee.
+  - MAY send a `commitment_signed` message that doesn't change the
+    commitment transaction aside from the new revocation number (due to dust,
+    identical HTLC replacement, or insignificant or multiple fee changes).
   - MUST include one `htlc_signature` for every HTLC transaction corresponding
     to the ordering of the commitment transaction (see [BOLT #3](03-transactions.md#transaction-input-and-output-ordering)).
+  - If `option_anchors` and `option_alternative_feerates` were negotiated:
+    - MUST send `alternative_feerates` containing other feerates for which it
+      will send signatures.
+    - MUST send one `alternative_commitment_signed` for each of those feerates.
+    - SHOULD include feerates that are higher than the current commitment feerate.
   - if it has not recently received a message from the remote node:
-      - SHOULD use `ping` and await the reply `pong` before sending `commitment_signed`.
+    - SHOULD use `ping` and await the reply `pong` before sending `commitment_signed`.
 
 A receiving node:
   - once all pending updates are applied:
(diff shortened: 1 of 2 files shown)


SomberNight:
How does this help with the HTLC txs? You still need to bring your own utxo - or are you assuming using the old `option_anchor_outputs` instead of `option_anchors_zero_fee_htlc_tx`? (If so, is the old feature supported by nodes? I thought it was pretty much deprecated in favour of `option_anchors_zero_fee_htlc_tx`)

> It can be especially useful for mobile wallets that don't want to manage on-chain fee-bumping reserves.

Yes, please. :) It would be great to be able to avoid reserving utxos...

t-bast:
> How does this help with the HTLC txs? You still need to bring your own utxo

No you won't need to, sorry if that was unclear, but the HTLC txs from `alternative_commitment_signed` use `SIGHASH_ALL`, not `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY`, and they do pay the feerate announced. If one of them has the feerate you want, you just add your local sigs and broadcast it without adding any new input. If none of your transactions have a good feerate, then you use the `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY` tx from `commitment_signed` and add your own inputs. One added benefit is a smaller HTLC tx since it has a single input, which is more fee-efficient.

> Doing this for HTLCs is also really complicated because we probably don't want to go back from the zero fee htlcs world for several reasons.

Can you detail what reasons? On the contrary, I want this _mostly_ for the HTLC txs (more than for the commit tx), because that's where we potentially get a big multiplier on the number of on-chain txs we have to confirm. For those HTLC txs, it is effectively broadcaster-pays (since the feerate is paid from the HTLC input).

It is slightly cumbersome for the commit tx, because you need to apply different trimming thresholds for each feerate, but it's not new complexity we didn't have before.

I agree it would make a lot of sense for those alternative commit txs to take the fee from the broadcaster's main output: this part would be new complexity, where you won't be able to reach the desired feerate if your peer's main output doesn't have enough value, but it doesn't feel overwhelmingly complex (especially in light of the benefits of sparing your utxos and the increased safety that comes with it - even though you still need some utxo management logic for the case where the pre-signed feerates are still not good enough).

I know it is definitely not a very satisfying solution :sweat_smile:, but I don't see how else we can get that extra-safety in the short term, and I think such extra-safety is quite desirable...

02-peer-protocol.md
```diff
@@ -1342,6 +1359,47 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+   * [`u16`:`num_htlcs`]
+   * [`num_htlcs*signature`:`htlc_signature`]
+
+#### Requirements
+
+A sending node:
+  - If `option_alternative_feerates` was negotiated:
+    - MUST send one `alternative_commitment_signed` for each feerate listed
+      in the corresponding `commitment_signed.alternative_feerates`.
+    - MUST use `SIGHASH_ALL` for the `htlc_signature`s.
```
SomberNight, line 1381: >> How does this help with the HTLC txs? You still need to bring your own utxo
>
> No you won't need to, sorry if that was unclear, but the HTLC txs from `alternative_commitment_signed` use `SIGHASH_ALL`, not `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY`, and they do pay the feerate announced.

I see. The fee part was not clear to me from the text. I guess I could have deduced it from using the sighash_all requirement. I think it would be better to explicitly mention this, e.g.:

```suggestion
    - MUST use `SIGHASH_ALL` for the `htlc_signature`s, and apply `feerate_per_kw` for each HTLC transaction even if `option_anchors_zero_fee_htlc_tx` applies to the channel.
```

(to counterbalance bolt-03 being very specific [about the contrary](https://github.com/lightning/bolts/blob/16973e2b857e853308cafd59e42fa830d75b1642/03-transactions.md#requirements-2))

t-bast: Yes definitely, if there is generally a good feeling about the proposal, I'll make that PR much more detailed, since there are quite a few subtleties that are currently hard to grasp.

02-peer-protocol.md
```diff
@@ -1342,6 +1359,47 @@ it's simplest to only allow it to set fee levels; however, as the same
...
+   * [`signature`:`signature`]
+   * [`u16`:`num_htlcs`]
+   * [`num_htlcs*signature`:`htlc_signature`]
+
+#### Requirements
+
+A sending node:
+  - If `option_alternative_feerates` was negotiated:
+    - MUST send one `alternative_commitment_signed` for each feerate listed
+      in the corresponding `commitment_signed.alternative_feerates`.
```
SomberNight, line 1380: Note that HTLC-timeout transactions are CLTV locked (with independent CLTVs between htlcs). A CLTV might be hundreds of blocks into the future. Broadcasting a commitment tx at `alternative_feerate1` locks in that same feerate for all HTLCs too. (because the sighash in the htlc tx's input commits to the txid of the parent commit tx)

Unless we make it quadratic and pre-sign htlc txs at each `alternative_feerate` for commit txs at each `alternative_feerate`. That would mean creating `num_htlcs * len(alternative_feerates)**2` signatures at every committed channel state update.

02-peer-protocol.md
```diff
@@ -1183,42 +1189,53 @@ sign the resulting transaction (as defined in [BOLT #3](03-transactions.md)), an
...
+  - MAY send a `commitment_signed` message that only alters the fee.
+  - MAY send a `commitment_signed` message that doesn't change the
+    commitment transaction aside from the new revocation number (due to dust,
+    identical HTLC replacement, or insignificant or multiple fee changes).
   - MUST include one `htlc_signature` for every HTLC transaction corresponding
     to the ordering of the commitment transaction (see [BOLT #3](03-transactions.md#transaction-input-and-output-ordering)).
+  - If `option_anchors` and `option_alternative_feerates` were negotiated:
+    - MUST send `alternative_feerates` containing other feerates for which it
+      will send signatures.
+    - MUST send one `alternative_commitment_signed` for each of those feerates.
```
SomberNight, line 1212: I find it weird that Alice decides what feerates will be available to Bob in Bob's commitment tx, and vice-versa. Ideally it would be Alice who can decide for herself that she wants feerates `[3,10,30,100,300]` (sat/vbyte). What if Alice sends some nice rates to Bob but Bob does not give any alternatives to Alice (`alternative_feerates=[]`)? Is Alice supposed to force-close because of not having enough feerate options?

02-peer-protocol.md
t-bast: This is a very good remark. I went back and forth on that, the other option I had in mind that doesn't have this issue is to only sign alternative HTLC txs and not the commit tx. We keep a single commit tx (the one signed in `commitment_signed`, which will likely need some CPFP using the anchor) and sign alternative HTLC txs for each pending HTLC (and all these alternatives would spend the same output of the commit tx).

That means we need to handle the case where an HTLC tx doesn't make sense to be signed at some feerates (because its value goes entirely to fees): there are different ways of doing that (either skip them entirely or provide a dummy signature like `0x00...00`), it's not necessarily hard but is a bit messy.

02-peer-protocol.md
t-bast: I agree, that's one of the decisions I wanted to have on this PR, thanks for raising it :wink:. What we do here depends on whether we do "broadcaster pays" or not and whether we re-sign commitment transactions or not.

One issue with Alice deciding what feerates Bob should sign for her commit transaction is that she ask Bob to pay a feerate so high that HTLCs all go to dust, and then she could broadcast the corresponding commitment once it's revoked and Bob won't be able to claim anything. Similarly to what we have with `update_fee`, Bob shouldn't sign feerates that it feels are absurdly high. But we can probably find a good middle ground that works well enough in practice.

t-bast:
> Don't we need to reintroduce all the dust complexity in https://github.com/lightning/bolts/pull/919 that we were able to basically get rid of?

I agree, that would be very annoying, but we don't necessarily have to, the design space is quite open here, it's a matter of what we want to do. A very simple solution is to decide that those `alternative_commitment_signed` don't apply any trimming and keep the exact same HTLC outputs as the main `commitment_signed`. Some of those HTLC outputs will be uneconomical to spend on-chain, but they're already uneconomical to spend without those additional signatures, so we're not making things worse?

I'm sorry if this is still quite hand-wavy, I didn't open this PR with a finalized idea of all the details, on the contrary I wanted to gather feedback to help decide what trade-offs seem best.

ariard:
Joining Matt's concern here, I think we're re-introducing all the risk aleas of dusty HTLCs either burnable by your counterparty as a straight damage or collectable by your counterparty if it behaves in collusion with a miner (an assumption lowered with the deployment of Stratum v2 in fact, where the block template can be selected at the endpoint). HTLC transactions can be both ways, as such even if you have the LSP always acting as a funder with the fee responsibility, you might give something like counterparty's `MAX_ACCEPTED_HTLC` * `max_feerate_presigned_commitment` * `average_htlc_transaction` of "trusted" fee surface to your counterparty. So assuming 50 (LDK default), 100 sat/vbytes, 200 bytes, that is 0.01  BTC of operational risk by connected spoke for a LSP.

I don't think the solution of relying on your reserve balance (from channel A) as a anchor input value on HTLC transactions (from channel B) is a viable solution either, as you cannot be certain in the worst-case scenario all your channels might go on-chain as the same time. The idea of alternative pre-signed feerates commitment transaction where the extra fee is taken from the broadcaster sounds more plausible. Though still you have to consider issue like withheld HTLC resolution from your counterparty to push you to broadcast your own version of the commitment. And if your fee-estimation is delegated to a third-party or external API, combining both for the highest fee altnerative to be broadcast. Appears that, it might be good.

t-bast:
> once the commitment transaction is on-chain you have your (at least reserve) balance lying there to use as anchor input value on HTLC transactions.

I don't think that really works though, unless you have a very long `cltv_expiry_delta` or a low enough `max-accepted-htlcs`...
If you use the default `max-accepted-htlcs` of 483, an attacker would simply make you send 483 outgoing HTLCs with expiries N+1, N+2, ..., N + 483 so you can't easily use a single utxo to claim them all on time.

At every block, you'd need to double-spend your previous batched transaction to include one more HTLC, and that requires increasing fees as well to follow RBF rules. It would be interesting to run the numbers and see how much extra fees you'd end up paying to "save" those utxos.

The alternative is to make a tree of unconfirmed transactions where you use the change output of the HTLC tx with expiry N+1 to fund the HTLC tx with expiry N+2, and repeat that for each HTLC, but that also ends up being an RBF nightmare: if the fees of the first HTLC tx are too low, RBF-ing it will be incredibly expensive because of the chain of unconfirmed descendants, so your only option is to use CPFP at the bottom of the tree, which means setting a high feerate for *all* the HTLC txs in the tree, whereas some of them still have some time before expiry. That also ends up greatly overpaying fees, which is a painful griefing attack, isn't it?

Those scenarios are why I'm focusing on HTLC txs, because getting commitment txs included is easier in my opinion since it doesn't have the transaction multiplier effect that HTLCs have, so you're probably ok always targeting a quick confirmation window which then frees another utxo to fee-bump another channel.

> Joining Matt's concern here, I think we're re-introducing all the risk aleas of dusty HTLCs either burnable by your counterparty as a straight damage

I agree that this is non-trivial: the signer cannot sign for feerates that are too high, because they otherwise potentially lose too much money to miners if a revoked commit is broadcast. The signer would need to choose the feerates it uses to sign to ensure that the amount burnt to fees never exceeds the channel reserve. I'll run some numbers to see if that could be reasonable or not.

PR #1039, opened 2022-11-07
BOLT#02: change SHOULD to MUST for fee-range mismatch

Crypt-iQ:
Addresses most of #1013, but without some warning codes the funder won't know what to change. This patch requires that the sender of `fee_range` doesn't send another until receiving a response via `closing_signed` or `warning`.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -704,6 +704,8 @@ The sending node:
   inclusion in a block.
   - SHOULD set `fee_range` according to the minimum and maximum fees it is
   prepared to pay for a close transaction.
+  - MUST NOT send another `fee_range` if it has already sent a `fee_range` and hasn't received
+  a `closing_signed` or `warning` in response from the peer.
   - if it doesn't receive a `closing_signed` response after a reasonable amount of time:
     - MUST fail the channel
   - if it is not the funder:
@@ -726,7 +728,7 @@ The receiving node:
     - MAY close the connection.
   - if the message contains a `fee_range`:
     - if there is no overlap between that and its own `fee_range`:
-      - SHOULD send a warning
+      - MUST send a warning
       - MUST fail the channel if it doesn't receive a satisfying `fee_range` after a reasonable amount of time
     - otherwise:
       - if it is the funder:


t-bast:
This looks like a hacky way of making the closing protocol compatible with Musig2 funding outputs (and failing to mention that this is the motivation for the change). I think we should instead work on creating a cleaner protocol for closing taproot channels. My personal preference is to make it an explicit turn-based protocol (instead of hijacking the unrelated `warning` message).

The new requirement doesn't even seem doable since there's no reestablish logic to know that when you sent a `fee_range` it was received by your peer, so you may end up with deadlocks?

Crypt-iQ:
> This looks like a hacky way of making the closing protocol compatible with Musig2 funding outputs

this isn't the motivation for the change though - the current protocol as written fails in an edge case

> The new requirement doesn't even seem doable since there's no reestablish logic

since things start over if the conn restarts, this doesn't seem necessary since you just start over

t-bast:
> since things start over if the conn restarts, this doesn't seem necessary since you just start over

Gotcha, that should be clarified, I read the requirement as something that must be applied throughout reconnects.

> the current protocol as written fails in an edge case

That's because it was designed that way, not being a turn-based protocol and requiring node operator intervention to resolve feerate conflicts / lack of progress. I now think that this was a mistake and I should have made it a strict turn-based protocol...

I don't think it makes any sense to require a node to send a `warning`. Warnings are optional hints that nodes send to each other when they detect weird behavior from their peer, and I believe they should stay that way. Using them to transform this protocol into a strict turn-based protocol is a red herring to me. If we want to make it a turn-based protocol, we should be more explicit about it (and as I previously stated in the related discussions, add an explicit reject message). I'm curious to get feedback from the other implementations though?

Crypt-iQ:
an explicit reject message is better imo, this patch was motivated by this comment https://github.com/lightning/bolts/issues/1013#issuecomment-1230772071

vincenzopalazzo (commented):
I'm also in favor of another message here but I'm not sure what kind of message in order to avoid making the protocol less verbose.

Also, encoding some data in the warning message could work, but I need to think more about this use case

t-bast:
Shall we change this PR to become the spec for a new `reject_closing_signed` message or should we close it and open a separate PR?

Crypt-iQ:
Can close this, the reject version of coop closing is in this PR https://github.com/lightning/bolts/pull/1016 -- I think that addresses what this PR is trying to fix

PR #1040, opened 2022-11-09
Use onion amount in MPP set calculation

t-bast:
The sender chooses the amounts that are set in the onion payload (`amt_to_forward`) but cannot predict what amounts will be set in the HTLCs (`amount_msat`) since intermediate nodes are allowed to overpay.

This was pointed out by @joostjager [here](https://github.com/lightning/bolts/pull/1031#discussion_r1016479991).

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1088,11 +1088,9 @@ An _intermediate hop_ MUST NOT, but the _final node_:
   - if the `cltv_expiry` value is unreasonably near the present:
     - MUST fail the HTLC.
     - MUST return an `incorrect_or_unknown_payment_details` error.
-  - if the `outgoing_cltv_value` does NOT correspond with the `cltv_expiry` from
-  the final node's HTLC:
+  - if the `cltv_expiry` from the final node's HTLC is below `outgoing_cltv_value`:
     - MUST return `final_incorrect_cltv_expiry` error.
-  - if the `amt_to_forward` does NOT correspond with the `incoming_htlc_amt` from the
-  final node's HTLC:
+  - if `amount_msat` from the final node's HTLC is below `amt_to_forward`:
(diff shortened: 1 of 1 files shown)


04-onion-routing.md
```diff
@@ -308,10 +309,10 @@ The final node:
     - MUST add it to the HTLC set corresponding to that `payment_hash`.
     - SHOULD fail the entire HTLC set if `total_msat` is not the same for
       all HTLCs in the set.
-    - if the total `amount_msat` of this HTLC set is equal to or greater than
-      `total_msat`:
+    - if the total `amt_to_forward` of this HTLC set is equal to or greater
+      than `total_msat`:
       - SHOULD fulfill all HTLCs in the HTLC set
-    - otherwise, if the total `amount_msat` of this HTLC set is less than
+    - otherwise, if the total `amt_to_forward` of this HTLC set is less than
```
joostjager, line 315: No comments on this change, but I did find this:

https://github.com/lightning/bolts/blob/b38156b9510c0562cf50f8758a64602cc0315c19/04-onion-routing.md?plain=1#L1093

Did we forget to change that line?

t-bast: Done in https://github.com/lightning/bolts/pull/1040/commits/648fcdc0d0ce61b9acb76abdeaee19549a4d7c59

cdecker:
ACK

carlaKC (approved):

halseth (approved):
As mentioned during the call, this was discussed earlier: https://github.com/lightning/bolts/pull/1032#discussion_r993310877. The conclusion then was that the change was not strictly needed (sender should always be able to handle a premature preimage release).

But I'm for it, this makes it simpler to reason about, and gives routing nodes less room to play around with the settlement.

ACK

PR #1042, opened 2022-11-19
a note on the maximum size of Tagged Fields

JssDWt:
Make it more obvious to readers of bolt 11 that the maximum length of a Tagged Field is 639 bytes.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -135,6 +135,8 @@ Each Tagged Field is of the form:
 1. `data_length` (10 bits, big-endian)
 1. `data` (`data_length` x 5 bits)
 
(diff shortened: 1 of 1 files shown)


cdecker:
It's good to call this limit out, however we should also discourage using large fields in the first place as they tend to degrade the readability in some mediums (QR codes).

ACK https://github.com/lightning/bolts/pull/1042/commits/5bdb7a0c1760c6e8be4c7eb7bcfdc0332fa0baf4

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1042/commits/5bdb7a0c1760c6e8be4c7eb7bcfdc0332fa0baf4, this is a minor clarification, merging now.

PR #1043, opened 2022-11-21
draft: Staking Credentials token issuance/redemption

ariard:
A formalization of a reputation-based scheme to solve channel jamming is proposed. The system relies on "credentials" issued by routing hops and requested to be attached to each HTLC forward request. The "credentials" can be used by a reputation algorithm to reward/punish payment senders and allocate channel liquidity resources efficiently. The "credentials"  initial distribution can be boostraped leveraging one-time upfront fees paid toward the routing hops. Afterwards, the "credentials" subsequent distribution can rely on previous HTLC traffic.

See corresponding mailing list thread for more context. There is also a PoC in LDK: https://github.com/lightningdevkit/rust-lightning/pull/1848 (though now code is behind the description)

Feedback very welcome,.This is a really rough sketch aiming to offer a sound reputation-based proposal to be compared against all the jamming mitigations alternatives: upfront fees/unconditional fees/forwarding pass/etc.

This proposal is inspired by [IETF's Privacy Pass](https://www.ietf.org/archive/id/draft-ietf-privacypass-architecture-08.html) from original @naumenkogs suggestion.

### DIFF
--- a/60-staking-credentials-archi.md
+++ b/60-staking-credentials-archi.md
@@ -1,0 +1,245 @@
+# Staking Credentials
+
+This document specifies the Staking Credentials architecture and requirements for its constituents
+protocols used for constructing anonymous credentials mechanisms in the context of Bitcoin financial
+contracts and Civkit functionary services.
+
+This draft is version 0.0.1 of the Staking Credentials architecture.
+
+## Introduction
+
+Staking Credentials is an architecture for Bitcoin financial contracts based on anonymous credentials.
+Those credentials are not only unlinking their issuance from their consumption, but also allows the
+user staking of credentials to downgrade the costs of entering in future Bitcoin financial contracts
+or accessing Civkit functionary services. The quality and thresholds of assets can be selected with
+granularity by the contract provider to achieve a satisfying level of risks, combining monetary and
+reputation strategies.
+
+Staking Credentials approach is the following: the clients presents a basket of scarce assets
+(Lightning preimage, stakes certificates) and blinded credentials to the issuance server, the credentials
+are authenticated and yeilded back to the client. Once unblinded the credentials can be used to fulfill
+a contract or service request towards a service provider server. The credentials are anonymous in
+the sense that a given credential cannot be linked to the protocol instance in which that credential
+was initially issued.
+
+The Staking Credentials architecture consists of two protocols: credentials issuance and redemption.
+The issuance protocol runs between two endpoints referred to as Requester and Issuer and one
+function: Credential Authentication. The entity that implements the Credential Authentication,
+referred to as the Issuer, is responsible for counter-signing the credentials in response to requests
+from Requester. Issuer might accept different base assets in function of their risk strategy, and
+the overhead cost they would like to transfer on the Committer for answering a contract or service
+request. Requester and Issuer can agree on an Authentication method in function of the cryptographic
+properties aimed for.
+
+The redemption protocol runs between two endpoint reffered to as Client and Provider and
+one function: Credential Consumption. The entity that implements the Credential Consumption, referred
+to as the Client, is responsible to provide an authenticated Credential covering the contract risk
+service request risk, as defined by the Provider. A Client can be a Provider, but those entities
+can be also dissociated. An Issuer can be a Provider, but those entities can be also dissociated.
+A Client can aggregate authenticated Credentials from multiple Requesters.
+
+The credentials issuance and redemption protocols operate in concert as shown in the figure below:
+
+```
+		 ___________						        ____________
+		|	    |      1. Scarce Asset + Blinded Credentials       |	    |
+		|	    |------------------------------------------------->|  	    |
+		|	    |						       |	    |
+		| Requester |						       |   Issuer   |
+          ------|	    |	   2. Authenticated Binded Credentials         |	    |
+          |	|	    |<-------------------------------------------------|	    |
+	  |	|___________|						       |____________|
+	  |										 ^
+	  |										 |
+	  |  3. Authenticated Unblinded Credentials					 |
+	  |										 |
+	  |				       5. Credentials Authentication Validation	 |
+	  |										 |
+	  |										 |
+	 _V__________								    _____|________
+	|	     |  4. Authenticated Unblinded Credentials + Service Request   |		  |
+	|	     |------------------------------------------------------------>|		  |
+	|	     |								   |		  |
+	|   Client   |								   |   Provider   |
+	|	     |         6. Contract Acceptance OR Contract Reject	   |		  |
+	|	     |<------------------------------------------------------------|		  |
+	|____________|								   |______________|
+
+```
+
+This documents describes requirements for both issuance and redemption protocols. It also provides
+recommendations on how the architecture should be deployed to ensure the economic effectiveness of
+assets, the compatibility of incentives, the conservation of the client UX, the privacy and security
+and the decentralization of the ecosystem.
+
+## Terminology
+
+The following terms are used throughout this document.
+
+- Client: An entity that pledges a Credential as assets to redeem a Contract or Service Request.
+- Requester: An entity that commit a set of scarce assets to redeem an authenticated Credential from an Issuer.
+- Issuer: An entity that accept a scarce asset in exchange of authenticating a Credential.
+- Provider: An entity that accept a set of Credentials as payment in case of Contract risk or Service providance to the client.
+
+## Architecture
+
+The Staking Credentials architecture consists of four logical entities -- Client, Issuer, Requester, Provider --
+that work in concert for credentials issuance and redemption.
+
+## Credentials Issuance
+
+The credentials issuance is an authorization protocol wherein the Requester presents scarce asset and
+blinded credentials to the Issuer for authorization. Usually, the rate of scarce assets to credentials
+as placed by the Issuer should have been discovered by the Committer in non-defined protocol phase.
+
+There are a number of scarce assets that can serve, including (non-limited):
+- proof-of-payment (e.g Lightning preimage or on-chain txid)
+- ecash token
+- stakes certificates
+- "faithful usage" reward
+
+Issuer should consider which base assets to accept in function of the risk-management strategy they adopt. If
+they aim for a 0-risk coverage, where all the contract or service execution risk is transferred on the counterparty,
+they should pick up the default of on-chain/off-chain payment, where each contract cost unit should be covered by
+a satoshi unit.
+
+The credential issuance can be deployed as "native" or "reward".
+
+In the "native" flow, a scarce asset for credential authentication should be exchanged a priori of a redemption phase:
+
+```
+		 ___________					    ____________
+		|	    |	  1. Scarce asset + Credentials    |	        |
+		|           |------------------------------------->|	        |
+		|           |					   |	        |
+		| Committer |					   |   Issuer   |
+		|	    |	  2. Authenticated Credentials	   |		|
+		|	    |<-------------------------------------|		|
+		|___________|					   |____________|
+
+```
+
+In the "reward" context, one type of scarce asset, i.e contract fees are paid by the Client to the Provider,
+a posteriori of a successful redemption phase. A quantity of authenticated credentials should be given
+from the linked Issuer, back to the Committer:
+
+```
+	         ____________										
+		|	     |						  ______________
+		|	     |					         |		|
+		|	     |						 |		|
+		|   Client   |      1. Contract or Service fees		 |		|
+		|	     |------------------------------------------>|   Provider   |
+		|	     |						 |		|
+		|____________|						 |		|
+									 |______________|
+										|
+							 ____________		|
+		 _____________				|	     |		|
+		|	      |				|	     |		|
+		|	      |				|	     |		|
+		|	      |	   3. Authenticated	|   Issuer   |<----------
+		|  Requester  |        Credentials	|	     |
+		|	      |<------------------------|	     |  2. Unauthenticated Credential Callback
+		|	      |				|____________|
+		|_____________|
+
+```
+
+During issuance, the credentials should be blinded to ensure future unlinkability during the redemption phase.
+
+Discovery of assets-to-credentials announced by the Issuer and consumed by the Client should be defined in
+its own document.
+
+A privacy-preserving communication protocol between Client and Issuer for credentials issuance should
+be defined in its own document.
+
+## Redemption
+
+The redemption protocol is an identification protocol wherein the Client presents authenticated credentials
+to the Provider to redeem the acceptance of a contract. The credentials should be unblinded before to be
+presented to the Provider. The quantity of credentials attached to the contract request should
+satisfy the contract liquidity units as enforced by the Contract Provider. Usually, the rate of credentials
+to contract/service unit announced by the Provider should have been discovered by the Committer in a non-defined
+protocol phase.
+
+The protocol works as in the figure below:
+
+```
+					1. Authenticated Unblinded Credential
+	 ____________			      		    +		      		      ______________
+	|	     |				Contract/Service Request    		     |		    |
+	|	     |---------------------------------------------------------------------->|		    |
+	|	     |									     |		    |
+	|   Client   |									     |   Provider   |
+	|	     |        2. Contract/Service Acceptance OR Contract/Service Reject	     |		    |
+	|	     |<----------------------------------------------------------------------|		    |
+	|____________|									     |______________|
+												    |
+												    | 3. Contract/Service
+												    | 	    Execution
+												    |------------------------>
+```
+
+Depending on the contract execution outcome (success or failure), an additional fee can be paid by the Client
+to the Provider. This fee should be determined in function of market forces constraining the type of contract/service
+executed.
+
+### Redemption Protocol Extensibility
+
+The Staking Credentials and redemption protocol are both intended to be receptive to extensions that expand
+the current set of functionalities through new types or modes of Bitcoin financial contracts covered. Among them,
+long-term held contracts based on a timelock exceeding the usual requirements of Lightning payments.
+Another type of flow is the correction of the opening asymmetries in multi-party Lightning channels funding transactions.
+
+## Deployment Considerations
+
+### Lightning HTLC routing
+
+In this model, a HTLC forwarder (Committer) send an off-chain payment and a list of blinded credentials to a Routing hop (Issuer).
+The Routing hop counter-sign the blinded credentials and send them back to the HTLC forwarder (Credentials Authentication phase).
+The HTLC forwarder (Client) unblinds the credential, attach a HTLC and send the request for acceptance to the Routing hop (Contract
+Provider). If the quantity of credentials attached satisfies the Routing hop, the HTLC is accepted and forward to the next hop
+in the payment path.
+
+This model is shown below:
+
+```
+		 _____________	   1. Off-chain payment + Blinded Credentials	      ____________
+		|	      |----------------------------------------------------->|		  |
+		|	      |							     |		  |
+		|     HTLC    |	      2. Counter-signed Blinded Credentials 	     |  Routing   |  4. HTLC relay
+		|  forwarder  |<-----------------------------------------------------|    hop     |--------------->
+		|	      |							     |		  |
+		|	      |	     3. Unblinded Credentials + HTLC forward 	     |		  |
+		|_____________|----------------------------------------------------->|____________|
+
+```
+
+
+## Security Considerations
+
+The major security risk from an Issuer perspective is the double-spend of the credentials by a Client. In
+the context of the Staking Credential architecture, a double-spend is the replay of credential for multiple
+Contract requests, therefore provoking a plain timevalue loss if contract fees are not paid by the Client for
+each Contract request instance.
+
+The Issuer should keep a private database to log every credential covering a Contract request. This private
+database should be accessible by the Contract Provider to validate all the credentials presented for a Contract
+request.
+
+A major security risk from a Committer perspective is the tampering of the credentials during transport between
+the Committer and Issuers hosts. Credentials could be hijacked based on credentials volume traffic monitoring. As
+such credentials transport should be authenticated and the packets obfuscate against network-level inspection.
+
+## References
+
+- [Privacy Pass Architecture](https://www.ietf.org/archive/id/draft-ietf-privacypass-architecture-09.html)
+- [Mitigating Channel Jamming with Stakes Certificates](https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-November/002884.html)
+- [Solving channel jamming issue of the lightning network](https://jamming-dev.github.io/book/about.html)
+- [Discreet Log Contracts specification](https://github.com/discreetlogcontracts/dlcspecs/)
+- [Lightning interactive transaction construction protocol](https://github.com/lightning/bolts/pull/851)
+
+## Copyright
+
+This document is placed in the public domain.
--- a/61-credentials-issuance.md
+++ b/61-credentials-issuance.md
@@ -1,0 +1,114 @@
+# Credential Issuance
+
+This document specifies the Credentials Issuance protocol inside the Staking Credentials framework.
+
+The messages data format, validation algorithms and implementation considerations are laid out.
+
+This draft is version 0.0.1 of the Credential Issuance protocol.
+
+## Credentials Issuance
+
+The Requester is sending a proof of scarce assets to the Issuer.
+
+The Issuer is sending back an authentication of the scarce assets to the Requester.
+
+	+-------+						  +-------+
+	|	|						  |	  |
+	|	|						  |	  |
+	|	|--(1)--- request_credentials_authentication ---->|	  |
+	|   A   |						  |   B	  |
+	|	|						  |	  |
+	|	|<-(2)--- reply_credentials_authentication -------|	  |
+	|	|						  |	  |
+	|	|						  |	  |
+	+-------+						  +-------+
+
+
+### The `request_credentials_authentication` message
+
+This message contains a proof of scarce asset for the credentials and a set of unsigned blinded credentials.
+
+1. type: 37560 (`request_credentials_authentication`)
+2. data:
+    * [`u32`: `session_id`]
+    * [`point`: `issuance_pubkey`]
+    * [`assetlen*byte`: `scarce_assets`]
+    * [`credentiallen*byte`: `blinded_credentials`]
+
+#### Requirements
+
+The sender:
+  - MUST set `session_id` to a pseudo-randomly generated 4-byte value.
+  - MUST set `issuance_pubkey` to a public key discovered from a `credential_policy` issued by the receiver.
+  - MUST NOT send `scarce_assets` if they have not been announced by a previous `credential_policy` issued by this node.
+  - MUST NOT send `blinded_credentials` of a format which not been previously announced by a `credential_policy` issued by this node.
+  - MUST set `blinded_credentials` to less than `max_onion_size`.
+
+The receiver:
+   - if `issuance_pubkey` has not been announced or has been rotated:
+      MUST reject this authentication request.
+  - if `scarce_asset` is not supported:
+    - MUST reject this authentication request.
+  - if `blinded_credentials` format is not supported:
+    - MUST reject this authentication request.
+  - if `blinded_credentials` size is equal or superior to `max_onion_size`:
+    - MUST reject this authentication request.
+  - if the `scarce_asset` amount is not covering the quantity of `blinded_credentials` requested for signature as announced by `credential_policy`:
+    - MUST reject this authentication request.
+
+#### Rationale
+
+The session id purpose is to enable concurrent authentication sessions where multiple set of credentials can be requested for
+authentication by the same Requester, and where the order of reception of concurrent request/reply is not guaranteed by the
+transport mechanism.
+
+The Issuer should be able to perform independent validation of a scarce asset to counter-sign the blinded credentials, and therefore
+apply a risk management policy.
+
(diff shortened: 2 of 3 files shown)


www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+to the reputation acquisition cost, a proportion sould be maintained between the credentials acquisition
+cost (either from upfront fees or honored routing fees) and the channel liquidity lockup rights.
+
+To preserve the confidentiality of the HTLC senders (or payers), the credentials should be anonymized,
+a routing hop should not be able to link between a credential issuance at dissemination and its usage
+at the reception of a HTLC forward request. A cryposystem satisfying those requirements could be to
+leverage EC [blinded signatures](https://sceweb.sce.uhcl.edu/yang/teaching/csci5234WebSecurityFall2011/Chaum-blind-signatures.PDF).
+
+Additionally, this proposal is extending the `channel_update` message to announce routing fees paid on the
+CLTV duration, therefore allowing compensation of the routing hops for the transport of long-term held
```
joostjager, line 87: Hold invoice may have a longer cltv, but I don't think that that is a suitable way to recognize and pay for them. Also quick-resolving htlcs on longer routes may have high cltv deltas. What makes hold invoices costly is their actual hold time.

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+In this diagram:
+1. Caroll settles the HTLC back to Bob with either a success or a failure.
+2a. If it's a success, Bob sign the blinded backward credentials stored during the previous phase.
+3a. If it's a failure, Bob slashes the forward credentials. There is no backward credentials return back to Alice.
+
+To prevent replay usage of slashed forward credentials or previously used forward credentials, the signature should be
+logged by the routing hop in an [accumulator data structure](https://link.springer.com/content/pdf/10.1007/3-540-48285-7_24.pdf),
+allowing efficient test membership. Merkle tree can constitute such accumulator. To limit the growing size of this accumulator,
```
joostjager, line 253: I believe this is a lossy data structure. Does that create issues with false positives or negatives?

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+	|	  |				            |	      |		  HTLC settlement	       |	  |
+	|  Alice  |				            |   Bob   |<---------------------------------------|  Caroll  |
+	|	  |         3b. HTLC 		            |	      |				               |	  |
+	|_________|<----------------------------------------|_________|					       |__________|
+
+```
+
+In this diagram:
+1. Caroll settles the HTLC back to Bob with either a success or a failure.
+2a. If it's a success, Bob sign the blinded backward credentials stored during the previous phase.
```
joostjager, line 248: What happens if upstream nodes along the path corrupt the signed credentials somehow? They are part of the `update_fulfill_htlc` message, right?

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+		       onions(credentials)
+		    4.
+
+```
+
+In this diagram:
+1. Alice discovers Caroll's `routing_policy` gossip.
+2. Alice fetches an offer from Caroll to pay the `credential_to_liquidity_unit` announced.
+3. Alice sends a HTLC to Caroll, the onion payload contains unsigned credentials.
+4. Caroll receives the HTLC, counter-signed the credentials and send them back to Alice by onions.
```
joostjager, line 139: Why are these send back via a separate onion message and not in the `update_fulfill_htlc` as described for ongoing signing below?

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+1. Caroll settles the HTLC back to Bob with either a success or a failure.
+2a. If it's a success, Bob sign the blinded backward credentials stored during the previous phase.
+3a. If it's a failure, Bob slashes the forward credentials. There is no backward credentials return back to Alice.
+
+To prevent replay usage of slashed forward credentials or previously used forward credentials, the signature should be
+logged by the routing hop in an [accumulator data structure](https://link.springer.com/content/pdf/10.1007/3-540-48285-7_24.pdf),
+allowing efficient test membership. Merkle tree can constitute such accumulator. To limit the growing size of this accumulator,
+all credentials issued can be expired according to a `routing_gossip::credential_expiration_height` field.
+
+Once Alice receives the countersigned and still blinded backward credentials, she can unblind them and rely on them
```
joostjager, line 256: Would it be possible for routing nodes to sign with different keys to do some traffic correlation? Or can Alice just verify the signatures and ensure herself that they are all signed with the routing node key?

joostjager (commented):
Thanks for making a start here with anti-jamming changes!

The paper linked in your rust-lightning pr about [privacy pass](https://www.petsymposium.org/2018/files/papers/issue3/popets-2018-0026.pdf) I also found useful for background info.

www-reputation-credentials-protocol.md
ariard: I think effectively we don't have a suitable, "quantifiable" way to dissociate between short-term held HTLCs and long-term held HTLCs, other than an arbitrary protocol/implementation-defined value. On the LDK-side, we reject a any outgoing HTLC with a CLTV superior to 2016 blocks from our tip. I don't think it's even opportune to introduce a "discrete" approach but rather design an anti-jamming solution covering both "continuously". 

Making the HTLC upfront fees/reputation cost actually paid on the "actual hold time" would be more valuable. A HTLC sender would only have to cover cost for effective duration, not worst-case CLTV scenario. I think this could be possible if `update_fulfill_htlc` paid back routing fee scaled on a new `fee_base_block` to encompass the time dimension.

www-reputation-credentials-protocol.md
ariard: I think this is correct, if we don't make the merkle branch to the root deterministic on the signature itself. E.g the X first bit of the signature could give you the left-right of merkle path from the root to the leaf. More efficient accumulators, with less studied assumptions, could be used as it can be upgraded in a non-coordinated fashion, this is purely internally to the node.

www-reputation-credentials-protocol.md
ariard: Yes I think the description should be fixed, the signed credentials should be wrapped into an onion. The onion can be attached to the `update_fulfill_htlc` or they could send back with a reply_path as the onion message proposal introduces. This reply_path could have been exchanged during the dissemination phase. I still have to catchup with the latest state of onion messages to see what works the best.

www-reputation-credentials-protocol.md
ariard: The quantity of signed credentials, and as such the size of the wrapping onion, could be a leak of Alice expected future usage of the payment paths. Therefore, dissociating the backward communication channel from the payment routing hops could improve privacy. The signed credentials could be fan-out in multiple 32k-byte onions.

www-reputation-credentials-protocol.md
ariard: I think this is a really good point and Alice should verify the signatures are all signed with the known routing node key. In the current PoC implem, the credential key is not tied to the node one though indeed to avoid this correlation issue this should. The key rotation should be visible by the HTLC sender.

www-reputation-credentials-protocol.md
joostjager: Yes, or in the credentials case let the routing node sign a number of credentials that varies based on the actual hold time? If the sender causes a lot of holding, they will keep sending out more credentials than they get back, bringing their credentials balance down. The onion payload is limited in size though, so it seems that it can't be very granular.

www-reputation-credentials-protocol.md
joostjager: This is indeed probably something that can be solved in the implementation phase. With the expiry, the set cannot grow indefinitely. Is that expiry actually encoded inside the 32 byte credential?

www-reputation-credentials-protocol.md
joostjager: Attached to `update_fulfill_htlc` as an onion, I'd think that nodes can still delete it? And stand-alone onion messages may not arrive either. If there is a way to somehow atomically deliver upstream together with the preimage, that seems to be better. Or use something like https://github.com/lightning/bolts/pull/1044 so that the sender can penalize the node that deleted the in-flight credentials.

www-reputation-credentials-protocol.md
joostjager: Is it necessary to do key rotation at all?

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+
+In contract, another type of reputation-based solution can be introduced by formalizing a new assumption
+"Reputation incentives HTLC senders to behave-well, if there is an adequate proportion between the
+acquisition cost and the amount of resources credited in function of said reputation". This assumption is
+in fact analoguous to the one underpining routing scoring algorithms incentiving the routing hops reliability.
+A reputation system attached to the HTLC forward, not only can protect against malicious jamming entity but
+also can offer the traffic shapping tooling for node operators to improve the quality of their inbound HTLC
+across time.
+
+# Proposed solution
```
joostjager, line 39: I think it may be useful too - if only as a thought experiment - to envision the simplest possible version of the scheme. So maybe just a single credential per htlc, no introduction of liquidity units and avoiding the use of onion messages?

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+there is an adequate compensation. Indeed, the routing fees are only paid in case of success.
+
+# Solution introduction
+
+This counterparty risk can be solved straightforwardly by the introduction of a premimum fee carried on by
+the HTLC sender. This jamming solution known as ["upfront fees"](https://github.com/t-bast/lightning-docs/blob/master/spam-prevention.md)
+in the Lightning community, aims to a fair distribution of the HTLC forward risk among all the chain of
+contracts counterparties. While there is a concern this risk distribution doesn't capture well the
+shenanigans of a HTLC forward (e.g offline routing hop, spontaneously congestioned channel), there is a
+major drawback as it introduces a permanent overhead fee for the network participants.
```
fiatjaf, line 29: From what I understand these reputation tokens are being sent as a form of upfront/unconditional monetary payment from the sender to each node in a route. Is this right?

Considering that, aside from these concerns you cite here, is there anything that makes paying these upfront fees with sats inherently harder or worse than paying with reputation tokens?

For example, in a route `A->B->C->D` is there something that prevents A from sending some extra msats unconditionally to B and C?

(I couldn't find anything that makes that a no-go on https://github.com/t-bast/lightning-docs/blob/master/spam-prevention.md)

www-reputation-credentials-protocol.md
ariard: If you assume there is a satoshi-cost behind each credential unit, I think you should scale the quantity of credentials requested on the outbound CLTV duration as expressed in the Script during the forward phase. There is no a priori knowledge of the HTLC actual hold time.

Then at settlement, if the observed actual hold time is above some average/median, the routing node can return less credentials, leading the sender to pre-pay for more credentials if it would like to keep doing a lot of holding.

I think this alternative also work, rather than paying routing fee scaled on the time dimension, however I think this more risky for the routing node, as the actual holding time of a HTLC might be far above your median. Requesting credentials to cover the worst-scenario is zero-risk for the routing node, at the trade-off of more expensive for the HTLC senders.

About the granularity, if the onion communication channel are detached from the payment paths, you could send multiple onions, overreaching beyond the 32kb limit. The granularity wished between msat/sat and credentials is a good question.

www-reputation-credentials-protocol.md
ariard: See the new gossip in BOLT7 `routing_policy` with the field `credential_expiration_height`. Encoding directly the height in the credentials could be a correlation vector, I believe. Failure to respect its own routing policy by a hop should be likely penalized by routing algorithms.

www-reputation-credentials-protocol.md
ariard: That's a good question, I prefer to say yes e.g if you would like to refresh the shares of your musig2 aggregated key. Note, I don't think we have a protocol-mechanism to inherit a LN node_id pubkey to a new one, with all its known `channel_announcement`s and as such conserve the reputation staked in routing algorithms.

www-reputation-credentials-protocol.md
ariard: That's a good point, stand-alone onion messages are not reliable. I think effectively you would like the credentials to be part of the hmac chain as proposed by #1044  and the sender should be able to detect tampering. Especially, if the credentials have a satoshi-cost, you open the way to damage by intermediary nodes.

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+
+If the HTLC is a failure, the reputation credentials are slashed. No new credentials are issued back
+to the HTLC sender as a punishment. To limit reputation whitewashing, where the jamming damage is superior
+to the reputation acquisition cost, a proportion sould be maintained between the credentials acquisition
+cost (either from upfront fees or honored routing fees) and the channel liquidity lockup rights.
+
+To preserve the confidentiality of the HTLC senders (or payers), the credentials should be anonymized,
+a routing hop should not be able to link between a credential issuance at dissemination and its usage
+at the reception of a HTLC forward request. A cryposystem satisfying those requirements could be to
+leverage EC [blinded signatures](https://sceweb.sce.uhcl.edu/yang/teaching/csci5234WebSecurityFall2011/Chaum-blind-signatures.PDF).
```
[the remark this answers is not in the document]
ariard: See the answer on this ML post: https://lists.linuxfoundation.org/pipermail/lightning-dev/2022-November/003756.html

For now, just thinking unlinkability between the blinded credentials and unblinded ones/signatures. In function of if we would like to have secondary transfers of the credentials, or the sender committing to the credentials, we study more advanced cryptosystem. From a UX viewpoint, it might be good to defer the credentials harvesting to the LSPs.

www-reputation-credentials-protocol.md
ariard: I think this is a good point, to work out the simplest possible version of the scheme.

A single credential per HTLC with a fixed cost-to-liquidity (10 sat of pre-paid for 1 BTC of liquidity for 2016 blocks). I wonder if in fact you really need liquidity unit if we can have global agreement on the credential granularity-level.

www-reputation-credentials-protocol.md
ariard: This is correct these reputation tokens are sent as a form of upfront/unconditional monetary payments. As noted on the mailing list, they might in fact belong more to the upfront/unconditional family of schemes rather than "pure reputation".

The major difference, minimizing the permanent overhead fee, is to stake the credentials across the HTLC sends. In the optimistic case (no HTLC failure), you only pay for few credentials once, then you should get more of them as your HTLC sends succeed, rather than spending a small upfront fee each time.

There is nothing that prevents A from sending extra msats unconditionally to B and C. The reason of the dissemination phase is for privacy reasons, otherwise you're disclosing to the routing hop this is the first time you're requesting a HTLC forward.

www-reputation-credentials-protocol.md
joostjager: Yes, the difference in hold time between a regular payment and a hold payment is massive. Hard to bridge that gap using credentials without requiring more trust.

Routing fee scaling on a time dimension - I think it only works for the success case? Seems we also need something for the slow-failed case.

www-reputation-credentials-protocol.md
joostjager: So `credential_expiration_height` is a delta height that both sender and router use to expire tokens? Can the router do this when it gets presented the unblinded token, or how does the router know when the token was produced?

www-reputation-credentials-protocol.md
joostjager: Nice to have another reason to work on #1044 😉

www-reputation-credentials-protocol.md
joostjager: Sounds like rotation is something for later reconsideration, with node keys currently being static.

www-reputation-credentials-protocol.md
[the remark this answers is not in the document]
joostjager: Ok. At least good to know that with VOPRF we have one option that seems reasonably well fitting.

www-reputation-credentials-protocol.md
joostjager: In a way, things look simpler without a dissemination phase. For each hop you either attach a valid token, or pay an unconditional upfront fee. Either way, you'll get a new free token back for the next payment attempt - assuming no misbehavior.

On the other hand this does require different changes to allow these unconditional fees.

joostjager: All though yes, this is probably the same as always requiring an upfront fee and lowering the routing fee slightly. In both cases, you pay a little bit for a failure and more for a success.

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+The exact ratio between a quantity of credentials and the channel liquidity lockup rights is
+a routing hop policy decision and can be adjusted in real-time in function of channel congestion
+rate and other factors. This routing policy can be announced towards the LN network through
+the public gossips mechanism.
+
+The credentials are disseminated from the routing hops towards the HTLC senders through onion
+communication channels. The credentials acquisition method and cost are defined by the routing
+hops. For this version of the protocol, we propose upfront fees as a credential acquisition
+method. Payment HTLC are sent to the routing hop and credentials are returned back to the sender
+with a onion reply path. The credentials are counter-signed by the issuer to authenticate their
```
naumenkogs, line 54: >The credentials are counter-signed by the issuer to authenticate their origin.

First I thought that a secret (without a signature) should be sufficient. But then I realized, because of onions, you get a signature "for free", right?

www-reputation-credentials-protocol.md
naumenkogs: This would make up a good section for this document :)

www-reputation-credentials-protocol.md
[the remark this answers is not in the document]
naumenkogs: I see how it can be easily done when credentials are mixed among a set of users (maybe private mixing is hard, but probably not — it's like Tor 2.0 stuff).
But this would be unfortunate, it likely converges to CoinJoin anonymity set stuff, which, again, relies on reputation... Seems like a recursive problem, but maybe I'm missing the fact that we can solve both at the same time somehow?

Now, fancier schemes, I still don't have a good sense of how they should be applied to our problem.
When I think about this, my first question is "what if there was only one client who received the credential, is there a way to protect them?"

1. This is probably solvable either with mixing (see above), or a trusted third party server (like a mix coordinator, but more flexible — and hopefully provably private towards everything) like a watchtower.
2. Maybe this is fine — we assume privacy only if a routing node served 100 clients a day or something. Then, maybe some minimal technology would be sufficient to help those 100 to make sure they get mixed among each other. Possibly something a routing node publishes every day, or maybe even something about burning the collateral of a routing node if cheating happens :P

www-reputation-credentials-protocol.md
naumenkogs: Another option is to make the payload same size for any number of creds somehow... Maybe a static big payload (part of which could be empty), or some fancy compressed data structure with the size independent from the number of creds.

www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+	|_________| \													|_________|
+		    Dave: 250 credentials
+							 ________
+							|	 |  channel_update::fee_base_msat: 10
+							|  Dave  |
+							|________|  routing_policy::credential_to_liquidity_unit: 1
+
+```
+
+In this diagram, Alice would like to send a 1 BTC HTLC to Eve. Bob, Caroll, Dave are all intermediary valid
```
naumenkogs, line 173: nit: eve is usually evil, so i was actually seeking for the attack description somewhere :)

naumenkogs:
Good stuff overall. I kinda feel this document is too in-depth for now, and should be simplified to:
1. Focus on the simplest version first.
2. Focus attention on high-level abstract justifications instead of debatable ideas like key rotation

www-reputation-credentials-protocol.md
ariard: For the slow-failed case, under the credentials paradigm, a routing node would have to scale the quantity of credentials requested to accept the HTLC forward. E.g, if the HTLC sender requests a 2-month outgoing CLTV, a routing node policy should be to have previously satisfied enough 4*2016 block period (2016 max standard allowed outgoing CLTV). The HTLC sender should have either "harvest" the credentials from previous satisfied routing success, or pre-paid 4 block period.

www-reputation-credentials-protocol.md
ariard: Yes, both the sender and the router should consider the credentials as expired once the `credential_expiration_height` is reached. Without assigning signing key per expiration period, I don't think there is a way for the router to dissociate credentials across expiration periods. However, rotating signing key might be undesirable as raised in another comment. Another direction could be to request the senders to "tag" all the credentials starting within said expiration period. For global visibility, the tag should be part of the `routing_policy` (without introducing traffic correlation).  

For now, I think the only clear rational to introduce a `credential_expiration_height` is effectively to bound the anti-replay accumulator size. Though key rotation, upgrading the credential format, or updating core fields of the routing policy could be justification. Sounds more for later consideration.

www-reputation-credentials-protocol.md
ariard: I think there is two constraint here, the bandwidth consumption if we use the max onion payload size, whatever the number of credentials delivered. Another constraint, the information efficiency of the credentials. Introducing multiple ranges of credentials could be a correlation traffic vector.

www-reputation-credentials-protocol.md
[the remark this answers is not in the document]
ariard: If the credentials do not bind to the sender key material, it should be pretty flexible to have them exchanged in a secondary-market, or even harvested by a LSP. 

I think this is a good point, if the credentials do not bind to the sender key material, or any secret with strict cost to share, they can be exchanged in a cleartext secondary-market. However a cleartext secondary-market would allow the initial credential issuer to collude with the routing hop to deanonymize the "second-hand" HTLC senders. Therefore, if credential transfer is allowed, they should be exchanged in a privacy-anonymizing way mixnet-style. I think this is good to be aware of but a subject for future considerations.

www-reputation-credentials-protocol.md
ariard: The main motivation of introducing "reputational" credentials is to stake/accumulate them across time, therefore making your future HTLC sends cheaper, optismistically without introducing traffic correlation between your HTLC forward requests. Instead of using a credential, a discount in function of actual hold time could be applied on the success fee to cover next payment's upfront fee, though I think you're limited to accumulate discounts across a period. For long-term held payments, this is an issue as the upfront fees should probably scale up to cover the corresponding routing hop risk.

There is also another advantage of introducing credentials, if they can be transferred, a LSP could pay for the upfront fees/unconditional fees of its spokes, therefore covering the jamming risk they represent for routing hops in the network.

www-reputation-credentials-protocol.md
ariard: The rational of the signature, it allows credentials authentication by the routing hop, otherwise one is free to forge credential to cover HTLC routing requests. The signature means "This credential has been endorsed by myself during a previous round in its blinded version". I can see how using the phrasing "authenticate their origin" is confusing, should say more to endorse their "endorsement".

www-reputation-credentials-protocol.md
ariard: Run out of my stack of English names starting by E. Eliot, Ethan, Eric, yeah...okay.

ariard:
Thanks for the feedback so far. Working on a simplified protocol description: "one credential per HTLC", without the complexity of the onions or accounting for accounting HTLC amount/time.

The main motivation of this "staking"/"reputational" credentials is to save on unconditional fees paid by HTLC senders. They benefit of their past HTLC routing success in term of more credentials allocated to them, and as such minimize the overhead cost of their future HTLC sends, or allow them to lock liquidity for longer periods. From a routing node viewpoint, a 0-risk HTLC forwarding acceptance can be maintained, by requesting strict binding between credentials acquisition cost and channel liquidity routed.

www-reputation-credentials-protocol.md
naumenkogs: But how this is different from issuing a secret only the issuer and the receiver knows?

www-reputation-credentials-protocol.md
joostjager: Can the router still verify the tag after unblinding? I'd think it has become unreadable for the router when it is presented as a forward credential?

ariard: The HTLC sender generates a credentials, concatenate the current tag for the period, blind and send it to the routing hop for signature. The counter-signed blinded credential is sent back to the HTLC sender, which can unblind both the credential/signature. I think the tag should still bind.

www-reputation-credentials-protocol.md
ariard: I understand, if the issuer can map the secret to a given receiver, you loss privacy of the credentials. Blinded signature is a simple scheme providing unlinkability. We might have other cryptosystems to consider.

ariard:
PR has been updated at c9f7b71 with: https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-January/003822.html
Oldest proposal should be available under `history/`.

history/www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+# Solution introduction
+
+This counterparty risk can be solved straightforwardly by the introduction of a premimum fee carried on by
+the HTLC sender. This jamming solution known as ["upfront fees"](https://github.com/t-bast/lightning-docs/blob/master/spam-prevention.md)
+in the Lightning community, aims to a fair distribution of the HTLC forward risk among all the chain of
+contracts counterparties. While there is a concern this risk distribution doesn't capture well the
+shenanigans of a HTLC forward (e.g offline routing hop, spontaneously congestioned channel), there is a
+major drawback as it introduces a permanent overhead fee for the network participants.
+
+In contract, another type of reputation-based solution can be introduced by formalizing a new assumption
```
saubyk, line 31: nit: do you mean 'In contrast'?

history/www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+"Reputation incentives HTLC senders to behave-well, if there is an adequate proportion between the
+acquisition cost and the amount of resources credited in function of said reputation". This assumption is
+in fact analoguous to the one underpining routing scoring algorithms incentiving the routing hops reliability.
+A reputation system attached to the HTLC forward, not only can protect against malicious jamming entity but
+also can offer the traffic shapping tooling for node operators to improve the quality of their inbound HTLC
```
saubyk, line 36: nit: traffic shaping tool

history/www-reputation-credentials-protocol.md
```diff
@@ -0,0 +1,265 @@
...
+In this diagram:
+1. Alice sends a HTLC+credentials to Bob.
+2. Bob extracts the credentials, verifies the signature and validate the credentials against his own `routing_policy`.
+3a. The routing checks are successful, Bob forwards the HTLC along the Bob-Caroll link.
+4a. Bob stores the backward credentials into the local state until the HTLC settlement.
+3b. The routing checks are unsuccessful, Bob rejects the HTLC.
+
+The credentials are wrapped inside the HTLC onion `credentials_payload`. There are two types of credentials: forward/backward.
+The forward credentials are covering the ongoing HTLC forward, they should have been counter-signed by the target routing hop
+during the dissemination phase or a previous HTLC forward/settlement pahse with the same target routing hop. The backward
```
saubyk, line 221: nit: HTLC forward/settlement phase

ariard:
Thanks for the review! Comments should have been fixed at aa8ac0e.

ariard:
Updated at 36553d1e with only the architecture / issuance / redemption flows documented. Working on a Rust implementation.

As a reminder Staking Credentials is an anonymous user-authentication mechanism based on “proof of scarce assets” challenges (e.g Bitcoin transaction or Lightning invoice, or successful previous redeemed service). The primary goal is to use those anti-DoS tokens to mitigate channel jamming though additional use-cases could be to protect LSPs service APIs.

Moving forward, I think I’ll split the documentation in many components:
- the token issuance / redemption flows as BOLT recommendations or extensions
- the blinding cryptography as BIPs
- the communication channels as BOLTs or NIPs (e.g tokens-over-websocket)

For reference, the Privacy Pass documentation (architecture, key discovery, rate-limitation, architecture) is available here: https://datatracker.ietf.org/wg/privacypass/documents/ I’ll look if some IETF draft can be reused as it is in the Bitcoin context, or if we have to submit modifications for the Bitcoin context.

ariard:
There is a working v0.0.1ish proof-of-concept of the Staking Credentials implemented here: https://github.com/civkit/civkit-node All the p2p msgs staking credentials code is implemented as a separate rust library to be re-used in Lightning and other Bitcoin systems. Next step will be probably to add the blinding crypto itself.

ariard:
@t-bast can you re-open it ? There has been an implementation by my care here during the past year:
https://github.com/civkit/staking-credentials

While it’s obviously relying on a lot of cryptography and most of the lightning protocol dev usually participating in this specification effort are usually incompetent to review cryptography, I can understand it takes time to make progess.

Thanks you.

ariard:
Thank you.

Yeah all those ideas we have worked with Gleb they have been inspired by Privacy Pass. And if my memory is correct, it took 5 years and more from the original idea to get something supported in a browser as an experiment. And I think Safari / Chrome (or what is another browser ?) have their own non-compatible versions. The Privacy Pass spec only got a non-draft IETF document in June of this year (RFC 9576).

PR #1052, opened 2023-01-18
draft: Upfront Fees to Mitigate Channel Jamming

carlaKC:
This PR is an early stage draft of the upfront fees portion of the jamming mitigations proposed in [Unjamming Lightning: A Systemic Approach](https://eprint.iacr.org/2022/1454.pdf).

It describes the _minimal_ set of changes for the _simplest_ iteration of upfront fees: 
* Channels advertise an upfront fee policy, which is charged on the outgoing edge for any routed htlcs
* Upfront fees are simply assigned to the `to_remote` balance, with no on-chain output or secret that proves the htlc was _actually_ forwarded
* Recipients advertise upfront fees in bolt 11 invoices, and pad this value to hide their position in the route (route hints TBD)
* The final hop's upfront fees contribute to the `total_amt` paid for a *htlc set*, so senders don't need to pay for recipient privacy

It is not ready for rigorous review yet, but has been opened here to assist discussion of various approaches to mitigating jamming in the lightning network. 

Proof of concept code for LND is available [here](https://github.com/lightningnetwork/lnd/pull/7339).

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -990,6 +990,12 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
    * [`u32`:`cltv_expiry`]
    * [`1366*byte`:`onion_routing_packet`]
 
+1. `tlv_stream`: `update_add_htlc_tlvs`
+2. types:
+    1. type: 2 (`upfront_fee_msat`)
+    2. data:
+        * [`tu64`:`upfront_fee_msat`]
+     
 #### Requirements
 
 A sending node:
@@ -1002,6 +1008,12 @@ A sending node:
     node is the funder:
       - MUST be able to additionally pay for `to_local_anchor` and 
       `to_remote_anchor` above its reserve.
+    - if `upfront_fee_msat` is included in the HTLC
+      - MUST be able to additionally pay for the `upfront_fee_msat` above its
+        reserve.
+      - MUST push the `upfront_fee_msat` amount to the remote party's balance,
+        by subtracting `upfront_fee_msat` from its expected `to_local` balance
+        and adding  `upfront_fee_msat` to its expected `to_remote` balance.
     - SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment
     transaction, its remaining balance doesn't allow it to pay the commitment
     transaction fee when receiving or sending a future additional non-dust HTLC
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1036,6 +1036,13 @@ reasonable time.
 The upfront fee amount was below that required by the outgoing channel from the
 processing node.
 
+1. type: 27 (`final_incorrect_upfront_fee`)
+2. data: 
+   * [`u64`:`upfront_fee`]
+
+The final hop's upfront fee in the HTLC doesn't match the value in the onion, 
+or does not meet the receiving node's expected upfront fee.
+
 ### Requirements
 
 An _erring node_:
@@ -1131,6 +1138,10 @@ An _intermediate hop_ MUST NOT, but the _final node_:
     - MUST return a `final_incorrect_htlc_amount` error.
   - if it returns a `channel_update`:
     - MUST set `short_channel_id` to the `short_channel_id` used by the incoming onion.
+  - if the `upfront_fee_msat` from the final node's HTLC is below `upfront_fee_to_forward`:
+    - MUST return `final_incorrect_upfront_fee` error.
+  - if the `upfront_fee_msat` does NOT pay sufficient fees: 
+    - MUST return `final_incorrect_upfront_fee` error.
 
 ### Rationale
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1027,6 +1027,15 @@ the decrypted byte stream.
 The complete amount of the multi-part payment was not received within a
 reasonable time.
 
+1. type: UPDATE|25 (`upfront_fee_insufficient`)
+2. data:
+   * [`u64`:`upfront_fee`]
+   * [`u16`:`len`]
+   * [`len*byte`:`channel_update`]
+
+The upfront fee amount was below that required by the outgoing channel from the
+processing node.
+
 ### Requirements
 
 An _erring node_:
@@ -1089,6 +1098,10 @@ A _forwarding node_ MAY, but a _final node_ MUST NOT:
   - if the channel is disabled:
     - report the current channel setting for the outgoing channel.
     - return a `channel_disabled` error.
+  - if the HTLC does NOT pay sufficient upfront fee:
+    - report the incoming upfront fee amounts and the current channel setting 
+      for the outgoing chanel.
+    - return a `upfront_fee_insufficient` error.
 
 An _intermediate hop_ MUST NOT, but the _final node_:
   - if the payment hash has already been paid:
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -201,6 +201,9 @@ This is formatted according to the Type-Length-Value format defined in [BOLT #1]
     1. type: 16 (`payment_metadata`)
     2. data:
         * [`...*byte`:`payment_metadata`]
+    1. type: 2 (`upfront_fee_to_forward`)
+    2. data:
+        * [`tu64`:`upfront_fee_to_forward`]
 
 `short_channel_id` is the ID of the outgoing channel used to route the
 message; the receiving peer should operate the other end of this channel.
@@ -227,11 +230,16 @@ The requirements ensure consistency in responding to an unexpected
 `outgoing_cltv_value`, whether it is the final node or not, to avoid
 leaking its position in the route.
 
+`upfront_fee_to_forward` is the upfront fee value, in millisatoshis, that 
+should be pushed to the next receiving peer specified within the routing
+information, or for the final destination.
+
 ### Requirements
 
 The writer:
   - For every node:
     - MUST include `amt_to_forward` and `outgoing_cltv_value`.
+    - MUST include `upfront_fee_to_forward` if every hop advertises `option_upfront_fee`, otherwise MUST NOT include `upfront_fee_to_forward`.
   - For every non-final node:
     - MUST include `short_channel_id`
     - MUST NOT include `payment_data`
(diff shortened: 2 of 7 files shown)


proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+    1. type: 1 (`upfront_fee_policy`)
+    2. data:
+        * [`tu32`:`upfront_fee_base_ppm`]
+        * [`tu32`:`upfront_fee_proportional_ppm`]
+
+```
+
+To save the network the bandwidth and storage required to transmit and save
+defaults, nodes that advertise `option_upfront_fee` in their node announcement
+should be assumed to have a base and proportional fee that is 1% of its success
```
joostjager, line 88: It's a magic value, but if it is just a hard-coded default to potentially save bytes, it hardly hurts. If it turns out that nobody wants 1% and every channel update contains explicit upfront fees, then the only cost is the code to support the not-used default.

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+case fees. Nodes also will not relay and senders will not utilize channels with
+a `channel_update` messages where upfront fees are > 10% of their success-case
```
joostjager, line 90: I am wondering if this does rule out certain use cases of upfront fees. As in very low success fees under the condition that you only request forwards that succeed. Otherwise you're penalized with a disproportionately high upfront fee.

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+equivalent. This eliminates the risk of nodes setting these fees so high that
+it becomes economically rational for forwarding nodes to fail payments for the
+sake of their upfront fees. 
+
+The upfront fee amount should simply be assigned to the receiving party's 
+balance when an incoming HTLC is added. These amounts are not expected to be
+enforceable on-chain (as they are likely to be dust), so there is no need to 
+include an output for them in the channel's commitment transactions.
```
joostjager, line 98: Suppose they wouldn't be dust, would you then create a separate output for it? I'd think that in that case it's still sufficient to just add to the receiver's balance?

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+1. type: 2 (`upfront_fee_to_forward`)
+    2. data:
+        * [`tu64`:`upfront_fee_to_forward`]
+```
+
+An upfront fee policy field is added to bolt-11 invoices to allow receiving 
+nodes to advertise the upfront fees that it will accept for the final hop 
+(since there is no outgoing link for the sender to obtain a policy from). 
+```
+`u` (17) `data_length` 26:
```
joostjager, line 129: For inbound fees, I also ran into the problem that multi-hop route hints aren't extensible: https://github.com/lightning/blips/pull/18/files#r988892046

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+- HTLC 1 arrives: amount = 15,000, upfront_fees=100
+- HTLC 1 fails with MPPTimeout
+- HTLC 2 arrives: amount = 15,000, upfront_fees=100
+- HTLC 3 arrives: amount = 14,700, upfront_fees=100
+
+The receiving node should reveal the preimage at this point, because it has
+received the total of 30,000 msat from the sender (29,700 paid via HTLCs in the
+set and 300 msat through upfront fees).
+
+Sending nodes can factor this upfront fee into the total amount they dispatch:
```
joostjager, line 153: One thing is that with htlcs in-flight, the sender doesn't know yet whether the htlc will reach the final destination or not. If it does, the upfront fee counts towards the total amount. If it doesn't they just pay intermediate nodes do not increase the total amount at the receiver. Maybe this complicates implementation.

04-onion-routing.md
```diff
@@ -1007,6 +1027,22 @@ the decrypted byte stream.
 The complete amount of the multi-part payment was not received within a
 reasonable time.
 
+1. type: UPDATE|25 (`upfront_fee_insufficient`)
+2. data:
+   * [`u64`:`upfront_fee`]
+   * [`u16`:`len`]
+   * [`len*byte`:`channel_update`]
```
joostjager, line 1034: We do have tlv failures now, so could make use of that.

02-peer-protocol.md
```diff
@@ -1002,6 +1008,12 @@ A sending node:
     node is the funder:
       - MUST be able to additionally pay for `to_local_anchor` and 
       `to_remote_anchor` above its reserve.
+    - if `upfront_fee_msat` is included in the HTLC
+      - MUST be able to additionally pay for the `upfront_fee_msat` above its
+        reserve.
+      - MUST push the `upfront_fee_msat` amount to the remote party's balance,
```
ariard, line 1014: If the sender is the channel initiation and as such encumbering the responsibility for fees, I don't know if you have too sum up the `upfront_fee_msat` to the offered `amount_msat` to determinate you're still in the bounds of the "fee spike buffer".

04-onion-routing.md
```diff
@@ -247,19 +255,23 @@ The writer:
...
   - MUST return an error if `amt_to_forward` or `outgoing_cltv_value` are not present.
+  - MUST return an error if `upfront_fee_msat` is present and `upfront_fee_to_forward` is not.
+  - MUST return an error if `upfront_fee_msat` is not present and `upfront_fee_to_forward` is.
   - if it is not the final node:
     - MUST return an error if:
       - `short_channel_id` is not present,
       - it cannot forward the HTLC to the peer indicated by the channel `short_channel_id`.
       - incoming `amount_msat` - `fee` < `amt_to_forward` (where `fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#htlc-fees))
       - `cltv_expiry` - `cltv_expiry_delta` < `outgoing_cltv_value`
+      - incoming `upfront_fee_msat` - `upfront_fee` < `upfront_fee_to_forward` (where `upfront_fee` is the advertised fee as described in [BOLT #7](07-routing-gossip.md#upfront-fees))
```
ariard, line 266: I think this is where the spec is unclear to me. 

`upfront_fee_msat` is unconditionally pushed on the `to_remote` balance from the viewpoint of the HTLC sender. At the same time, for this equality to hold incoming `upfront_fee_msat` must accumulate all the downstream `upfront_fee` to fulfil this equality. Otherwise you might have routing failures just due to hops setting their `upfront_fee_base_sat` and `upfront_fee_proportional_ppm` in an unequilibrated fashion.

E.g, you have Alice <-> Bob <-> Caroll <-> Dave. Let's say you have Bob requests 100 sat, Caroll requests 2000 sats and Dave requests 150 sats. The HTLC forward should fail at Bob, as 100 sat < 2000 sat, unless Alice sent an `upfront_fee_msat`= 2250 sats. However it also means Bob's balance is credited of 2250 sats, if understand the proposal correctly on this mechanism.

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+nodes' failed payments, which makes it difficult to identify. A solution that 
+addresses quick jamming therefore *must have a trivial impact on honest 
+traffic, but be sufficient to compensate the node(s) under attack*. A solution 
+that fails to do this will only magnify an attacker's ability to disrupt the 
+network, as they can leverage the jamming mitigation itself to further degrade 
+honest traffic.
+
+An upfront fee, paid regardless of the outcome of a payment attempt, is proposed
+to economically compensate nodes for providing traffic with access to liquidity
+and slots. Simulations in [2] show that a fee of as little as 1% of the success
```
ariard, line 69: I think the number holds as long as the simulation parameters hold. From my understanding of the paper model, some parameters are quite arbitrary (e.g the jam delay of 7 second), and beyond I think there is some assumption of periodicity of jamming and honest payment. Where the 1% unconditional fee doesn't make sense to me is in face of a jamming adversary perpetually succeeding to occupy the HTLC slots, in a way than never the success fees are paid. In that case, there is no equal compensation at all, I would say.

Additionally, I don't know about the methodological issue of substituting our own evaluation of the routing node liquidity opportunity cost, and if unconditional fees should not be floating to suit one's own evaluation of opportunity cost.

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+
+Sending nodes can factor this upfront fee into the total amount they dispatch:
+- Let a be the recipient's chosen `upfront_base_msat`.
+- Let b be the recipient's chosen `upfront_proportional_millionths`.
+- Let X be the total amount due to be paid. 
+- Let Y be the payment amount that the sender should dispatch.
+
+The sender can trivially solve for Y: 
+```
+X = Y + b + aY
```
saubyk, line 161: nit: shouldn't this equation be X = Y + a + bY?

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+2. types:
+    1. type: 2 (`upfront_fee_msat`)
+    2. data:
+        * [`tu64`:`upfront_fee_msat`]
+```
+
+The upfront fee that should be sent to the next peer in the route is provided
+in the onion's payload. As is currently the case with htlc forwarding amounts, 
+the difference between the incoming `upfront_fee_msat` amount in 
+`update_add_htlc_tlvs` and the outgoing `upfront_fee_to_forwad` is used to
```
GeorgeTsagk, line 116: nit: typo `upfront_fee_to_forwad` -> ``upfront_fee_to_forward`

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+validate that sufficient upfront fees are paid to the forwarding node. 
+```
+1. `tlv_stream`: `payload`
+1. type: 2 (`upfront_fee_to_forward`)
+    2. data:
+        * [`tu64`:`upfront_fee_to_forward`]
+```
+
+An upfront fee policy field is added to bolt-11 invoices to allow receiving 
+nodes to advertise the upfront fees that it will accept for the final hop
```
GeorgeTsagk, line 126: > to advertise the upfront fees that it will accept for the final hop

So IIUC: the receiver will also take an upfront fee, for the case where a failure occurs because e.g. they are not aware of the preimage that satisfies the proposed HTLC over that last hop

Is this introduced as a measure to prevent spam from direct peers that you have channels with?

proposals/channel-jamming-mitigation.md
GeorgeTsagk, line 161: Shouldn't we flip `X` and `Y`
> - Let X be the total amount due to be paid. 
> - Let Y be the payment amount that the sender should dispatch.

The way I understand this: X is amount the receiver should get  and Y is the inflated amount (with all kinds of fees included) that leaves the sender. Perhaps my interpretation of the above bullets is just not right.

proposals/channel-jamming-mitigation.md
```diff
@@ -0,0 +1,257 @@
...
+defaults, nodes that advertise `option_upfront_fee` in their node announcement
+should be assumed to have a base and proportional fee that is 1% of its success
+case fees. Nodes also will not relay and senders will not utilize channels with
+a `channel_update` messages where upfront fees are > 10% of their success-case
+equivalent. This eliminates the risk of nodes setting these fees so high that
+it becomes economically rational for forwarding nodes to fail payments for the
+sake of their upfront fees. 
+
+The upfront fee amount should simply be assigned to the receiving party's 
+balance when an incoming HTLC is added. These amounts are not expected to be
```
GeorgeTsagk, line 96: What I fail to understand is what exactly is added to the receiving end's balance directly when an HTLC with upfront fees is proposed.

Say we got this route `A` --> `B` --> `C` --> `D`
- `A` sends to `D`
- `B`,`C` and `D` ask for an upfront fee of `100sat` each

When `A` forwards the HTLC to `B` isn't the total sum of the upfront fees (`300sat`) forwarded to `B`?

So `B` gets `300sat` upfront fees from `A`, then `B` pays `200sat` upfront fees to `C` and `C` pays `100sat` upfront fees to the (receiver) `D`? Effectively, everybody gets their part of the `100sat` upfront fees, and this works as expected.

Can't `B` induce a failure and grab the full upfront fee amount (`300sat`) that would otherwise be delivered to the next hops?

carlaKC:
Closing this for now as we're focusing on endorsement first.

PR #1073, opened 2023-05-02
Bolt 1: Clarify the definition of tu fields

nepet:
The sentence `..., leading zeros in integers can be omitted` in the Bolt 1 fundamental type section: https://github.com/lightning/bolts/blob/50b2df24a27879e8329712c275db78876fd022fe/01-messaging.md#fundamental-types imposes that it is not necessary, but optional to truncate leading zeros in `tuXX` fields.

Having a look at the `tuXX` parsers across the ln implementations, it seems that it is expected to omit leading zeros.
This PR tries to find a better explanation/definition to avoid uncertainties.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -224,12 +224,12 @@ Various fundamental types are referred to in the message specifications:
 * `u32`: a 4 byte unsigned integer
 * `u64`: an 8 byte unsigned integer
 
-Inside TLV records which contain a single value, leading zeros in
-integers can be omitted:
+For the final value in TLV records, truncated integers may be used. Leading zeros in
+truncated integers MUST be omitted:
 
-* `tu16`: a 0 to 2 byte unsigned integer
-* `tu32`: a 0 to 4 byte unsigned integer
-* `tu64`: a 0 to 8 byte unsigned integer
+* `tu16`: a 0 to 2 byte truncated unsigned integer
+* `tu32`: a 0 to 4 byte truncated unsigned integer
+* `tu64`: a 0 to 8 byte truncated unsigned integer
 
 When used to encode amounts, the previous fields MUST comply with the upper
 bound of 21 million BTC:


morehouse:
Which implementations *require* leading zeros to be truncated?  From a quick look at CLN and LND, they appear to work fine for untruncated integers.

t-bast:
Eclair requires those values to be minimally-encoded, but I never realized the spec didn't make it a requirement...

I don't see any good reason to keep the redundant leading zeros, if we're using a truncated integer that's specifically because we want to remove them, right?

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1073/commits/60b77f6bc77a35a1da3e1cc2000df25228ae596e

01-messaging.md
```diff
@@ -225,11 +225,11 @@ Various fundamental types are referred to in the message specifications:
 * `u64`: an 8 byte unsigned integer
 
 Inside TLV records which contain a single value, leading zeros in
-integers can be omitted:
+truncated integers MUST be omitted:
```
morehouse, line 228: Truncated integers are not limited to TLV records with single values, and in fact we already use them in onion routing TLVs with multiple values.

Since we're making changes here, I would be in favor of also clarifying where truncated integers can be used.

```md
For the final value in TLV records, truncated integers may be used.  Leading zeros in
truncated integers MUST be omitted:
```

nepet:
> Which implementations _require_ leading zeros to be truncated? From a quick look at CLN and LND, they appear to work fine for untruncated integers.

Actually afaict, cln does check on leading zero bytes and fails if there are any. The following lines from `fromwire_tlv_uint(const u8 **cursor, size_t *max, size_t maxlen)` check that the first value read from stream that was stored in the bytes buffer in big-endian order is not zero.
```
	if (length > 0 && bytes[sizeof(bytes) - length] == 0) {
		SUPERVERBOSE("not minimal");
		fromwire_fail(cursor, max);
		return 0;
	}
``` 

Also, sorry for the (maybe) wrong terminology, I was in believe that the `t` in `tuXX` stands for _truncated_ instead of the obvious _tlv_.

Let me know if you would like me to change the wording.

morehouse:
> Actually afaict, cln does check on leading zero bytes and fails if there are any. The following lines from `fromwire_tlv_uint(const u8 **cursor, size_t *max, size_t maxlen)` check that the first value read from stream that was stored in the bytes buffer in big-endian order is not zero.
> 
> ```
> 	if (length > 0 && bytes[sizeof(bytes) - length] == 0) {
> 		SUPERVERBOSE("not minimal");
> 		fromwire_fail(cursor, max);
> 		return 0;
> 	}
> ```

Ah, you're right.  I completely skipped over that check.

> Also, sorry for the (maybe) wrong terminology, I was in believe that the `t` in `tuXX` stands for _truncated_ instead of the obvious _tlv_.
> 
> Let me know if you would like me to change the wording.

I also assumed it meant "truncated unsigned integer".

Roasbeef:
ACK on the changes, could also be an opportunity to add some basic test vectors here to tighten things up.

nepet:
If I understand the following lines from the `## Appendix B: Type-Length-Value Test Vectors` section correctly, the minimum encoding is already addressed.
https://github.com/lightning/bolts/blob/50b2df24a27879e8329712c275db78876fd022fe/01-messaging.md?plain=1#L709-L715
https://github.com/lightning/bolts/blob/50b2df24a27879e8329712c275db78876fd022fe/01-messaging.md?plain=1#L791-L799
Maybe we can further clarify it by providing separate test vectors for the fundamental `tuXX` type. But if people already are using the given set of test vectors the cases should be covered and the rewording of the definition might be sufficient to grasp the requirements of those fields.

01-messaging.md
```diff
@@ -224,12 +224,12 @@ Various fundamental types are referred to in the message specifications:
 * `u32`: a 4 byte unsigned integer
 * `u64`: an 8 byte unsigned integer
 
-Inside TLV records which contain a single value, leading zeros in
-integers can be omitted:
+For the final value in TLV records, truncated integers may be used. leading zeros in
```
t-bast, line 227: nit:
```suggestion
For the final value in TLV records, truncated integers may be used. Leading zeros in
```

nepet: Damn it ^^! I'll forcefully amend that to avoid a useless commit.

t-bast: Don't hesitate to just squash everything to a single commit :)

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1073/commits/bf97a7901040e1ce3a45ac89e62ffd37f8ab23b1

vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1073/commits/bf97a7901040e1ce3a45ac89e62ffd37f8ab23b1

rustyrussell:
Ack https://github.com/lightning/bolts/commit/bf97a7901040e1ce3a45ac89e62ffd37f8ab23b1

PR #1077, opened 2023-05-07
Bolt 4: fix onion test vector

ellemouton:
Fix the onion generation test vector to assume that the onion type being used is TLV and not the legacy onion type.

I think that implementations testing against this vector are defaulting to the legacy onion type when parsing the payloads instead of defaulting to the TLV type. This might be because in the previous test, the type was explicitly defined and so 
perhaps test vectors were not updated not to default to the legacy type.


Daddysgotthis:
The device  was malfunction during

t-bast:
I think I understand what's happening. The difference between your onion and the spec one is that you are pre-pending each `hop.payload` with its length, but they already contain their length.

For example, the first hop payload in the test vector is `12 02023a98 040205dc 06080000000000000001` where `12` is the length of the following tlv stream. In your onion, the first hop payload is `13 12 02023a98 040205dc 06080000000000000001`.

ellemouton:
ah! ok that makes way more sense! Thanks for the help here @t-bast 🙏 

I can update this PR so that it adds a clarifying comment to the test

t-bast (approved):
Thanks!

PR #1079, opened 2023-05-12
BOLT 8: make it clear that there are *two* chaining keys.

rustyrussell:
lnmessage got this wrong!  It would pass our test vectors, but actually fail in real usage, since it used the same `ck`.

Also, nonce rotation happens after 1000 encryptions, which is when the nonce reaches 1000 (since it's zero based!), not when it *exceeds* 1000.

### DIFF
--- a/08-transport.md
+++ b/08-transport.md
@@ -369,7 +369,9 @@ construction, and 16 bytes for a final authenticating tag.
        receiving messages for the duration of the session, are generated.
 7. `rn = 0, sn = 0`
      * The sending and receiving nonces are initialized to 0.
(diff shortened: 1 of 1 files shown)


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1079/commits/e485c4d193bae576d218fa725fd95f59336f9512

sstone:
ACK https://github.com/lightning/bolts/pull/1079/commits/e485c4d193bae576d218fa725fd95f59336f9512

t-bast (approved):
LGTM

PR #1086, opened 2023-06-05
BOLT4: Specify max HTLC nLocktime for expiry_too_far

ariard:
Currently BOLT#4 recommends to forwarding node to reject HTLC with `nLocktime` field to far in the future from chain tip height (type 21: `expiry_too_far`). This recommendation is reasonable to avoid liquidity griefing where a HTLC would be send so far in the future (e.g 20 000 blocks from chain tip), than in case of on-chain force-closure of the link, the forwarding node cannot confirm the HTLC-timeout and therefore recover the satoshis locked.

I think this recommendation is implemented by all the main Lightning implementations. For e.g, for LDK will reject HTLC more than 2016 block in the future (`CLTV_FAR_FAR_AWAY`). This comes with few downsides, a) a payer cannot build payment path with more than 14 hops requesting at least a `cltv_expiry_delta` of 144 blocks, such path while onerous in term of off-chain fees could be done for privacy purpose and b) the forwarding nodes are limited in the selection of their own `cltv_expiry_delta`  and as such the level of safety they wish for the HTLC forward (e.g in protection against time-dilation attacks). 

This proposal suggests to adopt a common value between all implementations, dubbed `max_htlc_cltv`. A default value of 4032 blocks (4 weeks) is proposed (though I didn’t check what is selected by LND/CLN/Eclair currently and if it makes sense to adapt in consequence). If adopted, such `max_htlc_cltv` will improve the safety of routing nodes network-wide and allow more advanced rebalancing like #780. 

Beyond, a new option could be introduced [`option_shrek`](https://shrek.fandom.com/wiki/Far_Far_Away) and a corresponding `channel_update` message bit where routing hops are announcing they accept “unsafe” packets beyond 4032 blocks. This level of flexibility can be used by use-cases like multi-hop submarine swaps.

cc https://github.com/lightningdevkit/rust-lightning/pull/2250

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -62,6 +62,7 @@ A node:
   * [Returning Errors](#returning-errors)
     * [Failure Messages](#failure-messages)
     * [Receiving Failure Codes](#receiving-failure-codes)
+  * [`max_htlc_cltv` Selection](#max-htlc-cltv-selection)
   * [Onion Messages](#onion-messages)
   * [Test Vector](#test-vector)
     * [Returning Errors](#returning-errors)
@@ -1333,7 +1334,7 @@ A _forwarding node_ MAY, but a _final node_ MUST NOT:
   - if the `cltv_expiry` is unreasonably near the present:
     - report the current channel setting for the outgoing channel.
     - return an `expiry_too_soon` error.
-  - if the `cltv_expiry` is unreasonably far in the future:
+  - if the `cltv_expiry` is more than `max_htlc_cltv` in the future:
     - return an `expiry_too_far` error.
   - if the channel is disabled:
     - report the current channel setting for the outgoing channel.
@@ -1580,6 +1581,10 @@ Onion messages don't explicitly require a channel, but for
 spam-reduction a node may choose to ratelimit such peers, especially
 messages it is asked to forward.
 
+## `max_htlc_cltv` Selection
+
+This `max_htlc_ctlv` value is defined as 2016 blocks, based on historical value
+deployed by Lightning implementations.
 
 # Test Vector
 


ProofOfKeags:
Implementations should specify what parameters they currently use for this: @rustyrussell @TheBlueMatt @t-bast @Roasbeef

thomash-acinq:
Eclair has increased its default to 2016 recently (https://github.com/ACINQ/eclair/pull/2677) to match LND and CLN. So now I believe all implementations use 2016 by default.
Given the current state of the network I think 2016 is enough and I only see downsides in doubling it to 4032. I disagree that having a limit of 14 hops is a downside, I challenge you to find a route this long that makes sense. Allowing longer routes makes jamming more potent and a higher `max_htlc_cltv` blocks liquidity longer in case of honest mistake (a node crashes while relaying a HTLC).

ProofOfKeags:
This appears to be [2016 in LND](https://github.com/lightningnetwork/lnd/blob/master/htlcswitch/link.go#L39-L51)

niftynei:
Spec meeting notes: should update 4032 to 2016.

rustyrussell:
OK, propose 2016 and everyone is happy.

04-onion-routing.md
```diff
@@ -1407,6 +1408,19 @@ The _origin node_:
...
+If the HTLC expires too far in the future, a forwarding node can return an `expiry_too_far` error.
+
+This BOLT4 processing requirement comes with two downsides:
+- a payer is constrained on the maximum size of its payment path
+- a forwarding node is constrained on the `cltv_expiry_delta` for the safety of its own channels
+
+A common value between implementations allow forwarding nodes to increase their `cltv_expiry_delta`
+without reducing the payers ability to route along the network.
+
+This new common value `max_htlc_cltv` is suggested to be 4032 blocks.
```
Roasbeef, line 1422: Looks like we all do 2016 today, so let's stick with that as the recommendation?

rustyrussell:
This should become a one liner.

Change:

```
  - if the `cltv_expiry` is unreasonably far in the future:
    - return an `expiry_too_far` error.
```

to:

```
  - if the `cltv_expiry` is more than 2016 blocks in the future:
    - return an `expiry_too_far` error.
```

04-onion-routing.md
ariard: Yep if we’re all doing 2016 today let’s stick to it. Good starting point then for a base value for routing hop gossips explicitly diverging from it.

ariard:
Updated at 9036d7c2 with the value of 2016 as it sounds common between all implementations. Rebased with onion changes.

> I disagree that having a limit of 14 hops is a downside, I challenge you to find a route this long that makes sense.

If you have sophisticated deanonymization adversaries targeting your flow of payments, and you’re looking to increase the number of hops to break common heuristics and you’re ready to pay the routing fees premium, potentially combined with things like https://github.com/lightning/bolts/issues/1008.

Note the `max-htlc-cltv` value to abstract from 2016 block and allow references by:
- `channel_update` extensions e.g long-term held (< 2016 blocks) packets for swaps
- mempool policy rules
- smart lightning signers policy control rules

04-onion-routing.md
```diff
@@ -1412,6 +1413,7 @@ The _origin node_:
   - MAY use the data specified in the various failure types for debugging
   purposes.
 
+<<<<<<< HEAD
```
t-bast, line 1416: nit: remove

ariard: corrected thanks

ariard:
Updated at ecfcf07.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1086/commits/ecfcf07bf353211fb6d4d8783ace493381267bc6

t-bast:
Spec feedback is that we should make this change a one-liner instead of defining a new section.

ariard:
> @ariard spec feedback is that we should make this change a one-liner instead of defining a new section.

I bind my reasoning exposed above. 

> Note the max-htlc-cltv value to abstract from 2016 block and allow references by:
>
> channel_update extensions e.g long-term held (< 2016 blocks) packets for swaps
> mempool policy rules
> smart lightning signers policy control rules

Especially for the first one `channel_update` extension, you wish to have a clear value you can reference from which to say “above max-htlc-cltv” you’re free to have custom routing feerates applied. Even for experimentation for now.

Apart keeping the spec succinct, what is the downside of more than one-line spec update ? Note if you look on TCP standards over the last decades they have been abstracting and naming magic constant over time.

04-onion-routing.md
```diff
@@ -1330,7 +1331,7 @@ A _forwarding node_ MAY, but a _final node_ MUST NOT:
     - report the `cltv_expiry` of the outgoing HTLC and the current channel setting for the outgoing
     channel.
     - return an `incorrect_cltv_expiry` error.
-  - if the `cltv_expiry` is unreasonably near the present:
+  - if the `cltv_expiry` is more than `max_htlc_cltv` near the present:
```
Roasbeef, line 1334: I think this should be modifying the line below?

ariard: Done at bccab9afc, thanks for the catch.

ariard:
Updated at bccab9a with introducing `max-htlc-cltv` on the right error line (the one above far in the future). The constify of the value is kept.

ProofOfKeags (approved):

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1086/commits/bccab9afc269a50a579850e71b06e58ef16c4e92

Roasbeef (approved):
LGTM 🦋

PR #1089, opened 2023-06-22
bolt07: enforce htlc_maximum_msat >= htlc_minimum_msat

carlaKC:
Bolt 7 specifies _how_ to set `hltc_minimum_msat`  and `htlc_maximum_msat` but does not explicitly enforce `hltc_minimum_msat` <= `htlc_maximum_msat`.

Since a channel where the largest HTLC that the advertising node will send through it is smaller than the minimum value the peer will accept is unusable, it seems reasonable to blacklist this type of gossip?

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -474,6 +474,7 @@ The origin node:
   - MUST set `htlc_maximum_msat` to the maximum value it will send through this channel for a single HTLC.
     - MUST set this to less than or equal to the channel capacity.
     - MUST set this to less than or equal to `max_htlc_value_in_flight_msat` it received from the peer.
+    - MUST set this to greater than or equal to `htlc_minimum_msat`.
   - MUST set `must_be_one` in `message_flags` to 1.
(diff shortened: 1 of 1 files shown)


carlaKC:
> I don't see why we should ban an entire node for such a thing, like I could see a bug on this resulting in this happening, but we should just ignore the channel, not ban the node and all their channels, no? 

I think ignore is also fine. Was following the pattern from "if `htlc_maximum_msat` is greater than channel capacity" since that's also a reaction to a bug?

> Did you come across this somewhere?

No, just noticed it when I was checking up on some route blinding stuff.

carlaKC:
> Yea, and arguably that also shouldn't result in us punishing a node entirely

Yeah fair point, bumped this one to should ignore during route construction.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1089/commits/d6b21774a172cfc8bd4b9bf3b5e6f1c51596d3a1

ProofOfKeags (approved):
ACK d6b21774a172cfc8bd4b9bf3b5e6f1c51596d3a1

PR #1092, opened 2023-06-30
Clean up: remove two unused features, assume four more

rustyrussell:
Removed:
- initial_routing_sync (only had an effect if gossip_queries not supported)
- option_anchor_outputs (only supported by older experimental-only CLN builds)

I looked at all node_announcements on my node.  There are 449 nodes apparently running a 4-year-old LND version (features hex `2200`), which have 3+ year old channels.  @Roasbeef points out that they already will have their channel_updates ignored due to lack of htlc_maximum_msat which is now required by LND and CLN, at least).

Features you can now assume (you should probably still set it for now, but you can stop checking it).
- var_onion_optin (all but 6 nodes)
- option_data_loss_protect (all but 11 nodes)
- option_static_remotekey (all but 16 nodes)

`gossip_queries` is now slightly repurposed: if you don't offer this, it means you won't give useful gossip (so you should not query such nodes, nor consider them when trying to figure out if you've got all the gossip).


09-features.md
```diff
@@ -30,28 +32,23 @@ The Context column decodes as follows:
...
-| 14/15 | `payment_secret`                  | Node supports `payment_secret` field                      | IN9      | `var_onion_optin`         | [Routing Onion Specification][bolt04]                                 |
+| 6/7   | `gossip_queries`                  | ASSUMED                                                   |          |                           |                                                                       |
+| 8/9   | `var_onion_optin`                 | ASSUMED                                                   |          |                           |                                                                       |
+| 10/11 | `gossip_queries_ex`               | Gossip queries can include additional information         | IN       |                           | [BOLT #7][bolt07-query]                                               |
+| 12/13 | `option_static_remotekey`         | ASSUMED                                                   |          |                           |                                                                       |
```
t-bast, line 40: I don't think we should make this one `ASSUMED`, as it removes the option to cleanly deprecate it. If I'm a node that only wants to support `anchor_outputs` (which we'll want soon), I want to be able to explicitly disable the `option_static_remotekey` feature bit from my `node_announcement` and `init`, to let other nodes know that they shouldn't connect to me if they want `option_static_remotekey` channels.

It is a good idea however to remove the dependency between `anchor_outputs` and `option_static_remotekey` (to allow setting only the `anchor_outputs` feature bit without the `option_static_remotekey` feature bit), and it would be useful to update all implementations to accept that.

07-routing-gossip.md
```diff
@@ -894,30 +889,10 @@ generated directly by the node itself, so they should ignore filters.
 If a node requires an initial sync of gossip messages, it will be flagged
```
t-bast, line 889: Shouldn't we remove the whole `Initial Sync` section? It doesn't really make sense anymore, at least in its current state.

05-onchain.md
```diff
@@ -608,20 +606,11 @@ to arrive at a reasonable fee.
 
 A node which broadcasts an HTLC-success or HTLC-timeout transaction for a
 commitment transaction:
-  1. if `option_anchor_outputs` applies:
-    - SHOULD combine it with inputs contributing sufficient fee to ensure
-      timely inclusion in a block.
-    - MAY combine it with other transactions.
-  2. if `option_anchors_zero_fee_htlc_tx` applies:
+  1. if `option_anchors` applies:
```
t-bast, line 609: nit:

```suggestion
  - if `option_anchors` applies:
```

03-transactions.md
```diff
@@ -631,8 +631,6 @@ also subtract two times the fixed anchor size of 330 sats from the funder
 Each commitment transaction uses a unique `localpubkey`, and a `remotepubkey`.
```
t-bast, line 631: This isn't true anymore if we assume `option_static_remotekey`?

02-peer-protocol.md
```diff
@@ -459,17 +453,16 @@ The recipient:
 
 #### Rationale
 
-We decide on `option_static_remotekey`, `option_anchor_outputs` or
-`option_anchors_zero_fee_htlc_tx` at this point when we first have to generate
+We decide on
+`option_anchors` at this point when we first have to generate
```
t-bast, line 457: nit:
```suggestion
We decide on `option_static_remotekey` or `option_anchors`
at this point when we first have to generate
```

t-bast (commented):
:broom: :broom: :rocket:

morehouse (commented):
Concept ACK

09-features.md
```diff
@@ -6,6 +6,8 @@ the `channel_announcement` and `node_announcement` messages ([BOLT
 #7](07-routing-gossip.md)).  The flags are tracked separately, since
 new flags will likely be added over time.
 
+Some features were introduced and became so widespread they are `ASSUMED` to be present by all nodes, and can be safely ignored (and the semantics are only defined in prior revisions of this spec).
```
Roasbeef, line 9: Wouldn't we better be served by just starting to set the feature bits to required? That alone might trigger some unexpected behavior in the network that shows maybe we have more stragglers than we thought. 

As an example, we _assumed_ that `max_htlc` was the new default, but then ended up inadvertently partitioning the network as (due to a desynchronized deployment across impls) lnd would still forward `channel_update` messages without that field, which caused CLN to hang up on it. In this instance, couldn't we have start to set the `max_htlc` field in a _new_ even TLV type, which then would have had the opposite effect: old nodes failing (in theory) to parse the new messages.

02-peer-protocol.md
```diff
@@ -439,10 +438,8 @@ Both peers:
       - the `channel_type` is `option_anchors_zero_fee_htlc_tx` and `option_static_remotekey` (bits 22 and 12)
     - otherwise, if `option_anchor_outputs` was negotiated:
       - the `channel_type` is `option_anchor_outputs` and `option_static_remotekey` (bits 20 and 12)
-    - otherwise, if `option_static_remotekey` was negotiated:
-      - the `channel_type` is `option_static_remotekey` (bit 12)
     - otherwise:
-      - the `channel_type` is empty
+      - the `channel_type` is `option_static_remotekey` (bit 12)
```
Roasbeef, line 439: Is this actually a functional change? IIUC, in the case there were no `channel_type` bits set, it would fall back to the existing "implicit" negotiation that selected based on feature bits (which maybe is actually under-defined?).

03-transactions.md
```diff
@@ -1207,377 +1203,6 @@ And, here are the keys needed to create the transactions:
 
 And, here are the test vectors themselves:
 
-    name: simple commitment tx with no HTLCs
```
Roasbeef, line 1210: 🫡

09-features.md
```diff
@@ -32,13 +32,13 @@ The Context column decodes as follows:
 
 | Bits  | Name                              | Description                                               | Context  | Dependencies              | Link                                                                  |
 |-------|-----------------------------------|-----------------------------------------------------------|----------|---------------------------|-----------------------------------------------------------------------|
-| 0/1   | `option_data_loss_protect`        | Requires or supports extra `channel_reestablish` fields   | IN       |                           | [BOLT #2][bolt02-retransmit]                                          |
+| 0/1   | `option_data_loss_protect`        | ASSUMED                                                   |          |                           |                                                                       |
```
Roasbeef, line 35: So does this mean we no longer set the bit? What about an intermediate state of just everything setting it to required? Seems strange to never exercise the flip from optional to required, then just assume everyone supports it. The assumption route would potentially lead to unclean failures, eg: one side has bit set, the other assume it's set.

09-features.md
Roasbeef: > that I hope can be implemented in ~1/2 years

Someone's feeling _very_ optimistic 😉

My lofty estimate is something more like 4x that, so ~2 years given the rollout that still needs to happen on the Bitcoin p2p network. 

I think everyone just setting it to be required achieves the same goal in practice? Those that don't support it will fail the feature bit connection check. IIRC in the past when we did this, we had some issues with peers that had old channels, but hadn't yet updated their feature bits: new and old were unable to reconnect as they bit check failed even though there was an existing channel open.

09-features.md
```diff
@@ -42,8 +42,7 @@ The Context column decodes as follows:
 | 14/15 | `payment_secret`                  | Node supports `payment_secret` field                      | IN9      | `var_onion_optin`         | [Routing Onion Specification][bolt04]                                 |
 | 16/17 | `basic_mpp`                       | Node can receive basic multi-part payments                | IN9      | `payment_secret`          | [BOLT #4][bolt04-mpp]                                                 |
 | 18/19 | `option_support_large_channel`    | Can create large channels                                 | IN       |                           | [BOLT #2](02-peer-protocol.md#the-open_channel-message)               |
-| 20/21 | `option_anchor_outputs`           | Anchor outputs                                            | IN       | `option_static_remotekey` | [BOLT #3](03-transactions.md)                                         |
```
Roasbeef, line 43: What's the expected behavior if we peer sends bits `20/21` for a channel type open? Do we assume that it's actually `option_anchors_zero_fee_htlc_tx`, or fail as we've all removed the feature bits from our codebases so that shows up as an undefined combination?

03-transactions.md
```diff
@@ -424,13 +424,13 @@ Note the reference to the "base fee" for a commitment transaction in the require
...
 The fee for an HTLC-timeout transaction:
-  - If `option_anchors_zero_fee_htlc_tx` applies:
+  - If `option_anchors` applies:
     1. MUST be 0.
   - Otherwise, MUST be calculated to match:
     1. Multiply `feerate_per_kw` by 663 and divide by 1000 (rounding down).
 
 The fee for an HTLC-success transaction:
-  - If `option_anchors_zero_fee_htlc_tx` applies:
+  - If `option_anchors` applies:
```
Roasbeef, line 433: Looks like left over from the last commit? Doesn't match the commit title of: 
> BOLT 9: assume var_onion_optin.

09-features.md
```diff
@@ -36,14 +36,14 @@ The Context column decodes as follows:
 | 3     | `initial_routing_sync`            | Sending node needs a complete routing information dump    | I        |                           | [BOLT #7][bolt07-sync]                                                |
 | 4/5   | `option_upfront_shutdown_script`  | Commits to a shutdown scriptpubkey when opening channel   | IN       |                           | [BOLT #2][bolt02-open]                                                |
 | 6/7   | `gossip_queries`                  | More sophisticated gossip control                         | IN       |                           | [BOLT #7][bolt07-query]                                               |
-| 8/9   | `var_onion_optin`                 | Requires/supports variable-length routing onion payloads  | IN9      |                           | [Routing Onion Specification][bolt04]                                 |
+| 8/9   | `var_onion_optin`                 | ASSUMED                                                   |          |                           |                                                                       |
```
Roasbeef, line 38: Similar comment here re also advising this is set to required. 

If a peer receives a legacy onion, should they fail back or proceed? Or are we assuming ppl have ripped out all the logic, so they'd just fail to even parse it in the first place (malformed onion)?

07-routing-gossip.md
```diff
@@ -829,16 +825,15 @@ The addition of timestamp and checksum fields allow a peer to omit querying for
...
     * [`chain_hash`:`chain_hash`]
     * [`u32`:`first_timestamp`]
     * [`u32`:`timestamp_range`]
 
 This message allows a node to constrain future gossip messages to
-a specific range.  A node which wants any gossip messages would have
-to send this, otherwise `gossip_queries` negotiation means no gossip
-messages would be received.
+a specific range.  A node which wants any gossip messages has
+to send this, otherwise no gossip messages would be received.
```
Roasbeef, line 836: Should we also codify the behavior we talked about re "all or nothing" re the timestamp range? So that implementations only ever use this to signal that they want to receive updates, or don't want to receive any updates at all (so no legacy announcement back fill).

morehouse: Seems reasonable to me, but I think it would fit better in a separate PR.

09-features.md
t-bast: > I think everyone just setting it to be required achieves the same goal in practice?

I'm not sure I follow, my goal here is to be able to easily **deprecate** `option_static_remotekey`. Setting it to required definitely does not allow that?

Maybe what you mean here is that making `option_static_remotekey` assumed instead of explicitly set lets us deprecate it cleanly? Which somewhat means we should break the feature link between `anchor_outputs` and `option_static_remotekey`?

09-features.md
t-bast: Agreed, this is a first step that implementation should probably take right now.

09-features.md
t-bast: I think we should fail in that case, because that is likely an old node that supports the non-zero-fee variant so this won't work. A node that wants `option_anchors_zero_fee_htlc_tx` will just set the correct set of bits for it, right?

07-routing-gossip.md
t-bast: This is unrelated to this PR, I agree it should be a separate one.

09-features.md
rustyrussell: Yeah, this PR is overly ambitious.  I think we go for stages:

1. Set it to even (i.e. old peers should go away), and simply assume everyone supports it.
2. Stop setting it, now everyone supports it (but allow it to be set).
3. In theory, recycle it now everyone ignores it.  (Far future)

Remember: you can simply set an even bit and *ignore whether they set it or not*.  It's their responsibility to disconnect if they have a problem, not yours!

09-features.md
rustyrussell: If you still support it, you'd accept it?  But pretty soon, nobody will...

09-features.md
rustyrussell: > This PR essentially reduces the basic channel types to anchors and no anchors. So if we want to require anchor channels, then for now we can set bit 22 (require `option_anchors_zero_fee_htlc_tx`).
> 
> But if there's a new channel type in the future that doesn't depend on anchors, and we want to support both anchors and the new type, we won't have a way to signal that.

You would offer both the new bit and the anchor bit if you allow both, surely?

rustyrussell: > > I think everyone just setting it to be required achieves the same goal in practice?
> 
> I'm not sure I follow, my goal here is to be able to easily **deprecate** `option_static_remotekey`. Setting it to required definitely does not allow that?

It actually does.  You set it required, and ignore the bit in the peer.  We should fix the definition of negotiated, however, to make this clear.  If you set it required, and they keep talking to you, you assume it's negotiated (this is how CLN works, for example).

t-bast: It is a bit convoluted though...deprecating `option_static_remotekey` by not setting the corresponding feature bit anymore would be way more explicit, wouldn't it? But maybe that ship has sailed since we initially created that dependency, and we just need to make sure we don't recreate the same kind of trap for future channel types.

09-features.md
```diff
@@ -35,9 +35,9 @@ The Context column decodes as follows:
 | 0/1   | `option_data_loss_protect`        | ASSUMED                                                   |          |                           |                                                                       |
 | 3     | `initial_routing_sync`            | Sending node needs a complete routing information dump    | I        |                           | [BOLT #7][bolt07-sync]                                                |
 | 4/5   | `option_upfront_shutdown_script`  | Commits to a shutdown scriptpubkey when opening channel   | IN       |                           | [BOLT #2][bolt02-open]                                                |
-| 6/7   | `gossip_queries`                  | More sophisticated gossip control                         | IN       |                           | [BOLT #7][bolt07-query]                                               |
+| 6/7   | `gossip_queries`                  | ASSUMED                                                   |          |                           |                                                                       |
```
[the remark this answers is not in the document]
rustyrussell: Yep, this has been changed.

t-bast (commented):
I think we should merge that PR, shouldn't we? I have fixed my initial comments and added a bit more clean-up in https://github.com/t-bast/bolts/commit/fdaf7d61949f2766cc59effa8f2279dd9d7deebd, you should be able to just grab this commit and I believe we should be ready to go @rustyrussell

Roasbeef (approved):
LGTM 🪓

vincenzopalazzo (approved):
ACK b5ce609697ef91b000f2a64a35c4b2220c6e4f31

t-bast:
It looks like this was merged without including the changes from https://github.com/lightning/bolts/pull/1092#pullrequestreview-2006373363 (which I believe are necessary otherwise this is incomplete). I'll open a PR for this commit.

PR #1093, opened 2023-07-03
Allow unset `onion_hash` in `invalid_onion_blinding`

t-bast:
In most cases the `onion_hash` isn't actionable in the case of blinded payments and it's wasteful to keep track of the incoming onion, so we allow setting it to an all zero value.


rustyrussell:
Ack!

carlaKC (approved):
ack @ 0185012

t-bast:
We got ACKs from 3 implementations, let's merge this! @valentinewallace this should make your life easier in LDK ;)

PR #1096, opened 2023-07-17
option_simple_close

rustyrussell:
This is a "can't fail!" close protocol, as discussed at the NY Summit, and on @Roasbeef's wishlist.  It's about as simple as I could make it: the only complexity comes from allowing each side to indicate whether they want to omit their own output.

It's "taproot ready"(TM) in the sense that `shutdown` is always sent to trigger it, so that can contain the nonces without any persistence requirement.

I split it into three commits for cleanliness:

1. Introduce the new protocol
2. Remove the requirement that shutdown not be sent multiple times (which was already nonsensical)
3. Remove the older protocols

I recommend reviewing it as separate commits, it'll make more sense!


ProofOfKeags:
Perhaps this is a misunderstanding on my part for how the document is organized, but wouldn't you still have to support the original `closing_signed` message in the main BOLT docs here? I understand that maybe you can enforce the new simple close protocol gets used if the channel was opened or upgraded to `option_simple_close` but the way I'm reading this is that it is deleting the old close method from the protocol completely and I'm pretty sure we can't do that unless this is an extension BOLT right?

Otherwise I tend to like this. The game that has been identified where the closer of the channel pays the fee has the side effect of keeping channels open longer which I believe to be a good thing.

ACK up to [aa6deb8](https://github.com/lightning/bolts/pull/1096/commits/aa6deb8161fad455e1d045b5a75860cf18592377)

rustyrussell:
There was an expressed desire in NY to more agressively clean up deprecated parts of the spec, even if they're still in use.  It keeps the document cleaner, and avoid having to revisit later (when?) to remove the older parts.  However, it's still in the git history.

ProofOfKeags:
Yeah I understand the desire to remove deprecated stuff. I just wonder what a "valid" implementation looks like here. If we implement this change, and essentially drop support for the old closing scheme from the spec, doesn't that mean that `option_simple_close` isn't really an option for a new impl? If it is an option but we remove the old closing scheme, doesn't that mean that there is no way to do a coop close anymore?

Maybe this is more of a meta-question of who the main BOLTs are for. My naive (and new!) perspective is that the BOLTs should be the document I should consult if I wanted to bootstrap a new LN impl and that as long as I implement everything in the main BOLTs without any of the optional stuff or extensions then I should be broadly compatible with the network. Through this lense, it would seem that this change would result in the removal of cooperative closes from the main specification. Is my understanding of the purpose of the spec wrong? If not, are we OK with removing coop closes from the table-stakes features of the lightning network?

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
 
+### Closing Negotiation: `closing_complete` and `closing_sig`
+
+Once shutdown is complete, the channel is empty of HTLCs, there are no commitments
+for which a revocation is owed, and all updates are included on both commitments,
+the final current commitment transactions will have no HTLCs.
+
+Each peer says what fee it will pay, and the other side simply signs that transaction.  The only complexity comes from allowing each side to omit its own output should it be uneconomic.
+
+This process will be repeated every time a `shutdown` message is received, which allows re-negotiation.
```
t-bast, line 669: Right now redundant `shutdown` messages are ignored: it won't be the case anymore (especially with taproot since we'll need to use the nonces from that message). Should we now allow the sender to use a different closing `scriptpubkey` than the one they used in previous `shutdown`? Or should we keep restricting it for simplicity?

t-bast (commented):
Huge concept ACK, I'll really enjoy removing the old-style mutual close from `eclair` :grin:

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
 No funds are at risk when that happens, but the channel must be force-closed as
 the closing transaction will likely never reach miners.
 
+### Closing Negotiation: `closing_complete` and `closing_sig`
+
+Once shutdown is complete, the channel is empty of HTLCs, there are no commitments
+for which a revocation is owed, and all updates are included on both commitments,
+the final current commitment transactions will have no HTLCs.
+
+Each peer says what fee it will pay, and the other side simply signs that transaction.  The only complexity comes from allowing each side to omit its own output should it be uneconomic.
```
vincenzopalazzo, line 667: small nit: we can split the line here in two

02-peer-protocol.md
vincenzopalazzo: I guess allowing different `scriptpubkey` should be possible and the closer want to redirect the funds in another `scriptpubkey`.

ofc at this point we should also send it back inside the `closing_complete` msg, imho

vincenzopalazzo (commented):
LGTM.

My two cents on what we should keep inside the BOLTs.

I'd like to suggest cleaning up the feature that is somewhat outdated because having an outdated ways to do the same operation in the BOLTs can make it more challenging to read and understand what is the recent version.

If someone is writing a lightning node from scratch (and I promise that I will do it someday), they should start reading from the current state of the BOLT.

Of course, there is the concern of what happens if the BOLTs change while I am implementing the node? I think we all assume that an implementor will stay updated on the development of the spec (at least we hope).

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+Multiple signatures are given, so each side can choose whether it wants to include its own output.  In the case of large fees and tiny channels, where neither side wants its output, the use of an OP_RETURN simply cleans up the dangling unspent transaction output.  This is expected to be extremely rare.
+
+Note that there is usually no reason to pay a high fee for rapid processing, since an urgent child could pay the fee on the closing transactions' behalf.
+
+If the closer proposes a transaction which will not relay (either its own output is dust, or the fee rate it proposes is too low), it doesn't harm the closee to sign the transaction.
+
+Similarly, if the closer proposes a high fee, it doesn't harm the closee to sign the transaction, as the closer is paying.
+
+There's a slight game where each side would prefer the other side pay the fee, and proposes a minimal fee.  If neither side proposes a fee which will relay, the negotiation can occur again, or the final commitment transaction can be spent.  In practice, the opener has an incentive to offer a reasonable closing fee, as they would pay the fee for the commitment transaction, which also costs more to spend.
```
whitslack, line 736: This discussion of the game theory hints at an issue I've always had with the "opener pays" rule. I do understand the pathological scenario that the rule is meant to avoid (where a malicious node can drain the funds of a victim node by repeatedly opening and abandoning channels to it), but this new proposal creates a perverse incentive.

Olav opens a channel to Kristján. Some time later, Kristján wants the channel closed. Under this new proposal, Kristján's fee-optimal channel-closing strategy would be to transmit a bogus `error` message to Olav so that Olav will be forced (by protocol convention) to unilaterally close the channel. Thusly, Kristján would obtain an immediately spendable output and would avoid paying any fees.

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+The sender of `closing_complete` (aka. "the closer"):
+  - MUST set `fee_satoshis` to a fee less than or equal to its outstanding balance, rounded down to whole satoshis.
+  - SHOULD set `has_closer_output` to 0 if it considers its own remaining balance to be uneconomic.
+  - Otherwise MUST set `has_closer_output` to 1.
+  - If it sets `has_closer_output` to 1:
+    - MUST set `signature_with_closee_output` to a valid signature of a transaction with both closer and closee outputs.
+	- MUST set `signature_without_closee_output` to a valid signature of a transaction with only a closer output.
+  - Otherwise: (`has_closer_output` is 0):
+    - MUST set `signature_with_closee_output` to a valid signature of a transaction with only the closee output.
+	- MUST set `signature_without_closee_output` to a valid signature of a transaction with only the null output.
```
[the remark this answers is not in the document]
instagibbs: Might be missing something but I don't think standardness will allow this case into mempool/relay.

1 native segwit input with 1 OP_RETURN output is ~61 bytes of non-witness data.

Bitcoin Core would allow 65+ starting with Bitcoin Core 25.0: https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-25.0.md#notable-changes

but sub-64 is still not allowed.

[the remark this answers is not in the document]
rustyrussell: It's defined in BOLT 3 which covers the actual tx construction...

02-peer-protocol.md
rustyrussell: Yes, let them change it. Why not?

02-peer-protocol.md
rustyrussell: That's always true.

But K can also mutual close at <1 sat per byte.

whitslack: How can any mutual closing transaction use <1 sat/vB fee rate? The transaction would not propagate.

And you're right that it's always true. It has always irked me. I swear I have seen some cases where nodes to which I have opened channels have sent me `error` messages for no reason other than to get me to unilaterally close. I think it was a design mistake to have commitment transactions reward the non-closing party with an immediately spendable output. I think commitment transactions should time-lock _both_ outputs to discourage baiting a peer into unilaterally closing.

02-peer-protocol.md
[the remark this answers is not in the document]
ajtowns: If you're in a cooperative close scenario, but you've decided you're happy to abandon all your channel funds, why not donate them to your counterparty rather than miners? ie, add what would have been your balance to your counterparty's output value. If your counterparty then wants to move some of that to fees, they can re-negotiate the cooperative close and do an rbf?

Adding null output handling doesn't seem in line with the goal of making things simpler and less special case-y.

If you are signing a "send all my funds to fees" tx, you should probably be doing it in an aggregatable way (`ANYONECANPAY|NONE` or `ANYONECANPAY|ALL`) so that in the event of a bunch of channels being closed because block space is scarce and expensive, the 10+ byte outputs of all those channels can be merged.

02-peer-protocol.md
t-bast: That will disappear once we have package relay, because at that point commitment transactions will pay no fees. The fees will have to be paid using CPFP on an ephemeral anchor. Also, if you send bogus `error`s, nodes will start blacklisting you at some point and won't accept new channels from you: this isn't the case today, but will definitely be in the future.

02-peer-protocol.md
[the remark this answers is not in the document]
t-bast: But why add any of this complexity in the first place? If we're in that scenario, that means the balance of the closer was very low (otherwise they wouldn't omit their output) which means that the balance on the closee side is large enough that they'll never want to omit it (it doesn't make sense to allow channels with a total capacity that is ~dust).

We should only send a signature for a transaction that contain the closee output, and omit the case where the transaction doesn't contain any real output, don't we?

02-peer-protocol.md
vincenzopalazzo: I guess we should have this because we can have concurrency around the network? and getting the confirmation on what scriptpubkey we close the channel can be useful to avoid unnecessary mess,  right?

02-peer-protocol.md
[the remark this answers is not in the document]
instagibbs: > why not donate them to your counterparty rather than miners?

Adding financial incentives to goad coop closes seems suboptimal.

> We should only send a signature for a transaction that contain the closee output, and omit the case where the transaction doesn't contain any real output, don't we?

The non-initiator can burn their dust later if they so choose, and in any manner they choose, without having to specify exactly how. I think that's a vote in favor of it.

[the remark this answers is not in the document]
ajtowns: > Adding additional financial incentives to goad coop closes seems suboptimal.

I think the logic here should be the same as for dust: "if this output will never be worth spending, don't create it". If done that way, there's no opportunity for goading -- first, you should never allow your balance to drop to that level, and in the event that it somehow does (your channel was down while fees were rising, you changed your configuration and increased your dust value), you simply unconditionally abandon the channel when it's closed.

If your threshold for a spendable output is X, you estimate a cooperative close fee as F, you estimate a unilateral close fee as U (U>F), and your balance is B, then the logic is:

 * if `B > X`, reject any forward that would drop B below X. (or, for balance privacy, if `B < X + htlc_maximum_msat`, reject any forward to this channel)
 * if `B < X`, always give the balance to your counterparty.
 * if `B < X+U` consider rebalancing to increase B above X+U (or, if supported, rebalancing(draining) the channel to 0 and closing it).
 * if `0 < B < X+F`, don't automatically close the channel. if the user insists anyway, give the balance to your counterparty.
 * if `X+F < B < X+U` and rebalancing isn't possible, consider proposing a mutual close.

If you initiate a mutual close in the last case, you get an output with (B-F > X) funds, and can choose a low fee rate that will be slow to confirm (it's only a small output anyway, so you probably don't care very much when it's confirmed), leaving your counterparty to feebump if they want access to their funds sooner.

(Also, burning funds to fees doesn't remove the incentive to game a channel close, it at best moves the incentive from your counterparty to miners, and, if your counterparty was going to pay lots of fees anyway, just funds your counterparty less directly)

02-peer-protocol.md
```diff
@@ -615,7 +615,6 @@ A sending node:
   - MAY send a `shutdown` before a `channel_ready`, i.e. before the funding transaction has reached `minimum_depth`.
   - if there are updates pending on the receiving node's commitment transaction:
     - MUST NOT send a `shutdown`.
-  - MUST NOT send multiple `shutdown` messages.
```
Roasbeef, line 606: Should't this instead be gated on awareness/negotiation of the new feature bit? Old nodes will still reject multiple shutdowns, so it should be feature gated.

02-peer-protocol.md
```diff
@@ -571,20 +570,6 @@ Closing happens in two stages:
 1. one side indicates it wants to clear the channel (and thus will accept no new HTLCs)
 2. once all HTLCs are resolved, the final channel close negotiation begins.
 
-        +-------+                              +-------+
```
Roasbeef: IMO we should have awareness of this be signalled via feature bit rather than remove the old version. Then in the future, we can start to make the bit required to enforce the switch over on the protocol level.

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+Each peer says what fee it will pay, and the other side simply signs that transaction.  The only complexity comes from allowing each side to omit its own output should it be uneconomic.
+
+This process will be repeated every time a `shutdown` message is received, which allows re-negotiation.
+
+1. type: 40 (`closing_complete`)
+2. data:
+   * [`channel_id`:`channel_id`]
+   * [`u64`:`fee_satoshis`]
+   * [`u8`: `has_closer_output`]
+   * [`signature`:`signature_with_closee_output`]
```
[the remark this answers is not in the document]
rustyrussell: The existing dust trimming logic is static, and applies to several different cases.  At this point, you know exactly what the spend cost is, so let's be explicit.

02-peer-protocol.md
rustyrussell: No, everyone always accepted multiple shutdown messages, because we *required* that they be retransmitted.  This requirement was always ignored!

02-peer-protocol.md
```diff
@@ -776,6 +788,83 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+   * [`signature`:`signature_without_closee_output`]
+
+1. type: 41 (`closing_sig`)
+2. data:
+   * [`channel_id`:`channel_id`]
+   * [`u8`: `closee_output`]
```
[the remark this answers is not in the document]
rustyrussell: I think we probably will, and make taproot dependent on this feature so it's easy...

(I haven't actually implemented this to see if there are any gotchas, which is needed anyway)

02-peer-protocol.md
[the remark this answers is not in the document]
rustyrussell: Two points:

- One day, splice-to-close will be a thing, and you'll get to carefully craft your close tx to do arbitrary things.
- I don't expect to ever see a "no output" close on mainnet, but we have to cover it.

09-features.md
```diff
@@ -48,6 +48,7 @@ The Context column decodes as follows:
 | 46/47 | `option_scid_alias`               | Supply channel aliases for routing                        | IN       |                           | [BOLT #2][bolt02-channel-ready]                                       |
 | 48/49 | `option_payment_metadata`         | Payment metadata in tlv record                            | 9        |                           | [BOLT #11](11-payment-encoding.md#tagged-fields)                      |
 | 50/51 | `option_zeroconf`                 | Understands zeroconf channel types                        | IN       | `option_scid_alias`       | [BOLT #2][bolt02-channel-ready]                                       |
+| 60/61 | `option_simple_close`             | Simplified closing negotiation                            | IN       |  `option_shutdown_anysegwit`  | [BOLT #2][bolt02-simple-close]                                    |
```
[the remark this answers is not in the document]
rustyrussell: Well, that's a separate discussion, right?  I think everyone should support anysegwit, and we can transition to ignoring it later.

If #1092 goes in first, then this would be implicitly removed...

02-peer-protocol.md
rustyrussell: No concurrency, reconnection requirements already say you re-xmit on every reconnect:

```
  - upon reconnection:
    - if it has sent a previous `shutdown`:
      - MUST retransmit `shutdown`.
```

02-peer-protocol.md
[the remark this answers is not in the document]
ajtowns: > I don't expect to ever see a "no output" close on mainnet, but we have to cover it.

mainnet is the only place where fees might plausibly render an entire channel to be dust making burning the entire balance to fees sensible. If the feature's not useful there, it isn't useful anywhere.

In the event that a pair of nodes really do want to abandon a channel balance, they can manually specify a scriptPubKey of  "OP_RETURN [padding]" or "p2pkh[G]" to either permanently burn the funds (ultrasound money!) or donate to some future miner that's got some free space in a block. Anyway, if you want to make your simple close unnecessarily complicated, I guess that's your call, so I won't harp on this further...

02-peer-protocol.md
vincenzopalazzo: Ah I was missing it! good point, but in this case, the concurrency can happen if I send two subsequent shutdown messages with different `scriptpubkey`. I do not know why people should do it, but the spec allows it

02-peer-protocol.md
```diff
@@ -662,119 +661,84 @@ of the receiving node to change the `scriptpubkey`.
...
+  - Otherwise MUST set `has_closer_output` to 1.
+  - If it sets `has_closer_output` to 1:
+    - MUST set `signature_with_closee_output` to a valid signature of a transaction with both closer and closee outputs.
+	- MUST set `signature_without_closee_output` to a valid signature of a transaction with only a closer output.
+  - Otherwise: (`has_closer_output` is 0):
+    - MUST set `signature_with_closee_output` to a valid signature of a transaction with only the closee output.
+	- MUST set `signature_without_closee_output` to a valid signature of a transaction with only the null output as described in [BOLT 3](03-transactions.md#closing-transaction).
+
+The receiver of `closing_complete` (aka. "the closee"):
+  - If `has_closee_output` is 1 and the closer's output would be dust (< 540 satoshis for P2SH, < 330 satoshis for P2WSH and P2TR):
```
morehouse, line 709:
```suggestion
  - If `has_closer_output` is 1 and the closer's output would be dust (< 540 satoshis for P2SH, < 330 satoshis for P2WSH and P2TR):
```

02-peer-protocol.md
```diff
@@ -662,119 +661,84 @@ of the receiving node to change the `scriptpubkey`.
...
+The sender of `closing_sig`:
+  - if it selected `signature_with_closee_output` to broadcast:
+    - MUST set `has_closee_output` to 1.
+	- MUST set `signature` to a valid signature on that transaction.
+  - otherwise (it selected `signature_without_closee_output`):
+    - MUST set `has_closee_output` to 0.
+	- MUST set `signature` to a valid signature on that transaction.
+
+The receiver of `closing_sig`:
+  - MUST broadcast the transaction indicated by `has_closee_output`.
```
morehouse, line 727: No signature verification before broadcast?

02-peer-protocol.md
```diff
@@ -577,17 +577,17 @@ Closing happens in two stages:
...
         |       |                              |       |
-        |       |--(3)-- closing_signed  F1--->|       |
-        |       |<-(4)-- closing_signed  F2----|       |
-        |       |              ...             |       |
-        |       |--(?)-- closing_signed  Fn--->|       |
-        |       |<-(?)-- closing_signed  Fn----|       |
+        |       |--(3a)- closing_complete Fee->|       |
+        |       |<-(3b)- closing_complete Fee--|       |
+        |       |<-(4a)- closing_sig ----------|       |
+        |       |--(4b)- closing_sig --------->|       |
```
morehouse, line 583: This doesn't seem like the typical use case where one side initiates the close and the other signs.  Some explanation somewhere is needed:

- Is the idea here that both sides can contribute to the closing fee if they want?
- Does B's `closing_complete` override A's?  And in that case, why does B send `closing_sig`?
- Did both A and B send `closing_complete` at the same time, so we have 2 flows happening at once?

02-peer-protocol.md
```diff
@@ -662,119 +661,84 @@ of the receiving node to change the `scriptpubkey`.
...
 
-The receiving node:
-  - if one of the outputs in the closing transaction is below the dust limit for its `scriptpubkey` (see [BOLT 3](03-transactions.md#dust-limits)):
-    - MUST fail the channel
+The close protocol is designed to avoid any failure scenarios caused by fee disagreement, since each side offers to pay its own desired fee.
 
-#### Rationale
+Multiple signatures are given, so each side can choose whether it wants to include its own output.  In the case of large fees and tiny channels, where neither side wants its output, the use of an OP_RETURN simply cleans up the dangling unspent transaction output.  This is expected to be extremely rare.
+
+Note that there is usually no reason to pay a high fee for rapid processing, since an urgent child could pay the fee on the closing transactions' behalf.
```
morehouse, line 735: We could also mention RBF of the closing transaction here.

03-transactions.md
```diff
@@ -349,44 +349,41 @@ To spend this via penalty, the remote node uses a witness stack `<revocationsig>
...
-* txout count: 0, 1 or 2
-   * `txout` amount: final balance to be paid to one node (minus `fee_satoshis` from `closing_signed`, if this peer funded the channel)
-   * `txout` script: as specified in that node's `scriptpubkey` in its `shutdown` message
+
+* txout count: 1 or 2 of the following
+  * The closer output:
+    * `txout` amount: the final balance for the closer, minus `closing_complete` `fee_satoshis`, rounded down to whole satoshis.
+	* `txout` script: as specified in that closer's `scriptpubkey` in its `shutdown` message
+  * The closee output:
+    * `txout` amount: the final balance for the closer, rounded down to whole satoshis.
```
morehouse, line 369:
```suggestion
    * `txout` amount: the final balance for the closee, rounded down to whole satoshis.
```

03-transactions.md
```diff
@@ -349,44 +349,41 @@ To spend this via penalty, the remote node uses a witness stack `<revocationsig>
...
-
-However, if one side chooses to eliminate its own output, there's no
-reason for the other side to fail the closing protocol; so this is
-explicitly allowed. The signature indicates which variant
-has been used.
-
-There will be at least one output, if the funding amount is greater
-than twice `dust_limit_satoshis`.
+The case where both sides omit their outputs due to a de-minimus channel is never expected to happen, however it is
+documented here for completeness, and serves to avoid an polluting the unspent outputs.  Unfortunately, the transaction
```
morehouse, line 385:
```suggestion
documented here for completeness, and serves to avoid polluting the unspent outputs.  Unfortunately, the transaction
```

morehouse (commented):
Concept ACK

rustyrussell:
Agreed at meeting that we should simply insist the larger output (or lower key as tiebreak) MUST have an output.  I guess if both are dust we cannot close, then.

ajtowns:
> Agreed at meeting that we should simply insist the larger output (or lower key as tiebreak) MUST have an output. I guess if both are dust we cannot close, then.

You could still do this in the implementation potentially -- allow `OP_RETURN` spks in shutdown(), require setting `has_closee_output` to false in `closing_sig` if you chose an `OP_RETURN` spk (and allow `signature_with_closee_output` to be invalid?). If you choose `OP_RETURN` and the other doesn't, wait for them to initiate the close; if you both choose `OP_RETURN`, wait a random amount of time, and send `closing_complete` with `fee_satoshis` equal to your entire balance.

(If you did this, could you move the "has_closee_output" decision to be whether or not the closee set the sPK to `OP_RETURN[something]`? If you change your mind, you can always just send shutdown again, aiui? Must set the sPK to `OP_RETURN` if balance is dust or below?)

In general, though, for a sufficiently small channel, you can't guarantee closure: a 100k sat channel closed to a single bare `OP_RETURN` via a musig2 signature would give you a fee rate of about 1300sat/vb; if miners have a backlog of txs above that fee rate (ie, making >13BTC per block in fees), your close tx won't confirm.

rustyrussell:
Maybe this whole situation is simpler if we move "do I want an output?" negotiation to shutdown.  This means we allow two new forms of scriptpubkey:

1. An empty scriptpubkey to indicate "I don't want my output at all".
2. OP_RETURN <6+ bytes of padding>.

And now, you SHOULD only set an empty scriptpubkey if you have the lesser output (if both are equal, neither can).  Note that the initial shutdown can happen while htlcs are still unsettled, but this will work itself out if you re-xmit shutdown afterwards, and have a rule not to send closing_complete if both sides opt for no output.

rustyrussell:
OK, *this time for sure*!

We only let one side (the "lesser in msats", if any) discard their output, so we always have one.  We make the signature fields a tlv, so we can simply share them across the two messages, and carefully dictate which one to use.

ajtowns:
(Avoid @'s in commit msgs to reduce github notification spam? cf https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md )

02-peer-protocol.md
```diff
@@ -662,119 +665,119 @@ of the receiving node to change the `scriptpubkey`.
...
-  and its previously-received `fee_satoshis`, UNLESS it has since reconnected:
-    - SHOULD send a `warning` and close the connection, or send an
-      `error` and fail the channel.
-  - otherwise, if the receiver agrees with the fee:
-    - SHOULD reply with a `closing_signed` with the same `fee_satoshis` value.
+      - MUST use `no_closer_closee`.
+  - If the selected signature field does not exist:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - If the signature field is not valid for the corresponding closing transaction specified in [BOLT #3](03-transactions.md#closing-transaction):
+    - MUST ignore `closing_complete`.
```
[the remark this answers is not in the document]
t-bast: I think I see, what you have in mind is that each side only remembers the last remote `shutdown` they received and the last `shutdown` they sent (and the corresponding scripts), so if there is a race between `shutdown` from one side and `closing_complete` from the other, there will be a mismatch in the scripts used.

It is important to highlight that this will work and converge because when sending `shutdown`, we expect the remote to answer with their `shutdown` (even if they haven't changed their script) before sending our `closing_complete`. Am I correct?

02-peer-protocol.md
```diff
@@ -662,119 +665,119 @@ of the receiving node to change the `scriptpubkey`.
...
+    - `no_closer_closee`: closing transaction with only the remote ("closee") output.
+    - `closer_and_closee`: closing transaction with both the closer and closee outputs.
+  - If the local outstanding balance (in millisatoshi) is less than the remote outstanding balance:
+    - MUST NOT set `closer_no_closee`.
+    - MUST set exactly one of `no_closer_closee` or `closer_and_closee`.
+    - MUST set `no_closer_closee` if the local output amount is dust.
+    - MAY set `no_closer_closee` if it considers the closee output amount uneconomic AND its `scriptpubkey` is not `OP_RETURN`.
+  - Otherwise (not lesser amount, cannot remove own output):
+    - MUST NOT set `no_closer_closee`.
+    - MUST set both `closer_no_closee` and `closer_and_closee`.
```
t-bast, line 729: Why require setting both here? If the closee's output is below dust, we should be allowed to only set `closer_no_closee`, because `closer_and_closee` would lead to a non-standard transaction?

I understand that when the closee's output is *not* below dust, we must still set `close_no_closee` to allow them to omit their output though.

So I think the requirements for that section should be:

```
 - Otherwise (not lesser amount, cannot remove own output):
    - MUST NOT set `no_closer_closee`.
    - If the closee's output amount is dust:
      - MUST set `closer_no_closee`.
      - SHOULD NOT set `closer_and_closee`.
    - Otherwise:
      - MUST set both `closer_no_closee` and `closer_and_closee`.
```

02-peer-protocol.md
```diff
@@ -662,119 +665,119 @@ of the receiving node to change the `scriptpubkey`.
...
+  - MUST send `closing_sig` with a single valid signature in the same tlv field as the `closing_complete`.
+
+The receiver of `closing_sig`:
+  - if `tlvs` does not contain exactly one signature:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - if `tlvs` does not contain one of the tlv fields sent in `closing_complete`:
+    - MUST ignore `closing_complete`.
+  - if the signature field is not valid for the corresponding closing transaction specified in [BOLT #3](03-transactions.md#closing-transaction):
+    - MUST ignore `closing_complete`.
+  - of the signature field is non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
```
t-bast, line 759: nit:
```suggestion
  - if the signature field is non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
```

rustyrussell:
> (Avoid @'s in commit msgs to reduce github notification spam? cf https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md )

I never understood that: what does it do?  But I get so many notifications I long ago started ignoring them.

There may be another way to mention someone in the commit message, which does not have the same effect?

02-peer-protocol.md
```diff
@@ -618,6 +617,10 @@ A sending node:
     3. if (and only if) `option_shutdown_anysegwit` is negotiated:
       * `OP_1` through `OP_16` inclusive, followed by a single push of 2 to 40 bytes
         (witness program versions 1 through 16)
+    4. if (and only if) `option_simple_close` is negotiated:
+      * `OP_RETURN` followed by one of:
+        * `6` to `75` inclusive followed by exactly that many bytes
+        * `76` followed by `76` to `80` followed by exactly that many bytes
```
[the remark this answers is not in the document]
rustyrussell: Script semantics: 
1. 6-75 push this many bytes on the stack.
2. 76 (aka OP_PUSHDATA1): next byte says how many bytes to push on the stack

02-peer-protocol.md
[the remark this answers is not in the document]
rustyrussell: Yes, I don't know if you received my last shutdown.  But as you say, it will work out, because you'll re-xmit closing_complete when you do.

02-peer-protocol.md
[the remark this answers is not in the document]
rustyrussell: Added that to Rationale.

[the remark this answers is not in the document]
t-bast: Thanks! Bitcoin script is weird :sweat_smile:

t-bast:
One thing that is not 100% clear for me yet is whether a new `shutdown` that changes the `scriptpubkey` requires the receiver to respond with their own `shutdown`? As currently spec-ed it doesn't, so when one side wants to change their `scriptpubkey` they will send `shutdown` and immediately a new `closing_complete`.

The receiver then may or may not send a new `closing_complete` using the new remote `scriptpubkey` (if they don't intend on paying a higher fee, they shouldn't send a new `closing_complete` because it will simply fail to replace the previous one).

I think that this is the right behavior, is that what you intended @rustyrussell? Maybe the diagram example line 573 should be expanded with more scenarios, that would help implementers. This is what I came up with:

```text
        +-------+                              +-------+
        |       | shutdown(scriptA1)           |       |
        |       |----------------------------->|       |
        |       |           shutdown(scriptB1) |       |
        |       |<-----------------------------|       |
        |       |                              |       |
        |       | <complete all pending HTLCs> |       |
        |   A   |             ....             |   B   |
        |       |                              |       |
        |       | closing_complete             |       |
        |       |----------------------------->|       |
        |       |             closing_complete |       |
        |       |<-----------------------------|       |
        |       |                  closing_sig |       |
        |       |<-----------------------------|       |
        |       | closing_sig                  |       |
        |       |----------------------------->|       |
        |       |                              |       |
        |       |   <A updates their script>   |       |
        |       |                              |       |
        |       | shutdown(scriptA2)           |       |
        |       |----------------------------->|       |
        |       | closing_complete             |       |
        |       |----------------------------->|       |
        |       |                  closing_sig |       |
        |       |<-----------------------------|       |
        |       |                              |       |
        |       |  <Both update their script>  |       | (*) This is a concurrent update
        |       |                              |       |
        |       | shutdown(scriptA3)           |       |
        |       |------------------->          |       |
        |       | closing_complete             |       |
        |       |------------------->          |       |
        |       |           shutdown(scriptB2) |       |
        |       |          <-------------------|       |
        |       |             closing_complete |       |
        |       |          <-------------------|       |
        |       |           shutdown(scriptA3) |       |
        |       |          ------------------->|       |
        |       |           closing_complete   |       |
        |       |          ------------------->|       | (*) B doesn't answer with closing_sig because A's sig doesn't use scriptB2
        |       | shutdown(scriptB2)           |       |
        |       |<-------------------          |       |
        |       |   closing_complete           |       |
        |       |<-------------------          |       | (*) A doesn't answer with closing_sig because B's sig doesn't use scriptA3
        |       | closing_complete             |       |
        |       |----------------------------->|       | (*) A now uses scriptB2 and scriptA3 for closing_complete 
        |       |             closing_complete |       |
        |       |<-----------------------------|       | (*) B now uses scriptB2 and scriptA3 for closing_complete 
        |       | closing_sig                  |       |
        |       |----------------------------->|       |
        |       |                  closing_sig |       |
        |       |<-----------------------------|       |
        +-------+                              +-------+
```

02-peer-protocol.md
t-bast: There is one flow happening in each direction, where the sender of `closing_complete` is paying the fee.
Let me know if the proposed diagram [here](https://github.com/lightning/bolts/pull/1096#issuecomment-1743283241) makes it easier to reason about.

02-peer-protocol.md
```diff
@@ -662,119 +665,122 @@ of the receiving node to change the `scriptpubkey`.
...
+  - MUST sign and broadcast the corresponding closing transaction.
+  - MUST send `closing_sig` with a single valid signature in the same tlv field as the `closing_complete`.
+
+The receiver of `closing_sig`:
+  - if `tlvs` does not contain exactly one signature:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - if `tlvs` does not contain one of the tlv fields sent in `closing_complete`:
+    - MUST ignore `closing_sig`.
+  - if the signature field is not valid for the corresponding closing transaction specified in [BOLT #3](03-transactions.md#closing-transaction):
+    - MUST ignore `closing_complete`.
```
t-bast, line 761:
```suggestion
    - MUST ignore `closing_sig`.
```

rustyrussell:
I wanted to quickly note here that Peter Todd points out that LND taproot closes still stand out.

At minimum, we should add nLocktime to the current proposal (rather than always 0).  nVersion is the other thing, but I worry about nodes being unable to sign future versions, and other semantic issues, so we'll do that by future negotiation and features I guess.

rustyrussell:
> One thing that is not 100% clear for me yet is whether a new `shutdown` that changes the `scriptpubkey` requires the receiver to respond with their own `shutdown`? As currently spec-ed it doesn't, so when one side wants to change their `scriptpubkey` they will send `shutdown` and immediately a new `closing_complete`.
> 
> The receiver then may or may not send a new `closing_complete` using the new remote `scriptpubkey` (if they don't intend on paying a higher fee, they shouldn't send a new `closing_complete` because it will simply fail to replace the previous one).

Yes, this is consistent.  It's *polite* to update your own closing tx when they change their scriptpubkey, but it's already uncertain that their change will be effective: presumably the old one could be mined.  So it's not really required.
 
> I think that this is the right behavior, is that what you intended @rustyrussell? Maybe the diagram example line 573 should be expanded with more scenarios, that would help implementers. This is what I came up with:

I like the diagram!

rustyrussell:
See also:

https://github.com/bitcoin/bips/blob/master/bip-0326.mediawiki

02-peer-protocol.md
```diff
@@ -776,6 +794,127 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+Each peer says what fee it will pay, and the other side simply signs that transaction.  The lesser-paid peer (if either is) can opt to omit their own output from the closing tx.
+
+This process will be repeated every time a `shutdown` message is received, which allows re-negotiation (and RBF).
+
+1. type: 40 (`closing_complete`)
+2. data:
+   * [`channel_id`:`channel_id`]
+   * [`u64`:`fee_satoshis`]
+   * [`u32`:`sequence`]
```
t-bast, line 813: What about also including `lockTime` like we do in `open_channel2`?

02-peer-protocol.md
```diff
@@ -776,6 +794,127 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - Select a signature for validation:
+    - if the local output amount is dust:
+      - MUST use `closer_no_closee`.
+    - otherwise, if it considers the closee output amount uneconomic AND its `scriptpubkey` is not `OP_RETURN`:
+      - MUST use `closer_no_closee`.
+    - otherwise, if `closer_and_closee` is present:
+      - MUST use `closer_and_closee`.
+    - otherwise:
+      - MUST use `no_closer_closee`.
```
t-bast, line 878: That is the gentleman's approach, but that isn't the most efficient one for the closee. If the closee receives a proposal that contains `no_closer_closee`, that is the option that pays the most fees, so the closee has an incentive to choose this one instead of `closer_and_closee`?

In practice though, if the sender sends `no_closer_closee`, they won't send `closer_and_closee`, so those cases are only theoretical?

t-bast (commented):
LGTM, there are still a few typos: I left my previous comments for them unresolved. I think we should add `lockTime` to `closing_complete`, apart from that it's looking good.

02-peer-protocol.md
```diff
@@ -776,6 +794,127 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+    - MUST NOT set `no_closer_closee`.
+    - If the closee's output amount is dust:
+      - MUST set `closer_no_closee`.
+      - SHOULD NOT set `closer_and_closee`.
+    - Otherwise:
+      - MUST set both `closer_no_closee` and `closer_and_closee`.
+
+The receiver of `closing_complete` (aka. "the closee"):
+  - If `fee_satoshis` is greater than the closer's outstanding balance:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
```
ProofOfKeags, line 867: Worth noting here that this introduces an edge case *if* we decide to rip out the original coop-close protocol. This would make it such that if you needed to coop-close the channel as the target of a channel-open that you would have no choice but to force close it. I think that everyone loses in this scenario.

```suggestion
The receiver of `closing_complete` (aka. "the closee"):
  - If `fee_satoshis` is greater than the closer's outstanding balance:
    - MAY either send a `warning` and close the connection, or send an `error` and fail the channel.
    - MAY choose to cover the deficit from their own output.
```

This way, if the closer tries to grief you by suggesting an insane fee, you can reject it, but otherwise if it's a fee you'd be willing to pay and it beats your fee liability if you have to force close, you come out ahead. Note that the only realistic scenario where this happens is if you just had a channel opened to you, received no inbound satsflow on the channel but you've decided to close the channel for whatever reason. In this case you may still want to be a good actor and coop-close the channel, but this construction (along with the removal of the existing coop-close) would force you to be more aggressive.

ProofOfKeags: Withdrawn. I misunderstood some of the other aspects of this protocol.

02-peer-protocol.md
morehouse: > There is one flow happening in each direction, where the sender of `closing_complete` is paying the fee. 

Do we really expect dual flows to happen like that in practice?  If the other party offers to pay for the closing transaction, why would we want to send `closing_complete` at all?

> Let me know if the proposed diagram [here](https://github.com/lightning/bolts/pull/1096#issuecomment-1743283241) makes it easier to reason about.

Your diagram helps to clarify corner cases but probably still needs some high-level explanation in the spec.

t-bast: > Do we really expect dual flows to happen like that in practice?

I think we do, because each side should independently decide what fee they'd like to pay and immediately send `closing_complete` (at least that's how I implemented it). Since the channel has already stopped relaying HTLCs at that point, both peers are likely eager to get their funds back asap to reinvest them into another channel.

Roasbeef:
So I'm finally ready for interop here @t-bast!

One thing that came up during my last debugging session was the fact that we allow the `Sequence` field to now be set to anything by either side. 

I'm not sure how y'all handle this in other implementations, but in `lnd` we use the sequence to decide if a funding output spend is either a: breach, local force close, remote force close, coop close, or an unknown state (SCB flow). 

Before this new flow, the sequence was _always_ set to the max possible sequence, which also disabled the RBF (which was want to enable) and also sequence locks semantics. However, if we allow a peer to set an _arbitrary_ sequence, then they can set a sequence that looks like a past breached state (due to extraction of the state hint) or even a future unknown state. So I ended up modifying our implementation to only set/allow a sequence of [` mempool.MaxRBFSequence`](https://github.com/btcsuite/btcd/blob/v0.24.0/mempool/mempool.go#L42-L45) (`0xfffffffd`). This is the largest sequence that still allows opting into the RBF rules. With this change, we're now able to deterministically detect type of spend we see on chain based on the sequence. 

IMO, there's no good reason to set your sequence to anything other than `0xfffffffd` for the co-op close transaction. Activating the BIP 112 semantics isn't very useful here as why would you want to delay your co-op close with a relative time lock? If we allow peers to set arbitrary values, then this also weakens the anonymity set of a coop close from a taproot channel as you'll stand out if you use anything other than `0xfffffffd`. 

So what do y'all think about dropping the sequence field from the new messages? We'd then always restrict it to be `0xfffffffd`, rejecting a sig if we get any other value.

t-bast:
> So I'm finally ready for interop here @t-bast!

:partying_face: 

> I'm not sure how y'all handle this in other implementations, but in lnd we use the sequence to decide if a funding output spend is either a: breach, local force close, remote force close, coop close, or an unknown state (SCB flow).

We don't rely too much on the `nSequence` field for that in eclair. When a transaction spends the funding transaction, we match it by `txid` against:

- splice transactions we signed
- the current local or remote commit
- the next remote commit if we sent `commit_sig` and didn't receive `revoke_and_ack`
- any mutual close transaction we signed
- otherwise, we check if it's a revoked commit by checking `nSequence` and `nLockTime` and the revocation secrets we have
- if that doesn't match, it's an unknown state (we lost data)

You always sign mutual close transactions, so you can remember the corresponding `txid`s and shouldn't need to rely on the `nSequence` field for that?

> With this change, we're now able to deterministically detect type of spend we see on chain based on the sequence.

I think this is dangerous to rely on `nSequence` only for that, it's much safer to rely on `txid`s IMO.

> IMO, there's no good reason to set your sequence to anything other than 0xfffffffd for the co-op close transaction. 

Why restrict it? There are a lot of discussions of leveraging `nSequence` for future bitcoin capabilities, so we should give that door open and allow anything IMO. This is the mutual close transaction paid for by our peer, so we should let them use whatever they want?

> If we allow peers to set arbitrary values, then this also weakens the anonymity set of a coop close from a taproot channel 

I don't think this argument really holds: if your peer wants to do something funny that stands out, they have many other ways to do this anyway, restricting the `nSequence` of the mutual close tx is not going to fix that.

Roasbeef:
> I think this is dangerous to rely on nSequence only for that, it's much safer to rely on txids IMO.

We don't only rely on it, but you might not always have the `txid` (eg: restore from seed, no state, partial data loss, etc). The sequence effectively serves as an authenticated sequence number since only in collaboration with the other party can a new one be signed. All implementations also _must_ use the sequence to detect _which_ date was broadcast assuming they aren't storing the `txid` of every commitment txn ever signed. 

In the past, we also based things _purely_ on `txid`, but discovered some edge cases, causing us to re-write that logic entirely all together. 

>  we match it by txid against any mutual close transaction we signed
> You always sign mutual close transactions, so you can remember the corresponding txids and shouldn't need to rely on the nSequence field for that?


How would you then handle the case of partial or complete data loss, but before that you signed a co-op close transaction, and that hit the chain? 

With the new protocol, do y'all also store very RBF'd close created by either party? Right now we only remember the very last one, as with the new flow, the other party can just continue to sign effectively useless updates to make the peer retain more state (soft DoS). 

> Why restrict it? There are a lot of discussions of leveraging nSequence for future bitcoin capabilities, so we should keep that door open and allow anything IMO. This is the mutual close transaction paid for by our peer, so we should let them use whatever they want?

Can think of two reasons not to: 
  1. You're spending the funding output for a co-op close, meaning you want to exit the channel, so why use a relative time lock? 
  2. If every implementation picks its own value, then it's easier to finger print. Today everyone uses the max sequence for co-op close. In a post taproot chans everywhere world, using a standard value can help co-op closes blend in with the broader transaction anonymity set. 
  3. One less free parameter. 

Any future protocol updates that repurpose the sequence for consensus rules would need to bump the transaction version, just like we did originally for the CSV package. If we follow the line of reasoning for allowing a custom sequence, then should we also allow a custom txn version?

t-bast:
> How would you then handle the case of partial or complete data loss, but before that you signed a co-op close transaction, and that hit the chain?

It simply falls into the "unknown transaction" case, where you cannot do anything and just let it confirm (which sends funds back to your wallet). I believe there is nothing to do in that case and it is harmless?

> With the new protocol, do y'all also store very RBF'd close created by either party?

Yes, we store the `txid` of every closing candidate, as it's quite useful for monitoring (it's also a good way to see how well RBF works in practice).

> Right now we only remember the very last one, as with the new flow, the other party can just continue to sign effectively useless updates to make the peer retain more state (soft DoS).

Since they have to increase the fees every time, that doesn't sound like a dangerous DoS (especially since you only need to store 32 bytes per attempt).

> Can think of two reasons not to:

Those are reasonable reasons, I don't feel very strongly either way (especially since right now, we'll always set the `sequence` to `0xfffffffd`). Let's discuss that during today's spec meeting!

rustyrussell:
> This was discussed during yesterday's spec meeting, and there was rough consensus that the `sequence` should always be set to `0xfffffffd`. Then I believe we should update the `closing_signed` message to be:
> 
> ```
> 1. type: 40 (`closing_complete`)
> 2. data:
>    * [`channel_id`:`channel_id`]
>    * [`u64`:`fee_satoshis`]
>    * [`u32`:`locktime`]
> ```
> 
> Since we want the ability to set the `nLockTime` for anti-fee sniping. We can add the ability to set a different `nSequence` in the future with a new TLV and a feature bit, if it becomes necessary.
> 
> @rustyrussell @Roasbeef does that sound good to you?

Yes, this looks good...

02-peer-protocol.md
ProofOfKeags: Also I fail to understand how we could ever have a transaction be urgent enough to push for confirmation if it involved burning all of the outputs in an OP_RETURN.

02-peer-protocol.md
```diff
@@ -776,6 +794,127 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+If one side has less funds than the other, it may choose to omit its own output, and in this case dust MUST be omitted, to ensure the resulting transaction can be broadcast.
+
+The corner case where fees are so high that both outputs are dust is addressed in two ways: paying a low fee to avoid the problem, or using an OP_RETURN (which is never "dust").
+
+Note that there is usually no reason to pay a high fee for rapid processing, since an urgent child could pay the fee on the closing transactions' behalf.
+
+However, sending a new `shutdown` message overrides previous ones, so you can negotiate again (even changing the output address) if you want: in this case there's a race where you could receive a `closing_complete` for the previous output address, and the signature won't validate.  In this case, ignoring the `closing_complete` is the correct behaviour, as the new `shutdown` will trigger a new `closing_complete` with the correct signature.  This assumption that we only remember the last-sent of any message is why so many cases of bad signatures are simply ignored.
```
Crypt-iQ, line 910: `so you can negotiate again (even changing the output address) if you want`

doesn't this override the safety feature of upfront shutdown?

t-bast: This is of course if you don't have `upfront_shutdown` activated, we can make that clearer.

02-peer-protocol.md
RocketNodeLN: The opener should at least be on the hook for whatever the commitment fee is. Otherwise, as the non-opener, it's cheaper to FC and wait out the time lock. Also, if there's nothing on the non-opener side, is FC the only options since there's nothing on that side to pay the fee.

ProofOfKeags: I want to challenge this idea of "it's cheaper to FC and wait out the time lock". Having money locked up carries opportunity cost of the entire output. Yes you may save in fees, but your outputs are tied up in chain resolution and that is a cost in its own right.

> Also, if there's nothing on the non-opener side, is FC the only options since there's nothing on that side to pay the fee.

Why would a node be concerned about aggressively closing a channel that has no capital tied up in it?

RocketNodeLN: Current coop close is ~168 vB (2 x `P2WPKH` outputs) to ~192 vB (2 x `P2TR` outputs). Sweeping the timelock is 121-123 vB.. so ~35-50% smaller tx. I agree with the time cost, but if you are closing due to months of inactivity, what's another up to two weeks. But the opener would be doubly screwed since they would pay the FC (+ anchors) and then have to pay to sweep. I supposed the opener can "deter" this by keeping max commit fees low.

Closing channels should be discouraged, but having dead/inactive channel is worse.

In my mind, game theory would have me coop close at a very low fee if the balance is mostly on the other side or FC if it's mostly on my side and I'm not the opener and commit fee is high enough. 

>Why would a node be concerned about aggressively closing a channel that has no capital tied up in it?

More of a hypothetical.. how does a closer pay if there's little to nothing on their side.

I agree a change to channel closing is needed, but 100% opener to 100% closer is a pretty drastic change. I would like to see the opener using (some of) the commit fee already reserved and/or splitting the fee based on balance.

PR #1097, opened 2023-07-24
Blinded Paths: clarify outgoing_cltv_value for final hop

carlaKC:
This PR replaces #1066, picking this up as discussed in the spec meeting!

Senders making payments to a blinded route know the total `cltv_delta` for the blinded portion of the route, but not the `min_final_cltv_delta` for the final hop. This PR updates the instructions for the sender to set the final hop's `outgoing_cltv_detla` to `current_height + any random offset added`.

See [gist](https://gist.github.com/t-bast/b1371d357a2c5f3e8c09514a62db7079) for additional context.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -254,7 +254,7 @@ The creator of `encrypted_recipient_data` (usually, the recipient of payment):
     - `total_fee_proportional_millionths(n+1) = ((total_fee_proportional_millionths(n) + fee_proportional_millionths(n+1)) * 1000000 + total_fee_proportional_millionths(n) * fee_proportional_millionths(n+1) + 1000000 - 1) / 1000000`
   - MUST create the `encrypted_recipient_data` from the `encrypted_data_tlv` as required in [Route Blinding](#route-blinding).
 
-The writer of `tlv_payload`:
+The writer of the TLV `payload`:
 
   - For every node inside a blinded route:
     - MUST include the `encrypted_recipient_data` provided by the recipient
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -262,6 +262,10 @@ The writer of the TLV `payload`:
       - MUST include the `blinding_point` provided by the recipient in `current_blinding_point`
     - If it is the final node:
       - MUST include `amt_to_forward`, `outgoing_cltv_value` and `total_amount_msat`.
+      - The value set for `outgoing_cltv_value`: 
+        - MUST use the current block height as a baseline value. 
+        - if a [random offset](07-routing-gossip.md#recommendations-for-routing) was added to improve privacy:
+          - SHOULD add the offset to the baseline value.
     - MUST NOT include any other tlv field.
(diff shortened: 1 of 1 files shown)


t-bast (approved):
LGTM :+1: 

People who implemented blinded paths should definitely take a look at https://gist.github.com/t-bast/b1371d357a2c5f3e8c09514a62db7079 and make sure I didn't miss anything (otherwise there may be a probing issue).

04-onion-routing.md
```diff
@@ -254,14 +254,18 @@ The creator of `encrypted_recipient_data` (usually, the recipient of payment):
...
+The writer of the TLV `payload`:
 
   - For every node inside a blinded route:
     - MUST include the `encrypted_recipient_data` provided by the recipient
     - For the first node in the blinded route:
       - MUST include the `blinding_point` provided by the recipient in `current_blinding_point`
     - If it is the final node:
       - MUST include `amt_to_forward`, `outgoing_cltv_value` and `total_amount_msat`.
+      - The value set for `outgoing_cltv_value`: 
+        - MUST use the current block height as a baseline value.
```
ProofOfKeags, line 266: If I understand this correctly, the `outgoing_cltv_value` is a replacement integrity check on the final onion packet to make sure that nodes along the route haven't laid claim to larger shares of the original CLTV budget than they were alotted.

Looking [here](https://github.com/lightning/bolts/blob/master/04-onion-routing.md#payload-format) it states that the final node:

> MUST return an error if incoming `cltv_expiry` < `current_block_height` + `min_final_cltv_expiry_delta`.

It seems to me that if I read your change as written then the final HTLC can be exactly equal to the current block height. This means that without the random offset (specified as SHOULD), then there is zero "headroom" to resolve the HTLC.

Maybe this edge case is allowable, but it does seem to me that we should EITHER: 
1. relax the baseline from MUST to SHOULD
2. constrict the offset recommendation from SHOULD to MUST
3. add a mandatory offset > 0 to the current block height as the baseline

Leaving this as is is also acceptable but it would mean that valid implementations of the spec can run into this weird edge case.

t-bast: > It seems to me that if I read your change as written then the final HTLC can be exactly equal to the current block height. This means that without the random offset (specified as SHOULD), then there is zero "headroom" to resolve the HTLC.

This is not an issue, for two different reasons:

1. The same issue exists with non-blinded payments: this is why implementations currently already add a small random offset to the current block height to avoid this issue
2. Even if the sender doesn't add a random offset, the recipient will add one in the blinded path's total `cltv_expiry_delta`, to make sure it covers its `min_final_cltv_expiry_delta` (to allow the recipient enough time before fulfilling this HTLC without risk of the downstream node using the HTLC-timeout path)

ProofOfKeags (approved):
ACK dd0af11a70f77b0b74f75e79708ac841a75f1749

valentinewallace (approved):

rustyrussell:
Ack!

PR #798, opened 2020-08-31
Offers

rustyrussell:
Offers are "static invoices", with many more features.  This is mostly implemented in c-lightning already, and is now ready for wider testing.

This PR is based on onion messages specified in #759

PR #1099, opened 2023-07-31
Reduce size of `offer` blinded paths

jkczyz:
Offers may contain blinded paths to allow for greater recipient privacy. However, they come at a cost of increased QR code size as each hop requires a 33-byte `point` for the `next_node_id`. Allow using `short_channel_id` instead, which only requires 8 bytes. Further, allow for specifying the introduction node as either a `point` or a `short_channel_id` in blinded paths via a new `node_id` fundamental type.

Based on #798.


t-bast:
While I definitely understand the appeal, I think that before depending on `scid`, we should think about what we want to do with `scid` in the future. They currently directly refer to the channel's funding utxo, which is simple enough. But we have two major features coming that may impact that:

- splicing: when we splice, should we keep referring to the channel using its **initial** `scid`? We probably do, because otherwise we'll invalidate routing maps. Are there any downsides to this approach?
- private gossip: once we stop advertising the real on-chain utxo for a channel, will we keep some concept of `scid` or will we get rid of them entirely? How will we avoid conflicts when generating fake `scid`s?

rustyrussell:
> > private gossip: once we stop advertising the real on-chain utxo for a channel, will we keep some concept of scid or will we get rid of them entirely? How will we avoid conflicts when generating fake scids?
> 
> Yea this is way more complicated. If we imagine a ZKP world there's no way to get SCIDs. I think this may be far enough away that we don't need to worry about it just yet - if we do simple overcommitment, there's at least always an SCID for one channel that the node has (or, we could make that a requirement), so this proposal doesn't impact that. The proposal does allow for pubkeys still (or it should if not) so at least it'll still work if we move to a ZKP world.

My rough plan for "non-proven channels" was to add shadow ids.  That is, if you prove 1111x22x3 as your txout, then you can also own 1111x22+Nx3 where N is the rounded-to-power-of-2 number of txs in block 1111 (thus you cannot merkle prove it).  This makes it easy to spot them, and figure out the UTXO whose proof they're sharing, and gives a fair bit of headroom (txnum is 16 bits, usually leaving 10 bits to play with).  You only need to remember the number of bits per block.

If we go full zkp, then this breaks down, and scids either become 32-bit local identifiers or are eliminated entirely.

12-offer-encoding.md
```diff
@@ -214,16 +214,27 @@ The human-readable prefix for offers is `lno`.
...
     2. data:
         * [`...*utf8`:`issuer`]
-    1. type: 20 (`offer_quantity_max`)
+    1. type: 22 (`offer_quantity_max`)
     2. data:
         * [`tu64`:`max`]
-    1. type: 22 (`offer_node_id`)
+    1. type: 24 (`offer_node_id`)
     2. data:
         * [`point`:`node_id`]
```
rustyrussell, line 228: I understand the desire for cleanliness, but this breaks everyone's test code.  Just make offer_compact_paths number 24 please!

12-offer-encoding.md
```diff
@@ -378,13 +395,16 @@ for [Signature Calculation](#signature-calculation).
...
+    1. type: 20 (`offer_issuer`)
     2. data:
         * [`...*utf8`:`issuer`]
-    1. type: 20 (`offer_quantity_max`)
+    1. type: 22 (`offer_quantity_max`)
     2. data:
         * [`tu64`:`max`]
-    1. type: 22 (`offer_node_id`)
+    1. type: 24 (`offer_node_id`)
     2. data:
```
rustyrussell, line 408: here too

12-offer-encoding.md
```diff
@@ -550,13 +570,16 @@ the `onion_message` `invoice` field.
...
+        * [`...*compact_blinded_path`:`paths`]
+    1. type: 20 (`offer_issuer`)
     2. data:
         * [`...*utf8`:`issuer`]
-    1. type: 20 (`offer_quantity_max`)
+    1. type: 22 (`offer_quantity_max`)
     2. data:
         * [`tu64`:`max`]
-    1. type: 22 (`offer_node_id`)
+    1. type: 24 (`offer_node_id`)
```
rustyrussell, line 582: here too

rustyrussell (commented):
All looks good except for renumbering...

jkczyz:
> For the introduction node, indeed, it's more of an issue...

Should we consider dropping SCID for the introduction node? It would make the offer format a little cleaner at very least by not having two separate path TLV records... along with removing the implementation complexity that comes along with it.

rustyrussell:
Radical idea: a new bolt fundamental type: sciddir_or_pubkey:

- If first byte is 0 or 1, that's the direction, and it's followed by 8 bytes of scid for 9 byte total.
- If first byte is 2 or 3, it's a pubkey, for 33 byte total.

There are several places where this could be used, IMHO.  And it's backwards compat with existing pubke fields if you *always* use a pubkey...

jkczyz:
> Radical idea: a new bolt fundamental type: sciddir_or_pubkey:
> 
> * If first byte is 0 or 1, that's the direction, and it's followed by 8 bytes of scid for 9 byte total.
> * If first byte is 2 or 3, it's a pubkey, for 33 byte total.
> 
> There are several places where this could be used, IMHO. And it's backwards compat with existing pubke fields if you _always_ use a pubkey...

Nice! I ended up naming it `node_id` and defined it in terms of `short_channel_id` and `point`. Decided against using "direction" in any way since the first byte simply determines whether `node_id_1` or `node_id_2` from the `channel_announcement` is used not which direction something applies to like `channel_update`. At least in how it is used in `blinded_path`. Maybe the same doesn't apply to other uses that you are envisioning?

Other alternatives could be:
- `node_id_or_ref`
- `point_or_ref`

Let me know if you have a preference. One argument against `node_id` is that it is already used in many field names for `point` types.

thomash-acinq:
I really like the `sciddir_or_pubkey`.
However I think the main problem here is that channels get opened and closed all the time and we shouldn't rely on an identifier that can disappear at any time. One solution would be to never forget channel ids but that's not very satisfying. Another solution would be for each node to chose one of their channel id and signal that this one should be remembered forever.

rustyrussell:
Will fold into #798

rustyrussell:
Yeah, "node_id" is counter-intuitive (internally, we have a node_id type already, it's a SECP1-encoded u8[33]).  sciddir_or_pubkey is clunky, but works, so I'll switch to that...

rustyrussell:
Hmm, since #798 has been rebased, there was a significant conflict with the first patch.  I think I got it right (though I insisted on exaclty one of scid or nodeid).

t-bast:
@jkczyz since this has been integrated into #798, I believe we can close this PR?

PR #1100, opened 2023-08-09
[QUESTION] bolt11: A writer MUST include "c" but not in examples

dni:
If the requirement is that a writer MUST include `c`, how come that the examples not all have `c`?

should there be a `c` in the examples? 
or is it really a MAY?

ran into this by implementing that writer exception and running the testsuite through the examples.

am i misunderstanding something?
thanks for the review.

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -177,7 +177,7 @@ A writer:
   - MAY include one `x` field.
     - if `x` is included:
       - SHOULD use the minimum `data_length` possible.
-  - MUST include one `c` field (`min_final_cltv_expiry_delta`).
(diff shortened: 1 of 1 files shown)


t-bast:
Good catch thanks, though I think a `SHOULD` would make more sense here.

dni:
is it really SHOULD? all the others on this levels are MAY.
https://github.com/lightning/bolts/blob/27af3bd1c0db4ad51763131cd7596efcb563edba/11-payment-encoding.md#requirements-3

i can change it if you want :)

t-bast:
Thanks!

t-bast (approved):

PR #1104, opened 2023-08-26
BOLT2: change "reject" to "fail" in channel opening requirements

ProofOfKeags:
Since "reject" isn't defined anywhere in the spec and "fail" does have a definition laid out in BOLT1 that includes special provisions for channels that have not yet been established, BOLT2 is amended to clarify the requirements for the receiver of `accept_channel`


t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1104/commits/87898753cff906300b7bcb0bd57917aed132d080

PR #1108, opened 2023-09-11
CONTRIBUTING.md: modern feature bit assignment.

rustyrussell:
1. Put it in the PR title so everyone can see.
2. Deploy with +100 while it's still unratified, in case it changes.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,11 @@ escape into the wild).
 also they can be used to find appropriate peers.
 
 Feature bits are always assigned in pairs, even if it doesn't make sense
-for them to ever be compulsory.
+for them to ever be compulsory.  The feature bit is self-assigned in the
+title of the PR, to make it easier for others to self-assign.  Until the PR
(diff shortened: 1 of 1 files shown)


t-bast (approved):
ACK

carlaKC (approved):

vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1108/commits/58473d03216d50b59c13be682bf2dbeef2e52f85

t-bast:
Should we extend that recommendation to messages and tlv fields introduced by experimental features?

This isn't strictly necessary: an implementation that wants to support both the experimental version and the official version could encode/decode messages and tlvs differently depending on what feature was negotiated (the experimental feature bit or the official one). But it's a bit cumbersome (at least in eclair we'd rather use separate messages for the experimental version of the feature). Maybe at least worth mentioning that this is a concern that implementers should worry about?

ProofOfKeags (approved):
ACK

PR #1109, opened 2023-09-13
BOLT 1: define what `offered` and `negotiated` mean.

rustyrussell:
i.e. it was present in the init feature bits.  We use this in several places, but assume everyone knows what it means.

It's particularly fraught with required features: it's explicitly legitimate to assume these are accepted if the node keeps talking to you after init!


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1109/commits/95f2bbf2e29e2f1c53e956a69b4c59c6e4b5d089

t-bast:
We don't have such code in eclair yet, if our peer sets a feature bit as mandatory that we don't set, we always disconnect.
For which feature does `cln` currently handle that mode?

ProofOfKeags (approved):

Roasbeef (approved):
LGTM 🦠

Roasbeef:
>  if our peer sets a feature bit as mandatory that we don't set, we always disconnect.

This is also lnd's behavior.

PR #1110, opened 2023-09-15
Peer storage for nodes to distribute small encrypted blobs.

adi2011:
This pull request proposes peer storage, inspired by @t-bast's [peer backup storage](https://github.com/lightning/bolts/pull/881). We create two separate messages, which users can use to distribute encrypted 'blobs' of up to 64 KB to their peers.

Users can distribute a 'blob' and update it whenever their configuration changes by sending a 'peer_storage' message. In return, they will always receive a 'your_peer_storage' message, which will serve as an acknowledgment.

Nodes can also decide whether they want to provide this service in exchange for some sats.

You can check out the implementation in CLN [here](https://github.com/ElementsProject/lightning/pull/5361)

Thank you so much, @rustyrussell, for all the help and guidance!

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -26,6 +26,8 @@ All data fields are unsigned big-endian unless otherwise specified.
     * [The `error` and `warning` Messages](#the-error-and-warning-messages)
   * [Control Messages](#control-messages)
     * [The `ping` and `pong` Messages](#the-ping-and-pong-messages)
+  * [Peer Storage](#peer-storage)
+    * [The `peer_storage` and `peer_storage_retrieval` Messages](#the-peer_storage-and-peer_storage_retrieval-messages)
   * [Appendix A: BigSize Test Vectors](#appendix-a-bigsize-test-vectors)
   * [Appendix B: Type-Length-Value Test Vectors](#appendix-b-type-length-value-test-vectors)
   * [Appendix C: Message Extension](#appendix-c-message-extension)
@@ -493,6 +495,79 @@ every message maximally).
 
 Finally, the usage of periodic `ping` messages serves to promote frequent key
 rotations as specified within [BOLT #8](08-transport.md).
+
+## Peer storage
+
+### The `peer_storage` and `peer_storage_retrieval` Messages
+
+Nodes that advertise the `option_provide_storage` feature offer storing
+arbitrary data for their peers. The data stored must not exceed 65531 bytes,
+which lets it fit in lightning messages.
+
+Nodes can verify that their `option_provide_storage` peers correctly store
+their data at each reconnection, by comparing the contents of the
+retrieved data with the last one they sent. However, nodes should not expect
+their peers to always have their latest data available.
+
+Nodes ask their peers to store data using the `peer_storage` message and expect
+peers to return the latest data to them using the `peer_storage_retrieval` message:
+
+1. type: 7 (`peer_storage`)
+2. data:
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+1. type: 9 (`peer_storage_retrieval`)
+2. data:
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+Requirements:
+
+The sender of `peer_storage`:
+  - MAY send `peer_storage` whenever necessary.
+  - MUST limit its `blob` to 65531 bytes.
+  - MUST encrypt the data in a manner that ensures its integrity upon receipt.
+  - SHOULD pad the `blob` to ensure its length is always exactly 65531 bytes.
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store the message anyway.
+
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per minute.
+  - MUST replace the old `blob` with the latest received.
+  - MUST send `peer_storage_retrieval` again after reconnection, after exchanging `init` messages.
+
+
+The sender of `peer_storage_retrieval`:
+  - MUST include the last `blob` it stored for that peer.
+  - when all channels with that peer are closed:
+    - SHOULD wait at least 2016 blocks before deleting the `blob`.
+
+The receiver of `peer_storage_retrieval`:
+  - when it receives `peer_storage_retrieval` with an outdated or irrelevant data:
+    - MAY send a warning.
+
+#### Rationale:
+The `peer_storage` and `peer_storage_retrieval` messages enable nodes to securely store
+and share data with other nodes in the network, serving as a backup mechanism for important
+information. By utilizing them, nodes can safeguard crucial data, enhancing the network's
+resilience and reliability. Additionally, even if we don't have an open channel, some nodes
+might provide this service in exchange for some sats so they may store `peer_storage`.
+
+Nodes should pad the `blob` to obscure its actual size, enhancing privacy by making size-based
+analysis more difficult for the receiver.
+
+`peer_storage_retrieval` should not be sent after `channel_reestablish` because then the user
+wouldn't have an option to recover the node and update its state in case they lost data.
+
+Nodes should send a `peer_storage` message whenever they wish to update the `blob` stored with their peers.
+This `blob` can be used to distribute encrypted data, which could be helpful in restoring the node.
 
 ## Appendix A: BigSize Test Vectors
 
--- a/09-features.md
+++ b/09-features.md
@@ -46,6 +46,7 @@ The Context column decodes as follows:
 | 26/27 | `option_shutdown_anysegwit`       | Future segwit versions allowed in `shutdown`              | IN       |                           | [BOLT #2][bolt02-shutdown]                                            |
 | 28/29 | `option_dual_fund`                | Use v2 of channel open, enables dual funding              | IN       |                           | [BOLT #2](02-peer-protocol.md)                                        |
 | 38/39 | `option_onion_messages`           | Can forward onion messages                                | IN       |                           | [BOLT #7](04-onion-routing.md#onion-messages)                         |
+| 42/43 | `option_provide_storage`          | Can store other nodes' encrypted backup data              | IN       |                           | [BOLT #1](01-messaging.md#peer-storage)                               |
 | 44/45 | `option_channel_type`             | Node supports the `channel_type` field in open/accept     | IN       |                           | [BOLT #2](02-peer-protocol.md#the-open_channel-message)               |
 | 46/47 | `option_scid_alias`               | Supply channel aliases for routing                        | IN       |                           | [BOLT #2][bolt02-channel-ready]                                       |
 | 48/49 | `option_payment_metadata`         | Payment metadata in tlv record                            | 9        |                           | [BOLT #11](11-payment-encoding.md#tagged-fields)                      |


01-messaging.md
```diff
@@ -476,6 +476,77 @@ every message maximally).
...
+The receiver of `your_peer_storage`:
+  - when it receives `your_peer_storage` with an outdated or irrelevant data:
+    - MAY send a warning.
+
+Rationale:
+If a peer supports `option_provide_storage` and we have a channel with it, we can send a `peer_storage` message with any `blob` so that peer can store it locally in their node. Everytime a node receives a `peer_storage` message, It needs to update the `blob` for the sender.
+
+Upon receiving `peer_storage` nodes should immediately respond with the latest `your_peer_storage`, which could work as an ack to the peer.
+
+Nodes with `option_provide_storage` should also regularly send `your_peer_storage` to verify that they store the latest and correct form of `blob`. Nodes may want to ignore `peer_storage` messages with peers they don’t have a channel with to avoid spam.
```
saubyk, line 546: This part is unclear to me. Based on my reading, there are two instances when a node (running `option_provide_storage`) will send `your_peer_storage` message:
- When it gets a `peer_storage` message and has stored the data (like an ack)
- When it reconnects with the peer for which it has stored data (to ensure that peer has a chance to update stale data)

The point about "regularly" sending `your_peer_storage` seem to be specified loosely.
- What's the right frequency to send this to your peer?
- Does it create a spam vector for your peer?

01-messaging.md
```diff
@@ -476,6 +476,77 @@ every message maximally).
...
+1. type: 9 (`your_peer_storage`)
+2. data:
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+Requirements:
+
+The sender of `peer_storage`:
+  - MUST support `option_want_storage`.
```
rustyrussell, line 512: s/support/offer/

01-messaging.md
```diff
@@ -476,6 +476,77 @@ every message maximally).
...
+  - MAY send `peer_storage` to only peers who support `option_provide_storage`.
```
rustyrussell, line 513: You can actually drop this...

01-messaging.md
```diff
@@ -476,6 +476,77 @@ every message maximally).
...
+  - MUST limit its `blob` to 65531 bytes.
+  - MUST encrypt the data in a way that can be validated when received.
+
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store anyway (e.g. might get paid for the service).
+    - MUST replace the old `blob` with the latest received.
```
rustyrussell, line 524: Move this under the "if it does store the message"?

01-messaging.md
```diff
@@ -476,6 +476,77 @@ every message maximally).
...
+
+- If it does store the message:
+  - MUST immediately respond with `your_peer_storage` containing `blob`
```
rustyrussell, line 527: Perhaps we can adapt this slightly to address spam?

```
- If it does store the message:
  - MAY delay storage to ratelimit peer to no more than one update per second.
  - MUST respond with `your_peer_storage` containing `blob` after it is stored.
``` 

?

01-messaging.md
rustyrussell: Agreed. You don't need to specify this, since you already said in requirements that "MAY send `peer_storage` whenever necessary (usually, everytime their configuration changes)".

And I made a suggestion for spam above...

adi2011: Yes, the two instances that you mentioned are the only time when a peer needs to send `your_peer_storage`.

01-messaging.md
adi2011: agreed

01-messaging.md
adi2011: Yes, a ratelimit should be there to prevent spam. Nice Idea!

01-messaging.md
adi2011: Hmm, that's make it more clear.

saubyk (approved):
Ack

t-bast:
As stated during the spec meeting, I think this should be moved to the bLIP repository: https://github.com/lightning/blips
This feels like an extension/add-on, not something that really needs to be in the BOLTs (especially if we're trying to somewhat reduce complexity here).
I wanted to move my encrypted storage proposal to a bLIP, but didn't do it because I was rather waiting for someone else to work on it first, which is happening now.

adi2011:
I think in order to make peer storage reliable, it should have maximum coverage across the network. This approach attempts to address the backup issue, which is essential for enhancing the usability of the Lightning Network. Many small node operators may be reluctant to invest in RAID systems to protect their nodes; thus, this could serve as one of the safety nets for the network.

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+  - MAY send `peer_storage` whenever necessary (usually, everytime their configuration changes)
+  - MUST limit its `blob` to 65531 bytes.
+  - MUST encrypt the data in a way that can be validated when received.
+
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store anyway (e.g. might get paid for the service).
```
GeorgeTsagk, line 522: > (e.g. might get paid for the service)

Could provide a bit more context here w.r.t being paid for the purpose of storing blobs

Is this why `if it has an open channel with the sender` is a requirement?

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+Requirements:
+
+The sender of `peer_storage`:
+  - MUST offer `option_want_storage`.
+  - MAY send `peer_storage` whenever necessary (usually, everytime their configuration changes)
+  - MUST limit its `blob` to 65531 bytes.
```
GeorgeTsagk, line 514: could also consider padding so that it is always fixed to 65531 bytes

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+their data at each reconnection, by comparing the contents of the
+retrieved data with the last one they sent. However, nodes should not expect
+their peers to always have their latest data available.
+
+Nodes ask their peers to store data using the `peer_storage` message & expect peers to return them latest data using `your_peer_storage` message:
+
+1. type: 7 (`peer_storage`)
+2. data:
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
```
GeorgeTsagk, line 500: Could also extend this in order to have key-values instead of one single storage blob?
```suggestion
   * [`u16`:`key`]
   * [`length*byte`:`blob`]
```

vincenzopalazzo (commented):
I concur with @t-bast that this feature could be a BLIP.

Since this is not a one-size-fits-all solution, we can create different BLIPs for various backup use cases (if any others are proposed in the future).

If we want to add this to the BOLTs, we should address some concerns like this one https://github.com/ElementsProject/lightning/issues/6652

01-messaging.md
adi2011: Yes, good idea, but I think that would be more useful in case we want to distribute multiple backups (or backup with >65kb data).

In that case, we'd also have to specify which `key` would be considered `ack`, and this could lead to spamming (i.e. Peer can send me many blobs that could fill up my storage) that is why the current scheme just replaces the old storage blob with a new one.

01-messaging.md
adi2011: Sure, this needs more explanation, I'll add to it when I get back to my laptop.

01-messaging.md
ProofOfKeags: Why would you want to pad?

01-messaging.md
ProofOfKeags: > Is this why if it has an open channel with the sender is a requirement?

I'd imagine this also helps with spam prevention. Since in order to do a drive-by to store data you'd have to commit capital.

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+
+Nodes that advertise the `option_want_storage` feature want to have their
+data stored by their peers that support `option_provide_storage`.
+
+Nodes can verify that their `option_provide_storage` peers correctly store
+their data at each reconnection, by comparing the contents of the
+retrieved data with the last one they sent. However, nodes should not expect
+their peers to always have their latest data available.
+
+Nodes ask their peers to store data using the `peer_storage` message & expect peers to return them latest data using `your_peer_storage` message:
```
ProofOfKeags, line 495: nit: I'm not a fan of these message names. Can I suggest

`blob_store` instead of `peer_storage` and `blob_retrieval`. Perhaps it's not easy to come up with truly satisfactory names, but I'd love to avoid using "your" anywhere because of the level of context required to reason about who is on what side of the connection etc.

adi2011: Yes, these are good suggestions but the names should suggest that they come from one of your peers. blob_* doesn't imply that.

ProofOfKeags: Don't all of the messages come from our peers?

adi2011: I meant peer_storage suggests that the blob inside the msg is coming from your peer's storage, `blob` is a bit abstract.

Although I agree using `your` is not a good idea. How about `peer_storage_retrieval`?

adi2011: Sorry, I meant "blob_store" might not necessarily imply that the data comes from a peer's storage. Unlike "peer_storage," which explicitly denotes that the data is stored by peers within the network.

Although I agree using `your` is not a good idea. How about `peer_storage_retrieval`?

ProofOfKeags: Yeah and maybe we can label the other one `peer_storage_request`. Makes it pretty clear the roles each one plays.

adi2011: Yes, this is a good suggestion, I'll modify the doc. Thanks! :)

01-messaging.md
ellemouton: I think the reasoning would be privacy related: ie so that receiver cant deduce anything about the peer sending the data based on size changes.

That being said: I think all of that detail doesnt need to be specced out on this layer. How the blob is structured is the responsibility of the layer above. 

Especially if the layer above is a "storage for payment" layer where they are charging per byte (or byte bucket) then we really dont want to force them to use the full 65k here.

01-messaging.md
ellemouton: > Could also extend this in order to have key-values instead of one single storage blob?

Under the hood it might be a key-value structure - but think that that is better left to the above layer.

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+1. type: 9 (`your_peer_storage`)
+2. data:
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+Requirements:
+
+The sender of `peer_storage`:
+  - MUST offer `option_want_storage`.
```
ellemouton, line 512: dont think i fully understand why we need the `option_want_storage` feature bit..? 

isnt it enough that im connecting to a node with `option_provide_storage` and sending `peer_storage` to it? Or is there a case where someone would want to seek out a node advertising `option_want_storage`?

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+data stored by their peers that support `option_provide_storage`.
+
+Nodes can verify that their `option_provide_storage` peers correctly store
+their data at each reconnection, by comparing the contents of the
+retrieved data with the last one they sent. However, nodes should not expect
+their peers to always have their latest data available.
+
+Nodes ask their peers to store data using the `peer_storage` message & expect peers to return them latest data using `your_peer_storage` message:
+
+1. type: 7 (`peer_storage`)
```
ellemouton, line 497: should we not use even types for these messages? since we are only sending these if the peer advertises `option_provide_storage`

01-messaging.md
ProofOfKeags: We already handle this with the randomization of ping/pong responses. The actual messages themselves are encrypted as well. Padding the message out to 65531 actually negatively impacts privacy as best I can tell since now we can do partial inference and filter out all non-max-length messages from the anonymity set.

01-messaging.md
ProofOfKeags: > Under the hood it might be a key-value structure - but think that that is better left to the above layer.

+1. Single Opaque Buffer is simplest to implement and all other data structures can be implemented atop this if we want. KV storage also introduces the question of what we are supposed to send back on reconnect.

01-messaging.md
ProofOfKeags: I'm inclined to agree. I think in practice it is very likely that people will implement this protocol wholesale or not at all (it's pretty simple). That said, I can imagine scenarios where you may specifically not want to provide this feature but still would like to consume it. At 65k per peer though, idk what the big deal would be with providing it anyway.

01-messaging.md
ProofOfKeags: does IOKTBO apply to message types?? I thought it was just TLVs

ellemouton: yeah it applies to messages too afaiu: https://github.com/lightning/bolts/blob/master/01-messaging.md#lightning-message-format

ProofOfKeags: I don't think this applies to the low message type numbers though afaict. If it did then things like `accept_channel` would be treated the same right? I would support using even messages, no issue with that at all, but I think I'm not quite understanding the intricacies of IOKTBO

ellemouton: yeah i think for some of the very first messages this was sort of overlooked but it wasnt a problem since nothing would work if a node didnt understand something fundamental like `accept_channel`

01-messaging.md
ProofOfKeags: Are you suggesting they will only want to offer it to some of their peers or rather in that situation they wouldn't want to offer it to any of them?

01-messaging.md
```diff
@@ -476,6 +476,75 @@ every message maximally).
...
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store anyway (e.g. might get paid for the service).
+
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per second.
+  - MUST respond with `your_peer_storage` containing `blob` after it is stored.
```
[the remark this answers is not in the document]
ProofOfKeags: we can also ACK with a sha256 of the data provided.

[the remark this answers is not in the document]
t-bast: I agree with @TheBlueMatt here, our proposal for this feature (https://github.com/lightning/bolts/pull/881) does not have an ACK because I didn't find that useful.

09-features.md
```diff
@@ -45,6 +45,8 @@ The Context column decodes as follows:
 | 24/25 | `option_route_blinding`           | Node supports blinded paths                               | IN9      | `var_onion_optin`         | [BOLT #4](bolt04-route-blinding)                                      |
 | 26/27 | `option_shutdown_anysegwit`       | Future segwit versions allowed in `shutdown`              | IN       |                           | [BOLT #2][bolt02-shutdown]                                            |
 | 38/39 | `option_onion_messages`           | Can forward onion messages                                | IN       |                           | [BOLT #7](04-onion-routing.md#onion-messages)                         |
+| 40/41 | `option_want_storage`             | Want to use other nodes to store encrypted backup data    | IN       |                           | [BOLT #1][bolt01-Peer-storage]                                        |
```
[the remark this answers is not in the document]
t-bast: That's a good point: nodes who want to store their data will look for nodes who offer the `option_provide_storage`, but we probably don't need anything more. Whenever the storing node has advertised `option_provide_storage`, anyone should be allowed to send them the corresponding backup messages.

01-messaging.md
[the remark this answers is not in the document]
adi2011: Thanks for the review. I'll add this to rationale.

01-messaging.md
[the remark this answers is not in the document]
adi2011: It serves as assurance to the sender, although not foolproof, that it won't be excessively wasteful. Additionally, updating peer_storage won't be a frequent occurrence.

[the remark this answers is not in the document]
adi2011: Although I agree that it's not necessary.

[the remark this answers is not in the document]
t-bast: > Additionally, updating peer_storage won't be a frequent occurrence.

That's not a good assumption at all! It will on the contrary be _very_ frequent. The way we use it today is to update the backup at every signed change in a channel, to ensure we're able to resume channels immediately after data loss.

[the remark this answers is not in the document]
ProofOfKeags: > Don't make me hash the data, either

Would you settle for md5? Or do we need to go all the way to CRC32? Like, on one hand, I understand a resistance to performing extra compute, but on the other hand...it's 65k of data. There are already provisions for rate limiting, so how much does this really matter?

If we want to drop the ACK, I can live with that, but I'm not sure I understand what is behind the strong resistance.

01-messaging.md
ellemouton: > Big nodes with large number of peers might not be interested in giving this service to all of their peers.

I dont think having/not having `option_want_storage` changes this though?

ellemouton: the only reason you would advertise `option_want_storage` is so that nodes offering the service could seek you out. but that doesnt really make sense to me. Makes more sense that if you want storage - you would seek out nodes that can offer it to you.

09-features.md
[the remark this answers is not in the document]
ellemouton: agreed - other thread discussing this here too: https://github.com/lightning/bolts/pull/1110/files#r1519349651

01-messaging.md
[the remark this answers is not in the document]
ellemouton: just to clarify: we would keep the blob response on re-start right? need someway of requesting our back-up from the peer. 

Also - if we do only send this on re-establish, would it maybe be a good idea to potentially put it back in the actual re-establish message as a tlv instead of adding a whole new message?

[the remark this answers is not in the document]
ellemouton: ah - actually.. just realised that then it would only work for peers we have channels with... which might not be a wanted restriction

09-features.md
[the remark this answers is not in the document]
adi2011: Yes, I think `option_want_storage` is redundant...

01-messaging.md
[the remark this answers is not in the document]
adi2011: The sole reason for ACK was to assure the sender, is their any other efficient method to do this?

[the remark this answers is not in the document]
t-bast: But does it really assure the sender of anything meaningful? If used to store every channel state change (which is what we've been doing for years), we don't want to pause the channel's operation waiting for an ACK, it adds unnecessary complexity and risk of force-closing. This is mostly being used by nodes who will regularly disconnect (e.g. mobile phones), who have an opportunity of checking the back-up returned on every connection reestablishment, which is IMO enough assurance that the storing node behaves correctly?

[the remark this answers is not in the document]
ProofOfKeags: I'd like to remind us that we didn't *need* ACKs for every channel update either, and yet we are nudging ourselves towards a simplified update protocol. Granted, the purpose of this is far less consequential than ACK'ing the channel updates themselves, so we can afford to be like "🤷🏻‍♂️ who knows if they got it, we'll just assume so".

My take here is that at minimum it helps with debugging, and the cost of the ACK seems rather trivial so avoiding it seems silly. Yes, it doesn't provide much, we can always send back the packet, or the hash, or some ID, and then just yeet the data into oblivion, but does it really cost us much to just ACK it?

09-features.md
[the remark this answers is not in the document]
ProofOfKeags: Yeah let's ditch it

01-messaging.md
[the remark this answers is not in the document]
adi2011: Removed ACK as a requirement. Still mentioned in the rationale so that implementations can decide for themselves.

01-messaging.md
ProofOfKeags: oops I misread this. Indeed the receiver may come to learn about the contents of the payload through size analysis. My commentary on ping/pong length randomization is about 3p privacy, not 2p privacy.

ProofOfKeags (approved):
ACK

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store anyway.
+
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per second.
```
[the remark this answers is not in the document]
ProofOfKeags: I think the idea is to replace it? Is it the IOPS themselves that will destroy the SSD or are you saying it'll just overrun the storage capacity.

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+   * [`u16`: `length`]
+   * [`length*byte`:`blob`]
+
+
+Requirements:
+
+The sender of `peer_storage`:
+  - MAY send `peer_storage` whenever necessary (usually, everytime their configuration changes)
+  - MUST limit its `blob` to 65531 bytes.
+  - MUST encrypt the data in a way that can be validated when received.
```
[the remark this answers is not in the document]
adi2011: MUST encrypt the data in a manner that ensures its integrity upon receipt?

01-messaging.md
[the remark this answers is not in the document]
adi2011: Yes, the peer is supposed to replace the received data every time. Although what do you suggest would be a good ratelimit?

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+
+
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store anyway.
```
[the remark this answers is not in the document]
adi2011: Done, thanks!

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per second.
+  - MUST replace the old `blob` with the latest received.
+  - MUST send `peer_storage_retrieval` again after reconnection.
```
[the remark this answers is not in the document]
adi2011: Should be mentioned in rationale?

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+  - MUST include the last `blob` it stored for that peer.
+  - when all channels with that peer are closed:
+    - SHOULD wait at least 2016 blocks before deleting the `blob`.
+
+The receiver of `peer_storage_retrieval`:
+  - when it receives `peer_storage_retrieval` with an outdated or irrelevant data:
+    - MAY send a warning.
+
+Rationale:
+If a peer supports `option_provide_storage` and we have a channel with it, we can send a `peer_storage` message with any `blob` so that peer can store it locally in their node. Everytime a node receives a `peer_storage` message, It needs to update the `blob` for the sender. Even if we don't have an open channel, some nodes might provide this service in exchange for some sats so they may store `peer_storage`.
```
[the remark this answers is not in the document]
adi2011: Rephrasing it. Thanks for pointing out.

01-messaging.md
```diff
@@ -476,6 +476,70 @@ every message maximally).
...
+
+Upon receiving `peer_storage` nodes should immediately respond with the latest `peer_storage_retrieval`, which could work as an ack to the peer.
```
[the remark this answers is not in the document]
adi2011: Agreed.

01-messaging.md
```diff
@@ -494,6 +494,76 @@ every message maximally).
 Finally, the usage of periodic `ping` messages serves to promote frequent key
 rotations as specified within [BOLT #8](08-transport.md).
 
+### Peer storage
```
ellemouton, line 497: I think the "Peer Storage" header should be larger that the sub-headers. Ie, main header should use "##" and subheaders "###". 

Also - should add an entry to the table of contents at the top of the doc

adi2011: Thanks for pointing this out. :)

01-messaging.md
```diff
@@ -494,6 +496,76 @@ every message maximally).
...
+The receiver of `peer_storage`:
+  - If it offered `option_provide_storage`:
+    - if it has an open channel with the sender:
+      - MUST store the message.
+    - MAY store the message anyway.
+
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per second.
+  - MUST replace the old `blob` with the latest received.
+  - MUST send `peer_storage_retrieval` again after reconnection, after exchaning `init` messages.
```
Chinwendu20, line 544:
```suggestion
  - MUST send `peer_storage_retrieval` again after reconnection, after exchanging `init` messages.
```

01-messaging.md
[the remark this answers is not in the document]
Chinwendu20: I think *storage* here means persistent storage? I am thinking this should not be in the spec, how often an implementation chooses to store the backup data should be up to them? There is really no need for a consensus on it?

The way this is set up, the peer is expected to replace the backup data with the latest one right? that means only the latest data is needed to be backed up. The data is not also retrieved throughout the connection with the peer only on reconnection. So that means there is really no need to persist this data until the peer connection is over, right? So no need for update/second, but then again this should be up to implementation?

01-messaging.md
adi2011: Thank you for pointing this out.

01-messaging.md
```diff
@@ -494,6 +496,76 @@ every message maximally).
...
+- If it does store the message:
+  - MAY delay storage to ratelimit peer to no more than one update per second.
+  - MUST replace the old `blob` with the latest received.
+  - MUST send `peer_storage_retrieval` again after reconnection, after exchanging `init` messages.
+
+
+The sender of `peer_storage_retrieval`:
+  - MUST include the last `blob` it stored for that peer.
+  - when all channels with that peer are closed:
+    - SHOULD wait at least 2016 blocks before deleting the `blob`.
```
GeorgeTsagk, line 550: what's the thinking here?

01-messaging.md
GeorgeTsagk: Yeah, not a great idea to go kv

01-messaging.md
GeorgeTsagk: Regardless, is it the BOLT's responsibility to document the need for padding?

01-messaging.md
```diff
@@ -494,6 +496,76 @@ every message maximally).
...
+
+Rationale:
+The `peer_storage` and `peer_storage_retrieval` messages enable nodes to securely store
+and share data with other nodes in the network, serving as a backup mechanism for important
+information. By utilizing them, nodes can safeguard crucial data, enhancing the network's
+resilience and reliability. Additionally, even if we don't have an open channel, some nodes
+might provide this service in exchange for some sats so they may store `peer_storage`.
+
+`peer_storage_retrieval` should not be sent after `channel_reestablish` because then the user
+wouldn't have an option to recover the node and update its state in case they lost data.
```
GeorgeTsagk, line 564: >wouldn't have an option to recover the node and update its state in case they lost data.

can you elaborate a bit more here? don't really get why sending `peer_storage_retrieval` hurts here

adi2011: If data is lost, a node would need to see their storage before they can process the reestablish message.

thomash-acinq (commented):
There is no word here about what this will be used for. If we intend to use it to store critical data that will allow us to recover from a loss of channel state on our side, then we should make that explicit.
For that use-case, it's important that we get the latest blob and we should ensure that an honest peer always gives us the latest blob.

01-messaging.md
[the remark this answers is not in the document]
thomash-acinq: > > rate limiting to 1 update per htlc that has been routed over the channel.
> 
> that's even worse :)

Assuming they pay at least one sat per htlc, that should easily cover the cost of replacing your SSD.

Roasbeef:
> There is no word here about what this will be used for. If we intend to use it to store critical data that will allow us to recover from a loss of channel state on our side, then we should make that explicit.

I think that's by design, at the surface level, it's just a blob that you send to the other party and can _maybe_ retrieve in the future. Any higher layer that uses this feature should make minimal assumptions. 

> For that use-case, it's important that we get the latest blob and we should ensure that an honest peer always gives us the latest blob.

There's no way to ensure that. All we can rely on is a best effort attempt by the peer to maybe retrieve the blob, or not.

09-features.md
```diff
@@ -46,6 +46,7 @@ The Context column decodes as follows:
 | 26/27 | `option_shutdown_anysegwit`       | Future segwit versions allowed in `shutdown`              | IN       |                           | [BOLT #2][bolt02-shutdown]                                            |
 | 28/29 | `option_dual_fund`                | Use v2 of channel open, enables dual funding              | IN       |                           | [BOLT #2](02-peer-protocol.md)                                        |
 | 38/39 | `option_onion_messages`           | Can forward onion messages                                | IN       |                           | [BOLT #7](04-onion-routing.md#onion-messages)                         |
+| 42/43 | `option_provide_storage`          | Can store other nodes' encrypted backup data              | IN       |                           | [BOLT #1][bolt01-Peer-storage]                                        |
```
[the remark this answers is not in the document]
adi2011: Thanks for pointing this out...

01-messaging.md
```diff
@@ -494,6 +496,76 @@ every message maximally).
...
+
+### The `peer_storage` and `peer_storage_retrieval` Messages
+
+Nodes that advertise the `option_provide_storage` feature offer storing
+arbitrary data for their peers. The data stored must not exceed 65531 bytes,
+which lets it fit in lightning messages.
+
+Nodes can verify that their `option_provide_storage` peers correctly store
+their data at each reconnection, by comparing the contents of the
+retrieved data with the last one they sent. However, nodes should not expect
```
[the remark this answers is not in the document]
adi2011: @TheBlueMatt  If a node is charging for this service, it would be helpful to compare the data with the latest version to ensure the peer is updating the peer storage correctly.

01-messaging.md
```diff
@@ -494,6 +496,79 @@ every message maximally).
...
+The sender of `peer_storage_retrieval`:
+  - MUST include the last `blob` it stored for that peer.
+  - when all channels with that peer are closed:
+    - SHOULD wait at least 2016 blocks before deleting the `blob`.
+
+The receiver of `peer_storage_retrieval`:
+  - when it receives `peer_storage_retrieval` with an outdated or irrelevant data:
+    - MAY send a warning.
+
+Rationale:
```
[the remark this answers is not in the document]
adi2011: Thanks, Fixed!

rustyrussell:
Agreed at meeting 2025-01-27.

PR #1131, opened 2024-01-29
Add `min_final_expiry_delta` to blinded route example

t-bast:
Payment recipients must ensure that they have a few blocks before fulfilling the payment, in case the inbound channel force-closes.


valentinewallace (approved):
Thanks, this clarifies it!

PR #1138, opened 2024-02-14
Compact representation for node id

thomash-acinq:
To save space in offers (#798) that need to fit on a QR code, we can identify public nodes by one of their public channel, using only 9 bytes instead of 33.
This was suggested by @jkczyz in https://github.com/lightning/bolts/pull/798/commits/3db064e4cb943885ff5557181ae655d03ac8ce79, I'm adding using it in the `next_node_id` field inside blinded routes.
Using it in the `next_node_id` field is needed for nodes that rely on trampoline and do not know the network graph. When such a node is given a compact blinded route, it can't find the corresponding public key for the introduction node id and needs to rely on its trampoline provider to do it.

### DIFF
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -247,9 +247,8 @@ The following convenience types are also defined:
 * `short_channel_id`: an 8 byte value identifying a channel (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
 * `sciddir_or_pubkey`: either 9 or 33 bytes referencing or identifying a node, respectively
     * if the first byte is 0 or 1, then an 8-byte `short_channel_id` follows for a total of 9 bytes
-        * 0 for the first byte indicates this refers to `node_id_1` in the `channel_announcement` for `short_channel_id`
-        * 1 for the first byte indicates this refers to `node_id_2` in the `channel_announcement` for `short_channel_id`
-          (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message)
+        * 0 for the first byte indicates this refers to `node_id_1` in the `channel_announcement` for `short_channel_id` (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message))
+        * 1 for the first byte indicates this refers to `node_id_2` in the `channel_announcement` for `short_channel_id` (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message))
     * if the first byte is 2 or 3, then the value is a 33-byte `point`
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](#appendix-a-bigsize-test-vectors).
 
--- a/01-messaging.md
+++ b/01-messaging.md
@@ -245,6 +245,12 @@ The following convenience types are also defined:
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
 * `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
 * `short_channel_id`: an 8 byte value identifying a channel (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `sciddir_or_pubkey`: either 9 or 33 bytes referencing or identifying a node, respectively
+    * if the first byte is 0 or 1, then an 8-byte `short_channel_id` follows for a total of 9 bytes
+        * 0 for the first byte indicates this refers to `node_id_1` in the `channel_announcement` for `short_channel_id`
+        * 1 for the first byte indicates this refers to `node_id_2` in the `channel_announcement` for `short_channel_id`
+          (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message)
+    * if the first byte is 2 or 3, then the value is a 33-byte `point`
 * `bigsize`: a variable-length, unsigned integer similar to Bitcoin's CompactSize encoding, but big-endian.  Described in [BigSize](#appendix-a-bigsize-test-vectors).
 
 ## Setup Messages
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -468,7 +468,7 @@ may contain the following TLV fields:
         * [`short_channel_id`:`short_channel_id`]
     1. type: 4 (`next_node_id`)
     2. data:
-        * [`point`:`node_id`]
+        * [`sciddir_or_pubkey`:`node_id`]
     1. type: 6 (`path_id`)
     2. data:
         * [`...*byte`:`data`]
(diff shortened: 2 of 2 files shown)


01-messaging.md
```diff
@@ -245,6 +245,12 @@ The following convenience types are also defined:
 * `signature`: a 64-byte bitcoin Elliptic Curve signature
 * `point`: a 33-byte Elliptic Curve point (compressed encoding as per [SEC 1 standard](http://www.secg.org/sec1-v2.pdf#subsubsection.2.3.3))
 * `short_channel_id`: an 8 byte value identifying a channel (see [BOLT #7](07-routing-gossip.md#definition-of-short-channel-id))
+* `sciddir_or_pubkey`: either 9 or 33 bytes referencing or identifying a node, respectively
+    * if the first byte is 0 or 1, then an 8-byte `short_channel_id` follows for a total of 9 bytes
+        * 0 for the first byte indicates this refers to `node_id_1` in the `channel_announcement` for `short_channel_id`
+        * 1 for the first byte indicates this refers to `node_id_2` in the `channel_announcement` for `short_channel_id`
+          (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message)
```
t-bast, line 252: nit: missing closing parenthesis and the link to bolt 7 applies equally to both cases

```suggestion
        * 0 for the first byte indicates this refers to `node_id_1` in the `channel_announcement` for `short_channel_id` (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message))
        * 1 for the first byte indicates this refers to `node_id_2` in the `channel_announcement` for `short_channel_id` (see [BOLT #7](07-routing-gossip.md#the-channel_announcement-message))
```

04-onion-routing.md
```diff
@@ -468,7 +468,7 @@ may contain the following TLV fields:
         * [`short_channel_id`:`short_channel_id`]
     1. type: 4 (`next_node_id`)
     2. data:
-        * [`point`:`node_id`]
+        * [`sciddir_or_pubkey`:`node_id`]
```
t-bast, line 471: Note for reviewers: this is the important part of that PR. This is a breaking change. But since nobody actually uses onion messages right now, we can afford to make that breaking change now (and avoid introducing a feature bit just for this).

This is necessary for nodes who want to send onion messages to blinded paths without an access to the graph data to resolve an `sciddir` into a `node_id`.

Also, it removes this small chunk from the Bolt 12 PR and thus makes it slightly smaller!

We're curious to know what you all think @rustyrussell @TheBlueMatt @jkczyz @valentinewallace

jkczyz: I'm not sure I understand the rationale in the PR description with regards to trampoline. Didn't we include ba2b95a98a415a24c27685a34648c1f64468a4bf in #798 for this purpose? The solution in this PR uses an extra byte per hop compared to that commit, IIUC.

t-bast: The description of the PR isn't very clear, let me try to detail that a bit more. Alice is a mobile wallet user who wants to pay an offer from Bob containing a blinded path that uses an `sciddir` for the introduction node. Alice doesn't have access to the graph, so she cannot resolve the `sciddir` to figure out the introduction node of the blinded path. Alice needs to send an onion message to that blinded path and relies on her LSP for that, so the only thing she can do is create an onion message that uses the following route:

```
Alice ---> LSP -- (sciddir) --> IntroductionNode -- (blinded) --> Bob
```

When creating her onion message path, the only value she can put in the `outgoing_node_id` of the payload for her LSP is the introduction node's `sciddir`, and let the LSP resolve that to a `node_id`. Does that make sense?

> Didn't we include https://github.com/lightning/bolts/commit/ba2b95a98a415a24c27685a34648c1f64468a4bf in https://github.com/lightning/bolts/pull/798 for this purpose?

This is different, because here the LSP and the introduction node aren't necessarily connected (and Alice cannot check that because she doesn't have the graph data). An scid would thus not be enough, because the LSP wouldn't know which node of that channel is the target.

t-bast: > Yea, I mean I still really hate having to look at the network graph to find an onion message first hop, but it saves enough space in the offer that its quite nice.

To be honest, I'm really not a fan of the `sciddir` idea in the first place, I find it really annoying to have to resolve it to a `node_id`...I'm not even convinced that the 24 bytes we're saving are worth the extra complexity, but it looks like LDK and CLN think it's worth it, so we're just following along.

t-bast: > Ugh, can y'all do that as a proprietary phoenix extension for now? At least for the forseeable future phoenix seems like the only wallet that is gonna go without a network graph, even on mobile. I'd rather not introduce any additional sciddir anywhere where we can avoid it, and if things don't evolve into many nodes not having a network graph then this would just be more sciddir resolution that we don't need.

Fair enough, we could do this, we figured this would be useful for other wallets out there and thus having it right now in the spec made sense (if we're introducing `sciddir_or_pubkey` anyway), but we can do our own thing.

I'd like @rustyrussell's opinion on this though before closing this PR!

jkczyz:
> Is this purely a space saving procedure? Is there any other benefit? If not, how consequential is it that we're saving 24 bytes? 🤔

Yes, for space. But it's per hop / introduction node and times 8/5 to account for bech32 encoding.

t-bast:
> Is this purely a space saving procedure? Is there any other benefit?

No, this isn't just for space saving, it's mostly for nodes who don't have access to the graph data (see the [comment above](https://github.com/lightning/bolts/pull/1138#discussion_r1489582868) and the PR description).

Unless you're talking more generally about the `sciddir_or_pubkey` feature that is introduced in https://github.com/lightning/bolts/commit/3db064e4cb943885ff5557181ae655d03ac8ce79? This was indeed added for space saving in minimal invoices, and while it is saving space, I must admit that it is quite annoying to support on the implementation side...but the space savings may be worth it :man_shrugging:

04-onion-routing.md
rustyrussell: It would have been cleaner to always have this type (it would have been v clean to have inside the onion for the next hop!), but here we are, better late than never.

I think it makes sense for smaller QRs, and I think yes, trampoline will need to understand it.  It's not generally hard to implement, though (but I haven't yet implemented it to be sure, that's today's job!).

t-bast: In the HTLC forwarding case, even if we're using an `sciddir_or_pubkey`, you wouldn't need a graph lookup, it has to be referring to one of _your_ channels?

thomash-acinq: This is not about HTLC forwarding but onion message forwarding which already requires the network graph if you're not already connected to the next node. For HTLC forwarding, we still use an outgoing channel id. We only need the network graph in the case of trampoline which is not what this PR is about.

jkczyz: How does this work without a network graph for the case when Alice wants to be paid?  Do the blinded paths in her offer simply use the LSP as the introduction node?

thomash-acinq: For the introduction node, Alice can choose any node to which she knows a path. Even without knowing the full network graph that includes herself and all the nodes she has a channel with.
To use the compact representation, she only needs to know one public channel of the introduction node. In case Alice only has private channels and only knows her own channels, then she can't use this compact representation, she would need her LSP to give her a compact node id (that would require a new message type that would only be useful between a wallet and the LSP).

PR #1140, opened 2024-02-16
Add `option_zero_reserve` (FEATURE 64/65)

t-bast:
Lightning channels require nodes to lock some of their channel funds into a channel reserve, which cannot be used for payments made on that channel. This guarantees that both nodes always have an output in the commitment transaction, which they will lose if they publish a revoked commitment.

While this requirement is generally useful, it creates inefficiencies since that liquidity can't be used to relay payments, and provides a bad user experience. In some settings, we may want to remove that channel reserve and allow nodes to use all of their channel funds.

The `option_zero_reserve` feature lets nodes advertise that they allow their peers to disable the channel reserve on their side, and may lift the same requirement in response.

This is an alternative to #1133

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -797,7 +797,10 @@ The sending node:
   - MUST set `push_msat` to equal or less than 1000 * `funding_satoshis`.
   - MUST set `funding_pubkey`, `revocation_basepoint`, `htlc_basepoint`, `payment_basepoint`, and `delayed_payment_basepoint` to valid secp256k1 pubkeys in compressed format.
   - MUST set `first_per_commitment_point` to the per-commitment point to be used for the initial commitment transaction, derived as specified in [BOLT #3](03-transactions.md#per-commitment-secret-requirements).
-  - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis`.
+  - If the receiver supports `option_zero_reserve`:
+    - MAY set `channel_reserve_satoshis` to `0`.
+  - Otherwise (`option_zero_reserve` cannot be used):
+    - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis`.
   - MUST set undefined bits in `channel_flags` to 0.
   - if both nodes advertised the `option_upfront_shutdown_script` feature:
     - MUST include `upfront_shutdown_script` with either a valid `shutdown_scriptpubkey` as required by `shutdown` `scriptpubkey`, or a zero-length `shutdown_scriptpubkey` (ie. `0x0000`).
@@ -847,10 +850,10 @@ The receiving node MUST fail the channel if:
   - it considers `feerate_per_kw` too small for timely processing or unreasonably large.
   - `funding_pubkey`, `revocation_basepoint`, `htlc_basepoint`, `payment_basepoint`, or `delayed_payment_basepoint`
 are not valid secp256k1 pubkeys in compressed format.
-  - `dust_limit_satoshis` is greater than `channel_reserve_satoshis`.
+  - it doesn't support `option_zero_reserve` and `dust_limit_satoshis` is greater than `channel_reserve_satoshis`.
   - `dust_limit_satoshis` is smaller than `354 satoshis` (see [BOLT 3](03-transactions.md#dust-limits)).
   - the funder's amount for the initial commitment transaction is not sufficient for full [fee payment](03-transactions.md#fee-payment).
-  - both `to_local` and `to_remote` amounts for the initial commitment transaction are less than or equal to `channel_reserve_satoshis` (see [BOLT 3](03-transactions.md#commitment-transaction-outputs)).
+  - it doesn't support `option_zero_reserve` and both `to_local` and `to_remote` amounts for the initial commitment transaction are less than or equal to `channel_reserve_satoshis` (see [BOLT 3](03-transactions.md#commitment-transaction-outputs)).
   - `funding_satoshis` is greater than or equal to 2^24 and the receiver does not support `option_support_large_channel`.
   - the `channel_type` is not suitable.
   - the `channel_type` includes `option_zeroconf` and it does not trust the sender to open an unconfirmed channel.
@@ -874,7 +877,7 @@ The requirement that `channel_reserve_satoshis` is not considered dust
 according to `dust_limit_satoshis` eliminates cases where all outputs
 would be eliminated as dust.  The similar requirements in
 `accept_channel` ensure that both sides' `channel_reserve_satoshis`
-are above both `dust_limit_satoshis`.
+are above both `dust_limit_satoshis`, unless `option_zero_reserve` is used.
 
 The receiver should not accept large `dust_limit_satoshis`, as this could be
 used in griefing attacks, where the peer publishes its commitment with a lot
@@ -928,16 +931,21 @@ The sender:
     - MUST set `minimum_depth` to zero.
   - otherwise:
     - SHOULD set `minimum_depth` to a number of blocks it considers reasonable to avoid double-spending of the funding transaction.
-  - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis` from the `open_channel` message.
-  - MUST set `dust_limit_satoshis` less than or equal to `channel_reserve_satoshis` from the `open_channel` message.
-  - MUST set `channel_type` to the `channel_type` from `open_channel`.
+  - If the receiver supports `option_zero_reserve` and `channel_reserve_satoshis` from the `open_channel` message is `0`:
+    - MAY set `channel_reserve_satoshis` to `0`.
+  - If the receiver supports `option_zero_reserve`:
+    - MAY set `channel_reserve_satoshis` to `0`.
+  - Otherwise (`option_zero_reserve` cannot be used):
+    - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis` from the `open_channel` message.
+    - MUST set `dust_limit_satoshis` less than or equal to `channel_reserve_satoshis` from the `open_channel` message.
+    - MUST set `channel_type` to the `channel_type` from `open_channel`.
 
 The receiver:
   - if `minimum_depth` is unreasonably large:
     - MAY fail the channel.
-  - if `channel_reserve_satoshis` is less than `dust_limit_satoshis` within the `open_channel` message:
+  - if `channel_reserve_satoshis` is less than `dust_limit_satoshis` within the `open_channel` message and `option_zero_reserve` is not set:
     - MUST fail the channel.
-  - if `channel_reserve_satoshis` from the `open_channel` message is less than `dust_limit_satoshis`:
+  - if `channel_reserve_satoshis` from the `open_channel` message is less than `dust_limit_satoshis` and `option_zero_reserve` is not set:
     - MUST fail the channel.
   - if the message doesn't include a `channel_type`:
     - MUST fail the channel.
@@ -1187,6 +1195,7 @@ This message initiates the v2 channel establishment workflow.
    2. data:
         * [`...*byte`:`type`]
    1. type: 2 (`require_confirmed_inputs`)
+   1. type: 4 (`disable_channel_reserve`)
 
 Rationale and Requirements are the same as for [`open_channel`](#the-open_channel-message),
 with the following additions.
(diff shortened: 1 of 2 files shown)


Impa10r:
Any traction on this? Will this option work with already opened channels, or only affect the future ones?

tnull (commented):
Recently, the question came up how zero-reserve channels would interact with maintaining zero Anchor reserve. The particular scenario we wondered about: currently, when a node maintains zero on-chain reserves to spend Anchors, it could at least theoretically use/redirect the `to_remote` funds to claim pending zero-fee HTLC transactions once the commitment transaction confirms. However, it seems that with `option_zero_reserve` the `to_remote` might very well way too little to actually get the claim confirmed. 

Do you have any thoughts on this? Wouldn't no-reserve users be at increased risk of losing HTLC funds to the counterparty?

02-peer-protocol.md
```diff
@@ -891,17 +894,22 @@ The sender:
...
+    - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis` from the `open_channel` message.
+    - MUST set `dust_limit_satoshis` less than or equal to `channel_reserve_satoshis` from the `open_channel` message.
   - if `option_channel_type` was negotiated:
     - MUST set `channel_type` to the `channel_type` from `open_channel`
 
 The receiver:
   - if `minimum_depth` is unreasonably large:
     - MAY fail the channel.
-  - if `channel_reserve_satoshis` is less than `dust_limit_satoshis` within the `open_channel` message:
+  - if `channel_reserve_satoshis` is less than `dust_limit_satoshis` within the `open_channel` message and `option_zero_conf` is not set:
```
valentinewallace: Here and in a few other places, I think `s/option_zero_conf/option_zero_reserve`?

07-routing-gossip.md
```diff
@@ -462,6 +462,8 @@ The origin node:
     reasons.
     - Note: such a `channel_update`, one not preceded by a
     `channel_announcement`, is invalid to any other peer and would be discarded.
+  - If the channel will not be announced (i.e. the `announce_channel` bit was not set) and `option_zero_conf` is used:
+    - MUST send a `channel_update` using the real channel `short_channel_id` once the channel is confirmed.
```
valentinewallace: I think this change is unrelated? :)

02-peer-protocol.md
t-bast: Good catch, thanks! Fixed in the latest rebase.

07-routing-gossip.md
t-bast: Interesting, no idea where that came from...this sounds generally useful, but was most likely part of another, unrelated PR. Removed in the latest rebase!

02-peer-protocol.md
```diff
@@ -889,16 +892,21 @@ The sender:
     - MUST set `minimum_depth` to zero.
   - otherwise:
     - SHOULD set `minimum_depth` to a number of blocks it considers reasonable to avoid double-spending of the funding transaction.
-  - MUST set `channel_reserve_satoshis` greater than or equal to `dust_limit_satoshis` from the `open_channel` message.
-  - MUST set `dust_limit_satoshis` less than or equal to `channel_reserve_satoshis` from the `open_channel` message.
-  - MUST set `channel_type` to the `channel_type` from `open_channel`.
+  - If the receiver supports `option_zero_reserve` and `channel_reserve_satoshis` from the `open_channel` message is `0`:
+    - SHOULD set `channel_reserve_satoshis` to `0`.
+  - If the receiver supports `option_zero_reserve`:
+    - MAY set `channel_reserve_satoshis` to `0`.
```
tankyleo, line 937: Hmm I'm not sure that the opener of the channel setting 0-reserve for the accepter should have any bearing on whether the accepter sets 0-reserve in return (in current LDK it does not); I would have used `MAY` in both cases here.

02-peer-protocol.md
```diff
@@ -1148,6 +1156,7 @@ This message initiates the v2 channel establishment workflow.
    2. data:
         * [`...*byte`:`type`]
    1. type: 2 (`require_confirmed_inputs`)
+   1. type: 4 (`disable_channel_reserve`)
```
tankyleo, line 1198: While we have not finished dual-funded channels, LDK currently use an odd TLV here:
- If the counterparty does not send an `update_fee` / `update_add_htlc` / `splice_out` we would have accepted, this is ok.
- In case the funder of the channel does not understand this field, we do not expect the funder of the channel to complain if the fundee adds an HTLC that bumps the funder's balance below its reserve; this is the fundee's problem, not the funder's.

02-peer-protocol.md
```diff
@@ -2261,6 +2280,9 @@ A sending node:
     its commitment transaction, it cannot pay the fee for the updated local or
     remote transaction at the current `feerate_per_kw` while maintaining its
     channel reserve.
+  - if, after adding that HTLC to its commitment transaction, that transaction
+    doesn't have any output:
+    - MUST NOT offer `amount_msat`.
```
tankyleo, line 2825: I want to confirm here that in 0FC channels, the "has at least one output" check goes away because of the permanent presence of the P2A output.

02-peer-protocol.md
tankyleo, line 2817: After adding an HTLC, LDK makes sure that, at 2x the feerate:
1. The commitments still have at least one output (without counting any additional non-dust HTLCs).
2. The funder of the channel can afford the fee for an additional inbound non-dust HTLC (we are ok if the funder drops below the dust limit here as the inbound non-dust HTLC will produce the single output on the commitment transaction).

So I would add this here:
- SHOULD NOT offer `amount_msat` if, after adding that HTLC to its commitment transaction, increasing the `feerate_per_kw` by a factor of two would produce a commitment transaction with no outputs.

09-features.md
```diff
@@ -55,6 +55,7 @@ The Context column decodes as follows:
 | 48/49 | `option_payment_metadata`         | Payment metadata in tlv record                            | 9        |                             | [BOLT #11](11-payment-encoding.md#tagged-fields)                      |
 | 50/51 | `option_zeroconf`                 | Understands zeroconf channel types                        | INT      | `option_scid_alias`         | [BOLT #2][bolt02-channel-ready]                                       |
 | 60/61 | `option_simple_close`             | Simplified closing negotiation                            | IN       | `option_shutdown_anysegwit` | [BOLT #2][bolt02-simple-close]                                        |
+| 64/65 | `option_zero_reserve`             | This node may accept zero-reserve channels                | IN       |                             | [BOLT #2][bolt02-open]                                                |
```
tankyleo, line 59: - LDK does not signal this feature at the moment, since we never expect channel accepters to grant 0-reserve to channel openers that they don't have some pre-existing, out-of-band relationship with.
- Since December 2021, LDK has not complained if some party opens a channel to LDK, and sets 0-reserve on LDK's side of the channel, see https://github.com/lightningdevkit/rust-lightning/pull/1163. If the channel opener sets 0-reserve for the channel accepter, the drop in security is entirely on the channel opener.

02-peer-protocol.md
tankyleo, line 1: I also have this diff against the `tx_complete` message, since it is possible to reach no-outputs in a 0-reserve channel with a "splice out".

```diff
diff --git a/02-peer-protocol.md b/02-peer-protocol.md
index 5957012..498f4dc 100644
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1796,6 +1796,8 @@ The receiving node:
     - Either side has added an output other than the channel funding output
       and the balance for that side is less than the channel reserve that
       matches the new channel capacity.
+    - Either side has added an output other than the channel funding output
+      and any resulting commitment transactions would have no outputs.

 ##### Rationale
```

tankyleo, line 1: Also should we clarify that after a channel has been spliced, the reserve remains 0 for zero-reserve channels ?

02-peer-protocol.md
t-bast: I'd like to keep it like that, because we should incentivize a tit-for-tat usage of 0-reserve. If you only set it on one side (for example on the mobile wallet only), then you don't get the full UX benefits: you're able to send all your balance, but you're not able to receive all the remaining channel amount.

There's no way to express: "I'm offering you 0-reserve but only if you do the same". Some nodes may choose to cancel the channel open if they offer 0-reserve but the remote peer doesn't offer it in response. So it's IMO a better idea to do tit-for-tat and use the `SHOULD` here in most cases.

Note that I'm biased towards the case where the opener is for example purchasing liquidity (mobile wallet), where as an acceptor you're being paid, so it's fine even if the other side ends up "cheating" and broadcasting a revoked commitment, because what you've been paid upfront will be more interesting than the force-close cost). But you're right that in the case where some random node opens a channel to you and lets you use zero-reserve, there's no reason to echo it back, so both options make sense (which is why this is not a `MUST`).

02-peer-protocol.md
t-bast: I don't understand, do you mean that you've introduced a similar TLV to the `open_channel` and `accept_channel` messages (used for single-funding)? That is completely unnecessary, you can rely on the existing `channel_reserve_satoshis` field as specified in this PR?

This TLV is necessary for the dual-funding case because the `channel_reserve_satoshis` field was removed (in favor of a default 1% reserve).

02-peer-protocol.md
t-bast: Yes, we should do that indeed! Note that it would only happen if all of the channel's balance goes to HTLCs that are below the `dust_limit`. Since nodes shouldn't allow channels to become ridiculously small, in practice this will never happen, but it's good to have the theoretical case covered.

02-peer-protocol.md
t-bast: I'll rebase to integrate that on top of the splicing changes.

t-bast: I think it's obvious, since there's no mechanism to re-negotiate a reserve during a splice? Do you really think it's useful?

t-bast: Added in https://github.com/lightning/bolts/pull/1140/commits/668c0f8c4c201c6369599f55a343a980b7ff3996

t-bast: Added in https://github.com/lightning/bolts/pull/1140/commits/668c0f8c4c201c6369599f55a343a980b7ff3996. I'm really eager to remove all of those very clunky requirements once everyone supports 0FC!

02-peer-protocol.md
t-bast: I've changed this in https://github.com/lightning/bolts/pull/1140/commits/668c0f8c4c201c6369599f55a343a980b7ff3996 so that:

- single-funding uses a `MAY`, since in that case there's no way to purchase a liquidity ads and the more general case would be to not use tit-for-tat with random nodes
- dual-funding uses a `SHOULD` when there is a liquidity purchase, and a `MAY` otherwise, to reflect the case we're using for Phoenix (and what other mobile wallets should do with their LSP)

02-peer-protocol.md
tankyleo: To me it's quite obvious yes. Feel free to resolve this thank you.

02-peer-protocol.md
tankyleo: I mean that for the `disable_channel_reserve` field (used in the dual-funding case), we use an odd TLV, `type: 3` instead of `type: 4`, as we don't mind if the receiver of the message does not understand that field for the two reasons I listed above.

02-peer-protocol.md
tankyleo: I made this review comment because @TheBlueMatt had this question when reviewing the 0-reserve PR in LDK IIRC.

TheBlueMatt: I dunno feels like it could be specified? Splicing mandates that the reserve changes to 10%, no?

t-bast: > Splicing mandates that the reserve changes to 10%, no?

What's that? What makes you say that?

t-bast:
Rebased to fix (trivial) conflicts.

02-peer-protocol.md
t-bast: Ok I think you meant 1% (10% is a lot :sweat_smile:), but indeed once we're on a 0-reserve channel, we should stay 0-reserve even if the channel is spliced. I'll add that to make sure it's clear. If we want an option to reintroduce a reserve to a 0-reserve channel, we'll add a negotiation mechanism for that, but I don't think that's useful.

t-bast: Clarified in https://github.com/lightning/bolts/pull/1140/commits/09e7deef3bcc87b510667a845f525ea451681a25

PR #1141, opened 2024-02-19
BOLT 4: It's Check Lock Time Verify not Check Time Lock Verify.

rustyrussell:
I always get this wrong too, so CLN actually has a source check for this, and it triggered when importing the latest spec!


rot13maxi:
concept ack. Gets me literally every time.

ProofOfKeags (approved):

t-bast (approved):

PR #1145, opened 2024-03-05
Advertise liquidity ads rates

t-bast:
Add the ability to advertise rates at which nodes wish to sell their liquidity. Buyers can then connect to sellers and request liquidity, at one of the advertised rates.

This can be used when creating a dual-funded channel, where both participants contribute to the funding transaction, and one of them is paid for (some of) their contribution. This can also be used to splice additional liquidity into existing channels, when splicing is supported.

We don't add any constraints to the commitment transaction. The lease duration isn't protected by script opcodes, which means that the seller could immediately close the channel, or splice funds out. It is up to the buyer to then blacklist that seller.

This is a subset of #878 and mostly matches what @niftynei proposed there, with some changes that were discussed during the review process and over threads on the mailing list (such as https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-December/004227.html). The goal of this PR is to be a first step that can be integrated before #878 and allows nodes to buy/sell liquidity without restrictions on the seller side, relying on incentives rather than scripts inside the commitment transactions. This will let us observe what behaviors emerge and the market dynamics. If we feel that adding restrictions on the seller side by modifying the commitment transactions can be useful (as proposed in #878), we can easily add it on top of this change using a dedicated feature bit. It would still allow unrestricted liquidity leases, by simply using a `lease_duration` of `0`. This provides an easy path for a more complex change, without sacrificing simplicity in the short term.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -276,11 +276,13 @@ nodes not associated with an already known channel are ignored.
 2. types:
    1. type: 1 (`option_will_fund`)
    2. data:
-       * [`...*16*byte`:`lease_rates`]
+       * [`...*24*byte`:`lease_rates`]
 
 1. subtype: `lease_rate`
 2. data:
    * [`u16`:`lease_duration`]
+   * [`u32`:`min_lease_amount_sat`]
+   * [`u32`:`max_lease_amount_sat`]
    * [`u16`:`funding_weight`]
    * [`u16`:`lease_fee_basis`]
    * [`u32`:`lease_fee_base_sat`]
@@ -360,7 +362,11 @@ The origin node:
   - If it includes `option_will_fund`:
     - MUST set `lease_duration` to the number of blocks during which the lease
       will be active.
+    - MUST set `min_lease_amount_sat` and `max_lease_amount_sat` to the minimum
+      and maximum amount it will contribute at this rate.
     - SHOULD include one `lease_rate` for each lease duration it supports.
+    - SHOULD include one `lease_rate` for each amount range it supports.
+    - MUST NOT include more than ten `lease_rate`s.
     - MUST set `lease_fee_base_sat` to the base fee (in satoshi) it will charge.
     - MUST set `lease_fee_basis` to the amount it will charge per contributed
       satoshi (in basis points, ie 1/10_000).
@@ -411,6 +417,8 @@ any future fields appended to the end):
   - SHOULD ignore Tor v2 onion services.
   - if more than one `type 5` address is announced:
     - SHOULD ignore the additional data.
+    - MUST not forward the `node_announcement`.
+  - if more than ten `lease_rate`s are included:
     - MUST not forward the `node_announcement`.
 
 ### Rationale
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -270,6 +270,32 @@ nodes not associated with an already known channel are ignored.
    * [`32*byte`:`alias`]
    * [`u16`:`addrlen`]
    * [`addrlen*byte`:`addresses`]
+   * [`node_announcement_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `node_announcement_tlvs`
+2. types:
+   1. type: 1 (`option_will_fund`)
+   2. data:
+       * [`...*16*byte`:`lease_rates`]
+
+1. subtype: `lease_rate`
+2. data:
+   * [`u16`:`lease_duration`]
+   * [`u16`:`funding_weight`]
+   * [`u16`:`lease_fee_basis`]
+   * [`u32`:`lease_fee_base_sat`]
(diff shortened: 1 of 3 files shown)


ProofOfKeags:
Is the goal here to get this in production while the details of #878 are still being ironed out? Is the intent to eventually replace #878? Trying to figure out what lense to review this with.

t-bast:
> Is the goal here to get this in production while the details of https://github.com/lightning/bolts/pull/878 are still being ironed out? Is the intent to eventually replace https://github.com/lightning/bolts/pull/878? Trying to figure out what lense to review this with.

That probably wasn't clear enough with the initial PR description, sorry! I've updated the description with more details, let me know if that's better.

niftynei:
Ok yeah, let's get this shipped! 

From a "building trustless protocols" perspective, I'm still holding onto hope that we can figure out an effective and useful way to enable optional commitments to lease enforcement, but I like the idea of moving forward with making this more widely available.

The biggest downside to shipping without a lease enforcement is that it skews the market in favor of a nodes with reputation; it may also retard adoption from buyers who are concerned about service fulfillment. A single bad actor could cause a market chill by behaving in a selfish manner, thereby making buyers less willing to participate in the protocol to buy inbound liquidity due to fear of being scalped/not having their channel liquidity lease term honored.

But given that:
- our current best lease enforcement protocol designs are widely agreed to be imperfect at best, and
- since this is an economic protocol, to some extent the downside risk can be 'priced' by buyers

I'm ok with moving forward with the just the lease negotiation protocol portion of the proposal, especially given that we may discover new information post launch that helps refine further versions.

t-bast:
> I'm ok with moving forward with the just the lease negotiation protocol portion of the proposal, especially given that we may discover new information post launch that helps refine further versions.

Awesome! Do you plan on updating `cln` to match this proposal to allow for cross-compatibility tests? This way once we have cross-compat on the format between `eclair` and `cln`, and a concept ACK from @ProofOfKeags for the lnd folks (and hopefully a concept ACK as well from LDK), we can get that PR merged and start observing what happens on mainnet :rocket:

saubyk:
> The lease duration isn't protected by script opcodes, which means that the seller could immediately close the channel, or splice funds out. It is up to the buyer to then blacklist that seller.

The burden here is really on the application developers to track the channel lease terms and alert the users for channel closures which do not respect the lease terms. Each app developer would need to come up with some sort of a reputation tracking mechanism to keep the users informed of the quality of liquidity providers.

07-routing-gossip.md
```diff
@@ -270,6 +270,32 @@ nodes not associated with an already known channel are ignored.
...
    * [`addrlen*byte`:`addresses`]
+   * [`node_announcement_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `node_announcement_tlvs`
+2. types:
+   1. type: 1 (`option_will_fund`)
+   2. data:
+       * [`...*16*byte`:`lease_rates`]
+
+1. subtype: `lease_rate`
```
t-bast, line 281: An open question here is whether this should also include a `min_amount_sat` and `max_amount_sat`? This is especially useful to adapt the `funding_weight` to the seller's utxo set: if the buyer purchases a very large amount, this may require more inputs in the seller side, so the seller side may want to use a higher `funding_weight`.

@niftynei what do you think?

remyers: I think @t-bast means "liquidity sellers to shape their utxo set".

ie. a seller might maintain a set of UTXOs which can effectively fund the amount(s) they advertise at different expected fee rates without needing to add extra inputs and producing no added change outputs.

t-bast: As discussed during the spec meeting, I have detailed that change in a follow-up commit here: https://github.com/t-bast/bolts/pull/1

Let me know what you think @rustyrussell @niftynei!

ProofOfKeags (commented):
Concept ACK

07-routing-gossip.md
```diff
@@ -360,7 +362,11 @@ The origin node:
   - If it includes `option_will_fund`:
     - MUST set `lease_duration` to the number of blocks during which the lease
       will be active.
+    - MUST set `min_lease_amount_sat` and `max_lease_amount_sat` to the minimum
+      and maximum amount it will contribute at this rate.
```
ProofOfKeags, line 366: Can this actually be enforced? I wonder if this should be "SHOULD"

t-bast:
After discussing using liquidity ads with the LSP specifications working group, I got some feedback that it would be useful to have more flexibility on how the fee is paid, and what constraints apply to the liquidity lease. That lead me to a different design, that allows extending liquidity ads with more lease types in #1153

I think #1153 is more versatile than this version, while also allowing the features offered by this version. I'm interested in reviewers feedback on those two different designs.

t-bast:
Closing in favor of #1153

PR #1149, opened 2024-03-19
Support async payments in BOLT 12

valentinewallace:
This builds on #989 by adding the ability to fetch an invoice from an
always-online node on behalf of an often-offline recipient, e.g. a mobile node.

The idea is that often-offline recipients will supply some always-online node
such as their wallet vendor with a static keysend (i.e. `payment_hash`-less)
invoice to return on its behalf. The recipient will then publish an offer
containing blinded paths that terminate at this always-online node, who payers
can request the invoice from if the recipient is offline at the time. After
receiving the keysend invoice, payers will commence the protocol outlined in
[1] to send the HTLC asynchronously. 

Some context on the top commit where we include the invoice request in the payment onion:
This definitely warrants discussion, but the idea is that this field may be useful for often-offline recipients who did not receive the invoice request when it was originally sent. Recipients may want to verify the invreq or be provided some other relevant data about the payment, while keeping the payment stateless until an HTLC is actually received. For example, future extensions have been proposed [2](https://docs.google.com/document/d/1MbOnzl-IiBBw95qnaLr9JeRqecfFCF9ehoQ8KZb9Tpk/edit#heading=h.1v99y4iw71z) that require the recipient to know a unique token for a payment, and this field would provide that to them. 

Seeking conceptual feedback! I'm also working on the implementation in LDK. 

Based on #798 and #989. 

[1]: https://lists.linuxfoundation.org/pipermail/lightning-dev/2021-October/003307.html

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1912,6 +1912,25 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
    * [`sha256`:`payment_hash`]
    * [`u32`:`cltv_expiry`]
    * [`1366*byte`:`onion_routing_packet`]
+   * [`update_add_htlc_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `update_add_htlc_tlvs`
+2. types:
+    1. type: 0 (`hold_htlc`)
+    2. data:
+        * [`32*bytes`:`payment_release_secret`]
+
+#### TLV fields for `held_htlc_available`
+1. `tlv_stream`: `held_htlc_available`
+2. types:
+
+#### TLV fields for `release_held_htlc`
+
+1. `tlv_stream`: `release_held_htlc`
+2. types:
+    1. type: 0 (`payment_release_secret`)
+    2. data:
+        * [`32*bytes`:`payment_release_secret`]
 
 1. `tlv_stream`: `update_add_htlc_tlvs`
 2. types:
@@ -1956,6 +1975,19 @@ A sending node:
   - MUST increase the value of `id` by 1 for each successive offer.
   - if it is relaying a payment inside a blinded route:
     - MUST set `blinding_point` (see [Route Blinding](04-onion-routing.md#route-blinding))
+  - MUST NOT include a `hold_htlc` TLV unless the sending node expects the
+    final recipient of the HTLC to be offline at the time the HTLC would arrive
+  - MUST NOT include a `hold_htlc` TLV unless the sending node expects to be
+    offline for an extended duration starting soon.
+  - If the `hold_htlc` TLV is present:
+    - MUST immediately send at least two onion messages across at least two
+      different paths to the final HTLC recipient.
+    - Each onion message MUST contain a `held_htlc_available` TLV.
+    - Each onion message MUST contain a unique `reply_path`s which terminates
+      at the reciever of the `update_add_htlc` message.
+    - Each `reply_path` MUST contain a `release_held_htlc` TLV for the
+      `update_add_htlc` recipient in the `encrypted_data_tlvs` with a
+      `payment_release_secret` matching that in the `hold_htlc` TLV.
 
 `id` MUST NOT be reset to 0 after the update is complete (i.e. after `revoke_and_ack` has
 been received). It MUST continue incrementing instead.
@@ -1982,6 +2014,12 @@ A receiving node:
       `error` and fail the channel.
   - if `blinding_point` is provided:
     - MUST use the corresponding blinded private key to decrypt the `onion_routing_packet` (see [Route Blinding](04-onion-routing.md#route-blinding))
+  - if the `hold_htlc` TLV is present:
+    - MUST NOT forward the HTLC until a corresponding `release_held_htlc` onion
+      message is received with a matching `payment_release_secret`.
+    - Upon receipt of a `release_held_htlc` onion message with a matching
+      `payment_release_secret` the HTLC SHOULD be treated as any HTLC without
+      the `hold_htlc` TLV and forwarded as usual.
 
 The `onion_routing_packet` contains an obfuscated list of hops and instructions for each hop along the path.
 It commits to the HTLC by setting the `payment_hash` as associated data, i.e. includes the `payment_hash` in the computation of HMACs.
@@ -2016,6 +2054,19 @@ reach a state where it is unable to send or receive any non-dust HTLC while
 maintaining its channel reserve (because of the increased weight of the
 commitment transaction), resulting in a degraded channel. See [#728](https://github.com/lightningnetwork/lightning-rfc/issues/728)
 for more details.
+
+For often-offline recipients, e.g. mobile clients, nodes can use the
+`hold_htlc` TLV to prevent further forwarding of an HTLC until the recipient
+comes online. As long as the final recipients' counterparty is online and
+storing onion messages for the recipient, the recipient can reply to the onion
+message when they come online, unblock the HTLC, and expect to receive it
+quickly thereafter.
+
+Note that if the sender expects to be online when the recipient comes online,
+they can utilize the `release_held_htlc` onion message without utilizing the
+`hold_htlc` TLV - they can simply send a `held_htlc_available` onion message
+to the final recipient and wait to send any HTLC at all until they receive a
+`release_held_htlc` message back.
 
 ### Removing an HTLC: `update_fulfill_htlc`, `update_fail_htlc`, and `update_fail_malformed_htlc`
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1473,6 +1473,15 @@ even, of course!).
     1. type: 4 (`encrypted_recipient_data`)
     2. data:
         * [`...*byte`:`encrypted_recipient_data`]
+    1. type: 64 (`invoice_request`)
+    2. data:
+        * [`tlv_invoice_request`:`invreq`]
+    1. type: 66 (`invoice`)
+    2. data:
+        * [`tlv_invoice`:`inv`]
+    1. type: 68 (`invoice_error`)
+    2. data:
+        * [`tlv_invoice_error`:`inverr`]
 
 1. subtype: `blinded_path`
 2. data:
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -213,6 +213,9 @@ This is formatted according to the Type-Length-Value format defined in [BOLT #1]
     1. type: 18 (`total_amount_msat`)
     2. data:
         * [`tu64`:`total_msat`]
+    1. type: 5482373484 (`keysend_payment_preimage`)
+    2. data:
+        * [`32*byte`:`payment_preimage`]
 
 `short_channel_id` is the ID of the outgoing channel used to route the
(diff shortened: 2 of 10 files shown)


valentinewallace:
Changed up the feature bits, required single-chain offers, and disallowed setting `invoice_amount_msat` for static invoices. Also moved away from the "keysend" language for clarity. These changes were based on implementing this in LDK in https://github.com/lightningdevkit/rust-lightning/pull/3082, will take this out of draft once the design solidifies a bit more in that PR.

valentinewallace:
Added a fixup requiring the invreq to be included in the payment onion per spec meeting discussion a week ago!

valentinewallace:
Rebased after merge of #798 🎉

02-peer-protocol.md
```diff
@@ -2082,6 +2120,19 @@ maintaining its channel reserve (because of the increased weight of the
...
 for more details.
 
+For often-offline recipients, e.g. mobile clients, nodes can use the
+`hold_htlc` TLV to prevent further forwarding of an HTLC until the recipient
+comes online. As long as the final recipients' counterparty is online and
+storing onion messages for the recipient, the recipient can reply to the onion
+message when they come online, unblock the HTLC, and expect to receive it
+quickly thereafter.
+
+Note that if the sender expects to be online when the recipient comes online,
```
joostjager, line 2130: Is this an expectation that can be had? I am wondering what this means to the user of a mobile app.

12-offer-encoding.md
```diff
@@ -58,6 +58,19 @@ The merchant-pays-user flow (e.g. ATM or refund):
 3. The merchant confirms the *invoice_node_id* to ensure it's about to pay the correct
    person, and makes a payment to the invoice.
 
+The pay-mobile-user flow (e.g. paying a friend back to their mobile node):
+1. The mobile user supplies some always-online node with a static (i.e.
```
joostjager, line 62: How does this relate to the PTLC requirement that was mentioned in the original ML post, is that still a requirement?

12-offer-encoding.md
```diff
@@ -58,6 +58,19 @@ The merchant-pays-user flow (e.g. ATM or refund):
...
+1. The mobile user supplies some always-online node with a static (i.e. 
+   `payment_hash`-less) invoice to return on its behalf. This always-online node may 
+   be the mobile user's channel counterparty, wallet vendor, or another node on the 
+   network that it has an out-of-band relationship with.
+2. The mobile user publishes an offer that contains blinded paths that terminate
+   at the always-online node.
+3. The payer sends an `invoice_request` to the always-online node, who replies
+   with the static invoice previously provided by the mobile user if the mobile user 
+   is offline. If they are online, the `invoice_request` is forwarded to the mobile 
+   user as usual.
```
joostjager, line 71: Why would you distinguish between the offline and online case - can't the offline flow be used always for simplicity?

12-offer-encoding.md
```diff
@@ -441,6 +460,20 @@ while still allowing signature validation.
...
 
+## Invoice Request Features
+
+| Bits | Description                      | Name                        |
+|------|----------------------------------|-----------------------------|
+| 59   | Supports paying static invoices  | static_invoice_pay/optional |
+
+Setting `static_invoice_pay` indicates that the payer supports receiving a
+`payment_hash`-less invoice in response to their `invoice_request`, and
+subsequently setting `sender_provided_payment_preimage` in their payment onion.
```
joostjager, line 471: Is this 'keysend' in LND terms?

12-offer-encoding.md
```diff
@@ -58,6 +58,19 @@ The merchant-pays-user flow (e.g. ATM or refund):
 3. The merchant confirms the *invoice_node_id* to ensure it's about to pay the correct
    person, and makes a payment to the invoice.
 
+The pay-mobile-user flow (e.g. paying a friend back to their mobile node):
+1. The mobile user supplies some always-online node with a static (i.e. 
+   `payment_hash`-less) invoice to return on its behalf. This always-online node may 
+   be the mobile user's channel counterparty, wallet vendor, or another node on the 
+   network that it has an out-of-band relationship with.
+2. The mobile user publishes an offer that contains blinded paths that terminate
+   at the always-online node.
```
joostjager, line 67: The blinded paths, that isn't a strict requirement, or is it?

Also wondering how pathfinding works for the sender. They only get to do one hold htlc without retries, so it must be first time right?

02-peer-protocol.md
```diff
@@ -1972,6 +1972,25 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
+    2. data:
+        * [`32*bytes`:`payment_release_secret`]
+
+#### TLV fields for `held_htlc_available`
+1. `tlv_stream`: `held_htlc_available`
+2. types:
+
+#### TLV fields for `release_held_htlc`
+
+1. `tlv_stream`: `release_held_htlc`
```
joostjager, line 1989: Are there interesting edge cases where the release is signaled, but then the receiver quickly goes offline after all?

02-peer-protocol.md
```diff
@@ -1972,6 +1972,25 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
    * [`sha256`:`payment_hash`]
    * [`u32`:`cltv_expiry`]
    * [`1366*byte`:`onion_routing_packet`]
+   * [`update_add_htlc_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `update_add_htlc_tlvs`
+2. types:
+    1. type: 0 (`hold_htlc`)
```
joostjager, line 1979: Just to get a better feel for the design space: In a model where the sender LSP does the routing (I think Breez works like that, and also Phoenix with trampoline?), what extra options would that give?

Holding the htlc at the LSP makes sense of course, so that the sender can go offline. When the LSP knows the final destination though, they could just keep trying to complete the payment until is succeeds, and then claim the incoming htlc with the preimage. This would cut out all (not reliable?) onion message communication for hold and release.

02-peer-protocol.md
```diff
@@ -1972,6 +1972,25 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
    * [`sha256`:`payment_hash`]
    * [`u32`:`cltv_expiry`]
    * [`1366*byte`:`onion_routing_packet`]
+   * [`update_add_htlc_tlvs`:`tlvs`]
+
+1. `tlv_stream`: `update_add_htlc_tlvs`
```
joostjager, line 1977: Discussed offline: maybe signal `hold_htlc` in the onion, so that the option remains open to park htlcs at a more remote node? Potentially as a way to unburden the sender LSP from async payments.

12-offer-encoding.md
```diff
@@ -441,6 +460,20 @@ while still allowing signature validation.
...
+| Bits | Description                      | Name                        |
+|------|----------------------------------|-----------------------------|
+| 59   | Supports paying static invoices  | static_invoice_pay/optional |
+
+Setting `static_invoice_pay` indicates that the payer supports receiving a
+`payment_hash`-less invoice in response to their `invoice_request`, and
+subsequently setting `sender_provided_payment_preimage` in their payment onion.
+
+Useful if the payee is often offline and the invoice is being returned on
+their behalf by another node, to avoid trusting that other node to not reuse a
```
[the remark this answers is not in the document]
valentinewallace: I was thinking it would be good to explain the reasoning behind needing a `payment_hash`-less invoice, not sure I see what you're getting at with the first sentence though 🤔

02-peer-protocol.md
valentinewallace: It may justify the sender's LSP resending the `held_htlc_available` message, otherwise I believe it's just like any other HTLC where the next-hop forward is offline.

02-peer-protocol.md
valentinewallace: Discussed offline, but we currently expect async recipient wallets to come online infrequently and briefly. So it wouldn't be ideal for the sender's LSP to send a bunch of HTLCs and lock up liquidity repeatedly over a period of days until the recipient comes online. Onion message reliability is indeed a bit tbd though, although with direct-connect it becomes a lot more reliable today at the cost of privacy.

02-peer-protocol.md
```diff
@@ -2047,6 +2079,12 @@ A receiving node:
     - MUST respond with an error as detailed in [Failure Messages](04-onion-routing.md#failure-messages)
   - Otherwise:
     - MUST follow the requirements for the reader of `payload` in [Payload Format](04-onion-routing.md#payload-format)
+  - if the `hold_htlc` TLV is present:
+    - MUST NOT forward the HTLC until a corresponding `release_held_htlc` onion
```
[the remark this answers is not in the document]
valentinewallace: Good point, although maybe worth noting that the HTLC slot is only taken up on the `mobile sender <> LSP` channel, which isn't typically being used for any other payment forwards. LSPs may indeed want to limit the number of outstanding async payments though as you say.

12-offer-encoding.md
joostjager: Yes that was indeed my question. Thanks for the link, understood.

02-peer-protocol.md
joostjager: In this case trampoline is also going to help right? Because if there's just a single sender-determined route a retry is not possible because of onion reuse prevention?

02-peer-protocol.md
joostjager: Yes that makes sense. Didn't realize that the online periods may be so short that it might be impossible to hit.

02-peer-protocol.md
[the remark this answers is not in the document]
joostjager: For small tips it may not be a problem because those do not add to the commit tx weight and fee risk for the LSP. Although the risk then is that the LSP might lose the htlc values in a close event.

02-peer-protocol.md
```diff
@@ -2016,6 +2035,19 @@ A sending node:
   - MUST increase the value of `id` by 1 for each successive offer.
   - if it is relaying a payment inside a blinded route:
     - MUST set `path_key` (see [Route Blinding](04-onion-routing.md#route-blinding))
+  - MUST NOT include a `hold_htlc` TLV unless the sending node expects the
+    final recipient of the HTLC to be offline at the time the HTLC would arrive
+  - MUST NOT include a `hold_htlc` TLV unless the sending node expects to be
+    offline for an extended duration starting soon.
+  - If the `hold_htlc` TLV is present:
+    - MUST immediately send at least two onion messages across at least two
```
joostjager: Where does the intermediate node get these paths from? Maybe from `invoice_request` that is added to the onion payload?

04-onion-routing.md
```diff
@@ -189,6 +189,9 @@ This is formatted according to the Type-Length-Value format defined in [BOLT #1]
 
 1. `tlv_stream`: `payload`
 2. types:
+    1. type: 1 (`invoice_request`)
```
joostjager, line 192: Does this contain the paths that an LSP uses to send the `held_htlc_available` onion message to?

PR #1153, opened 2024-04-15
Extensible liquidity ads rates

t-bast:
Add the ability to advertise rates at which nodes wish to sell their liquidity. Buyers can then connect to sellers and request liquidity, at one of the advertised rates.

This can be used when creating a dual-funded channel, where both participants contribute to the funding transaction, and one of them is paid for (some of) their contribution. This can also be used to splice additional liquidity into existing channels, when splicing is supported.

We support multiple types of liquidity leases, and use an extensible format to allow adding new types of leases in the future. This can be useful to allow different ways of paying the lease fees, and different constraints on the seller or buyer side.

LSPs may for example want to offer leases where fees are not paid immediately when creating the transaction, but instead taken from every future HTLC relayed on that channel using something similar to https://github.com/lightning/blips/pull/25. This also creates that ability for bLIPs to introduce new types of liquidity leases, without affecting the BOLTs.

The existing lease types don't add any constraints to the commitment transactions. The lease duration isn't protected by script opcodes, which means that the seller could immediately close the channel, or splice funds out. It is up to the buyer to then blacklist that seller. We can in the future modify the commitment transactions to restrict the seller: that can be done by introducing a new `lease_type` if it is considered useful.

This introduces a few subtypes, that can be hard to read in the spec format: that's why a test vector is included that makes it easier to see how data is nested, and how it is encoded.

This PR is a concurrent design for #1145.


07-routing-gossip.md
```diff
@@ -1136,6 +1157,200 @@ A->D's `update_add_htlc` message would be:
...
+    * [`byte`:`lease_type`]
+    * [`funding_lease`:`funding_lease`]
+
+1. `lease_type`: 1 (`basic_funding_lease`)
+2. data:
+    * [`u32`:`min_lease_amount_sat`]
+    * [`u32`:`max_lease_amount_sat`]
+    * [`funding_lease_fee`:`funding_lease_fee`]
+
+1. `lease_type`: 3 (`duration_based_funding_lease`)
```
t-bast, line 1191: I included this second lease type to show how supporting multiple lease types is done, but I'm not convinced it should be included in the first version. Should we leave it out for now?

t-bast:
@TheBlueMatt I believe this kind of design leaves enough room for introducing different types of leases and fee mechanisms, and would make it easy to support all LSP use-cases. It also makes it easy for bLIPs to introduce lease types, that would keep using most of the liquidity ads mechanisms. Let me know what you think!

07-routing-gossip.md
```diff
@@ -1136,6 +1157,200 @@ A->D's `update_add_htlc` message would be:
...
+
+Liquidity ads allow nodes to announce their willingness to provide funding to
+other nodes for a fee. Sellers advertise various rates for the liquidity they
+are selling depending on the guarantees offered to buyers.
+
+1. subtype: `will_fund_lease_rates`
+2. data:
+    * [`will_fund_lease_rates_tlvs`:`lease_rates`]
+
+1. `tlv_stream`: `will_fund_lease_rates_tlvs`
```
t-bast, line 1170: Note how a TLV stream is used here to allow sellers to offer various types of leases. Buyers will only read the lease types they support and ignore the lease types they don't support (as long as we odd numbers for the TLVs).

07-routing-gossip.md
```diff
@@ -1136,6 +1157,200 @@ A->D's `update_add_htlc` message would be:
...
+    points to millionths) the maximum `fee_proportional_millionths` it will
+    use in its `channel_update` while the lease is active.
+
+### Paying the `funding_lease_fee`
+
+When `request_funds` and `will_fund` have been exchanged, the buyer must pay
+fees to the seller for the funding they provide to the channel based on the
+agreed upon `funding_lease_fee`.
+
+The lease fee is taken from the buyer's funding inputs and added to the
```
JssDWt, line 1295: In this case the client already needs funds available onchain to receive the channel. If the way the lease fee is taken would be defined by TLV types as well, there could be multiple options for payment methods. I'm thinking for example JIT channels, where the fee is taken from htlcs.

EDIT: I suppose in that case the `funding_lease_fee` would be zero and the payment method can go in a TLV elsewhere.

02-peer-protocol.md
```diff
@@ -1185,11 +1188,18 @@ The sending node:
   - MUST set `funding_feerate_perkw` to the feerate for this transaction
   - If it requires the receiving node to only use confirmed inputs:
     - MUST set `require_confirmed_inputs`
+  - If it wants the receiving node to contribute to the funding transaction
+    using `option_will_fund`:
+    - MUST send `request_funding` containing one of the `lease_type`s
+      advertised by the receiving node in their [liquidity ads](07-routing-gossip.md#liquidity-ads).
```
JssDWt, line 1194: Thinking about the wording here. Does this imply the `lease_type` used MUST be advertised in gossip? Or can it also be supplied out-of-band?

JssDWt (commented):
From the first look this looks like a good extensible model for channel leases. I didn't find any obvious gaps.

07-routing-gossip.md
t-bast: In the case where the buyer doesn't have any funds (either on-chain or in an existing channel) to pay the fees, you wouldn't use one of the lease types provided here. That's where an on-the-fly liquidity bLIP would introduce a new lease type that would pay the fees by subtracting it from the HTLC amount (or from all future HTLCs), most likely based on https://github.com/lightning/blips/pull/25

02-peer-protocol.md
t-bast: We should allow `lease_type`s to be obtained out-of-band, that should be rephrased to something like:

```
- MUST send `request_funding` containing one of the `lease_type`s supported by the receiving node
```

JssDWt: > We should allow `lease_type`s to be obtained out-of-band, that should be rephrased to something like:
> 
> ```
> - MUST send `request_funding` containing one of the `lease_type`s supported by the receiving node
> ```

Sounds good

07-routing-gossip.md
```diff
@@ -1136,6 +1157,200 @@ A->D's `update_add_htlc` message would be:
...
+    * [`u32`:`max_channel_fee_base_msat`]
+
+Sellers may offer multiple `lease_type`s, described in the following sections.
+Buyers select a specific lease offered by the seller and use `request_funds`
+to purchase that lease. Sellers answer with `will_fund` containing a signature
+that commits to the lease parameters included in the `lease_witness`.
+
+### The `basic_funding_lease` type
+
+A `basic_funding_lease` does not provide any guarantee that the seller won't
```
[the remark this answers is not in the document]
t-bast: I'm not sure what you mean, can you rephrase?

07-routing-gossip.md
t-bast: > My main point is that independently from the lease type, the client will always need to be able to calculate the withheld fee

Not necessarily: in a model where the LSP only applies a proportional fee on all HTLCs received by the wallet and automatically provides liquidity as needed, there may be no fee applied at each funding/splicing event at all. The actual fee will depend on future payments received by the user, which we don't know at funding time. I don't think we need to require that the fee has to be known beforehand, this would limit the ways this can be used?

> for any additional lease types we add we should also clearly define how the fee will be calculated for them, i.e. we'll need specific `Paying the ..` sections for each of them

Yes, I agree with that: each type of lease will need to explain the fee structure associated with it. That's why I currently created one section per lease type, in which that can be detailed. Since the two lease types I added share the same fee mechanism, I factored that in its own section, maybe that's what created some confusion? If we remove the `duration_based_funding_lease` from this version, I'll nest the `Paying the fee` section inside the `The basic_funding_lease type` section, which may be clearer?

07-routing-gossip.md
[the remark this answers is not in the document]
t-bast: We did discuss ways of creating guarantees for the buyer in https://lists.linuxfoundation.org/pipermail/lightning-dev/2023-December/004227.html

We can create lease types today that modify scripts to ensure that the seller cannot get their funds back immediately. I don't think that's a good idea, because of the drawbacks discussed in this email thread, but other people might feel differently and may want to ship such lease types.

So I'm not sure this deserves to be in a separate section, especially if we drop the `duration_based_funding_lease` from this initial version (which I'll do in a second commit, so that people can still see what it looks like to define another lease type in the git history).

07-routing-gossip.md
t-bast: Done in https://github.com/lightning/bolts/pull/1153/commits/d30aa2b2ab0980e3a58a58bd3b8341c028a67bd3: readers should have a look at the first commit to see how multiple `lease_types` can be defined.

02-peer-protocol.md
t-bast: Done in https://github.com/lightning/bolts/pull/1153/commits/357a8b01ae4d63023d8e65f644ead966e367800c

02-peer-protocol.md
```diff
@@ -1270,10 +1283,30 @@ The accepting node:
   - MAY respond with a `funding_satoshis` value of zero.
   - If it requires the opening node to only use confirmed inputs:
     - MUST set `require_confirmed_inputs`
+  - If the `request_funding` tlv was sent in `open_channel2`:
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/lightning/bolts/pull/1153/commits/357a8b01ae4d63023d8e65f644ead966e367800c

02-peer-protocol.md
```diff
@@ -1270,10 +1283,30 @@ The accepting node:
...
+    - Otherwise, if it decides to be paid for its contributions:
+      - MUST include the `provide_funding` tlv.
+      - MUST set `funding_satoshis` to a value greater than `0`.
+      - MAY set `funding_satoshis` less or more than `requested_sats`.
+      - MUST fill the `provide_funding` tlv with a lease witness that
+        matches the lease from `request_funds` ([liquidity ads](07-routing-gossip.md#liquidity-ads)).
 
 The receiving node:
+  - MAY fail the negotiation if:
+    - It sent `request_funding` and `provide_funding` is set and:
```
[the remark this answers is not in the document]
t-bast: Done in https://github.com/lightning/bolts/pull/1153/commits/357a8b01ae4d63023d8e65f644ead966e367800c

07-routing-gossip.md
niftynei: > Where the buyer doesn't pay any fee when the channel is created

One current limitation with channel buys using liquidity ads is that the buyer initiates the channel open, which means they're also responsible for the onchain closure fees. So the buyer *has* to contribute at least some funds to the channel at start in order for it to be operational (some CLN users have run into problems where they created basically unusable channels because they didn't put enough sats in at start for a liquidity buy)

If a buyer doesn't have any sats to begin with, they won't be able to use liqudity ad buys.

One way to solve this would be to add a flag that allows for flipping which node is considered the "opener" for the calculation of closing fees. This would put the cost/risk of channel closure onto the liquidity seller, and allows for an opener to essentially do a zero-funds (no utxo) channel open. I'd presume that the seller could price this option independently (maybe add it as a new type of lease?) as it does add a potential cost risk to them.

When you pair this with the JIT/HTLC payment plan bLIP proposals, I think you end up with a pretty nice protocol for allowing zero-UTXO parties to buy inbound liquidity at will. Maybe I'm out of date on the cost models for onchain fees in commitment transactions though?

niftynei (commented):
While this is a considerable evolution of the original proposal, I think it's fair to ask that that I get at least a by-line credit the commit description for this. For example something like

```
Credit to: @niftynei for the original liquidity ads proposal
```

t-bast:
> While this is a considerable evolution of the original proposal, I think it's fair to ask that that I get at least a by-line credit the commit description for this. For example something like

Of course, I'd like to re-arrange the commits once reviewed (the first commit is only here to help guide the review and shouldn't be included on `master` IMO), once we get closer to a final version I'll clean-up the commits and add credits!

07-routing-gossip.md
t-bast: > One current limitation with channel buys using liquidity ads is that the buyer initiates the channel open, which means they're also responsible for the onchain closure fees. So the buyer has to contribute at least some funds to the channel at start in order for it to be operational (some CLN users have run into problems where they created basically unusable channels because they didn't put enough sats in at start for a liquidity buy)

Indeed, that's why I've been experimenting with https://github.com/ACINQ/eclair/pull/2845 as a short-term solution until we move to 0-fee commit txs thanks to v3 transactions. I planned to make that option a bLIP if we want to ship before v3 transactions.

07-routing-gossip.md
```diff
@@ -1136,6 +1152,181 @@ A->D's `update_add_htlc` message would be:
...
+other nodes for a fee. Sellers advertise various rates for the liquidity they
+are selling.
+
+1. `lease_type`: 0 (`basic_funding_lease`)
+2. data:
+    * [`u32`:`min_lease_amount_sat`]
+    * [`u32`:`max_lease_amount_sat`]
+    * [`u16`:`funding_weight`]
+    * [`u16`:`lease_fee_basis`]
+    * [`tu32`:`lease_fee_base_sat`]
```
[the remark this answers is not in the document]
t-bast: No, this is a truncated int on purpose.

niftynei:
Thinking about this a bit more, in focusing on allowing flexibility for LSPs, we're at risk of fracturing the liquidity market across a variety of lease-types and contracts means that the interfaces for participating will be more difficult to implement and less accessible as a result.

One design goal of the original liquidity ads specification was to allow for a "thick/uniform" market for liquidity on lightning, which is to say that we were aiming to maximize the ability of *any* seller in the market to make an offer that would be maximally competitive across the network of ~all potential buyers.

We did a number of things to encourage this, such as fixing the length of the contract to 4032 blocks (~4wks) and building in some assurances for buyers that their spending on the inbound liquidity would be honored by the contract they were signing (putting a lock on the seller's liquidity that matches the length of the lease). Notably, we only offered one way of selling liquidity, which meant buyers were more easily able to compare offers from all sellers.

This proposal, by adding extensibility, removes the uniformity from the protocol which makes it more difficult to build market thickness (large number of buyers or sellers competing around a standard contract). It also increases the implementation surface required to support buying + selling from nodes that offer "liquidity ads". In fact, I'd say that this proposal makes the term "supporting liquidity ads" a fairly minimal assertion, with little to no guarantees of interoperability or accessibility to buyers or sellers depending on what implementation they're running.

I'm not sure how to resolve this, but wanted to highlight some of the tradeoffs we're making by adding extensibility and flexibility to this protocol.

07-routing-gossip.md
niftynei: > Indeed, that's why I've been experimenting with https://github.com/ACINQ/eclair/pull/2845 as a short-term solution until we move to 0-fee commit txs thanks to v3 transactions. I planned to make that option a bLIP if we want to ship before v3 transactions.

Is that a good reason to not add it to the spec currently though? I presume there's a lot of other work to be done before v3 transactions are widely supported. The fact that you're experimenting with it in eclair already is a good indication that this is in fact a useful/necessary feature to add more widely?

This is a side conversation for sure but I'm still a bit unclear about the proposal for how non-utxo owning lightning nodes will manage channel closes, even in a v3 regime.

t-bast: I don't think it would belong to the BOLTs, but a bLIP is definitely needed! That's on my todo-list once I've finished the on-the-fly funding prototype, I'll open a PR for it soon. Would a bLIP work for you?

t-bast:
That's true, adding extensibility means that the market offers will be more diverse (which can also be a good thing, because sellers don't necessarily all want the same thing). One important thing to note that may help us decide on some core design decisions is that for on-the-fly liquidity, we only need flexibility around the `payment_type` (how the fees are paid), which doesn't impact the uniformity of the market. The `from_channel_balance` payment type will always be the best choice because it has no trust trade-off, and the additional payment types only exist to cover the cases where the buyer doesn't have enough channel balance to pay the fees. Flexibility around the lease type may not be necessary if we agree on a single lease type that we're all happy with.

That said, I still believe that having a lease duration doesn't make sense and that the restrictions on routing fees don't make much sense either (and both of these technically don't really work). That's what lead me to the currently proposed lease type, which is the most basic one could think of: no duration, no guarantees, only economic incentives. So I see two paths forward:

- If we all think this basic lease type is sufficient for the foreseeable future, we can remove extensibility around the lease type entirely, and probably even remove the lease witness concept, which simplifies the proposal: the only extensibility point is on how the fees are paid (`payment_type`)
- Otherwise, if some people feel that enforcing a lease duration is an important option to have, we do need extensibility around the lease type. We can then either:
  - Use the extensibility options currently spec-ed
  - Or we could ignore this option for now and remove the lease type extensibility, and introduce it later via a completely separate TLV field, which would potentially only reuse the `payment_type` mechanism

WDYT?

t-bast:
I have simplified this PR based on past discussions and recommendations from the last spec meeting. This now has a smaller scope and simpler, more consistent types.

We only allow extending *how* the funding fees will be paid, by adding new `payment_type` options, like what is done in https://github.com/lightning/blips/pull/36.

The "proof" that the seller provides to the buyer is a simple signature covering the funding rates and funding script: this doesn't reveal exactly how much was purchased and how the fees were paid, and can be used to point to the channel utxo in case of force-close (even though I believe this is a very weak mechanism, since there may be valid reasons for force-closing if the buyer misbehaves and lets HTLCs timeout for example).

We discussed several times with @TheBlueMatt an alternative design for on-the-fly funding where instead of paying liquidity fees for each funding/splice transaction, a wallet user would agree to skim X% of every incoming HTLC in exchange for always-available liquidity. I tried to make this model fit within liquidity ads, but I'm now convinced that this wouldn't be the right design. If a wallet user uses such a model, they wouldn't mix it with "occasional" standard liquidity purchases: it would rather be an agreement between the wallet user and its LSP that for the lifetime of the channel, the LSP ensures that there is enough liquidity for every incoming payment. I don't think this should use liquidity ads at all: it would rather use a dedicated feature bit and one custom "init" message from the wallet to the LSP to "sign" the agreement, and after that the LSP would open a channel and regularly splice funds in whenever necessary. Such a model is quite easy to specify and implement without any interaction with liquidity ads (it only requires dual-funding and splicing).

t-bast:
Thanks for the comment @SeverinAlexB.

> I suggest this to be an independent blip/LSPS and not to tie it to a funding method like LiqAds.

I agree, this is independent of liquidity ads since it doesn't involve doing an on-chain transaction, while liquidity ads is a tool for negotiating the parameter of a collaborative on-chain transaction. For the use-case you describe, a Bolt 12 offer with recurrence seems to be what makes the most sense.

For actually creating that channel in the first place, I believe you could use liquidity ads though, and a tool like on-the-fly funding!

SeverinAlexB:
> a Bolt 12 offer with recurrence seems to be what makes the most sense.

Yes, that could be a great fit.

> For actually creating that channel in the first place, I believe you could use liquidity ads though, and a tool like on-the-fly funding!

Fully agree.

02-peer-protocol.md
```diff
@@ -453,6 +453,9 @@ completed.
     2. data:
         * [`s64`:`satoshis`]
    1. type: 2 (`require_confirmed_inputs`)
+   1. type: 5 (`request_funding`)
```
dunxen, line 456: nit: If we want an odd type here, why not `3`? Doesn't seem claimed as far as I can see in other PRs, unless I'm missing one?

07-routing-gossip.md
```diff
@@ -1107,6 +1122,187 @@ A->D's `update_add_htlc` message would be:
...
+  - When receiving `request_funds`:
+    - If the `funding_rate` does not match a rate it offers:
+      - MUST reject the funding attempt.
+    - If the `payment_type` is not supported:
+      - MUST reject the funding attempt.
+    - Otherwise:
+      - MUST send `will_fund` with the received `funding_rate`, the funding
+        script and a signature covering those fields (see below).
+      - MUST keep the channel open for at least a month to honor the
+        liquidity purchase.
```
dunxen, line 1192: I know the initial proposal by nifty included timelocks on the seller's output, but was maybe complex and bit much to do right now.

As I understand from the PR description, a buyer should blacklist a seller that does not honour the period of the liquidity purchase.

I was wondering whether there could be something in-between, maybe not in this PR, but a seller could "commit" to terms by signing some message with the proposed lease conditions and sending it to the buyer during lease negotiation. If they violate those conditions, this can be proved to the rest of the network via gossip with that signature and the commitment transaction (txid) from force close. (I assume co-op close within the 1 month period indicates the buyer is okay with closing the channel...). This way other nodes on the network seeking to buy liquidity can proactively "blacklist" the seller or downrank them when seeking sellers.

I haven't considered DoSy behaviour or efficiency around this. Just a thought?

02-peer-protocol.md
dunxen: Ah, sorry.

07-routing-gossip.md
dunxen: Thanks for the tl;dr. It does seem more trouble/gameable than value it would add.

02-peer-protocol.md
```diff
@@ -463,6 +466,12 @@ The sender:
     - MUST set `funding_output_contribution`
   - If it requires the receiving node to only use confirmed inputs:
     - MUST set `require_confirmed_inputs`
+  - If it wants the receiving node to contribute to the funding transaction
+    using `option_will_fund`:
+    - MUST send `request_funding` containing one of the funding rates and
+      `payment_type`s supported by the receiving node.
+  - If the previous transaction included `request_funding`:
```
remyers, line 473: If [PR 1236](https://github.com/lightning/bolts/pull/1236) is merged, and either side can send `tx_init_rbf` then this line will need some clarification.

Only the initiator of the previous transaction should set `request_funding` for `tx_init_rbf`. The non-initiator must only set `funding_output_contribution` (as mentioned above) if they will provide an output contribution.

Same for the "The recipient" section below; `request_funding` is only expected by the non-initiator, but `funding_output_contribution` should be equal to or greater than the previous transaction that included a `request_funding`.

Note: this should also apply when the previous transaction is a splice.

t-bast: Indeed, that was discussed during spec meetings for #1236: when someone is purchasing liquidity from you, you shouldn't be the one initiating an RBF, because you would then lose this purchase. I'll update the PR to clarify that after splicing and #1236 are added to the BOLTs.

Issue #1125, opened 2023-12-18
Error decryption - adopt requirement also for new tlv package format.


ziggie1984:
Currently the spec restrict the error decryption to 20 (hops):

```SHOULD continue decrypting, until the loop has been repeated 20 times.```

I think this springs from the old legacy format which had a maximum of 20 hops in a path. I think the new tlv package format can have up to 27 hops, so I wonder if this was kept deliberately or whether we should adopt this to the new tlv package structure ?

How I came to 27:

```
// NumMaxHops is the maximum path length. There is a maximum of 1300 bytes in
// the routing info block. Legacy hop payloads are always 65 bytes, while tlv
// payloads are at least 47 bytes (tlvlen 1, amt 2, timelock 2, nextchan 10,
// hmac 32) for the intermediate hops and 37 bytes (tlvlen 1, amt 2, timelock 2,
// hmac 32) for the exit hop. The maximum path length can therefore only be
// reached by using tlv payloads only. With that, the maximum number of
// intermediate hops is: Floor((1300 - 37) / 47) = 26. Including the exit hop,
// the maximum path length is 27 hops.
```
https://github.com/lightningnetwork/lightning-onion/blob/06182b1d7d2f9bd15ba1ebb404d026604758e78b/path.go#L10-L20

**Comment by t-bast** [comment]
I believe that's just a leftover that we should remove from the spec, feel free to open a PR! We should loop until we've exhausted the shared secrets (which matches the number of hops we actually used for the payment path), not until a specific hard-coded value.

**Comment by t-bast** [comment]
Maybe we're not talking about the same steps in the onion process. I thought @ziggie1984 meant what is done by the sender of a payment when it receives an `update_fail_htlc` message and tries to decrypt the error contained in that message. In that case I don't think there is any timing-related attacks that applies?

**Comment by ziggie1984** [comment]
I am also confused about the last part:
```SHOULD use constant ammag and um keys to obfuscate the route length.```

I am not sure what is meant by obfuscate, because the origin node only cares about de-obfuscating, the ammag and um keys are locked in at this point for the routing hops (defined when building the route?). So not sure why this is important. I thought the erring nodes decides the size of the `failure message` and all the following nodes should keep the size constant?

And not sure about the first requirement as well:
```
SHOULD store a copy of the message.
```
Should we really require this, it's up to the sender to decide whether he wants to keep the data or not, I think relay protection is not really relevant for the Erring Messages?

t-bast:
[...]

PR #1154, opened 2024-04-16
BOLT04: Add rationale for constant error decryption.

ziggie1984:
Fixes https://github.com/lightning/bolts/issues/1125

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1063,9 +1063,10 @@ The _origin node_:
 
 ### Rationale
 
-The requirements for the _origin node_ should help hide the payment sender. By
-continuing decrypting 27 times (dummy decryption cycles after the error is found)
-no information can be obtained via timing analysis.
+The requirements for the _origin node_ should help hide the payment sender.
+By continuing decrypting 27 times (dummy decryption cycles after the error is found)
+the erroring node cannot learn its relative position in the route by performing
+a timing analysis if the sender were to retry the same route multiple times.
 
 ## Failure Messages
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1057,8 +1057,15 @@ The _erring node_:
 The _origin node_:
   - once the return message has been decrypted:
     - SHOULD store a copy of the message.
-    - SHOULD continue decrypting, until the loop has been repeated 20 times.
+    - SHOULD continue decrypting, until the loop has been repeated 27 times
+    (maximum route length of tlv payload type).
     - SHOULD use constant `ammag` and `um` keys to obfuscate the route length.
+
+### Rationale
+
+The requirements for the _origin node_ should help hide the payment sender. By
+continuing decrypting 27 times (dummy decryption cycles after the error is found)
+no information can be obtained via timing analysis.
 
 ## Failure Messages
 


04-onion-routing.md
```diff
@@ -1057,9 +1057,16 @@ The _erring node_:
...
-    - SHOULD continue decrypting, until the loop has been repeated 20 times.
+    - SHOULD continue decrypting, until the loop has been repeated 27 times
+    (maximum route length of tlv payload type).
     - SHOULD use constant `ammag` and `um` keys to obfuscate the route length.
 
+### Rationale
+
+The requirements for the _origin node_ should help hide the payment sender. By
+continuing decrypting 27 times (dummy decryption cycles after the error is found)
+no information can be obtained via timing analysis.
```
ProofOfKeags, line 1068: Doesn't the effort for Attributable Errors directly contradict this? This is definitely the weakest part of my spec knowledge, but afaict, the choice to continue decrypting beyond the required amount doesn't actually materially affect privacy right?

I understand this is a rationale expansion and not a fundamental change to the spec, but I'm just trying to make sure I actually understand the change to be able to evaluate the correctness of the added rationale section.

ziggie1984: Good question. Yes as stated in the issue it's just a should and tcp-network latency is probably way more dominant than decrypting some more hops. But what I understand is that we don't want to reveal the `erring node` its position in the route, which could on the one hand reveil the `sender` but on the other hand  also the destination. See also this comment: https://github.com/lightningnetwork/lightning-onion/blob/master/crypto.go#L351-L354

Maybe we need to be even more specific in the rationale how this timing analysis can harm the sender/receiver.

ziggie1984: hmm yes but we don't want the erring_node to learn its position in the route when it's deliberately failing the route. Which could be done if the erring_node fails some payments and the sender retries them with a smaller amount maybe ?

ProofOfKeags: Let's maybe amend the rationale section as follows:

```
...decrypting 27 times (...) the erroring node cannot learn its relative in position in the route by performing a timing analysis if the sender were to retry the same route multiple times
```

because with the current proposed wording I wasn't able to tell who would be doing the timing analysis to learn what.

ziggie1984:
Agree, either we get rid of the whole section or we introduce the rationale because when I was reading this part of the spec I had a hard time understanding the reasoning behind it. LND has implemented it it actually does the 27 decryption cycle so maybe keep it? But open for both approaches, wdyt @morehouse

t-bast:
> LND has implemented it, it actually does the 27 decryption cycle

Then it's probably a good idea to simplify lnd by removing that extra code if it's useless :wink: 
This timing information seems completely unusable by an attacker to me, but maybe I'm missing something.

ProofOfKeags:
I would like to second the opinion that it seems entirely useless. Two reasons  for this.

1.  Like y'all have already stated, network delay middle 50% variance probably dwarfs the entirety of the decryption process.
2. Learning your relative position in the route is pretty useless beyond 3 hops. It is important to make it hard to distinguish whether your upstream is the sender or a forwarder, but let's say you **know** you're hop number 6... what then?! I'm supposedly 6 degrees of separation from everyone on the planet, so you've successfully ascertained it could be....anyone...

All that said, this PR aims to clarify and for as long as we continue to have the normative requirement to do that extra decryption. I support merging this as the rationale to clarify the purpose of this (seemingly ridiculous) practice.

t-bast:
This was discussed during the last two spec meetings, and it was agreed that this specific decryption step isn't really a fingerprinting vector. But there are other fingerprinting vectors when retrying payments, so this section should be re-worked to explain that generally, as a payment sender, you should ensure you're protecting against fingerprinting by adding random delays before retries and trying as much as possible to be constant time in most operations.

Roasbeef (approved):
LGTM 🦎

ProofOfKeags (approved):
This immediate commit is better for consistency and clarity of the existing spec. However, the "real solution" to this is to generalize the recommendation to say that the sender SHOULD implement reasonable defenses against timing attacks by inserting a random delay before retrying.

t-bast (approved):

PR #1158, opened 2024-04-29
BOLT-04: replace quotation marks used in math expressions

dple:
In this commit, we propose a change of quotation marks being used in math expression. Rather than using " (e.g., $"rho_i"$), we use $``rho''$ that will be rendered better in Latex

PR #1169, opened 2024-06-11
Typo fixes

rustyrussell:


vincenzopalazzo (approved):
ACK https://github.com/lightning/bolts/pull/1169/commits/1583ada3295cbabe649f4e28756caf64aacfd68f

I trust you with the LaTeX fixes, but I feel that we need something in the CI to check for this kind of typo. It seems really easy to make this kind of mistake.

t-bast:
The LateX fixes are a duplicate subset of https://github.com/lightning/bolts/pull/1158, which is stuck because the author is unresponsive. Can you grab the other changes from #1158 and fix the remaining comments on that PR, and we should be good to go?

t-bast:
The LateX fixes were included in #1181, so it looks like you can rebase to just keep the tiny changes to Bolt 3?

Issue #1174, opened 2024-06-25
Discussion: some confusion re blinded path `max_cltv`


ellemouton:
Going through the example outlines [in the route blinding proposal doc](https://github.com/lightning/bolts/blob/master/proposals/route-blinding.md#blinded-payments)
And I'm just a bit confused about the values used for CLTVs. 

To recap the example a bit: 
- current block height is 1000
- Alice, the receiver, wants the path to expire in 200 blocks. Meaning that that the maximum absolute incoming CLTV for the final hop can be 1200. She encodes this value in the `payment_constraints.max_cltv` to herself. 
- She also chooses a `min_final_cltv_expiry` of 12 which iiuc, she does _not_ explicitly communicate to the recipient. Instead, she will add this to the accumulated route CLTV delta. 
- She chooses a blinded route to herself: C -> B -> A.  She rounds up their advertised CLTV deltas to 144 each. She now calculates the total blinded path CLTV delta to be: 12 + 144 + 144 = 300.
- For the encoded payloads she creates for C and B, she also indicates what they should each consider the `max_expiry` so that they can fail the payment early. The example shows that she sets this value to `1356` for B which seems to indicate that this was calculated by taking the maximum cltv: 1200, and adding Alice's `min_final_cltv_expiry` of 12 and finally adding B's cltv delta of 144 to get a final value of 1356. Then for C, she takes the previous value (1356) and adds C's delta of 144 to get 1500. 
- She communicates to the receiver: the total route CLTV delta of 300, and the final `max_cltv` of 1200. 

The rest of the example, where Erin pays via the route, because Erin chooses a final CLTV that is significantly below Alice's given maximum. 

But let's say that Erin wants to play on the edge and instead chooses a final CLTV value of 1200. Erin would then add the total blinded route delta of 300 to get the CLTV value to use on C's incoming link: 1500. Ok so let's say C get's this incoming CLTV of 1300, then C will check it's encoded `max_cltv` from Alice, see that it is 1500 so the incoming value is ok. C then subtracts the 144 delta and sends B a CLTV of 1356. This is also ok given C's encoded `max_cltv` so he subtracts 144 and sends Alice a CLTV of 1212. But this then violates Alice's initial chosen `max_cltv` value of 1200. 

[...]

PR #1176, opened 2024-06-26
bolt4+proposals: fix `max_cltv_expiry` calculation

ellemouton:
- This PR adds some clarification in BOLT 4 around how to calculate the `max_ctlv_expiry` that is added to the `payment_constraints` field for
hops in a blinded path. 
- It also adds a formula for calculating the total 
blinded path cltv delta so that it is clear that this should include the `min_final_cltv_expiry_delta`. 
- Finally, the route blinding payments example in the proposals doc is
  fixed so that the correct `max_cltv_expiry` is included in the final 
  hop's payload. 

Fixes https://github.com/lightning/bolts/issues/1174

The one other thing from #1174 that I didnt cover here was re the shadow route suggestions: for blinded paths. See comment [here](https://github.com/lightning/bolts/issues/1174#issuecomment-2189795809) for more details. If it is agreed that it no longer makes sense for the sender to be responsible for adding the shadow route buffer, then maybe we should add that to the spec

### DIFF
--- a/proposals/route-blinding.md
+++ b/proposals/route-blinding.md
@@ -261,7 +261,9 @@ total `cltv_expiry_delta`. This yields the following values:
 * `route_cltv_expiry_delta`: 300
 
 Let's assume the current block height is 1000. Alice wants the route to be used in the next 200
-blocks, so she sets `max_cltv_expiry = 1200` and adds `cltv_expiry_delta` for each hop. Alice then
+blocks, meaning that the `max_cltv_expiry` she will communicate to the payer will be 1200. She 
+also wants a `min_final_cltv_expiry_delta` of 12 though and so in the encrypted payload to herself, 
+she sets `max_cltv_expiry = 1212` and adds `cltv_expiry_delta` for each hop after that. Alice then
 transmits the following information to the sender (most likely via an invoice):
 
 * Blinded route: `[N(carol), B(bob), B(alice)]`
@@ -276,7 +278,7 @@ transmits the following information to the sender (most likely via an invoice):
 * Encrypted data for blinded nodes:
   * `encrypted_payload(alice)`:
     * `path_id`: `payment_preimage`
-    * `max_cltv_expiry`: 1200
+    * `max_cltv_expiry`: 1212
   * `encrypted_payload(bob)`:
     * `outgoing_channel_id`: `scid_bob_alice`
     * `fee_base_msat`: 100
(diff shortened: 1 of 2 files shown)


t-bast (commented):
Thanks for writing this up! It looks good to me, @TheBlueMatt @jkczyz does this match what is implemented by LDK?

proposals/route-blinding.md
```diff
@@ -261,7 +261,9 @@ total `cltv_expiry_delta`. This yields the following values:
 * `route_cltv_expiry_delta`: 300
 
 Let's assume the current block height is 1000. Alice wants the route to be used in the next 200
-blocks, so she sets `max_cltv_expiry = 1200` and adds `cltv_expiry_delta` for each hop. Alice then
+blocks, meaning that the `max_cltv_expiry` she will communicate to the payer will be 1200. She
```
jkczyz, line 264: What is meant by "the `max_cltv_expiry` she will communicate to the payer"? Does that mean through the invoice's `payinfo.cltv_expiry_delta`? If so, why is that a delta while `max_cltv_expiry` is absolute? 🤔 

AFAICT, LDK doesn't include `min_final_cltv_expiry_delta` in the encrypted payload calculation, but *does* include it in what Alice communicates to the payer (i.e., in `payinfo.cltv_expiry_delta`), which seems like the opposite of what is stated here.

@valentinewallace Could confirm / provide some insight here? You have a better understanding of this than I do. Background: https://github.com/lightning/bolts/issues/1174#issue-2371364610

valentinewallace: ^ That's all accurate from my understanding. 

>What is meant by "the max_cltv_expiry she will communicate to the payer"?

I tried to clarify this in the past via https://github.com/lightning/bolts/pull/1069, i.e the `max_cltv_expiry` should be conveyed via the invoice absolute expiry. But LDK doesn't do this lol, we just choose a very buffered absolute CLTV expiry.

jkczyz: Is my understanding right, though, that LDK doesn't include `min_final_cltv_expiry_delta` in the encrypted data calculation? That seems at odds with the change in the first commit.

valentinewallace: I believe we don't need to *explicitly* include `min_final_cltv_expiry_delta` in our encrypted data calculation because we add so much buffer as mentioned above. Because senders only see the invoice expiry (2 hours), the final expiry that they set shouldn't come close to violating our min final delta. Does that make sense to you?

PR #1179, opened 2024-07-09
Clarify onionmessage decryption

rustyrussell:
I recently re-implemented this to get our code production ready, and I found the spec unclear and even misleading.  This series seeks to rework it to be clearer and more coherent.

Most pointedly, it's now clear that blinding applies to the onion, not the internal encrypted blob: the spec was confused about this at various points.  The onion decoding is now specified exactly, and is general.  The old references to "realms" (from legacy onions) is removed.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -435,18 +435,55 @@ intermediary nodes could simply claim the remaining ones.
 ### Route Blinding
 
 Nodes receiving onion packets may hide their identity from senders by
-"blinding" an arbitrary amount of hops at the end of an onion path.
+providing a `blinded_path`.  To do this the recipient needs to hide both
+the real node ids, and the instructions to those nodes on where to forward
+the onion.
 
-When using route blinding, nodes find a route to themselves from a given
-"introduction node" and initial "blinding point". They then use ECDH with
-each node in that route to create a "blinded" node ID and an encrypted blob
-(`encrypted_data`) for each one of the blinded nodes.
+1. subtype: `blinded_path`
+2. data:
+   * [`point`:`first_node_id`]
+   * [`point`:`blinding`]
+   * [`byte`:`num_hops`]
+   * [`num_hops*onionmsg_hop`:`path`]
 
-They communicate this blinded route and the encrypted blobs to the sender.
-The sender finds a route to the introduction node and extends it with the
-blinded route provided by the recipient. The sender includes the encrypted
-blobs in the corresponding onion payloads: they allow nodes in the blinded
-part of the route to "unblind" the next node and correctly forward the packet.
+1. subtype: `onionmsg_hop`
+2. data:
+    * [`point`:`blinded_node_id`]
+    * [`u16`:`enclen`]
+    * [`enclen*byte`:`encrypted_recipient_data`]
+
+A blinded path consists of:
+1. an initial introduction point (`first_node_id`)
+2. an initial tweak to modify the first node_id to decrypt the onion (`blinding`)
+3. a series of tweaked node ids (`path.blinded_node_id`)
+4. a series binary blobs encrypted to the real node ids (`path.encrypted_recipient_data`)
+   to tell them the next hop.
+
+For example, Dave wants Alice to reach him via public node Bob then
+Carol.  He creates and encrypts three `encrypted_data_tlv`s:
+1. blob_bob: For Bob to tell him to forward to Carol
+2. blob_carol: For Carol to tell her to forward to him
+3. blob_dave: For himself to indicate the path was used, and any metadata he wants.
+
+To mask the node ids, he derives three blinding factors, which turn
+Bob into Bob', Carol into Carol' and Dave into Dave'.  These are a simple
+chain, so Bob can derive Carols without having to be told explicitly.
+
+So this is the `blinded_path` he hands to Alice.
+
+1. `first_node_id`: Bob
+2. `blinding`: to turn Bob into Bob'
+3. `path`: [Bob', bob_blob], [Carol', carol_blob], [Dave', dave_blob]
+
+Alice encrypts an onion to Bob', Carol', Dave' and gives it to Bob
+with the first blinding factor `blinding`.
+
+Bob uses the blinding and his private key to decrypt the first layer
+of the onion (created by Alice), and uses his normal private key to
+decrypt "bob_blob" (created by Dave).  The blob decrypts into a
+`encrypted_data_tlv` which indicates where the onion is to be
+forwarded (i.e. Carol).  Bob derives the next `blinding` and sends it
+an the onion to Carol.
 
 Note that there are two ways for the sender to reach the introduction
 point: one is to create a normal (unblinded) payment, and place the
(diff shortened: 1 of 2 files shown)


04-onion-routing.md
```diff
@@ -290,13 +295,11 @@ The reader:
   - If `encrypted_recipient_data` is present:
     - If `blinding_point` is set in the incoming `update_add_htlc`:
       - MUST return an error if `current_blinding_point` is present.
-      - MUST use that `blinding_point` as the blinding point for decryption.
```
thomash-acinq, line 293: Why did you remove this? `blinding_point` needs to be used as $E_i$ for the Route Blinding Requirements section.

04-onion-routing.md
```diff
@@ -290,13 +295,11 @@ The reader:
...
     - Otherwise:
       - MUST return an error if `current_blinding_point` is not present.
-      - MUST use that `current_blinding_point` as the blinding point for decryption.
+      - MUST use that `current_blinding_point` as `E_i` to derive the following blinding point.
```
thomash-acinq, line 300: I don't understand what you mean by "following blinding" point here.

04-onion-routing.md
```diff
@@ -855,76 +883,66 @@ func NewOnionPacket(paymentPath []*btcec.PublicKey, sessionKey *btcec.PrivateKey
...
-    - MUST drop the packet.
-    - MUST signal a route failure.
-  - MUST address the packet to another peer that is its direct neighbor.
-  - if the processing node does not have a peer with the matching address:
-    - MUST drop the packet.
-    - MUST signal a route failure.
+      - MUST abort processing the packet and fail.
+  - if `blinding` is specified:
+    - Calculate the `blinding_ss` as ECDH(`blinding`, `node-privkey`)
+    - Tweak `public_key` by multiplying by $`HMAC256(\text{"blinded\_node\_id}", blinding_ss)`$
```
thomash-acinq, line 909: This is confusing. We need to tweak the **private** key. And then we need to use the tweaked key instead of the public one.

thomash-acinq (commented):
> blinding applies to the onion, not the internal encrypted blob.

That's not what's implemented in eclair and phoenix, and it contradicts the Route Blinding Requirements section.
I guess the error comes from the fact that there are two ways to use the blinding:
- For the onion itself we use $`B_i = HMAC256(\text{"blinded\_node\_id"}, ss_i) * N_i`$
- For the encrypted blob inside the onion we use $`rho_i = HMAC256(\text{"rho"}, ss_i)`$

04-onion-routing.md
```diff
@@ -432,32 +435,143 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+The writer of `blinded_path`:
+
+- MUST create a viable path to itself ($`N_r`$) i.e. $`N_0 \rightarrow N_1 \rightarrow ... \rightarrow N_r`$.
+- MUST set `first_node_id` to $`N_0`$
+- MUST create a series of ECDH shares secrets for each node in the route using the following algorithm:
+  - $`e_0 /leftarrow {0;1}^256`$ ($`e_0`$ SHOULD be obtained via CSPRG)
+  - $`E_0 = e_0 \cdot G`$
+  - For every node in the route:
+    - let $`N_i = k_i * G`$ be the `node_id` ($`k_i`$ is $`N_i`$'s private key)
+    - $`ss_i = SHA256(e_i * N_i) = SHA256(k_i * E_i)$` (ECDH shared secret known only by $`N_r`$ and $`N_i`$)
```
thomash-acinq, line 504:
```suggestion
    - $ss_i = SHA256(e_i * N_i) = SHA256(k_i * E_i)$ (ECDH shared secret known only by $N_r$ and $N_i$)
```
Markdown rendering is broken because of a misplaced \`, we should probably remove all the \` when we already have \$.

04-onion-routing.md
```diff
@@ -432,32 +435,143 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+2. data:
+    * [`point`:`blinded_node_id`]
+    * [`u16`:`enclen`]
+    * [`enclen*byte`:`encrypted_recipient_data`]
+
+A blinded path consists of:
+1. an initial introduction point (`first_node_id`)
+2. an initial tweak to modify the first node_id to decrypt the onion (`blinding`)
+3. a series of tweaked node ids (`path.blinded_node_id`)
+4. a series binary blobs encrypted to the real node ids (`path.encrypted_recipient_data`)
```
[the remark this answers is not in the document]
valentinewallace: I believe this is what we have in LDK as well, although I'm not too familiar with the notation being used. E.g., for `ss_0`, we compute `ecdh(N_0, session_priv)`, which I *think* is equivalent to `SHA256(e_0 * N_0)` as in the above comment. Eclair is the most interop tested with us so far so pretty sure we're on the same page.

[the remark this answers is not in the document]
rustyrussell: > This is incorrect, the blob is not encrypted to the real node id, but with a key that is derived from both the blinding and the node's private key: ssi=SHA256(ei∗Ni)=SHA256(ki∗Ei) (ECDH shared secret known only by Nr and Ni) rhoi=HMAC256("rho",ssi) (key used to encrypt the payload for Ni by Nr)

Well, yes.  I should spell that out, indeed you use that same blinding as the ECDH ephemeral key.

What you *don't* do, is `ssi=SHA256(ei*Bi)`.  That would be a valid construction (which I would call "encrypting to the blinded node pubkey").  i.e. you don't use `blinding` as blinding at all here...

i.e. the onion is encrypted to the blinded node id, the inner encrypted_recipient_data is not.  This is important!

04-onion-routing.md
rustyrussell: This is wrong, it's used both to derive the  ss and to derive the next blinding point. Will fix...

04-onion-routing.md
rustyrussell: Yes, but it's not used as a blinding point.  I changed this and the next one:

```
    - If `blinding_point` is set in the incoming `update_add_htlc`:
      - MUST return an error if `current_blinding_point` is present.
	  - MUST use `blinding_point` as $`E_i`$
    - Otherwise:
      - MUST return an error if `current_blinding_point` is not present.
      - MUST use `current_blinding_point` as $`E_i`$
      - SHOULD add a random delay before returning errors.
    - MUST return an error if `encrypted_recipient_data` does not decrypt to a valid `encrypted_data_tlv` as described in [Route Blinding](#route-blinding).
```
?

04-onion-routing.md
rustyrussell: This is the alternative construction, where we tweak the onion ephemeral key (which is equivalent).  (CLN needs to do this because we have an HSM and we didn't want to teach it to tweak, but it's also easier in the text).

We could specify both, but we do note the private key method alternative in the rationale?

04-onion-routing.md
rustyrussell: Wow, I didn't even know that worked.  But if we take out ` the formatting changes from monospaced to normal, which would be confusing unless we do it everywhere.

Fixed for now...

rustyrussell:
OK, so I rolled in Thomas' feedback, which lead me to think that we should simply avoid calling the point "blinding" (as that's *not* all it's used for).  Example commit which changes it to "path_key" at the end?

I think this is clearer.  WDYT?

04-onion-routing.md
```diff
@@ -432,32 +434,149 @@ otherwise meets the amount criterion (eg. some other failure, or
...
+into Dave'.
+
+So this is the `blinded_path` he hands to Alice.
+
+1. `first_node_id`: Bob
+2. `first_path_key`: the first path key for Bob 
+3. `path`: [Bob', bob_blob], [Carol', carol_blob], [Dave', dave_blob]
+
+Alice encrypts an onion to Bob', Carol', Dave' and gives it to Bob
+with the `first_path_key`.
```
t-bast, line 486: When used for payments, Alice doesn't encrypt the onion to Bob', she encrypts it to Bob and includes the `first_path_key` in the onion. That's what allows nodes between Alice and Bob to be unaware that a blinded path is used downstream. If the onion was encrypted to Bob', Bob would have no way of obtaining the `path_key` to decrypt it.

Alice uses Bob' only when she combines a blinded path from herself to Bob to the blinded path from Bob to Dave (which is what is done for onion messages).

In both cases, Bob only uses that path key to decrypt the `encrypted_data_tlv`: it's only nodes that are *not* the introduction node that use the path key to decrypt the onion.

I clarified that in https://github.com/t-bast/bolts/commit/c073ef815152e1f523c39a3ff6d7a17ebbfc8734

t-bast (commented):
I stopped reviewing this half-way, when I reached the incorrect statement that Alice encrypts the onion to (Bob', Carol', Dave'). As detailed in my comment below, this is not how it works, and leads me to think that this updated version is actually less clear than the previous one.

But I do see where the confusion comes from: there is a fundamental difference between onion messages and payments, because onion messages combine two blinded paths while payments only use the recipient path. That's why I think the way it is specified currently on `master` makes more sense: how you decrypt the onion only depends on whether you received a path key or not. Then, once you have a decrypted onion, if it contains an `encrypted_data_tlv`, you will use the path key you must have received *somehow* to derive the shared secret to decrypt it.

I think we can create a smaller patch to clear up those confusions than what this PR does, I'd like to try a different approach for that.

I like the renaming of blinding points to path keys though!

t-bast:
I think that one of the things that makes this PR hard to review is that https://github.com/lightning/bolts/pull/1179/commits/5983830c81782a2f5454e23961ddbf0a68c0cbb5 is unrelated and could be extracted to its own PR. It's easier to review on its own and we should reach agreement quickly on it and merge it as a first step.

04-onion-routing.md
rustyrussell: I agree on the details, just not the framing.  

I think of the broad case which I'm describing here as "encrypt to the entire blinded chain", with a special exception for the payment case.  

This is spelled out clearly in the new requirements (which I just noticed did not replace `blinding` so uses the old terminology):

```
The reader of the `blinded_path`:
- MUST create its own onion to reach the `first_node_id`
- For the first entry in `path`:
  - if it is sending a payment:
    - MAY encrypt the first blinded path onion to `first_node_id` and include `blinding` as `current_blinding_point`
  - if it does not do that:
    - MUST encrypt the first blinded path onion to the first `blinded_node_id`.
    - MUST set `next_blinding_override` in the prior onion payload to `blinding`.
  - MUST include the first `path` `encrypted_recipient_data` in each onion payload within the blinded path.
```

And in the rationale:

```
Note that there are two ways for the sender to reach the introduction
point: one is to create a normal (unblinded) payment, and place the
initial blinding point in `current_path_key` along with the
`encrypted_recipient_data` in the onion payload for the introduction point to
start the blinded path. The second way (which is the only way for
onion messages) is to create a blinded path to
the introduction point, set `next_path_key_override` inside the
`encrypted_data_tlv` on the hop prior to the introduction point to the
`first_path_key`, so it is sent to the introduction node.  However, this
only works if that prior node supports blinded paths.
```

rustyrussell:
> I think that one of the things that makes this PR hard to review is that [5983830](https://github.com/lightning/bolts/commit/5983830c81782a2f5454e23961ddbf0a68c0cbb5) and [570f6fa](https://github.com/lightning/bolts/commit/570f6fa69eeea9add90167fc67d1d00c5edc5059) are unrelated to route blinding and could be extracted to its own PR. It's easier to review on its own and we should reach agreement quickly on it and merge it as a first step.

Agreed.  It's only when I started editing I found these.

I will open a new PR for those, then rebase on that.

PR #1186, opened 2024-07-28
BOLT 7: `gossip_status` (feature 66/67)

rustyrussell:
This is a simple sync message, so vastly out-of-sync nodes can be handed the gossip they need.  It's a stopgap until gossip v2.

I'll be implementing this (as feature 100+66/67, and message 30000+267) for testing.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -898,6 +898,48 @@ generated directly by the node itself, so they should ignore filters.
 A node:
   - MUST NOT relay any gossip messages it did not generate itself, unless explicitly requested.
 
+### The `gossip_status` Message
+
+`gossip_status` is a simple option to detect if nodes are
+significantly out of sync with each other.
+
+1. type: 267 (`gossip_status`)
+2. data:
+    * [`chain_hash`:`chain_hash`]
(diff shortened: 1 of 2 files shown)


morehouse:
At a spec level it seems query messages should already provide this functionality.

Do implementations other than CLN encounter issues with the current query mechanism?  Perhaps it would be quicker to fix any implementation bugs with the existing mechanism rather than create a whole new spec that not all implementations may implement any time soon (or ever).

rustyrussell:
Well, LDK kind of gave up and just asks for all the gossip.  This is gentler.  Not sure what the status of LND is, but without something like this you can't even *know* if there's stuff you don't know.   If someone wants to write up a good way of doing this, I'd be delighted!

With gossipv2 it's possible to use set reconciliation, but there's a lot of work needed to do that, so this is a simple bandaid in the meantime.

morehouse:
Isn't this what extended gossip queries are for?

You send a series of `query_channel_range` messages with the timestamp bit set, checking the replies against your `channel_update` timestamps.  Then you can request the specific `channel_update`s you're missing.

rustyrussell:
Dropping this due to lack of enthusiasm, and other ways of implementing.

PR #1188, opened 2024-08-09
BOLT 2: introduce alt_addr feature for private reconnections

maxrantil:
This PR adds documentation for a new feature: the `alt_addr` message in Bolt 02. The `alt_addr` message allows nodes to specify a private alternative connection address for peers, improving privacy, reliability, and latency in established connections.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -44,6 +44,7 @@ operation, and closing.
       * [Committing Updates So Far: `commitment_signed`](#committing-updates-so-far-commitment_signed)
       * [Completing the Transition to the Updated State: `revoke_and_ack`](#completing-the-transition-to-the-updated-state-revoke_and_ack)
       * [Updating Fees: `update_fee`](#updating-fees-update_fee)
+      * [Alternative Addresses: `alt_addr`](#alternative-addresses-alt_addr)
     * [Message Retransmission: `channel_reestablish` message](#message-retransmission)
   * [Authors](#authors)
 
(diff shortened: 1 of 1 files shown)


t-bast:
I'm not sure that this is generally useful, and thus probably doesn't meet the bar for inclusion in the BOLTs. Maybe a [bLIP](https://github.com/lightning/blips) would make more sense?

maxrantil:
Thanks for the input, @t-bast , I’m a Summer of Bitcoin intern, and this is my first implementation here. The project proposal, authored by @EndotermicDev , originally included a BOLT proposal as part of the plan. But honestly, I didn’t know about the bLIP process until you brought it up.

That said, I believe that enhancing privacy is generally useful, so I figured it was worth adding. However, I’m totally open to hearing where others think this should go—whether it’s a BOLT or a bLIP.

02-peer-protocol.md
```diff
@@ -2377,6 +2378,52 @@ be trimmed at the updated feerate, this could overflow the configured
 `max_dust_htlc_exposure_msat`. Whether to close the channel preemptively
 or not is left as a matter of node policy.
 
+### Alternative Addresses: `alt_addr`
```
yyforyongyu, line 2381: Think this should be done in either a BOLT01 `init` msg or a BOLT07 `node_announcement` msg, and not in the channel state machine.

yyforyongyu (commented):
Agree if we wanna include this, it's better to be put in a bLIP.

t-bast:
> That said, I believe that enhancing privacy is generally useful

It's really unclear to me how this _generally_ enhances privacy. It rather looks like it enhances privacy only for some very specific use-cases and nodes, and isn't a general mechanism that we'd expect all nodes to use. That's why it's rather a candidate for a bLIP than the BOLTs, which we don't want to clutter with too many optional features.

maxrantil:
this is now moved to bLIP 43 [https://github.com/lightning/blips/pull/43](url) . Thanks everyone for your input! Closing this out.

PR #1198, opened 2024-09-18
BOlT07: clarify dont_forward bit use cases.

ziggie1984:


ziggie1984:
Changed it into the parenthesis, did not really fit as its own point.

PR #1205, opened 2024-10-11
`option_simple_close` (features 60/61)

t-bast:
This PR is a continuation of #1096, that @rustyrussell asked me to take over. The original description was:

```
This is a "can't fail!" close protocol, as discussed at the NY Summit, and on @Roasbeef's wishlist.  It's about as simple as I could make it: the only complexity comes from allowing each side to indicate whether they want to omit their own output.

It's "taproot ready"(TM) in the sense that `shutdown` is always sent to trigger it, so that can contain the nonces without any persistence requirement.

I split it into three commits for cleanliness:

1. Introduce the new protocol
2. Remove the requirement that shutdown not be sent multiple times (which was already nonsensical)
3. Remove the older protocols

I recommend reviewing it as separate commits, it'll make more sense!
```

I believe it is still useful to review as separate commits: however, we initially allowed setting `nSequence`, which we removed in favor of setting `nLockTime`. That part can probably be skipped. I squashed the fixup commits from the previous PR, but kept the rest.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -1892,7 +1892,7 @@ The receiver of `closing_complete` (aka. "the closee"):
     - SHOULD send a `warning`.
     - SHOULD close the connection.
   - If `closer_scriptpubkey` is a valid `OP_RETURN` script:
-    - MUST set its output amount to zero so that all funds go to fees, as specified in [BOLT #3](03-transactions.md#closing-transaction).
+    - MUST set the closer's output amount to zero so that all funds go to fees, as specified in [BOLT #3](03-transactions.md#closing-transaction).
   - MUST generate the remote closing transaction as specified in [BOLT #3](03-transactions.md#closing-transaction).
   - Select a signature for validation:
     - If the local output amount is dust:


02-peer-protocol.md
```diff
@@ -1713,6 +1762,161 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+If the closer proposes a transaction which will not relay (an output is dust, or the fee rate it
+proposes is too low), it doesn't harm the closee to sign the transaction.
+
+Similarly, if the closer proposes a high fee, it doesn't harm the closee to sign the transaction,
+as the closer is paying.
+
+There's a slight game where each side would prefer the other side pay the fee, and proposes a
+minimal fee. If neither side proposes a fee which will relay, the negotiation can occur again,
+or the final commitment transaction can be spent. In practice, the opener has an incentive to
```
tankyleo, line 1916: Nit:

When creating the mutual close, the final commitment transaction hasn't been broadcast yet, so it cannot be spent yet right?

I suggest something like: "or the final commitment transaction can be broadcast and spent."

Or is that what is meant by "to fail the channel" elsewhere in the spec? Thank you.

dunxen: Yeah, it means force close by signing and broadcasting commitment transaction in this case.

The explanation is in the introduction (bolt 0):
https://github.com/lightning/bolts/blob/aa5207aeaa32d841353dd2df3ce725a4046d528d/00-introduction.md?plain=1#L135-L141

02-peer-protocol.md
```diff
@@ -1508,23 +1509,68 @@ Closing happens in two stages:
 2. once all HTLCs are resolved, the final channel close negotiation begins.
 
         +-------+                              +-------+
-        |       |--(1)-----  shutdown  ------->|       |
-        |       |<-(2)-----  shutdown  --------|       |
+        |       | shutdown(scriptA1)           |       |
```
yyforyongyu, line 1512: Nice diagram! And I think it may be more helpful if we could break it down case by case, also addresses the [comment here](https://github.com/lightning/bolts/pull/1096#discussion_r1483925909), to make it clear that,
1. a pair of `closing_complete` and `closing_sig` completes the flow, and,
2. whoever sends the `closing_complete` pays the fees.

In a hindsight it probably makes more sense to name `closing_complete` `closing_params` and `closing_sig` `closing_complete`, well, but the code is written so nvm.

yyforyongyu: During normal flow, once the shutdown scripts are exchanged via `shutdown` messages, Alice sends a `closing_complete` to specify the fees she is willing to pay, and Bob replies with a `closing_sig` to consent.

```
 +-------+                              +-------+
 |       | shutdown(scriptA1)           |       |
 |       |----------------------------->|       |
 |       |           shutdown(scriptB1) |       |
 |       |<-----------------------------|       |
 |       |                              |       |
 |       | <complete all pending HTLCs> |       |
 |   A   |             ....             |   B   |
 |       |                              |       |
 |       | closing_complete             |       |
 |       |----------------------------->|       |
 |       |                  closing_sig |       |
 |       |<-----------------------------|       |
 +-------+                              +-------+
 ```

If Bob disagrees with closing fees, he can send a `closing_complete` to specify the fees he is willing to pay. Alice should reply with a `closing_sig` if she agrees.

 ```
 +-------+                              +-------+
 |       | shutdown(scriptA1)           |       |
 |       |----------------------------->|       |
 |       |           shutdown(scriptB1) |       |
 |       |<-----------------------------|       |
 |       |                              |       |
 |       | <complete all pending HTLCs> |       |
 |   A   |             ....             |   B   |
 |       |                              |       |
 |       | closing_complete             |       |
 |       |----------------------------->|       |
 |       |             closing_complete |       |
 |       |<-----------------------------|       |
 |       | closing_sig                  |       |
 |       |----------------------------->|       |
 +-------+                              +-------+
 ```

It's allowed to specify shutdown scripts multiple times.

 ```
 +-------+                              +-------+
 |       | shutdown(scriptA1)           |       |
 |       |----------------------------->|       |
 |       |           shutdown(scriptB1) |       |
 |       |<-----------------------------|       |
 |       |                              |       |
 |       | <complete all pending HTLCs> |       |
 |   A   |             ....             |   B   |
 |       |                              |       |
 |       |   <A updates their script>   |       |
 |       |                              |       |
 |       | shutdown(scriptA2)           |       |
 |       |----------------------------->|       |
 |       |           shutdown(scriptB1) |       | (*) Bob doesn't update his script
 |       |<-----------------------------|       |
 |       | closing_complete             |       |
 |       |----------------------------->|       |
 |       |                  closing_sig |       |
 |       |<-----------------------------|       |
 +-------+                              +-------+
 ```

It's possible to encounter a race condition when updating the shutdown script. This is handled by always sending an updated `closing_complete` with the latest shutdown script used.

 ```
 +-------+                              +-------+
 |       | shutdown(scriptA1)           |       |
 |       |----------------------------->|       |
 |       |           shutdown(scriptB1) |       |
 |       |<-----------------------------|       |
 |       |                              |       |
 |       | <complete all pending HTLCs> |       |
 |   A   |             ....             |   B   |
 |       |                              |       |
 |       |   <Bob updates his script>   |       |
 |       |                              |       |
 |       |           shutdown(scriptB2) |       |
 |       |<-----------------------------|       |
 |       | shutdown(scriptA2)           |       |
 |       |----------------------------->|       |
 |       |             closing_complete |       |
 |       |          <-------------------|       |
 |       |                              |       |
 |       |  <Alice updates her script>  |       | (*) This is a concurrent update while Bob is sending closing_complete
 |       |                              |       |
 |       | shutdown(scriptA3)           |       |
 |       |------------------->          |       |
 |       |   closing_complete           |       |
 |       |<-------------------          |       | (*) A doesn't answer with closing_sig because B's sig doesn't use scriptA3
 |       |           shutdown(scriptA3) |       |
 |       |          ------------------->|       |
 |       |           shutdown(scriptB2) |       |
 |       |<-----------------------------|       | (*) B answers A's shutdown with his own shutdown, without any changes
 |       |             closing_complete |       |
 |       |<-----------------------------|       | (*) B now uses scriptB2 and scriptA3 for closing_complete 
 |       | closing_sig                  |       |
 |       |----------------------------->|       | (*) A now uses scriptB2 and scriptA3 for closing_sig
 +-------+                              +-------+
 ```

File: 02-peer-protocol.md [hunk 3]
```diff
@@ -1713,6 +1762,161 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+Both nodes:
+  - After a `shutdown` has been sent and received, AND no HTLCs remain in either commitment transaction:
+    - SHOULD send a `closing_complete` message.
+  - When receiving `shutdown` again, if it did not send `shutdown` first:
+    - MUST respond with `shutdown`.
+    - MAY send `closing_complete` afterwards.
+
+The sender of `closing_complete` (aka. "the closer"):
+  - MUST set `fee_satoshis` to a fee less than or equal to its outstanding balance, rounded down to whole satoshis.
```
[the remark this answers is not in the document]
yyforyongyu: I guess it's easier to look at this question - if I don't have local balance, am I allowed to start a shutdown session to kick off the coop close flow?

Based on this requirement here, since I am required to set the `fee_satoshi` to be no greater than my local balance in my `closing_complete`, it means the value I choose here is 0?

02-peer-protocol.md
```diff
@@ -1713,6 +1762,161 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+      - MUST set `closer_output_only`.
+      - SHOULD NOT set `closer_and_closee_outputs`.
+    - Otherwise:
+      - MUST set both `closer_output_only` and `closer_and_closee_outputs`.
+  - If it wants to send another `closing_complete` (e.g. with a different `fee_satoshis`):
+    - MUST send `shutdown` first.
+    - MUST receive `shutdown` from the remote node in response.
+    - MUST use the `scriptpubkey`s from those `shutdown` messages.
+
+The receiver of `closing_complete` (aka. "the closee"):
```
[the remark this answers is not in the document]
yyforyongyu: Cool thanks for clarity!

02-peer-protocol.md
```diff
@@ -1713,6 +1762,161 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+      - MUST use `closer_and_closee_outputs`.
+    - otherwise:
+      - MUST use `closee_output_only`.
+  - If the selected signature field does not exist:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - If the signature field is not valid for the corresponding closing transaction specified in [BOLT #3](03-transactions.md#closing-transaction):
+    - MUST ignore `closing_complete`.
+  - If the signature field is non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - MUST sign and broadcast the corresponding closing transaction.
```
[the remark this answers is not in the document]
yyforyongyu: I referring to the race condition here,
```
 |       |             closing_complete |       |
 |       |          <-------------------|       |
 |       |                              |       |
 |       |  <Alice updates her script>  |       | (*) This is a concurrent update while Bob is sending closing_complete
 |       |                              |       |
 |       | shutdown(scriptA3)           |       |
 |       |------------------->          |       |
 |       |   closing_complete           |       |
 |       |<-------------------          |       | (*) A doesn't answer with closing_sig because B's sig doesn't use scriptA3
```

But now I see Alice MUST ignore the `closing_complete` due to sig being invalid so nvm.

02-peer-protocol.md
yyforyongyu: Yeah I did realize it's a dual flow - updated the diagram and description, could you verify if it's correct?

02-peer-protocol.md
```diff
@@ -1508,23 +1509,68 @@ Closing happens in two stages:
...
+        |       |----------------------------->|       |
+        |       |             closing_complete |       |
+        |       |          <-------------------|       |
+        |       |                              |       |
+        |       |  <Alice updates her script>  |       | (*) This is a concurrent update while Bob is sending closing_complete
+        |       |                              |       |
+        |       | shutdown(scriptA3)           |       |
+        |       |------------------->          |       |
+        |       |   closing_complete           |       |
+        |       |<-------------------          |       | (*) A doesn't answer with closing_sig because B's sig doesn't use scriptA3
```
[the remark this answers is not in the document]
yyforyongyu: > but it's more complex because that means Alice must remember the state she had when she sent shutdown(scriptA2), whereas it's simpler to drop the old state as soon as you send shutdown.

Yeah this does look cleaner.

> That seems dangerous, because it introduces a risk of deadlock: if Bob never sends closing_complete, Alice would never be allowed to send a new shutdown?

Could you clarity a bit? What prevents Alice from sending a new `shutdown` given that node can now send multiple `shutdown` msgs?

Roasbeef:
> Whenever you want to sign a new version of your closing transaction, you just send closing_complete. This is what removes rounds in the protocol and gets rid of the state machine.

Gotcha, this makes sense. If it's still a state machine or not is somewhat subjective and an implementation detail, but I think what's concrete here is that it no longer needs to loop back upon itself to re-enter the shutdown phase. As the responder you can just send the rely, but as the initiator you still shift between sending the reply and waiting for it to complete. 

Will work on updating my implementation so we can finally get this thing merged!

02-peer-protocol.md
```diff
@@ -1713,6 +1784,179 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - If the signature field is not valid for the corresponding closing transaction specified in [BOLT #3](03-transactions.md#closing-transaction):
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - If the signature field is non-compliant with LOW-S-standard rule<sup>[LOWS](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - otherwise:
+    - MUST broadcast the corresponding closing transaction.
+  - MAY send another `closing_complete` (e.g. with a different `fee_satoshis` or `closer_scriptpubkey`).
+
+### Rationale
```
Roasbeef, line 1922: Can we add rationale here for why both of the new messages carry the closing scripts for both parties?

Roasbeef: One thing that isn't immediately clear is: why does `closing_singed` need to replicate these fields? As a peer you MUST remember the last address you sent to make sure they don't trick you into sending all your funds away. Therefore, isn't it redundant to send the scripts _again_ for the response message?

02-peer-protocol.md
```diff
@@ -1713,6 +1784,181 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+The close protocol is designed to avoid any failure scenarios caused by fee disagreement,
+since each side offers to pay its own desired fee.
+
+If one side has less funds than the other, it may choose to omit its own output, and in this case
+dust MUST be omitted, to ensure that the resulting transaction can be broadcast.
+
+The corner case where fees are so high that both outputs are dust is addressed in two ways: paying
```
Roasbeef, line 1930: Reading this again, doesn't this allow one party to just entirely remove the other party's output on the closing transaction? If Bob can unilaterally decide that both outputs are dust, then he can burn my output by sending an `OP_RETURN` script.

02-peer-protocol.md
```diff
@@ -1713,6 +1784,181 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+  - If the local outstanding balance (in millisatoshi) is less than the remote outstanding balance:
+    - MUST NOT set `closer_output_only`.
+    - MUST set `closee_output_only` if the local output amount is dust.
+    - MAY set `closee_output_only` if it considers the local output amount uneconomical AND its `closer_scriptpubkey` is not `OP_RETURN`.
+  - Otherwise (not lesser amount, cannot remove its own output):
+    - MUST NOT set `closee_output_only`.
+    - If the closee's output amount is dust:
+      - MUST set `closer_output_only`.
+      - SHOULD NOT set `closer_and_closee_outputs`.
+    - Otherwise:
```
Roasbeef, line 1874: Right now the `OP_RETURN` logic is only detailed at the end in the rationale section. I think we should add a requirement here to make it clearer _when_ someone can/should send a script with an `OP_RETURN`. So maybe something like:
```
- If the closer considers their output uneconomical, and the output of the closee is dust:
  - The closer MAY send a `closer_scriptpubkey` that is a valid `OP_RETURN` script. 
    - If sent, then the closer MUST set their output value to zero, ensuring that only a single output exists on the closing transaction (all funds go to fees). 

```

02-peer-protocol.md
```diff
@@ -1713,6 +1784,181 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+  - After a `shutdown` has been sent and received, AND no HTLCs remain in either commitment transaction:
+    - SHOULD send a `closing_complete` message.
+
+The sender of `closing_complete` (aka. "the closer"):
+  - MUST set `fee_satoshis` to a fee less than or equal to its outstanding balance, rounded down to whole satoshis.
+  - MUST set `fee_satoshis` so that at least one output is not dust.
+  - MUST set `closer_scriptpubkey` to its desired output script.
+  - MUST set `closee_scriptpubkey` to the last script it received from its peer (from `closing_complete` or from the initial `shutdown`).
+  - MUST set `locktime` to the desired `nLockTime` of the closing transaction.
+  - MUST generate its closing transaction as specified in [BOLT #3](03-transactions.md#closing-transaction).
```
Roasbeef, line 1860: Perhaps we should move this section to closer to the end, as the intermediate sections are concerned with which outputs are removed/dust. You must determine which outputs will exist on the closing transaction, before you can construct it.

02-peer-protocol.md
```diff
@@ -1713,6 +1784,181 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+    - SHOULD close the connection if it doesn't receive `closing_sig`.
+
+The receiver of `closing_complete` (aka. "the closee"):
+  - If `fee_satoshis` is greater than the closer's outstanding balance:
+    - MUST either send a `warning` and close the connection, or send an `error` and fail the channel.
+  - If `closee_scriptpubkey` does not match the last script it sent (from `closing_complete` or from the initial `shutdown`):
+    - SHOULD ignore `closing_complete`.
+    - SHOULD send a `warning`.
+    - SHOULD close the connection.
+  - MUST generate the remote closing transaction as specified in [BOLT #3](03-transactions.md#closing-transaction).
```
Roasbeef, line 1887: Following from the above, here's a suggestion to add under here:
```
- If `close_scriptpubkey` is a valid `OP_RETURN` script:
   - If the closee's output is not dust:
      - SHOULD send a `warning`
      - SHOULD close the connection. 
   - If the closee's output is greater than the closer's output:
       - (send warning + disconnect)
   - If the closee's output is dust, and the closer's output is greater then the closee's:
       - MUST construct a closing transaction where the sole output has zero value and pays to the valid `OP_RETURN` script. 
```

Roasbeef (changes requested):
Thanks for clearing up the semantics related to `OP_RETURN`!

Before it was a bit buried in the rational section with only a passing reference in the requirements section. I propose we make things a bit more explicit in the requirements section, add some suggestions as in-line comments.

02-peer-protocol.md
```diff
@@ -1713,6 +1784,186 @@ satoshis, which is possible if `dust_limit_satoshis` is below 546 satoshis).
...
+
+An output is *dust* if the amount is less than the [Bitcoin Core Dust Thresholds](03-transactions.md#dust-limits).
+
+Both nodes:
+  - After a `shutdown` has been sent and received, AND no HTLCs remain in either commitment transaction:
+    - SHOULD send a `closing_complete` message.
+
+The sender of `closing_complete` (aka. "the closer"):
+  - MUST set `fee_satoshis` to a fee less than or equal to its outstanding balance, rounded down to whole satoshis.
+  - MUST set `fee_satoshis` so that at least one output is not dust.
```
yyforyongyu, line 1856: I still find this requirement a bit weird...since I can only influence my local output amount, and following this strictly means I need to choose a fee such that my local output is not dust. Meanwhile we allow it to be dust, as long as `OP_RETURN` is used. And by using `OP_RETURN` it means there's nothing to do with the fees I set.

Roasbeef (approved):
LGTM 🥳

PR #1215, opened 2024-12-19
Remove explicit 6-blocks delay for `announcement_signatures`

t-bast:
We previously required waiting for 6 confirmations before sending `announcement_signatures`, but the real criteria is that it should only be sent once you are confident that a reorg will not invalidate it.

Most nodes only wait for 3 confirmations before sending `channel_ready`, which means that they are confident that a malicious 3-blocks reorg will not happen, otherwise their peer may steal the entire channel amount. If they're ready to put their own money at stake after 3 confirmations, then they're also ready to announce the channel after 3 confirmations, right?

In eclair we scale the number of confirmations with the size of the channel, with a minimum amount of 3 confirmations for small channels. I believe that every implementation should do something similar, in which case we don't need this explicit 6-blocks delay? For context, I'm trying to simplify our code in preparation for announcing splices, and it's much simpler to rely only on the number of confirmations we feel safe with than mixing it with a separate 6-blocks delay for announcements.

What implementations should check before ack-ing this PR:

- how they behave when receiving `announcement_signatures` too early: they should stash it and replay it once the funding tx has reached enough confirmations
- how they behave when receiving `channel_announcement` before 6-confirmations: IMO the channel should be added to the routing graph as long as it's confirmed and unspent, and we should simply watch for it to be spent or reorg-ed (which is already what the spec mandates)

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -84,8 +84,8 @@ The `announcement_signatures` message is created by constructing a `channel_anno
 A node:
   - if the `open_channel` message has the `announce_channel` bit set AND a `shutdown` message has not been sent:
     - MUST send the `announcement_signatures` message.
-      - MUST NOT send `announcement_signatures` messages until `channel_ready`
-      has been sent and received AND the funding transaction has at least six confirmations.
+      - MUST NOT send `announcement_signatures` until `channel_ready` has been sent and received.
+      - MUST NOT send `announcement_signatures` until the funding transaction has enough confirmations to ensure that it won't be reorganized.
   - otherwise:
     - MUST NOT send the `announcement_signatures` message.
   - upon reconnection (once the above timing requirements have been met):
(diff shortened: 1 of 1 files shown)


yyforyongyu:
For `lnd`, when sending `channel_ready`,
- for normal txns, we require 3 to 6 confirmations based on the channel amount, or the user's own preference via `accept_channel.minimum_depth`.
- for coinbase txns, we require `max(100, user_specified)` confirmations.

When sending `announcement_signatures`, we require at least 6 confirmations.


> how they behave when receiving announcement_signatures too early

We require the corresponding `channel_announcement` and `channel_update` to be received first, otherwise we cache and process them later.


> how they behave when receiving channel_announcement before 6-confirmations: IMO the channel should be added to the routing graph as long as it's confirmed and unspent, and we should simply watch for it to be spent or reorg-ed (which is already what the spec mandates)

Yeah that's exactly how we handle it.


> If they're ready to put their own money at stake after 3 confirmations, then they're also ready to announce the channel after 3 confirmations, right?

I think so, tho there's a slight difference between their willingness to announce after X confirmations vs another node's willingness to include an edge in its graph after Y confirmations, maybe this can be a config option for users when building their graphs.

IMO we should make a default of 3, and allow the users to customize if needed.

rustyrussell:
CLN stores channels under 6 confirms, and we also use 3 blocks to make active by default (on mainnet, 1 on testnets).

But we do not actually throw out reorged channels (though I should check!). We should, of course, but at 6 deep it's never been a problem. Will fix.

rustyrussell:
I'm happy to change this, FWIW. It's a nice simplification.

Roasbeef (approved):
LGTM 💐

PR #1220, opened 2025-01-15
More clarifications around `channel_announcement` handling

t-bast:
This is a follow-up of #1215, where we keep reworking the announcement requirements.

We remove the rationale around deferring after `channel_ready` since that has been changed years ago and shouldn't be an issue anymore.

We slightly rework the wording to minimize future conflicts with the splicing PR.

While nodes are free to send `announcement_signatures` whenever they feel that the channel is safe from reorg, we disallow broadcasting the `channel_announcement` before at least 6 confirmations: this gives nodes a simple heuristic to ignore `channel_announcement`s for remote channels that aren't spec-compliant, without having to deal with the extra cost of managing reorgs.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -78,42 +78,36 @@ The willingness of the initiating node to announce the channel is signaled durin
 
 ### Requirements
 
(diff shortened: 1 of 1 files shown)


yyforyongyu (approved):
LGTM🌹

Roasbeef (approved):
LGTM 💎

Issue #1221, opened 2025-01-22
Lightning Specification Meeting 2025/01/27


t-bast:
The meeting will take place on Monday 2025/01/13 at 7pm UTC (5:30am Adelaide time) on Libera Chat IRC [#lightning-dev](irc://irc.libera.chat/lightning-dev). It is open to the public.

A video link is available for higher bandwidth communication: https://meet.jit.si/Lightning-Spec-Meeting

## Recently Updated Proposals / Seeking Review

This section contains changes that have been opened or updated recently and need feedback from the meeting participants.

- [x] Simplified mutual close #1205
- [x] Announcement confirmation follow-up #1220
- [x] Dual funding reconnect `commit_sig` retransmission #1214
- [x] `splice_locked` retransmission #1223
- [x] Offer mixed-case test vector #1217
- [x] Bolt 12 in index #1200
- [x] Mailing list discussion #1170
- [x] Splicing #1160
- [x] Taproot #995
- [x] Taproot gossip #1059
- [x] Dynamic Commitments #1117
- [x] Bolt 12 contacts: https://github.com/lightning/blips/pull/42
- [ ] ~Liquidity ads #1153~
- [x] Peer storage backup #1110
- [ ] ~Channel jamming #1071 #1218~
- [ ] ~Trampoline routing #829 and #836~
- [ ] ~Async payments #1149~

## Stale Proposals

This section contains pending changes that may not need feedback from the meeting participants, unless someone explicitly asks for it during the meeting. These changes are usually waiting for implementation work to happen to drive more feedback.

- [ ] ~Inbound fees https://github.com/lightning/blips/pull/18 and https://github.com/lightning/blips/pull/22~

## Waiting for interop

This section contains changes that have been conceptually ACKed and are waiting for at least two implementations to fully interoperate.
They most likely don't need to be covered during the meeting, unless someone asks for updates.

- [ ] ~Zero reserve #1140 (follow-up on #1133 to include a feature bit)~
- [ ] ~Attributable errors #1044~
- [ ] ~Don't force close until error is received after `channel_reestablish` #934~

## Long Term Updates

This section contains long-term changes that need review, but require a substantial implementation effort.

- [ ] ~Simplified commitment #867~
- [ ] ~lnprototest (https://github.com/rustyrussell/lnprototest)~

Roasbeef:
rbf coop close:
  * interop between lnd and eclair 
[...]

PR #1228, opened 2025-02-14
Zero-fee commitments using v3 transactions (feature 40/41)

t-bast:
We introduce a new `channel_type` that leverages v3 (TRUC) transactions, pay-to-anchor outputs and ephemeral dust. With this change, commitment transactions don't pay any mining fee (in most cases), which gets rid of `update_fee` and all of the related channel reserve issues. It also gets rid of the undesired channel force-closes that happen when the mempool feerate spikes and channel participants disagree on what feerate to use, which has been a major source of wasted on-chain space.

It also offers better protection against pinning attacks (thanks to package relay) and reduces the on-chain footprint compared to anchor output channels.

We use a single anchor output whose amount is the sum of all trimmed outputs (outputs below the `dust_limit`, which are thus not included in the commitment transaction) and all millisatoshi amounts that have been rounded down to satoshis, capped at `240 sats` (and may thus be `0 sat` when there are no trimmed outputs or rounded down millisatoshi amounts). Note that when the sum of trimmed outputs and millisatoshi rounding exceeds `240 sats`, the commitment transaction won't by 0-fees anymore: but the fees aren't deduced from any of the commitment transaction's outputs.

This anchor output is unkeyed: it simply uses the standard P2A output introduced in https://github.com/bitcoin/bitcoin/pull/30352.

We remove the 1-block relative delay used by anchor output channels: this allows using our channel outputs (main balance or pending HTLCs) to CPFP a remote commitment transaction (no need to add external inputs, which simplifies wallet management and liquidity requirements).

Note that v3 transactions and pay-to-anchor outputs have been standard since the release of Bitcoin Core v28.0. Ephemeral dust has become standard in the Bitcoin Core v29.0 release. All of this is thus ready to use on the network now!


05-onchain.md
```diff
@@ -100,9 +100,11 @@ delayed_pubkey.
 funding_pubkey.
 4. _remote node's anchor output_: one output paying to the *remote node's*
 funding_pubkey.
-5. _local node's offered HTLCs_: Zero or more pending payments (*HTLCs*), to pay
+5. _shared ephemeral anchor_: one output paying to the *local* or *remote*
+node's funding_pubkey.
```
esneider, line 104: * The ephemeral in the output name isn't used anywhere else, probably just "shared anchor"?
* Description doesn't match the fact that the shared anchor is using P2A

saubyk:
> We use a single anchor output whose amount is the sum of all trimmed outputs (and may thus be 0 sat).

Hi @t-bast can you elaborate this a bit?

saubyk:
> > We use a single anchor output whose amount is the sum of all trimmed outputs (and may thus be 0 sat).
> 
> Hi @t-bast can you elaborate this a bit?

Never mind. This delving post has the details, makes for a good supplemental reading:
https://delvingbitcoin.org/t/which-ephemeral-anchor-script-should-lightning-use/1412

03-transactions.md
```diff
@@ -280,60 +298,73 @@ To redeem the HTLC, the HTLC-success transaction is used as detailed below. This
...
 
 These HTLC transactions are almost identical, except the HTLC-timeout transaction is timelocked. Both HTLC-timeout/HTLC-success transactions can be spent by a valid penalty transaction.
 
-* version: 2
+* version: 3 when using `zero_fee_commitments`, otherwise 2
 * locktime: `0` for HTLC-success, `cltv_expiry` for HTLC-timeout
 * txin count: 1
    * `txin[0]` outpoint: `txid` of the commitment transaction and `output_index` of the matching HTLC output for the HTLC transaction
-   * `txin[0]` sequence: `0` (set to `1` for `option_anchors`)
+   * `txin[0]` sequence: `1` for `option_anchors`, `0xFFFFFFFD` for `zero_fee_commitments`, `0` otherwise
```
SomberNight, line 361: What is the difference between using `0` and `0xFFFFFFFD` here?
Neither imposes a bip68 timelock, and both explicitly opt-in to bip125 RBF.

I am just trying to understand the motivation for why the old-style txs were using `0` and why the newer style proposed here uses a different value.

SomberNight: I see, so there is no interesting technical difference I missed. Thanks.
I understand the potential privacy aspect, but I don't think it's worth the bikeshedding in this case, so feel free to mark this as resolved :)

05-onchain.md
```diff
@@ -148,9 +150,10 @@ A node:
         - otherwise:
           - MUST broadcast the *last commitment transaction*, for which it has a
           signature, to perform a *unilateral close*.
-          - MUST spend any `to_local_anchor` output, providing sufficient fees as incentive to include the commitment transaction in a block.
-          Special care must be taken when spending to a third-party, because this re-introduces the vulnerability that was
-          addressed by adding the CSV delay to the non-anchor outputs.
+          - MUST spend any `to_local_anchor` or `shared_anchor` output, providing sufficient fees
+          as incentive to include the commitment transaction in a block. When using `nVersion=2`,
+          special care must be taken when spending to a third-party, because this re-introduces
+          the vulnerability that was addressed by adding the CSV delay to the non-anchor outputs.
```
morehouse, line 156: Bringing up `nVersion=2` here is a bit confusing and may give the reader the idea they can set `nVersion=3` for `to_local_anchor` spends to avoid this issue (which won't relay).

```suggestion
          - MUST spend any `to_local_anchor` or `shared_anchor` output, providing sufficient fees
          as incentive to include the commitment transaction in a block. For `to_local_anchor`,
          special care must be taken when spending to a third-party, because this re-introduces
          the pinning vulnerability that was addressed by adding the CSV delay to the non-anchor outputs.
```

05-onchain.md
```diff
@@ -515,7 +517,7 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - if `option_anchors` applies:
+  - if `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions:
```
morehouse, line 520:
```suggestion
  - if `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions (i.e. `option_anchors` or `zero_fee_commitments` applies):
```

05-onchain.md
```diff
@@ -532,8 +534,8 @@ A single transaction that resolves all the outputs will be under the
 standard size limit because of the 483 HTLC-per-party limit (see
 [BOLT #2](02-peer-protocol.md#the-open_channel-message)).
 
-Note: if `option_anchors` applies, the cheating node can pin spends of its
-HTLC-timeout/HTLC-success outputs thanks to SIGHASH_SINGLE malleability.
+Note: when `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions,
+the cheating node can pin spends of its HTLC-timeout/HTLC-success outputs.
```
morehouse, line 538:
```suggestion
Note: when `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions (i.e. `option_anchors` or `zero_fee_commitments`),
the cheating node can pin spends of its HTLC-timeout/HTLC-success outputs.
```

05-onchain.md
```diff
@@ -588,25 +590,25 @@ Note: even if the `to_remote` output is not swept, the resulting
 
 # Generation of HTLC Transactions
 
-If `option_anchors` does not apply to the commitment transaction, then
-HTLC-timeout and HTLC-success transactions are complete transactions with
-(hopefully!) reasonable fees and must be used directly.
+If `SIGHASH_ALL` is used for HTLC transactions, then HTLC-timeout and
+HTLC-success transactions are complete transactions with (hopefully!)
+reasonable fees and must be used directly.
```
morehouse, line 595: This section describes *when* to use `SIGHASH_ALL` vs `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY`.  So we shouldn't assume the reader already knows which sighashes to use for each channel type.

```suggestion
If `option_anchors` and `zero_fee_commitments` do not apply, then HTLC-timeout and
HTLC-success transactions are signed with SIGHASH_ALL and are complete transactions with (hopefully!)
reasonable fees and must be used directly.
```

morehouse (commented):
Approach ACK.

Really eager to see this get implemented and used across the network.  Also hope we can follow up soon with the [changes needed](https://github.com/lightning/bolts/issues/1221#issuecomment-2626983064) to fix HTLC pinning.

05-onchain.md
morehouse: This wasn't fully addressed.  It still refers to a keyed anchor instead of P2A.

03-transactions.md
instagibbs: I *suspect* anyone using TRUC tx will keep this 0 since it's not a meaningful signal, especially post-fullrbf world?

I don't think we have enough data to say exactly.

TheBlueMatt (commented):
This was actually a good bit easier to implement than I thought.

Roasbeef:
> Just want to node that it was discussed in yesterday's spec call that as part of this work we'll probably also want to make option_simple_close transactions (optionally) v3, i.e., require v3 close transactions for v3 channels.

Compared to the HTLC case, are there any pinning concerns when it comes to coop close? The channel is no longer active, and the only outputs on the coop close transaction send funds directly to either party. 

I point this out as the TRUC constraints are passed on to all unconfirmed children. As a result, if the coop close transaction is v3, then a user wouldn't be able to CPFP with a non-v3 transaction to bump the fee (imagine the peer is just no longer offline). Another scenario that would be prevented is a non-v3 aware wallet spending that unconfirmed change to create a normal transaction, or fund a new channel.

02-peer-protocol.md
```diff
@@ -745,6 +744,7 @@ affect the channel operation).
 The currently defined basic types are:
   - `option_static_remotekey` (bit 12)
   - `option_anchors` and `option_static_remotekey` (bits 22 and 12)
+  - `zero_fee_commitments` (bit 40)
```
carlaKC, line 747: Why doesn't this channel type also have `option_static_remotekey`?

03-transactions.md
```diff
@@ -171,9 +174,25 @@ After 16 blocks, anyone can sweep the anchor with witness:
...
+#### `shared_anchor` Output (zero_fee_commitments)
+
+This output can be spent by anyone to provide incentive to mine the transaction, using child-pays-for-parent (since the commitment transaction doesn't pay any mining fees).
+This output is only added when the commitment transaction is using version 3 (`zero_fee_commitments`) and prevents pinning attacks on the commitment transaction.
+It is using the following standard P2A (pay-to-anchor) script:
+
+    OP_1 <0x4e73>
+
+The amount of this output is usually 0 sats, in which case it must be spent by the child transaction paying the mining fees.
+When the commitment contains [trimmed outputs](#trimmed-outputs), their amount is added to this `shared_anchor` output until it reaches 240 sats.
```
carlaKC, line 186: nit: I think it would be helpful to make a note here explaining why need to switch between the anchor and fees
* Ephemeral dust requires zero fees for values <= dust limit to be standard
* The dust limit for P2A is 240

Perhaps link to this issue: https://github.com/bitcoin/bitcoin/issues/31938 (helped me understand it better)

02-peer-protocol.md
carlaKC: > Why do you want to include the option_static_remotekey feature bit?

I thought that channel type is supposed to be all the features that a channel includes, so static remote key should be included here (because zero fee commitment channels will also use static remote keys).

It wasn't obvious to me that now that `option_static_remote_key` is assumed, it would also be assumed for channel types as well, but makes sense so that we can deprecate completely 👌

carlaKC: > The issue with thinking that way is that we conflated channel types and lightning feature bits, so if we consider that 0-fee commitments use option_static_remotekey we end up being force to continue supporting the plain option_static_remotekey channel type, which we want to deprecate at some point (ideally, soon).

Yeah makes sense 👍  I was indeed conflating the two ideas.

t-bast:
I have rebased to fix the merge conflicts, and squashed into a single commit. I have also:

- changed the explanation of the `240 sats` cap in Bolt 3
- added a warning in Bolt 5 about batching HTLCs transactions and the 10kvB limit

I'm working on adding a set of Bolt 3 test vectors.

03-transactions.md
t-bast: Done in the latest rebase: I haven't linked to the bitcoin issue though, I'm reluctant to link to remote repositories (even bitcoin) since issues can be closed and even deleted, I'd rather add more rationale directly inside the BOLTs to explain why we're doing it that way. Let me know if the change is clear enough, otherwise feel free to suggest some changes!

bolt03/zero-fee-commitments-test.json
```diff
@@ -0,0 +1,206 @@
...
+  "funding_amount_satoshis": 10000000,
+  "commitment_number": 42,
+  "to_self_delay": 720,
+  "local_payment_basepoint_secret": "94f29d20a225ea2f7093331ba0f0f28a9382d8ed08e1fd121329925cd0c01b6d",
+  "local_delayed_payment_basepoint_secret": "e9d4e1935bf16e948d76ad007baf0646df023af38f41bcf2c8799336949d291e",
+  "local_htlc_basepoint_secret": "f699038ef4f95b6b16b22a5c04fcb3c508d68d02cd2f86cf197e0fac451681b0",
+  "per_commitment_point": "0275d12130c276b4274358a328901f8fc47e6c72629102e4b46c9f27dd2c1dda98",
+  "remote_payment_basepoint_secret": "580bff39085f3a6ae8b1f32905e67366c522ea8f2418391145b2e98f1a7cb3f2",
+  "remote_htlc_basepoint_secret": "32df9c4dd46ab6210e74e81e15282106f8db883f45674eabb3324166c6513062",
+  "revocation_pubkey": "03e1ea1b8f5e4cae2dbac454a341ad5c3323b5827054b7e8b0497600010b31f860",
```
carlaKC: Could you provide the `revocation_basepoint` rather than the `revocation_pubkey` that's already been created from `revocation_basepoint` + `per_commitment_point`?

It's a bit hacky for me to override with the already derived pubkey in LDK.

t-bast: That should be doable, I'll do that early next week.

t-bast: Done in https://github.com/lightning/bolts/pull/1228/commits/3213327aaefdc0e5402ba9ee323d5b46914c69de, let me know if that works for you.

carlaKC: Works for LDK - thanks!

t-bast: That's great news! Do you have a branch of `ldk-server` that I can use for cross-compat tests?

carlaKC: All passing in https://github.com/lightningdevkit/rust-lightning/pull/4202 🎉 

> Do you have a branch of ldk-server that I can use for cross-compat tests?

@tankyleo should be able to help you out with interop testing!

tankyleo: > That's great news! Do you have a branch of ldk-server that I can use for cross-compat tests?

So far have tested on mainnet with the PR in `ldk-node`, but will get a `ldk-server` branch ready.

tankyleo: @t-bast I just published an `ldk-server` branch for 0FC here:

https://github.com/lightningdevkit/ldk-server/pull/75

have done some testing with it on mainnet, all good so far. Would recommend you configure the chain source with a http esplora backend.

For 0FC channels, we currently set feature bit 141 is that what you have too ? Edit: checked your code, yes sounds like we are on the same page there.

t-bast: Thanks, I'm able to do some testing with this branch, this is helpful!

03-transactions.md
```diff
@@ -172,9 +176,29 @@ After 16 blocks, anyone can sweep the anchor with witness:
...
+
+The standard dust limit for this type of script is `240 sat`.
+However, if the parent transaction doesn't pay any on-chain fees, using values below this dust limit is allowed under the "ephemeral dust" rule.
+
+We set the amount of this `shared_anchor` based on the [trimmed outputs](#trimmed-outputs) of the commitment transaction.
+When there are no trimmed outputs, it is set to `0 sat`.
+When there are trimmed outputs, their amount is added to this `shared_anchor` until it reaches `240 sats`: the remaining amount (if any) is not added
+to any output and thus directly contributes to the transaction's mining fees (which is standard since the `shared_anchor` has reached its dust limit).
+
+Spending of the output requires the following (emtpy) witness:
```
carlaKC: nit: emtpy/empty

02-peer-protocol.md
```diff
@@ -2241,26 +2253,27 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-    transaction fee when receiving or sending a future additional non-dust HTLC
-    while maintaining its channel reserve. It is recommended that this "fee spike
-    buffer" can handle twice the current `feerate_per_kw` to ensure predictability
-    between implementations.
-  - if it is _not responsible_ for paying the Bitcoin fee:
-    - SHOULD NOT offer `amount_msat` if, once the remote node adds that HTLC to
-    its commitment transaction, it cannot pay the fee for the updated local or
-    remote transaction at the current `feerate_per_kw` while maintaining its
-    channel reserve.
+  - if `zero_fees_commitments` has not been negotiated:
```
carlaKC: nit: zero_fees_commitments/zero_fee_commitments, we use the singular everywhere else

05-onchain.md
```diff
@@ -537,8 +541,12 @@ A local node:
   using the revocation private key.
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - if `option_anchors` applies:
+  - if `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions
+  (i.e. `option_anchors` or `zero_fee_commitments` applies):
     - MAY use a single transaction to *resolve* all the outputs.
+    - If `zero_fee_commitments` applies:
+      - MUST NOT exceed the 10kvB limit of v3 transactions, otherwise the
+        transaction may not reach miners.
```
carlaKC: Could we add a more general warning about V3 restrictions somewhere? You can also run into some V3 restrictions if you batch second stage transactions (eg, ~60 success claims will hit 10kvb).

carlaKC (approved):
> everything is working like a charm

🔥 

LGTM! Last test vector passes in LDK, just some nits and a request for a general "beware V3 gremlins" warning 🙏

bolt03/zero-fee-commitments-test.json
```diff
@@ -0,0 +1,255 @@
...
+        },
+        {
+          "id": 1,
+          "amount_msat": 1000000,
+          "payment_hash": "29a74a69c5941d402838f7e1a95c2b2ec534d79524b2582f48df7bc519ebaecf",
+          "cltv_expiry": 920130
+        },
+        {
+          "id": 2,
+          "amount_msat": 10000000,
```
tankyleo, line 358: Seems to me we are missing coverage for the case where non-zero "msat tails" get rounded down, and together can cause the anchor output to increase in value ?

03-transactions.md
t-bast: Thanks, fixed in https://github.com/lightning/bolts/pull/1228/commits/a0bc7bc462927ebe6a01a5644208a36519a98ca6

02-peer-protocol.md
t-bast: Thanks, fixed in https://github.com/lightning/bolts/pull/1228/commits/a0bc7bc462927ebe6a01a5644208a36519a98ca6

05-onchain.md
t-bast: I added a section in https://github.com/lightning/bolts/pull/1228/commits/a0bc7bc462927ebe6a01a5644208a36519a98ca6 with a link to the BIP. If there's specific content you'd like me to include, please suggest some text!

bolt03/zero-fee-commitments-test.json
t-bast: Great point, this is worth adding! I'll add one soon.

t-bast: I added three test vectors in https://github.com/lightning/bolts/pull/1228/commits/a7862935df307fe165255f1cf92fc24ef062e942, thanks for the suggestion.

bolt03/zero-fee-commitments-test.json
```diff
@@ -197,6 +197,81 @@
...
+      ],
+      "outgoing_htlcs": [
+        {
+          "id": 0,
+          "amount_msat": 21474,
+          "payment_hash": "29a74a69c5941d402838f7e1a95c2b2ec534d79524b2582f48df7bc519ebaecf",
+          "cltv_expiry": 920125
+        }
+      ],
+      "signed_commit_tx": "03000000000101ae1e3c841378cf9e4c383bfdd033d4b3c6945e0587ff16635a00b347eea2704b0100000000340fef800332000000000000000451024e7362841e0000000000160014f2123f1a4b67887f2e5f02eda73e6327010152eaea117a0000000000220020f2d298ffcfd6d899a3abada37bfc6f42ce0b7b66f3e39e903e8419ac97dca75a04004830450221008bc280fdd03504d216662ed63e3c9e48fbe1d9e7be293ec01822f721646d6c4e022004ea9032d5f13113ad46502d12b9dc64a18d51ee6c8152bee61b73ed51f06ad10147304402203ecdb06caa332381e7a07f0254b4242f4e8d2379a0e873f214782147d7a9cd71022058c2a0e82ceef47a682bd823af2fe110a82c62693b92084d6b194d8e27f7357801475221027eb9596a68740445fb151ff37d5422e7f65f2c497c90fda63e738eb606c15bd62103bbc16dc8851bece603322f06b3c8da329401b7be7e9fdd3f3090ad19aed0807052aec50fbb20",
```
tankyleo: I expect an anchor value of 52 sats on this transaction: `10000000 - 1999970 - 7999978 = 52`, not 50

bolt03/zero-fee-commitments-test.json
```diff
@@ -197,6 +197,81 @@
...
+      ],
+      "outgoing_htlcs": [
+        {
+          "id": 0,
+          "amount_msat": 21475,
+          "payment_hash": "29a74a69c5941d402838f7e1a95c2b2ec534d79524b2582f48df7bc519ebaecf",
+          "cltv_expiry": 920125
+        }
+      ],
+      "signed_commit_tx": "03000000000101ae1e3c841378cf9e4c383bfdd033d4b3c6945e0587ff16635a00b347eea2704b0100000000340fef800333000000000000000451024e7362841e0000000000160014f2123f1a4b67887f2e5f02eda73e6327010152eaea117a0000000000220020f2d298ffcfd6d899a3abada37bfc6f42ce0b7b66f3e39e903e8419ac97dca75a0400473044022033e3e272f752dd0f8eeb526f40b5c18dc44a528ba4a1257758baadee2a5e824b02207eeda638931de228f36fa4d07d97441d72f98afc1d0b27c9d22242f7d22e8c92014830450221008c50a570216ef1cc4f9c59a65ce06d4b0b6d2bfa72243116259c1c87a453913c02202808c8add90f5c2a0d3f7ca5c6a00ed6ec40557ea55edf14fafc138a8d05897801475221027eb9596a68740445fb151ff37d5422e7f65f2c497c90fda63e738eb606c15bd62103bbc16dc8851bece603322f06b3c8da329401b7be7e9fdd3f3090ad19aed0807052aec50fbb20",
```
tankyleo: Same as before, I expect 52 sats on the anchor, not 51

05-onchain.md
carlaKC: Thanks - looks great!

bolt03/zero-fee-commitments-test.json
```diff
@@ -197,6 +197,81 @@
...
+      ],
+      "outgoing_htlcs": [
+        {
+          "id": 0,
+          "amount_msat": 21712,
+          "payment_hash": "29a74a69c5941d402838f7e1a95c2b2ec534d79524b2582f48df7bc519ebaecf",
+          "cltv_expiry": 920125
+        }
+      ],
+      "signed_commit_tx": "03000000000101ae1e3c841378cf9e4c383bfdd033d4b3c6945e0587ff16635a00b347eea2704b0100000000340fef800333000000000000000451024e7362841e0000000000160014f2123f1a4b67887f2e5f02eda73e6327010152eaea117a0000000000220020f2d298ffcfd6d899a3abada37bfc6f42ce0b7b66f3e39e903e8419ac97dca75a0400473044022033e3e272f752dd0f8eeb526f40b5c18dc44a528ba4a1257758baadee2a5e824b02207eeda638931de228f36fa4d07d97441d72f98afc1d0b27c9d22242f7d22e8c92014830450221008c50a570216ef1cc4f9c59a65ce06d4b0b6d2bfa72243116259c1c87a453913c02202808c8add90f5c2a0d3f7ca5c6a00ed6ec40557ea55edf14fafc138a8d05897801475221027eb9596a68740445fb151ff37d5422e7f65f2c497c90fda63e738eb606c15bd62103bbc16dc8851bece603322f06b3c8da329401b7be7e9fdd3f3090ad19aed0807052aec50fbb20",
```
tankyleo: 52 sats on the anchor here too :)

bolt03/zero-fee-commitments-test.json
t-bast: Hah, good thing you're reviewing this seriously, because the code I wrote was wrong, I didn't spend enough time on unit tests for millisatoshi rounding, and even the spec was wrong :facepalm:, thanks a lot for spotting this!

The anchor amount isn't just set based on trimmed outputs, but also on any millisatoshi amount that is rounded down to satoshis. I changed the spec in https://github.com/lightning/bolts/pull/1228/commits/fc83fef80763f5a7917ae320f41420a33e7dad96 to fix that, fix those 3 test vectors, and add another test vector for millisatoshi rounding without dust outputs.

Let me know if those test vectors work with LDK, and if I finally got this right this time!

tankyleo: test vectors pass ! thank you for those, will be making another pass on the spec today

02-peer-protocol.md
```diff
@@ -2241,26 +2253,27 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
...
-    channel reserve.
+  - if `zero_fee_commitments` has not been negotiated:
+    - if it is _responsible_ for paying the Bitcoin fee:
+      - MUST NOT offer `amount_msat` if, after adding that HTLC to its commitment
+      transaction, it cannot pay the fee for either the local or remote commitment
+      transaction at the current `feerate_per_kw` while maintaining its channel
+      reserve (see [Updating Fees](#updating-fees-update_fee)).
+      - if `option_anchors` applies to this commitment transaction and the sending
+      node is the funder:
+        - MUST be able to additionally pay for `to_local_anchor` and
```
tankyleo: nit: some trailing whitespace here

tankyleo (approved):

02-peer-protocol.md
t-bast: Thanks, fixed. I squashed all commits together and updated the commit message to reflect the latest changes, this should be ready to go!

02-peer-protocol.md
Roasbeef, line 3141: This is still the case, why remove this section?

03-transactions.md
```diff
@@ -281,60 +306,68 @@ To redeem the HTLC, the HTLC-success transaction is used as detailed below. This
...
 #### Requirements
 
-The base fee and anchor output values:
-  - before the commitment transaction outputs are determined:
-    - MUST be subtracted from the `to_local` or `to_remote`
-    outputs, as specified in [Fee Calculation](#fee-calculation).
+Before the commitment transaction outputs are determined:
+  - the base fee MUST be subtracted from the `to_local` or `to_remote` output,
+    as specified in [Fee Calculation](#fee-calculation).
+  - if `option_anchors` applies:
```
Roasbeef, line 324: Should this actually be `zero_fee_commitments`? Don't understand this portion of the diff.

03-transactions.md
```diff
@@ -281,60 +306,68 @@ To redeem the HTLC, the HTLC-success transaction is used as detailed below. This
...
 The commitment transaction:
   - if the amount of the commitment transaction `to_local` output would be
-less than `dust_limit_satoshis` set by the transaction owner:
+    less than `dust_limit_satoshis` set by the transaction owner:
     - MUST NOT contain that output.
   - otherwise:
     - MUST be generated as specified in [`to_local` Output](#to_local-output).
   - if the amount of the commitment transaction `to_remote` output would be
-less than `dust_limit_satoshis` set by the transaction owner:
+    less than `dust_limit_satoshis` set by the transaction owner:
```
Roasbeef, line 335: White space/format changes like this littered all over the PR adds noise, and makes the diff as a whole harder to review. 

I sincerely hope that this is the last PR written in the style of an ever expanding document which specifies All Things. The extension BOLT format is a superior way to make spec changes, as the documents are self contained, carrying nearly all the context needed for review.

05-onchain.md
```diff
@@ -537,8 +542,13 @@ A local node:
...
   - SHOULD extract the payment preimage from the transaction input witness, if
   it's not already known.
-  - if `option_anchors` applies:
+  - if `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used for HTLC transactions
+  (i.e. `option_anchors` or `zero_fee_commitments` applies):
     - MAY use a single transaction to *resolve* all the outputs.
+    - If `zero_fee_commitments` applies:
+      - MUST NOT exceed the 10kvB limit of v3 transactions, otherwise the
+        transaction may not reach miners (as explained in the section
+        about [v3 limitations](#limitations-of-v3-transactions) below).
```
Roasbeef, line 551: IIUC, this doesn't apply if the transaction has already been confirmed before a sweep is attempted. AFAIK, we all currently wait for confirmation before attempting to sweep a channel's output.

05-onchain.md
```diff
@@ -610,25 +621,25 @@ Note: even if the `to_remote` output is not swept, the resulting
...
+HTLC-success transactions are signed with the input and output having the
+same value. This means they have a zero fee and MUST be combined with other
+inputs to arrive at a reasonable fee.
 
 ## Requirements
 
 A node which broadcasts an HTLC-success or HTLC-timeout transaction for a
 commitment transaction:
-  - if `option_anchors` applies:
+  - if `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY` is used:
```
Roasbeef, line 642: It would be more precise to just enumerate the channel types here so `option_anchors` and `zero_fee_commitments`. Or in the glossary make a new entry for `anchor_channels` or something like that, and ref as necessary.

05-onchain.md
```diff
@@ -649,6 +660,20 @@ A node:
     - Note: watching for mempool transactions should result in lower latency
     HTLC redemptions.
 
+## Limitations of v3 transactions
```
Roasbeef, line 663: Let's say a user has a zero conf channel, the funding transaction is currently unconfirmed and a v3 transaction. Are they able to do a coop close with a v2 transaction, or will the bitcoind logic reject that as its considered an ancestor of the funding transaction?

03-transactions.md
t-bast: No, this isn't for `zero_fee_commitments`, this is for existing anchor output channels. The spec here was missing this part, which was in other sections but should have been in this one as well. The introduction of the shared anchor forced me to "refactor" a bit this file to properly distinguish the shared anchor case from the local/remote anchor case.

05-onchain.md
t-bast: I'm not sure I understand your comment here. HTLC transactions are using v3, so if you batch a bunch of HTLC transactions together (using the fact that their input uses `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY`), the resulting transaction is a batched HTLC transaction that is using v3. So regardless of whether the commitment transaction is confirmed or not, the batched HTLC transaction cannot exceed 10kvB. Does that make sense?

05-onchain.md
t-bast: I disagree, I think it makes more sense to state the condition that makes adding another input possible (it's only possible because of those sighash flags on the input). Also, it's more future-proof because as we add more channel types that also use `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY`, we don't have to edit all the places where we list channel types, the requirements are automatically inherited by the fact that we specify what should happen for `SIGHASH_SINGLE | SIGHASH_ANYONECANPAY` inputs.

05-onchain.md
t-bast: It will be rejected since its parent transaction is an unconfirmed v3 transaction. That's one of the reasons we use "high" feerates for 0-conf transactions in Phoenix, especially when there are no change outputs available to CPFP.

02-peer-protocol.md
t-bast: I can't see which section you're referring to here, github doesn't link that comment to any code change...

t-bast:
As discussed during yesterday's spec meeting, rebased to fix minor conflicts with splicing. This is ready to integrate as we have cross-compatibility between LDK and eclair. @tankyleo @TheBlueMatt, can you give a final ACK before I merge?

02-peer-protocol.md
```diff
@@ -809,15 +809,19 @@ The sending node:
...
   - set `to_self_delay` sufficient to ensure the sender can irreversibly spend a commitment transaction output, in case of misbehavior by the receiver.
-  - set `feerate_per_kw` to at least the rate it estimates would cause the transaction to be immediately included in a block.
-  - set `dust_limit_satoshis` to a sufficient value to allow commitment transactions to propagate through the Bitcoin network.
-  - set `dust_limit_satoshis` to a sufficient value to allow HTLC transactions to propagate through the Bitcoin network.
   - set `htlc_minimum_msat` to the minimum value HTLC it's willing to accept from this peer.
+  - when using `zero_fee_commitments` or `option_anchors`:
+    - set `dust_limit_satoshis` to a sufficient value to ensure that it is economical to spend, taking the cost of HTLC transactions into account.
+  - otherwise (not using `zero_fee_commitments` or `option_anchors`):
+    - set `feerate_per_kw` to at least the rate it estimates would cause the transaction to be immediately included in a block.
+    - set `dust_limit_satoshis` to a sufficient value to allow commitment transactions to propagate through the Bitcoin network.
```
TheBlueMatt: this applies in both cases.

02-peer-protocol.md
```diff
@@ -3253,6 +3270,9 @@ A sending node:
     - MUST ensure that requirements are met for all commmitment transactions.
 
 A receiving node:
+  - if `zero_fee_commitments` is used:
+    - MUST ignore the message.
+    - SHOULD send a `warning` and disconnect.
```
TheBlueMatt: I don't think this is useful tbh. Many nodes will send a commitment-update-message in a batch with the actual `commitment_signed`. As a result, on reconnect they have to resend the commitment update (they already signed it!) and we just enter a reconnect loop. We should instead fail the channel - the peer is hopelessly broken.

02-peer-protocol.md
t-bast: Indeed, moved back above where it applies to both cases in https://github.com/lightning/bolts/pull/1228/commits/e59e30ca504f9ff6999c80f66e0cddf132cd2197.

02-peer-protocol.md
t-bast: Makes sense, changed that to a channel fail in https://github.com/lightning/bolts/pull/1228/commits/e59e30ca504f9ff6999c80f66e0cddf132cd2197.

t-bast:
@tankyleo @TheBlueMatt, shall we? :innocent:

PR #1232, opened 2025-02-24
Assume `option_channel_type`

t-bast:
This feature has become more and more useful as we've introduced various channel commitment formats, and has been introduced more than 3 years ago in #880. It should be safe to assume that every implementation sets this TLV fields?

We also add a new context type `T` for features that can be included in the `channel_type` field when opening channels in the second commit.

### DIFF
--- a/09-features.md
+++ b/09-features.md
@@ -29,6 +29,7 @@ The Context column decodes as follows:
 * `C+`: presented in the `channel_announcement` message, but always even (required).
 * `9`: presented in [BOLT 11](11-payment-encoding.md) invoices.
 * `B`: presented in the `allowed_features` field of a blinded path.
+* `T`: used in the `channel_type` field [when opening channels](02-peer-protocol.md#the-open_channel-message).
 
 | Bits  | Name                              | Description                                               | Context  | Dependencies                | Link                                                                  |
 |-------|-----------------------------------|-----------------------------------------------------------|----------|-----------------------------|-----------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


morehouse (commented):
There's a bunch of "default channel type" logic that we should also remove from the `funding_signed` requirements.

Roasbeef:
Wouldn't the first step here be for everyone to _first_ start to set the required feature bit? That's the existing mechanism we have in the protocol to change something from optional to mandatory. Compared to this PR, it's explicit and unambiguous.

rustyrussell:
I'm happy to set it as compulsory in the CLN 25.05 release.  We added this in 0.12.0 (2022-08-23), though there were fixes as late as 24.02.  Either way that's well outside our support window.

rustyrussell:
However, there are about 2600 nodes which still don't advertize support.  Those versions are all well out-of-date though, so those results may be false.  Further analysis required.

For example, some nodes haven't sent out a channel update in *years*.  Their channels are kept alive by their peers retransmitting (in most cases, with the disabled flag set).  CLN has logic to *not* create refresh channel_update messages for disabled channels, maybe this should be in the spec?  That would allow these channels (and thus nodes) to vanish from the graph naturally.

Though really, closing uncontactable channels after a year would also be a decent spec update!

Consider:

https://mempool.space/lightning/node/03c918ba3e2fd8207bf4c54a546e5b7e60bd4148a805905e7f701b1095318def0c

It seems like its peers are keeping it alive by refreshing the channel_update.  But I'd guess they haven't heard from it in months / years, so they shouldn't?

Roasbeef (approved):
LGTM 🏷️

PR #1236, opened 2025-03-10
Allow either node to send `tx_init_rbf` in v2 establishment

jkczyz:
Splicing allows either node to send `tx_init_rbf` as they may want to take the opportunity to contribute or withdraw additional funds to / from the channel. Allow the same for v2 channel establishment for consistency.


02-peer-protocol.md
```diff
@@ -1407,7 +1407,7 @@ a transaction paying more fees to make the channel confirm faster.
 #### Requirements
 
 The sender of `tx_init_rbf`:
-  - MUST be the *initiator*
+  - MAY be the either the *initiator* or the *accepter*
   - MUST NOT have sent or received a `channel_ready` message.
```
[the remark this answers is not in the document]
jkczyz: Updated accordingly

vincenzopalazzo (approved):
LGTM

ddustin:
I haven't tested it but this should work fine in CLN. We essentially treat RBF as just a `splice_init` anyway.

02-peer-protocol.md
```diff
@@ -1407,7 +1407,10 @@ a transaction paying more fees to make the channel confirm faster.
 #### Requirements
 
 The sender of `tx_init_rbf`:
-  - MUST be the *initiator*
+  - MAY be either the *initiator* or the *accepter*
+    - If the sender is the accepter, it becomes the initiator of the `interactive-tx` session and thus:
```
niftynei, line 1419: Hello! Looking at updating CLN to accommodate this, and ran into a question which seems obvious but wanted to confirm.

I'm taking this to mean that the serial_ids and all other role-related indicators are swapped for the entirety of the tx construction phase?

niftynei:
Is it an implementation detail to have the OG opening peer re-adding their original inputs to the transaction or is that something we should specify?

02-peer-protocol.md
jkczyz: Correct

jkczyz:
> Is it an implementation detail to have the OG opening peer re-adding their original inputs to the transaction or is that something we should specify?

I'd say it's an implementation detail. They could choose to not add them / take the opportunity to use entirely different inputs / amounts.

jkczyz:
> I'd say it's an implementation detail. They could choose to not add them / take the opportunity to use entirely different inputs / amounts.

Oh, and since they are no longer the initiator, they wouldn't be paying for the common fields nor the funding output. So they may decide to choose different UTXOs to spend as their fee responsibility may decrease.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1236/commits/90ae27568c2a2c54a931803e9397cff4ad51d87a

02-peer-protocol.md
```diff
@@ -464,6 +464,10 @@ The sender:
     - MUST set `funding_output_contribution`
   - If it requires the receiving node to only use confirmed inputs:
     - MUST set `require_confirmed_inputs`
+  - If it contributed to previous transactions:
+    - MUST send `tx_add_input` with at least one input that it used in previous
+      transaction construction attempts, to ensure that the new transaction
+      double-spends all other attempts.
```
SomberNight, line 470: This is phrased in an ambiguous way.
I guess what is meant is:
> MUST send `tx_add_input` with at least one input from *each* previous transaction construction attempt

For example, if
- the original tx1 attempt has inputs A and B,
- and then replacement tx2 has inputs B and C,
- then replacement tx3 should not use inputs C and D,

as tx1 and tx3 would not conflict. tx3 satisfies one interpretation of the current text: it shares at least one input it used in a previous attempt (C was used in tx2).

However as the rationale is clearly stated ("to ensure that the new transaction double-spends all other attempts."), perhaps this is a non-issue.

I guess the simplest implementation is to always re-add all previous inputs, which side-steps most of the complexity.

t-bast: I see how that can be misinterpreted indeed! But thankfully `to ensure that the new transaction double-spends all other attempts` clears it up I believe? I don't think it's a good idea to require reusing all previous inputs, this is limiting for no good reason.

For example, if:

- the original tx1 has inputs A and B
- to increase the feerate, the replacement tx2 has inputs B and C where C is larger than A

This is more efficient (smaller tx) than having to keep input A and add yet another input?

SomberNight: > I don't think it's a good idea to require reusing all previous inputs, this is limiting for no good reason.

Right, I did not mean requiring that, just remarked that is one way to write a simple and safe implementation. (e.g. that's how user-facing RBF works in Electrum. As the user might have the same seed on both a laptop and a phone at the same time, and might do bumpfee on a random tx first on one machine and then on another machine, any given machine might be missing information about some previous tx variants, so the only safe way of changing the input set is by never removing inputs. Otherwise we would have to sync previous tx replacements between machines via nostr or something :D  In the case of LN, it seems unlikely multiple machines try to RBF the funding tx, so that specific concern does not apply.)

What I meant is that it might be worth it to rephrase that text (without changing its intent):
```
  - If it contributed to previous transactions:
    - MUST ensure that the new transaction double-spends all other attempts,
      for example by sending `tx_add_input` with at least one input
      from each previous transaction construction attempt.
```

t-bast: Sounds good, feel free to open a clarifying PR!

PR #1240, opened 2025-03-24
BOLT 12: re-add recurrence support.

rustyrussell:
This is taken from the old draft, and updated with modern field numbers and names.

I've made some minor changes:
1. The `proportional_amount` field comes before `seconds_before` which seems more logical to me.
2. The `offer_absolute_expiry` is clarified to only apply to the *initial* invoice request: once you're recurring the other fields take over.

The complexities mainly come from two sources:

1. A requirement to be precise about when recurrence happens.  This is because we want a push system, not a pull, since only the former allows for real authorization by the user, and there are many different reasonable ways to do recurrence.
2. This is the first wallet feature which isn't just "fire and forget", but requires them to keep state and do unprompted actions.

One proposal I would like to add is a courtesy onion message when a user cancels a periodic payment.  This is generally useful to tell the difference between a deliberate cancellation and a user-related failure.

### DIFF
--- a/12-offer-encoding.md
+++ b/12-offer-encoding.md
@@ -364,8 +364,8 @@ A writer of an offer:
     - MUST NOT set `offer_quantity_max`.
   - If an offer MAY trigger time-spaced invoice requests:
     - MUST include exactly one of `offer_recurrence_optional` or `offer_recurrence_compulsory`.
-	- MAY use `offer_recurrence_optional` if a payment of a single period would be useful (compatibility with pre-recurrence readers).
-	- MUST NOT use `offer_recurrence_optional` if `offer_recurrence_base` is present.
+    - MAY use `offer_recurrence_optional` if a payment of a single period would be useful (compatibility with pre-recurrence readers).
+    - MUST NOT use `offer_recurrence_optional` if `offer_recurrence_base` is present.
   - if it includes either `offer_recurrence_optional` or `offer_recurrence_compulsory`:
       - MUST set `time_unit` to 0 (seconds), 1 (days), 2 (months) or 3 (years).
       - MUST set `period` to how often (in `time-unit`) it wants to be paid.
@@ -470,7 +470,7 @@ painful to have to special-case the "only one left" offer generation.
 
 Offers can be used to simply send money without expecting anything in return (tips, kudos, donations, etc), which means the description field is optional (the `offer_issuer` field is very useful for this case!); if you are charging for something specific, the description is vital for the user to know what it was they paid for.
 
-We insist that recurring requests be in order (thus, if you pay an invoice for #34 of a recurring offer, it implicitly commits to the successful payment of #0 through #33).  The `invreq_recurrence_paywindow` constrains how far you can pay in advance precisely, and if it isn't in the offer the defaults provide some slack, without allowing commitments into the far future.
+We insist that recurring requests be in order (thus, if you pay an invoice for #34 of a recurring offer, it implicitly commits to the successful payment of #0 through #33).  The `offer_recurrence_paywindow` constrains how far you can pay in advance precisely, and if it isn't in the offer the defaults provide some slack, without allowing commitments into the far future.
 
 # Invoice Requests
 
--- a/12-offer-encoding.md
+++ b/12-offer-encoding.md
@@ -592,6 +592,7 @@ while still allowing signature validation.
     1. type: 93 (`invreq_recurrence_start`)
     2. data:
         * [`tu32`:`period_offset`]
+    1. type: 94 (`invreq_recurrence_cancel`)
     1. type: 240 (`signature`)
     2. data:
         * [`bip340sig`:`sig`]
@@ -606,6 +607,9 @@ The writer:
     - otherwise:
       - if it sets `invreq_chain` it MUST set it to bitcoin.
     - MUST set `signature`.`sig` as detailed in [Signature Calculation](#signature-calculation) using the `invreq_payer_id`.
+      - if it sets `invreq_recurrence_cancel`:
+      - SHOULD NOT send any future `invoice_request` for this offer with this `invreq_payer_id` without `invreq_recurrence_cancel` set.
+      - MAY omit `invreq_amount`, `invreq_quantity` and `invreq_paths`.
     - if `offer_amount` is not present:
       - MUST specify `invreq_amount`.
     - otherwise:
@@ -624,9 +628,11 @@ The writer:
       - for the initial request:
         - MUST use a unique `invreq_payer_id`.
         - MUST set `invreq_recurrence_counter` `counter` to 0.
+        - MUST NOT set `invreq_recurrence_cancel`.
       - for any successive requests:
         - MUST use the same `invreq_payer_id` as the initial request.
         - MUST set `invreq_recurrence_counter` `counter` to one greater than the highest-paid invoice.
+        - MAY set `invreq_recurrence_cancel`.
       - if `offer_recurrence_base` is present:
         - MUST include `invreq_recurrence_start`
         - MUST set `period_offset` to the period the sender wants for the initial request
@@ -645,6 +651,7 @@ The writer:
     - otherwise:
       - MUST NOT set `invreq_recurrence_counter`.
       - MUST NOT set `invreq_recurrence_start`
+      - MUST NOT set `invreq_recurrence_cancel`
   - otherwise (not responding to an offer):
     - MUST set `offer_description` to a complete description of the purpose of the payment.
     - MUST set (or not set) `offer_absolute_expiry` and `offer_issuer` as it would for an offer.
@@ -676,6 +683,9 @@ The reader:
   - MUST reject the invoice request if `signature` is not correct as detailed in [Signature Calculation](#signature-calculation) using the `invreq_payer_id`.
   - if `num_hops` is 0 in any `blinded_path` in `invreq_paths`:
     - MUST reject the invoice request.
+  - if `invreq_recurrence_cancel` is present:
+    - SHOULD reject any successive `invoice_request` for this offer and `invreq_payer_id`.
+    - MUST NOT reply.
   - if `offer_issuer_id` is present, and `invreq_metadata` and `invreq_recurrence_counter` (if any) are identical to a previous `invoice_request`:
     - MAY simply reply with the previous invoice.
   - otherwise:
@@ -1043,6 +1053,10 @@ A reader of an invoice:
     - MUST have authorization for the payment purpose, recipient and amount.
     - if `offer_recurrence_optional` or `offer_recurrence_compulsory` are present:
       - SHOULD obtain single authorization which covers all expected payments.
+      - SHOULD send the next `invoice_request` once the current period has expired (unless `offer_recurrence_limit` is reached).
+      - If authorization is cancelled:
+        - SHOULD send that `invoice_request` with `invreq_recurrence_cancel` set.
+        - SHOULD NOT send another `invoice_request` for this offer with the same `invreq_payer_id`.
 
 ## Rationale
 
@@ -1097,6 +1111,8 @@ Authorization can be very general ("install this app to spray sats to the world!
 For example, consider an offer with weekly recurrence (`time_unit`=1, `period`=7), `amount` 500, `currency` `AUD` ($5 Australian dollars).  An implementation may present this to the user as USD $3.53 (max $3.71), to allow up to 5% exchange slippage, and receive their authorization.  As it received each invoice, it would convert the `msat` into USD to check that it was below the maximum authorization of USD$3.71.  If it was, it would simply pay the invoice without user interaction.  If not, it requires re-authorization.
 
 Note that the implementation of a trusted exchange rate service is left to the reader.
+
+Users can cancel their recurring payment at any time, but as a courtesy to the offer issuer (to differentiate from technical or unintended missing payments), the following invoice_request *is sent*, but it explicitly indicates the intention to cancel (optionally using `invreq_payer_note`).  The delay in sending this until the next payment is due also discourages early cancellation of service.  Note that this cancellation message is best-effort, since no reply is received.
 
 # Invoice Errors
 
--- a/12-offer-encoding.md
+++ b/12-offer-encoding.md
@@ -618,9 +618,9 @@ The writer:
     - otherwise:
       - if it sets `invreq_chain` it MUST set it to bitcoin.
     - MUST set `signature`.`sig` as detailed in [Signature Calculation](#signature-calculation) using the `invreq_payer_id`.
-      - if it sets `invreq_recurrence_cancel`:
+    - if it sets `invreq_recurrence_cancel`:
       - SHOULD NOT send any future `invoice_request` for this offer with this `invreq_payer_id` without `invreq_recurrence_cancel` set.
-      - MAY omit `invreq_amount`, `invreq_quantity` and `invreq_paths`.
+      - MAY omit `invreq_amount` and `invreq_quantity`.
     - if `offer_amount` is not present:
       - MUST specify `invreq_amount`.
     - otherwise:
@@ -672,6 +672,7 @@ The writer:
     - if the chain for the invoice is not solely bitcoin:
       - MUST specify `invreq_chain` the offer is valid for.
     - MUST set `invreq_amount`.
+    - MUST NOT set `invreq_recurrence_cancel`.
   - MUST NOT set any non-signature TLV fields outside the inclusive ranges: 0 to 159 and 1000000000 to 2999999999
   - MUST set `invreq_metadata` to an unpredictable series of bytes.
   - if it sets `invreq_amount`:
@@ -694,9 +695,6 @@ The reader:
   - MUST reject the invoice request if `signature` is not correct as detailed in [Signature Calculation](#signature-calculation) using the `invreq_payer_id`.
   - if `num_hops` is 0 in any `blinded_path` in `invreq_paths`:
     - MUST reject the invoice request.
-  - if `invreq_recurrence_cancel` is present:
-    - SHOULD reject any successive `invoice_request` for this offer and `invreq_payer_id`.
-    - MUST NOT reply.
   - if `offer_issuer_id` is present, and `invreq_metadata` and `invreq_recurrence_counter` (if any) are identical to a previous `invoice_request`:
     - MAY simply reply with the previous invoice.
   - otherwise:
@@ -708,7 +706,7 @@ The reader:
     - otherwise:
       - MUST ignore any invoice_request if it arrived via a blinded path.
     - if `offer_quantity_max` is present:
-      - MUST reject the invoice request if there is no `invreq_quantity` field.
+      - MUST reject the invoice request if `invreq_recurrence_cancel` is not present and there is no `invreq_quantity` field.
       - if `offer_quantity_max` is non-zero:
         - MUST reject the invoice request if `invreq_quantity` is zero, OR greater than `offer_quantity_max`.
     - otherwise:
@@ -725,7 +723,9 @@ The reader:
         - MUST reject the invoice request if `invreq_amount`.`msat` is less than the *expected amount*.
         - MAY reject the invoice request if `invreq_amount`.`msat` greatly exceeds the *expected amount*.
     - otherwise (no `offer_amount`):
-      - MUST reject the invoice request if it does not contain `invreq_amount`.
+      - MUST reject the invoice request if `invreq_recurrence_cancel` is not present and it does not contain `invreq_amount`.
+    - if `offer_absolute_expiry` is present, and `invreq_recurrence_counter` is either not present or equal to 0:
+      - MUST reject the invoice request if the current time is after `offer_absolute_expiry`.
     - if `offer_recurrence_optional` or `offer_recurrence_compulsory` are present:
       - MUST reject the invoice request if there is no `invreq_recurrence_counter` field.
       - if `offer_recurrence_base` is present:
@@ -737,16 +737,23 @@ The reader:
       - if `offer_recurrence_limit` is present:
         - MUST reject the invoice request if the period index is greater than `max_period`.
       - MUST calculate the period using the period index as detailed in [Period Calculation](#offer-period-calculation).
-      - if `invreq_recurrence_counter` is non-zero:
+      - MUST reject the invoice request if an invoice has prevously been paid for this offer, `offer_issuer_id`, `invreq_metadata` and `invreq_recurrence_counter`.
+      - if `invreq_recurrence_counter` is zero (initial request):
+        - MUST reject the invoice request if `invreq_recurrence_cancel` is present.
+      - otherwise: (successive requests)
         - MUST reject the invoice request if no invoice for the previous period has been paid.
+        - MUST reject the invoice request if it has previously received a valid invoice request with `invreq_recurrence_cancel`.
         - if `offer_recurrence_paywindow` is present:
           - SHOULD reject the invoice request if the current time is before the start of the period minus `seconds_before`.
           - SHOULD reject the invoice request if the current time is equal to or after the start of the period plus `seconds_after`.
         - otherwise:
           - SHOULD reject the invoice request if the current time is prior to the start of the previous period.
+        - if `invreq_recurrence_cancel` is present:
+          - MUST NOT send an invoice in reply.
     - otherwise (no recurrence):
       - MUST reject the invoice request if there is a `invreq_recurrence_counter` field.
       - MUST reject the invoice request if there is a `invreq_recurrence_start` field.
+      - MUST reject the invoice request if there is a `invreq_recurrence_cancel` field.
     - SHOULD send an invoice in response using the `onionmsg_tlv` `reply_path`.
   - otherwise (no `offer_issuer_id` or `offer_paths`, not a response to our offer):
     - MUST reject the invoice request if any of the following are present:
--- a/12-offer-encoding.md
+++ b/12-offer-encoding.md
@@ -235,7 +235,6 @@ The human-readable prefix for offers is `lno`.
         * [`tu32`:`period`]
     1. type: 25 (`offer_recurrence_paywindow`)
     2. data:
-        * [`byte`:`proportional_amount`]
         * [`u32`:`seconds_before`]
         * [`tu32`:`seconds_after`]
     1. type: 26 (`offer_recurrence_limit`)
@@ -243,7 +242,7 @@ The human-readable prefix for offers is `lno`.
         * [`tu32`:`max_period`]
     1. type: 27 (`offer_recurrence_base`)
     2. data:
-        * [`byte`:`start_any_period`]
+        * [`byte`:`proportional_amount`]
         * [`tu64`:`basetime`]
 
 ### Recurrence
@@ -260,15 +259,15 @@ Some offers are *periodic*, such as a subscription service or monthly dues, in t
    * which repeat from whenever you made the first payment
 
 Thus, each offer containing a recurring payment has:
-1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months), 3 (years).
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months).
 2. A `period`, defining how often (in `time_unit`) it has to be paid.
 3. An optional `offer_recurrence_limit` of total payments to be paid.
 4. An optional `offer_recurrence_base`:
    * `basetime`, defining when the first period starts in seconds since 1970-01-01 UTC.
-   * `start_any_period` if non-zero, meaning you don't have to start paying at the period indicated by `basetime`, but can use `offer_recurrence_start` to indicate what period you are starting at.
+   * The payer uses `invreq_recurrence_start` to indicate what period they are starting at.  If you don't want them to start at arbitrary periods, use `offer_absolute_expiry`.
+   * A `proportional_amount` flag: if set indicating that a payment made during the period itself will be charged proportionally to the remaining time in the period (e.g. 150 seconds into a 1500 second period gives a 10% discount).
 5. An optional `offer_recurrence_paywindow`:
    * `seconds_before`, defining how many seconds prior to the start of the period a payment will be accepted.
-   * `proportional_amount`, if set indicating that a payment made during the period itself will be charged proportionally to the remaining time in the period (e.g. 150 seconds into a 1500 second period gives a 10% discount).
    * `seconds_after`, defining how many seconds after the start of the period a payment will be accepted.
   If this field is missing, payment will be accepted during the prior period and the paid-for period.
 
@@ -284,31 +283,24 @@ Each period has a zero-based index, and a start time and an end time.  Because t
   - the start of period #0 is the `invoice_created_at`.`timestamp` of the first `invoice` for this particular offer and `invreq_payer_id`.
 
 To calculate UTC time of the start of period #`N` for `N` > 0:
+- if the period #0 start is on a leap second:
+  - the *period base* is one second before period #0 start.
+- otherwise:
+  - the *period base* is the period #0 start.
 - if `time_unit` is 0 (seconds):
-  - period `N` starts at period #0 start plus `period` multiplied by `N`, in seconds.
+  - period `N` starts at *period base* plus `period` multiplied by `N`, in seconds.
 - otherwise, if `time_unit` is 1 (days):
-  - calculate the offset in seconds within the day of period #0 start.
+  - calculate the offset in seconds within the day of *period base*.
   - add `period` multiplied by `N` days to get the day of the period start.
   - add the offset in seconds to get the period end in seconds.
 - otherwise, if `time_unit` is 2 (months):
-  - calculate the offset in days within the month of period #0 start.
-  - calculate the offset in seconds within the day of period #0 start.
+  - calculate the offset in days within the month of *period base*.
+  - calculate the offset in seconds within the day of *period base*.
   - add `period` multiplied by `N` months to get the month of the period start.
   - add the offset days to get the day of the period start.
     - if the day is not within the month, use the last day within the month.
   - add the offset seconds to get the period start in seconds.
-- otherwise, if `time_unit` is 3 (years):
-  - calculate the offset in months within the year of period #0 start.
-  - calculate the offset in days within the month of period #0 start.
-  - calculate the offset in seconds within the day of period #0 start.
-  - add `period` multiplied by `N` years to get the year of the period start.
-  - add the offset months to get the month of the period start.
-  - add the offset days to get the day of the period start.
-    - if the day is not within the month, use the last day within the month.
-  - add the offset seconds to get the period start in seconds.
 - otherwise, the time is invalid.
-
-Note that offset seconds can overflow only if the period start is in a leap second; we ignore this!
 
 See [offer-period-test.json](bolt12/offer-period-test.json).
 
@@ -364,7 +356,7 @@ A writer of an offer:
       - MUST set `offer_quantity_max` to 0.
   - otherwise:
     - MUST NOT set `offer_quantity_max`.
-  - MAY include `offer_recurrence` to indicate offer should trigger time-spaced invoices.
+  - MAY include `offer_recurrence` to indicate offer should trigger time-spaced invoice requests.
   - if it includes `offer_recurrence`:
       - MUST set `time_unit` to 0 (seconds), 1 (days), 2 (months) or 3 (years).
       - MUST set `period` to how often (in `time-unit`) it wants to be paid.
@@ -376,10 +368,10 @@ A writer of an offer:
       - if periods are always at specific time offsets:
         - MUST include `offer_recurrence_base`
         - MUST set `basetime` to the initial period time in number of seconds after midnight 1 January 1970, UTC
-        - if the first paid-for-period does not have to be the initial period:
-          - MUST set `start_any_period` to 1.
+        - if `amount` is specified and the node will proportionally reduce the amount charged for a period payed after the start of the period:
+          - MUST set `proportional_amount` to 1 
         - otherwise:
-          - MUST set `start_any_period` to 0.
+          - MUST set `proportional_amount` to 0
       - otherwise:
         - MUST NOT include `offer_recurrence_base`.
       - if payments will be accepted for any time in the current or next period:
@@ -391,10 +383,6 @@ A writer of an offer:
         - MUST set `seconds_after` to the maximum number of seconds into to a period for which it will accept payment or invoice_request for that period.
         - MAY NOT enforce this for the initial period for offers without `offer_recurrence_base`
         - SHOULD NOT set `seconds_after` to greater than the maximum number of seconds in a period.
-        - if `amount` is specified and the node will proportionally reduce the amount charged for a period payed after the start of the period:
-          - MUST set `proportional_amount` to 1 
-        - otherwise:
-          - MUST set `proportional_amount` to 0
   - otherwise:
     - MUST NOT include `offer_recurrence_base`.
     - MUST NOT include `offer_recurrence_paywindow`.
@@ -549,7 +537,6 @@ while still allowing signature validation.
         * [`tu32`:`period`]
     1. type: 25 (`offer_recurrence_paywindow`)
     2. data:
-        * [`byte`:`proportional_amount`]
         * [`u32`:`seconds_before`]
         * [`tu32`:`seconds_after`]
     1. type: 26 (`offer_recurrence_limit`)
@@ -557,7 +544,7 @@ while still allowing signature validation.
         * [`tu32`:`max_period`]
     1. type: 27 (`offer_recurrence_base`)
     2. data:
-        * [`byte`:`start_any_period`]
+        * [`byte`:`proportional_amount`]
         * [`tu64`:`basetime`]
     1. type: 80 (`invreq_chain`)
     2. data:
@@ -627,7 +614,7 @@ The writer:
       - for any successive requests:
         - MUST use the same `invreq_payer_id` as the initial request.
         - MUST set `invreq_recurrence_counter` `counter` to one greater than the highest-paid invoice.
-      - if `offer_recurrence_base` is present with `start_any_period` non-zero:
+      - if `offer_recurrence_base` is present:
         - MUST include `invreq_recurrence_start`
         - MUST set `period_offset` to the period the sender wants for the initial request
         - MUST set `period_offset` to the same value on all following requests.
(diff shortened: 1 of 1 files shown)


12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
...
+* Payments due on every Monday, 1pm Pacific Standard Time.
+* Payments due once a year:
+   * which must be made on January 1st, or
+   * which are only valid if started January 1st 2021, or
+   * which if paid after January 1st you (over) pay the full rate first year, or
+   * which if paid after January 1st are paid pro-rata for the first year, or
+   * which repeat from whenever you made the first payment
+
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months), 3 (years).
```
thomash-acinq, line 263: I understand the need for seconds, days and months but years seem redundant as 1 year is always 12 months.

12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
...
+2. A `period`, defining how often (in `time_unit`) it has to be paid.
+3. An optional `offer_recurrence_limit` of total payments to be paid.
+4. An optional `offer_recurrence_base`:
+   * `basetime`, defining when the first period starts in seconds since 1970-01-01 UTC.
+   * `start_any_period` if non-zero, meaning you don't have to start paying at the period indicated by `basetime`, but can use `offer_recurrence_start` to indicate what period you are starting at.
```
thomash-acinq, line 268:
```suggestion
   * `start_any_period` if non-zero, meaning you don't have to start paying at the period indicated by `basetime`, but can use `invreq_recurrence_start` to indicate what period you are starting at.
```

12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
...
+  - calculate the offset in seconds within the day of period #0 start.
+  - add `period` multiplied by `N` days to get the day of the period start.
+  - add the offset in seconds to get the period end in seconds.
+- otherwise, if `time_unit` is 2 (months):
+  - calculate the offset in days within the month of period #0 start.
+  - calculate the offset in seconds within the day of period #0 start.
+  - add `period` multiplied by `N` months to get the month of the period start.
+  - add the offset days to get the day of the period start.
+    - if the day is not within the month, use the last day within the month.
+  - add the offset seconds to get the period start in seconds.
```
thomash-acinq, line 299: If the day is not within the month, we should use the last day of the month **and the last second of that day**.

We want a one month period starting on January 30th at 2AM to end before a one month period starting on January 31th at 1AM as it starts almost a full day later. However with your proposal, the former would end an hour after the latter.

12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
...
+  - calculate the offset in days within the month of period #0 start.
+  - calculate the offset in seconds within the day of period #0 start.
+  - add `period` multiplied by `N` years to get the year of the period start.
+  - add the offset months to get the month of the period start.
+  - add the offset days to get the day of the period start.
+    - if the day is not within the month, use the last day within the month.
+  - add the offset seconds to get the period start in seconds.
+- otherwise, the time is invalid.
+
+Note that offset seconds can overflow only if the period start is in a leap second; we ignore this!
```
thomash-acinq, line 311: If we ignore leap seconds, then there is no need for days as a time unit as seconds can be used instead.

12-offer-encoding.md
```diff
@@ -282,6 +364,41 @@ A writer of an offer:
       - MUST set `offer_quantity_max` to 0.
   - otherwise:
     - MUST NOT set `offer_quantity_max`.
+  - MAY include `offer_recurrence` to indicate offer should trigger time-spaced invoices.
```
thomash-acinq, line 367:
```suggestion
  - MAY include `offer_recurrence` to indicate offer should trigger time-spaced invoice requests.
```

12-offer-encoding.md
thomash-acinq: Do we ever want to force the payer to pay for all the previous periods? It seems to me that `start_any_period` should always be non-zero.

12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
     1. type: 22 (`offer_issuer_id`)
     2. data:
         * [`point`:`id`]
+    1. type: 24 (`offer_recurrence`)
```
thomash-acinq, line 232: I think we should use odd TLVs so that a payer that doesn't support recurrence can pay for one period.

12-offer-encoding.md
rustyrussell: True.  I could argue that 4 years is better UX than 48 months, but the same argument would apply for minutes and hours.  I'll drop it!

12-offer-encoding.md
rustyrussell: No, it would mean if you don't start at the beginning you can't join at all.

But perhaps we should rely on the expiration of the offer for this, instead?

12-offer-encoding.md
rustyrussell: If you're having monthly recurrences, you don't care about +/- a day,  BUT we have to define it somehow.  I think you'll find this easier to implement as written, so I think that's better.

12-offer-encoding.md
rustyrussell: We don't ignore, we simply ignore the case where period start *is on* a leap second.  We should probably round it down, for consistency in that case?

12-offer-encoding.md
rustyrussell: I wondered about that!  I think that may be worse UX, depending on the use case?  They will think it's a one-off payment ("hey that's cheap!") and then get upset when it's not.

12-offer-encoding.md
```diff
@@ -348,6 +473,8 @@ painful to have to special-case the "only one left" offer generation.
 
 Offers can be used to simply send money without expecting anything in return (tips, kudos, donations, etc), which means the description field is optional (the `offer_issuer` field is very useful for this case!); if you are charging for something specific, the description is vital for the user to know what it was they paid for.
 
+We insist that recurring requests be in order (thus, if you pay an invoice for #34 of a recurring offer, it implicitly commits to the successful payment of #0 through #33).  The `invreq_recurrence_paywindow` constrains how far you can pay in advance precisely, and if it isn't in the offer the defaults provide some slack, without allowing commitments into the far future.
```
a-mpch, line 476: I think `invreq_recurrence_paywindow` is a typo to `offer_recurrence_paywindow`.

if not is missing in the TLVs and I don't get how it would work 👀 . `invreq_recurrence_paywindow` would be set by the payer but it would be controlled by the issuer

12-offer-encoding.md
```diff
@@ -229,6 +229,88 @@ The human-readable prefix for offers is `lno`.
...
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months), 3 (years).
+2. A `period`, defining how often (in `time_unit`) it has to be paid.
+3. An optional `offer_recurrence_limit` of total payments to be paid.
+4. An optional `offer_recurrence_base`:
+   * `basetime`, defining when the first period starts in seconds since 1970-01-01 UTC.
+   * `start_any_period` if non-zero, meaning you don't have to start paying at the period indicated by `basetime`, but can use `offer_recurrence_start` to indicate what period you are starting at.
+5. An optional `offer_recurrence_paywindow`:
+   * `seconds_before`, defining how many seconds prior to the start of the period a payment will be accepted.
+   * `proportional_amount`, if set indicating that a payment made during the period itself will be charged proportionally to the remaining time in the period (e.g. 150 seconds into a 1500 second period gives a 10% discount).
```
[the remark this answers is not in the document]
rustyrussell: Indeed.  Let's give it a separate TLV entry.

12-offer-encoding.md
rustyrussell: Good catch.

12-offer-encoding.md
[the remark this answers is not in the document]
rustyrussell: Actually, it belongs in "offer_recurrence_base".  If that is not present, the period starts when you request the invoice, so it's always "the entire period".

12-offer-encoding.md
rustyrussell: Ok, now I'm using both.  That way the user can decide...

rustyrussell:
OK, updated.  I will need to rework the test vectors, however!

12-offer-encoding.md
a-mpch: @rustyrussell I think this one might slipped out from your changes

vincenzopalazzo (commented):
LGTM overall, I want to take another look later in the month, but probably in this PR we can also remove the FIXME N 7 in the https://github.com/lightning/bolts/blob/master/12-offer-encoding.md#fixme-possible-future-extensions

GBKS:
> One proposal I would like to add is a courtesy onion message when a user cancels a periodic payment. This is generally useful to tell the difference between a deliberate cancellation and a user-related failure.

Subscriptions likely require more communication. From the initial sharing of the offer, to price changes, to notifying a user about a missed payment, providing a receipt, etc. Maybe best to leave this type of communication up to the participants to handle in their medium of choice?

sbddesign:
> Subscriptions likely require more communication. From the initial sharing of the offer, to price changes, to notifying a user about a missed payment, providing a receipt, etc. Maybe best to leave this type of communication up to the participants to handle in their medium of choice?

@GBKS I agree that from a UX perspective there is always going to be more communication required between the subscriber and the recipient. However, if there is a protocol-native way to _create_ a subscription, I think there also needs to be a protocol-native way to _cancel_ the subscription.

GBKS:
> However, if there is a protocol-native way to create a subscription, I think there also needs to be a protocol-native way to cancel the subscription.

Maybe. You're creating a subscription proposal, but it's not active until the payee accepts it and acts on it. It's more like a subscription request, like we have payment requests. So if you want a "subscription cancelled" message, you may also want a "subscription accepted" message. But TBH, I don't have a super strong opinion. Just doesn't seem 100% right to exclusively have the cancel message.

rustyrussell:
Note: @TheBlueMatt pointed out that the correct behavior is to send the courtesy "cancel" message in place of the next request.  Slightly more complex to implement (your wallet now needs to remember cancelled subscriptions), but it's a nice balance between letting the vendor know, and not letting them cut you off early.  Apparently Apple Pay does something similar, and I can see the logic.

From an implementation POV, this may simply look like a stub invoice_request?

rustyrussell:
> > One proposal I would like to add is a courtesy onion message when a user cancels a periodic payment. This is generally useful to tell the difference between a deliberate cancellation and a user-related failure.
> 
> Subscriptions likely require more communication. From the initial sharing of the offer, to price changes, to notifying a user about a missed payment, providing a receipt, etc. Maybe best to leave this type of communication up to the participants to handle in their medium of choice?

That, however, has proven to be a centralized disaster, and severely limits interoperation.  You cannot automate this unspecified stuff...

Eventually we will probably want offer replacement (where we reject an invoice_request and optionally refers to an updated offer), but one step at a time!

12-offer-encoding.md
rustyrussell: "This time for sure!"

rustyrussell:
Added the "cancel" flag.  You basically send an invreq like normal, but it says "cancel" instead of "send me the next invoice".

Also added the requirement to fetch the next invoice explicitly to the requirements for recurring payments!

Ramin078 (commented):
rustyrussell:guilt/offers-recurrence

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+* Payments due once a year:
+   * which must be made on January 1st, or
+   * which are only valid if started January 1st 2021, or
+   * which if paid after January 1st you (over) pay the full rate first year, or
+   * which if paid after January 1st are paid pro-rata for the first year, or
+   * which repeat from whenever you made the first payment
+
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months).
+2. A `period`, defining how often (in `time_unit`) it has to be paid.
```
Ramin078, line 278: raminasadi009@gmail.com

Ramin078 (commented):
rustyrussell:guilt/offers-recurrence 69099

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+        * [`recurrence`:`recurrence`]
+    1. type: 26 (`offer_recurrence_base`)
+    2. data:
+        * [`recurrence_base`:`recurrence_base`]
+    1. type: 27 (`offer_recurrence_paywindow`)
+    2. data:
+        * [`recurrence_paywindow`:`recurrence_paywindow`]
+    1. type: 29 (`offer_recurrence_limit`)
+    2. data:
+        * [`tu32`:`max_period`]
```
shaavan: `max_period` indicates the maximum number of periods to trigger/pay the offer, but could the name be mistaken for “maximum time period”? Do you think that could cause confusion?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+
+1. subtype: `recurrence_base`
+2. data:
+    * [`byte`:`proportional_amount`]
+    * [`tu64`:`basetime`]
+
+1. subtype: `recurrence_paywindow`
+2. data:
+    * [`u32`:`seconds_before`]
+    * [`tu32`:`seconds_after`]
```
shaavan, line 261: Why is `seconds_before` in u32 (full-length) while `seconds_after` is in truncated u32?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+* Payments due on every Monday, 1pm Pacific Standard Time.
+* Payments due once a year:
+   * which must be made on January 1st, or
+   * which are only valid if started January 1st 2021, or
+   * which if paid after January 1st you (over) pay the full rate first year, or
+   * which if paid after January 1st are paid pro-rata for the first year, or
+   * which repeat from whenever you made the first payment
+
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months).
```
shaavan, line 277: Why does `time_unit` have to be a single byte? Would we be able to represent a recurrence with an unusual composite interval, e.g., every 1 month + 15 days (Jan 1 → Feb 16 → Apr 1, etc.)? This doesn’t seem possible without allowing composite time units.

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+2. A `period`, defining how often (in `time_unit`) it has to be paid.
+3. An optional `offer_recurrence_limit` of total payments to be paid.
+4. An optional `offer_recurrence_base`:
```
shaavan, line 280: If `offer_recurrence_base` is optional, why is it assigned to field 26?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+   * `basetime`, defining when the first period starts in seconds since 1970-01-01 UTC.
+   * The payer uses `invreq_recurrence_start` to indicate what period they are starting at.  If you don't want them to start at arbitrary periods, use `offer_absolute_expiry`.
+   * A `proportional_amount` flag: if set indicating that a payment made during the period itself will be charged proportionally to the remaining time in the period (e.g. 150 seconds into a 1500 second period gives a 10% discount).
+5. An optional `offer_recurrence_paywindow`:
+   * `seconds_before`, defining how many seconds prior to the start of the period a payment will be accepted.
+   * `seconds_after`, defining how many seconds after the start of the period a payment will be accepted.
```
shaavan, line 286: Should we explicitly state whether `seconds_before` and `seconds_after` are inclusive? Is the range `[seconds_before, seconds_after]` or `(seconds_before, seconds_after)`?

Also, what if the period and range are large enough that expressing them in seconds is inefficient? For example, an offer with a 3-year period and a pay_window from 1 year before to 1 year after the payment date.

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+  If this field is missing, payment will be accepted during the prior period and the paid-for period.
+
+Note that the `offer_absolute_expiry` field already covers the case where an offer is no longer valid after January 1st 2021.
```
shaavan: I’m not sure if I understand the context this line is referencing to. It would be awesome if you can help me understand the reference here. Thanks a lot! :)

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+Each period has a zero-based index, and a start time and an end time.  Because the periods can be in non-seconds units, the duration of a period can depend on when it starts.  The period with index N+1 begins immediately following the end of period with index N.
+
+- if an offer contains `offer_recurrence_base`:
+  - the start of period #0 is `basetime` seconds since 1970-01-01 UTC.
+- otherwise:
+  - the start of period #0 is the `invoice_created_at`.`timestamp` of the first `invoice` for this particular offer and `invreq_payer_id`.
+
+To calculate UTC time of the start of period #`N` for `N` > 0:
+- if the period #0 start is on a leap second:
+  - the *period base* is one second before period #0 start.
```
shaavan, line 304: I am curious as to why we chose “minus one second” over “plus one second”? Is it done so that, we maintain the month, date, hour, and minute of beginning of period, if the leap second was absent (For example, December 31st 23:59:60 to 23:59:59)?

12-offer-encoding.md
```diff
@@ -282,6 +373,40 @@ A writer of an offer:
...
+      - otherwise:
+        - MUST NOT include `offer_recurrence_base`.
+      - if payments will be accepted for any time in the current or next period:
+        - SHOULD NOT include `offer_recurrence_paywindow`
+      - otherwise:
+        - MUST include `offer_recurrence_paywindow`
+      - if it includes `offer_recurrence_paywindow`:
+        - MUST set `seconds_before` to the maximum number of seconds prior to a period for which it will accept payment or invoice_request for that period.
+        - MUST set `seconds_after` to the maximum number of seconds into to a period for which it will accept payment or invoice_request for that period.
+        - MAY NOT enforce this for the initial period for offers without `offer_recurrence_base`
```
shaavan, line 405: Could this even be enforced? If `offer_recurrence_base` is not set, the period start-time isn’t defined until the first invoice is created. Wouldn’t that make the pay_window meaningless in this case? So should this be “MUST NOT” instead of “MAY NOT”?

12-offer-encoding.md
```diff
@@ -282,6 +373,40 @@ A writer of an offer:
...
+        - SHOULD NOT set `seconds_after` to greater than the maximum number of seconds in a period.
```
shaavan, line 406: What if it was set to be longer than the period, would this allow the user to pay for a period’s amount in the next period as well?

12-offer-encoding.md
```diff
@@ -439,14 +589,25 @@ while still allowing signature validation.
...
-        * [`u8`:`domain_len`]
-        * [`domain_len*byte`:`domain`]
+        * [`bip_353_name`:`bip_353_name`]
+    1. type: 92 (`invreq_recurrence_counter`)
+    2. data:
+        * [`tu32`:`counter`]
+    1. type: 93 (`invreq_recurrence_start`)
+    2. data:
+        * [`tu32`:`period_offset`]
+    1. type: 94 (`invreq_recurrence_cancel`)
```
shaavan, line 602: Why is the data for type 94 (invreq_recurrence_cancel) not specified?

12-offer-encoding.md
```diff
@@ -457,6 +618,9 @@ The writer:
     - otherwise:
       - if it sets `invreq_chain` it MUST set it to bitcoin.
     - MUST set `signature`.`sig` as detailed in [Signature Calculation](#signature-calculation) using the `invreq_payer_id`.
+    - if it sets `invreq_recurrence_cancel`:
+      - SHOULD NOT send any future `invoice_request` for this offer with this `invreq_payer_id` without `invreq_recurrence_cancel` set.
+      - MAY omit `invreq_amount` and `invreq_quantity`.
```
shaavan, line 629: Since `invreq_recurrence_cancel` is meant to indicate immediate cancellation (line 751: “MUST NOT send an invoice in reply”), shouldn’t `invreq_amount` and `invreq_quantity` be “MUST NOT present” as they would be meaningless in this context?

12-offer-encoding.md
```diff
@@ -471,6 +635,34 @@ The writer:
...
+        - MUST use a unique `invreq_payer_id`.
+        - MUST set `invreq_recurrence_counter` `counter` to 0.
+        - MUST NOT set `invreq_recurrence_cancel`.
+      - for any successive requests:
+        - MUST use the same `invreq_payer_id` as the initial request.
+        - MUST set `invreq_recurrence_counter` `counter` to one greater than the highest-paid invoice.
+        - MAY set `invreq_recurrence_cancel`.
+      - if `offer_recurrence_base` is present:
+        - MUST include `invreq_recurrence_start`
+        - MUST set `period_offset` to the period the sender wants for the initial request
```
shaavan, line 656: Is the period offset from `offer_recurrence_base`? If so, could we clarify that? In general, would it make sense to have a dedicated section (like the recurrence section) defining the exact meaning of `invreq_recurrence_counter`, `invreq_recurrence_start`, and `invreq_recurrence_cancel`?

12-offer-encoding.md
```diff
@@ -740,6 +987,12 @@ A writer of an invoice:
     - for the bitcoin chain, it MUST set each `fallback_address` with
       `version` as a valid witness version and `address` as a valid witness
       program
+  - if `offer_recurrence_optional` or `offer_recurrence_compulsory` are present:
+    - MUST set `invoice_recurrence_basetime`.`basetime` to the start of period #0 as calculated by [Period Calculation](#offer-period-calculation).
```
shaavan, line 1002: Here, it’s not specified whether this check applies to the first invoice, successive invoices, or all invoices. However, lines 1055–1056 clearly say to reject successive invoices where `invoice_recurrence_base` mismatches that of previous period’s invoice. This seems like a reader/writer instruction mismatch.

shaavan (commented):
Looks great on the first pass! 🚀
I’ve got a few questions and points I’d like to discuss. If some (or all 😄) have already been covered in earlier conversations I may have missed, please excuse the repetition. And in case I’ve misunderstood anything, please feel free to correct me. Thanks!

12-offer-encoding.md
thomash-acinq: I don't think we should allow such composite interval, it gets too messy. Actually the example you gave shows why it's bad: the third date is not Apr 1 but Mar 31.

12-offer-encoding.md
thomash-acinq: I'm assuming you're asking because of the rule "it's OK to be odd".
This rule means that odd types can be safely ignored if you don't understand them, which is different from being optional. Here `offer_recurrence_base` is optional, but if it is set it's not OK to ignore it.

12-offer-encoding.md
thomash-acinq: I understand it as inclusive, however payments rarely settle in less than one second so it doesn't really matter.

I would guess that `seconds_after` would usually not be that high so it's OK to be "inefficient".

12-offer-encoding.md
thomash-acinq: Because we need to know how many bytes each of them uses. So only can use a truncated encoding, and it has to be the last one.

12-offer-encoding.md
thomash-acinq: Because it's empty. It's just a boolean flag without any data associated to it.

12-offer-encoding.md
rustyrussell: Agreed!  max_period_index is better, since we define "period index".

12-offer-encoding.md
rustyrussell: Agreed.  Also see: "Third Thursday of the Month" as things we don't support!

12-offer-encoding.md
rustyrussell: We do in fact define this in the actual requirements (Spoiler: it's inclusive!):

```markdown
      - if it includes `offer_recurrence_paywindow`:
        - MUST set `seconds_before` to the maximum number of seconds prior to a period for which it will accept payment or invoice_request for that period.
        - MUST set `seconds_after` to the maximum number of seconds into to a period for which it will accept payment or invoice_request for that period.
```

12-offer-encoding.md
rustyrussell: Updated:

Note: you can create an recurring offer with a fixed start which can only be used from the first period, using `offer_absolute_expiry`.  For example, my weekly book club starts January 1st 2026, for $5 a week: by setting `offer_absolute_expiry` to January 8th 2026, I ensure nobody can join after the first week.  Now, if only I could require them to actually read this week's book!

12-offer-encoding.md
rustyrussell: No, if you said "you must pay within 10 seconds of the start of a period" you could use the invoice expiry to do this, even for the first period.  You probably don't want to though!

You cannot enforce `seconds_before`, and should be tolerant with `seconds_after`.

12-offer-encoding.md
rustyrussell: I wanted to leave implementations free: they might produce a normal invoice and slap a cancel on it?

12-offer-encoding.md
rustyrussell: No, this is the writer side.  You always need to set it to this value:

```
- if an offer contains `offer_recurrence_base`:
  - the start of period #0 is `basetime` seconds since 1970-01-01 UTC.
- otherwise:
  - the start of period #0 is the `invoice_created_at`.`timestamp` of the first `invoice` for this particular offer and `invreq_payer_id`.
```

The reader side doesn't know the value (except for invoice 0) so it simply checks it's unchanged (which is usually simpler than going back to invoice 0).

12-offer-encoding.md
rustyrussell: Yes.  Also, it's arbitrary.

12-offer-encoding.md
rustyrussell: Yes.  And in fact this is possible, since periods can be in, say, months, and this is in seconds.  So 30 * 84600 is greater than February...

12-offer-encoding.md
rustyrussell: The offer_recurrence_base says "Period 0 always starts Feb 1st" for example.  Say it's monthly.

The user wants to start in August.  So, she sets `invreq_recurrence_start`'s `period_offset` to 6.  But her first invoice request, *as always* is for `invreq_recurrence_counter` 0.  That means it's trivial for the receiver to spot the first in a series of invoice requests....

I prefer to avoid nomenclature sections if we can make it clear in-situ.  Nobody (else!) ever reads them.

12-offer-encoding.md
shaavan: Ah, I didn’t know about this, and it looks like this is common with every field. Will look into it more. Thanks!

12-offer-encoding.md
shaavan: > I don't think we should allow such composite interval, it gets too messy. Actually the example you gave shows why it's bad: the third date is not Apr 1 but Mar 31.

Good catch! 😅 I think the point does come across.

12-offer-encoding.md
shaavan: Thanks, that clears it up! And it makes sense because `offer_recurrence_base` defines the period and proportionality behavior, which only really applies if the offer’s meant to be paid on a recurring basis, not just once. So yeah, assigning it to an even-numbered field fits that logic.

12-offer-encoding.md
shaavan: > We do in fact define this in the actual requirements (Spoiler: it's inclusive!):

Yep, Found it!

12-offer-encoding.md
shaavan: Looks good! (Also, wish there were a `book_reading_absolute_expiry` too 😄)

12-offer-encoding.md
shaavan: Okay, that makes sense! Keeping it flexible might reduce the transformation effort when adapting a regular invoice request into a recurrence cancellation.

12-offer-encoding.md
shaavan: Got it, thanks for the clarification! So it’s the writer’s responsibility to always set `invoice_recurrence_base` to the correct basetime at period #0, and the reader’s job to ensure it stays consistent by comparing it with the previous invoice.

12-offer-encoding.md
shaavan: > I would guess that seconds_after would usually not be that high so it's OK to be "inefficient".

I see, that makes sense. Though I imagine it’s still a valid use case in some scenarios.

Also, defining a pay window as something like ±3 days feels more natural than specifying it as ±259,200 seconds. @rustyrussell Do you think it might be more consistent (and readable) to express the pay window in terms of `(time_unit, window)`, similar to how `recurrence` is defined?

12-offer-encoding.md
shaavan: @rustyrussell

Okay, that clears it up! On a related note, do you think it would make sense to rename it to `period_index_offset`? Since it refers to the index rather than an actual time offset, it would better align with the `max_period_index` naming we landed on earlier, and could help avoid confusion.

shaavan (commented):
Thanks for all the clarifications, Rusty and Thomas!

12-offer-encoding.md
```diff
@@ -282,6 +373,40 @@ A writer of an offer:
       - MUST set `offer_quantity_max` to 0.
   - otherwise:
     - MUST NOT set `offer_quantity_max`.
+  - If an offer MAY trigger time-spaced invoice requests:
+    - MUST include exactly one of `offer_recurrence_optional` or `offer_recurrence_compulsory`.
+    - MAY use `offer_recurrence_optional` if a payment of a single period would be useful (compatibility with pre-recurrence readers).
+    - MUST NOT use `offer_recurrence_optional` if `offer_recurrence_base` is present.
+  - if it includes either `offer_recurrence_optional` or `offer_recurrence_compulsory`:
+      - MUST set `time_unit` to 0 (seconds), 1 (days), 2 (months) or 3 (years).
```
shaavan: `time_unit = 3 (years)` seems to have been removed from the spec, should we also remove it from the writer/reader instructions for consistency?

12-offer-encoding.md
```diff
@@ -282,6 +373,40 @@ A writer of an offer:
...
+      - MUST set `period` to how often (in `time-unit`) it wants to be paid.
```
shaavan, line 382: Should we add an explicit constraint for `period != 0` to avoid ambiguity? Something like:

Writer → MUST NOT set `period` to 0
Reader → MUST NOT respond to offers with `period` set to 0

12-offer-encoding.md
rustyrussell: Yes, good point.  Done.

12-offer-encoding.md
```diff
@@ -480,6 +641,35 @@ The writer:
         - MUST set `invreq_quantity` less than or equal to `offer_quantity_max`.
     - otherwise:
       - MUST NOT set `invreq_quantity`
+    - if `offer_recurrence_optional` or `offer_recurrence_compulsory` are present:
+      - for the initial request:
+        - MUST use a unique `invreq_payer_id`.
+        - MUST set `invreq_recurrence_counter` `counter` to 0.
+        - MUST NOT set `invreq_recurrence_cancel`.
+      - for any successive requests:
+        - MUST use the same `invreq_payer_id` as the initial request.
```
shaavan, line 650: Why does the payer need to maintain a stable `invreq_payer_id` across recurrence invoice requests? This appears to introduce public linkability for the payer across the lifetime of a recurrence session.

If the goal is to allow the payee to associate invoice requests with a particular recurrence session, could the recently added `invoice_recurrence_prev_state` and `invoice_recurrence_next_state` fields serve that purpose instead, avoiding the need for a stable payer identifier?

12-offer-encoding.md
shaavan, line 983: Recurrence sessions may be very long-lived, potentially lasting months or years.

When `offer_paths` is present, subsequent recurrence invoice requests are required to use the same blinded paths from the original offer. This appears to require those paths to remain usable for the entire lifetime of the recurrence session, potentially long after the original offer itself has expired.

If those paths become invalid over time, how is the payer expected to continue the recurrence session?

More generally, should long-lived recurrence sessions depend on communication paths established at offer creation time, or is some mechanism for path refresh expected?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+   * which are only valid if started January 1st 2021, or
+   * which if paid after January 1st you (over) pay the full rate first year, or
+   * which if paid after January 1st are paid pro-rata for the first year, or
+   * which repeat from whenever you made the first payment
+
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months).
+2. A `period`, defining how often (in `time_unit`) it has to be paid.
+3. An optional `offer_recurrence_limit` of total payments to be paid.
+4. An optional `offer_recurrence_base`:
```
shaavan, line 280: The current design allows a recurrence to be anchored either via `offer_recurrence_base` in the offer or via `invoice_recurrence_basetime` in the first invoice.

What advantages does allowing recurrence anchoring to originate from either the offer or the first invoice provide compared to requiring `period #0` to always be defined at offer creation time and making the offer the sole source of truth for recurrence anchoring?

### Additional Thoughts

I can see the value of payer-driven recurrence basetimes. However, I am curious whether that flexibility could be moved to the application layer while keeping recurrence anchoring entirely offer-defined.

For example:

* A merchant could support generating recurrence-on-demand offers, with the basetime chosen when the payer decides to begin the recurrence.
* For a long-lived static offer, a payer could begin participating after several recurrence periods have already elapsed, with the recurrence offset indicating the period at which participation began.

Would an approach like this preserve the same flexibility while keeping a single source of truth for recurrence anchoring?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+5. An optional `offer_recurrence_paywindow`:
+   * `seconds_before`, defining how many seconds prior to the start of the period a payment will be accepted.
+   * `seconds_after`, defining how many seconds after the start of the period a payment will be accepted.
+  If this field is missing, payment will be accepted during the prior period and the paid-for period.
+
+Note: you can create an recurring offer with a fixed start which can only be used from the first period, using `offer_absolute_expiry`.  For example, my weekly book club starts January 1st 2026, for $5 a week: by setting `offer_absolute_expiry` to January 8th 2026, I ensure nobody can join after the first week.  Now, if only I could require them to actually read this week's book!
+
+For wallets which don't (yet) support recurrence, the basic recurrence field comes in two variants: `offer_recurrence_compulsory` if they should not attempt a payment, and `offer_recurrence_optional` if it still makes sense for them to attempt a single payment.
```
shaavan, line 291: The current design distinguishes between recurrence being optional or compulsory at the protocol level.

What advantages are gained by encoding this distinction in the recurrence specification itself?

Could recurrence instead be treated as a protocol extension where recurrence is always optional at the spec level, with applications advertising and deciding whether one-time payments are accepted or rejected?

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+* Payments due on every Monday, 1pm Pacific Standard Time.
+* Payments due once a year:
+   * which must be made on January 1st, or
+   * which are only valid if started January 1st 2021, or
+   * which if paid after January 1st you (over) pay the full rate first year, or
+   * which if paid after January 1st are paid pro-rata for the first year, or
+   * which repeat from whenever you made the first payment
+
+Thus, each offer containing a recurring payment has:
+1. A `time_unit` defining 0 (seconds), 1 (days), 2 (months).
```
jharveyb, line 277: From previous discussion: the smallest 'reasonable' unit here is likely hours, not seconds (for example, if mobile wallets need to wake up).

martinsaposnic:
as a merchant/payee there's no way to clearly signal that a recurrence is terminated. only tool it has is rejecting the next invoice_request with a generic `invoice_error`. the payer's wallet can't tell "subscription terminated" from a transient failure, so it has to guess whether to retry next period or drop its session state

merchants discontinue products, change prices, and drop customers, so this will be hit constantly

vincenzopalazzo:
>merchants discontinue products, change prices, and drop customers, so this will be hit constantly

Wondering if we can just leverage the error message when rejecting the invoice request? Make a special one to say: Wait, baby, this is the end of our relationship!

12-offer-encoding.md
benthecarman: yes I think we explicitly want this to be larger, otherwise people might attempt to use this for streaming payments which seems incorrect / a misuse

12-offer-encoding.md
```diff
@@ -532,11 +723,45 @@ The reader:
         - if `offer_currency` is not the `invreq_chain` currency, convert to the
           `invreq_chain` currency.
         - if `invreq_quantity` is present, multiply by `invreq_quantity`.`quantity`.
+        - if `offer_recurrence_base` is present and `proportional_amount` is 1:
```
benthecarman, line 726: it's not defined for what to do if `offer_recurrence_base` is not present when `proportional_amount` is. I assume this is invalid?

12-offer-encoding.md
```diff
@@ -283,6 +373,41 @@ A writer of an offer:
...
+        - MUST NOT include `offer_recurrence_base`.
+      - if payments will be accepted for any time in the current or next period:
+        - SHOULD NOT include `offer_recurrence_paywindow`
+      - otherwise:
+        - MUST include `offer_recurrence_paywindow`
+      - if it includes `offer_recurrence_paywindow`:
+        - MUST set `seconds_before` to the maximum number of seconds prior to a period for which it will accept payment or invoice_request for that period.
+        - MUST set `seconds_after` to the maximum number of seconds into to a period for which it will accept payment or invoice_request for that period.
+        - MAY NOT enforce this for the initial period for offers without `offer_recurrence_base`
+        - SHOULD NOT set `seconds_after` to greater than the maximum number of seconds in a period.
```
jharveyb, line 406: Related to previous comments here: should `seconds_before` also be clamped to the size of a period? Such that a payer can only pay in advance for a maximum of one recurrence.

Without clamping, a payer could pre-pay for a full recurrence (or a significant portion of it) at once, which may be undesirable behavior from the perspective of the payee. From what I've seen of subscriptions in the wild, usually the price of a recurrence is tied to the period, and making multiple pre-payments is disallowed.

Concrete example:

A monthly subscription of 10 USD, for 1 year. `seconds_before` = 1 year, `seconds_after` = 1 month.

IIUC, with the spec as written the payer can pay for all 12 periods upfront, effectively 'locking in' the current exchange rate. I would *expect* the payee here to want to provide a different offer for a 12 month prepayment, likely with a lower price (100 USD for 1 year of service, or 100 USD per year, etc.)

tl;dr not clamping BOTH `seconds_before` and `seconds_after` seems like leaving a potential footgun around. Happy to hear counterarguments though.

12-offer-encoding.md
```diff
@@ -229,6 +229,97 @@ The human-readable prefix for offers is `lno`.
...
+
+1. subtype: `recurrence`
+2. data:
+    * [`byte`:`time_unit`]
+    * [`tu32`:`period`]
+
+1. subtype: `recurrence_base`
+2. data:
+    * [`byte`:`proportional_amount`]
+    * [`tu64`:`basetime`]
```
shaavan, line 256: I was wondering whether `basetime` and `proportional_amount` should be separate subtypes?

For example, a merchant may want to let the payer choose when to start a recurrence (i.e., by not setting `basetime`) while also allowing them to pay a proportional amount beginning with the first successive period.

Also, I think @jharveyb raised an interesting point (offline) about proportional amounts. IIUC, the current interpretation effectively acts as a discount. For example, if you pay 7 days late in a month-long period, you only pay for the remaining ~75% of the period.

In practice, though, merchants often don't terminate service immediately when a payment is late. For example, Spotify gives users a grace period before suspending service.

Jonathan's suggestion, if I understood it correctly, was that proportional amounts could instead be interpreted as a late-payment penalty. In other words, the payer would owe *more* if they paid after the due date, rather than less.

That might also help address another issue. Today, once the payment window is missed, the recurrence becomes void. If proportional amounts represented a penalty instead, it might allow payments even after the payment window, with the additional amount serving as the late-payment penalty.

I'd love to hear @jharveyb's thoughts on whether that's what he had in mind. Thanks!

jharveyb: Hey, yes that interpretation mostly matches what I meant in my prior example. I'll rephrase a bit, but generally:

- The effects of the `basetime` and `proportional_amount` fields are separate, so there's no need to pair them in the spec (as long as we don't introduce new, obvious footguns). IMO, when you start a recurrence should be totally separate from any discounts or penalties applied to some payment _during_ the recurrence (including the initial payment).

Concretely:

- Payees may want to allow a grace period for late payments, and/or allow early payments. This is already representable using `paywindow`.
- Payees may _also_ want to incentivize or punish early or late payments. E.x. "Your VOD subscription has payment due every month. You can pay early or late by 7 days, but the amount due increases proportional to how late your payment is."

So, a payment late by 7 days would be ~37/30*(original amount) -> 23% larger than an on-time payment.

Thinking about it now, I'm not sure if discounting early payment makes sense for every pay period, as a payer could end up spending a total amount significantly below the original recurrence value.

One option for 'discounts' is to have the payee application use the refund information provided by the payer, to send the payer a 'rebate' after their recurrence payment settles.  E.x. "Save up to $5 on your $50/month recurrence if you pay early! I, the payee, will send you this bonus amount after your payment deadline each month."

Would be happy to hear other input on how we could support discounts & per-payment penalties.

jharveyb:
Following up on offline conversation about how the spec would handle missed payments and payment ordering, e.x. "I forgot to pay rent the last 2 months, but I haven't been evicted yet".

AFAICT, the invoices used are made sequential by way of `invreq_reccurence_counter` and `invreq_recurrence_prev_state`. For handling missed paywindows, its up to the payee to serve invoices in the (correct) sequential order for a payer to catch-up on their recurrence:

https://github.com/lightning/bolts/pull/1240/changes#diff-cff1283072dc12a9b5d36cb2086aa18fbed57df0852ca1e2372c9e69329f7f9eR751

The language around paywindow uses SHOULD, so its already permissible to accept payments outside of the original paywindow. And it's explicitly mentioned that a payee can just serve a previously-generated invoice based on `invreq_recurrence_counter`:

https://github.com/lightning/bolts/pull/1240/changes#diff-cff1283072dc12a9b5d36cb2086aa18fbed57df0852ca1e2372c9e69329f7f9eL506

So I don't any language needs to change to support tracking missed payments, nor 'catch-up' payments. If a payee wants to strictly enforce paywindows by rejecting catch-up payments and cancelling the recurrence early, the `invoice_error` is flexible enough to communicate that.

PR #1242, opened 2025-04-01
Make payment_secret mandatory and ASSUMED

erickcestari:
## Background
The BOLT 11 payment secret (`s` field) was previously required for writers but not for readers (as implemented in #887), likely to maintain backward compatibility with unupgraded nodes during the initial rollout period. Now that several years have passed and implementations have widely adopted this requirement, this PR updates the spec to make the payment secret field mandatory for both writer and reader, and updates the `payment_secret` feature to ASSUMED status in BOLT 9.

## Changes

1. Added requirement for readers to fail payment if the `s` field is missing
2. Remove requirement to skip 's' fields with incorrect length
3. Added a test vector demonstrating an invalid invoice that's missing the `s` field
4. Changed the `payment_secret` feature to ASSUMED in BOLT 9, since all nodes are now required to support it

## Motivation
This change improves protocol privacy by ensuring that all Lightning invoices contain a payment secret. The payment secret prevents intermediate nodes in the payment path from probing for the destination by generating their own payment onions. Since implementations like LDK have already implemented this requirement (refusing to pay invoices missing a payment secret) and it's been ~4 years since the writer-side requirement was introduced, formalizing it in the specification makes sense. At this point, all implementations now always write payment secrets, and other implementations should follow LDK's example in refusing to process invoices without them.

Making the `payment_secret` feature ASSUMED in BOLT 9 reflects the reality that this feature has become standard across the Lightning Network and should be considered a core part of the protocol rather than an optional feature.

### DIFF
--- a/09-features.md
+++ b/09-features.md
@@ -38,7 +38,7 @@ The Context column decodes as follows:
 | 8/9   | `var_onion_optin`                 | ASSUMED                                                   |          |                             |                                                                       |
 | 10/11 | `gossip_queries_ex`               | Gossip queries can include additional information         | IN       |                             | [BOLT #7][bolt07-query]                                               |
 | 12/13 | `option_static_remotekey`         | ASSUMED                                                   |          |                             |                                                                       |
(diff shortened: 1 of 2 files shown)


Roasbeef (approved):
LGTM 🥬

Issue #1235, opened 2025-03-07
BOLT #11 Ambiguity: p Field Skipping vs. Mandatory Payment Hash


nGoline:
I’ve noticed a potential ambiguity in [BOLT 11](https://github.com/lightning/bolts/blob/master/11-payment-encoding.md) regarding the p field:
- The spec states that a writer MUST include exactly one p field.
- However, under the “Tagged Fields” → “Requirements” section, it also says a reader MUST skip:
“… unknown fields, OR an f field with unknown version, OR p, h, s or n fields that do not have data_lengths of 52, 52, 52 or 53, respectively.”

This can be misread to mean that if a p field doesn’t match the required length (e.g. zero length), the invoice parser should skip it rather than fail.

However, skipping a zero-length p field effectively means “no payment hash field is present,” which conflicts with the writer's requirement that p is mandatory. We know from the spec that the invoice must have a valid p field to be payable.

### Why is this confusing?

A literal reading of “MUST skip over p fields that do not have data_length of 52” could be interpreted as:

	“If data_length is not 52 for p, just skip it. Move on.”

But skipping a mandatory field is the same as omitting it. This leads to the question:

	Is an invoice valid if we skip the p field because its data_length is 0 or invalid?

### Why am I asking this now?

While running some [Differential Fuzz Tests](https://github.com/brunoerg/bitcoinfuzz/pull/87) between LND and NLightning, we found out that for the decoding of the following invoice, LND fails and NLightning succeeds:
`lnbc1pxtpcnsdz5cw9huz3z7wctpfpnptxcsdz5w9chzcw9huz3z7wc309t5cw9huz3z7wc3tpfpxpcnsdz5w9chzu3z7qqqqqqqqqqqqqqqqqqqqqqqquq44p9`

*A live demo for NLightning's Decoder can be found [here](https://bolt11.ipms.io/bolt11-decoder)*

### Potential Fix / Clarification

It might help to clarify in BOLT 11 that while you can skip unknown field types, you must fail if a mandatory known field (like p) doesn’t meet its specified data length. Perhaps something like:

	“A reader MUST fail the invoice if a mandatory field (p, h, s, n) is present with an incorrect length, rather than skipping it.”

This would remove any possible interpretation that a zero-length p field can be skipped and the invoice somehow remain valid.

nGoline:
I had some issues while testing against the provided vectors, specifically on the signatures of the invoices.
[...]

PR #1243, opened 2025-04-01
Clarify Mandatory Field Length Requirements and Add Note on Low R Signatures in BOLT 11

nGoline:
This PR addresses the ambiguity in BOLT 11 regarding the handling of mandatory fields (p, h, s) when their lengths are incorrect. Specifically, it clarifies that:

- A reader **MUST fail** the invoice if a mandatory field (p, h, s) is present with an incorrect length, rather than skipping it. This ensures that invoices with invalid mandatory fields are not misinterpreted as valid.

Additionally, this PR adds a note to the examples section regarding the use of Low R signatures. While Low R signatures save 1 byte in DER-encoded signatures, they are not enforced in the specification. This clarification is added to help implementers avoid confusion when testing against provided vectors, as the test vectors do not enforce Low R.

## Changes

1. Updated the "Requirements" section to explicitly state that invoices must fail if mandatory fields have incorrect lengths.
2. Added a note to the examples section explaining that Low R signatures are not enforced in the specification, even though they can save space.

## Context

This PR resolves the ambiguity raised in #1235.  The issue highlighted a potential misinterpretation where a p field with an invalid length could be skipped, leading to an invalid invoice being treated as valid. It also addresses confusion caused by test vectors not enforcing Low R signatures, which caused discrepancies during testing with libraries like NBitcoin.

## Testing

- Verified that the clarification aligns with the behavior of major implementations by running a [Differential Fuzzing](https://github.com/brunoerg/bitcoinfuzz?tab=readme-ov-file#lightning-modules).
- Updated documentation to ensure consistency with the clarified requirements.


11-payment-encoding.md
```diff
@@ -207,6 +207,10 @@ A writer:
 A reader:
   - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h`, `s` or
   `n` fields that do NOT have `data_length`s of 52, 52, 52 or 53, respectively.
+    - MUST fail the payment if a mandatory field (`p` or `s`) was skipped due to an incorrect
+    length.
+    - MUST fail the payment if a `d` field is not present and a `h` field was skipped due to an
+    incorrect length.
```
morehouse, line 213: Is there a good reason to have this skipping behavior at all?

Can we just require exactly zero or one such fields, always with the correct length?

nGoline: I agree it should be as straightforward as possible, but I didn't want to change the default expected behavior, so I added it to the existing spec.
It seems to me that most implementations are already enforcing what "it should have been" and not the spec line-by-line. Maybe we should open an issue to address a cleanup on the BOLT11 spec.

morehouse: I think it's worth cleaning this up from the spec, especially since some implementations already do the simpler thing.  

I know @erickcestari has run into this inconsistency already with differential fuzzing.  From that point of view, it would be nice to get all implementations doing the simple thing so ugly workarounds aren't needed for bitcoinfuzz.

nGoline: @t-bast Done!

11-payment-encoding.md
```diff
@@ -209,8 +209,8 @@ A writer:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
 
 A reader:
-  - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h` or
-  `n` fields that do NOT have `data_length`s of 52, 52 or 53, respectively.
+  - MUST fail the payment if any mandatory field (`p`, `h`, `s`, `n`) has an incorrect length (52, 52, 52, 53).
+  - MUST fail the payment if neither a `d` field or a `h` field are present.
```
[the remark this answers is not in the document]
nGoline: Done!

11-payment-encoding.md
```diff
@@ -209,8 +209,9 @@ A writer:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
 
 A reader:
-  - MUST skip over unknown fields, OR an `f` field with unknown `version`, OR  `p`, `h` or
-  `n` fields that do NOT have `data_length`s of 52, 52 or 53, respectively.
+  - MUST skip over `f` fields that use an unknown `version`.
+  - MUST fail the payment if any mandatory field (`p`, `h`, `s`, `n`) does not have the correct length (52, 52, 52, 53).
+  - MUST fail the payment if neither a `d` field nor a `h` field is present.
```
morehouse, line 214: On the writer side, the requirement is to have exactly one `d` or `h` field present, so if we make this change on the reader side we should also fail the payment if *both* `d` and `h` are present.

vincenzopalazzo (approved):
LGMT after the https://github.com/lightning/bolts/pull/1243/files#r2096059297 is addressed

11-payment-encoding.md
nGoline: I believe it's ok to keep the current phrase while adding that `if both are present`.

illuminated-shade (approved):

PR #1250, opened 2025-04-12
BOLT 4: fix broken link

elnosh:
fix broken `max_htlc_cltv` link


04-onion-routing.md
```diff
@@ -63,7 +63,7 @@ A node:
   * [Returning Errors](#returning-errors)
     * [Failure Messages](#failure-messages)
     * [Receiving Failure Codes](#receiving-failure-codes)
-  * [`max_htlc_cltv` Selection](#max-htlc-cltv-selection)
+  * [`max_htlc_cltv` Selection](#max_htlc_cltv-selection)
```
carlaKC, line 66: While we're here, the ordering is wrong - this section comes after onion messages so might as well move it to the right place.

Otherwise LGTM!

PR #1256, opened 2025-05-02
Gossip Infinite Loop Fix

ddustin:
Update the spec to prevent an infinite message loop.

As previously specified, both nodes must send `announcement_signatures` back and forth to each other forever.

Update it to specify “once per connection” to prevent this loop.


ddustin:
Sounds like we're moving towards putting info about a `channel_announcement` in the channel reestablish

PR #1261, opened 2025-05-27
Success hold times

joostjager:
Extending the idea proposed in #1044 to the success case.


02-peer-protocol.md
```diff
@@ -2350,6 +2350,12 @@ To supply the preimage:
    * [`channel_id`:`channel_id`]
    * [`u64`:`id`]
    * [`32*byte`:`payment_preimage`]
+1. `tlv_stream`: `uupdate_fulfill_htlc_tlvs`
```
thomash-acinq, line 2353: typo
```suggestion
1. `tlv_stream`: `update_fulfill_htlc_tlvs`
```

joostjager:
Would you use a single feature bit for both? In that case, I think it would delay deployment of attributable failures. It's more work to also extend the success path.

thomash-acinq:
Yes I would use the same feature bit (which should be renamed if it applies to fulfill too?). It's barely more work and attributable failures is already delayed anyways.

joostjager:
Maybe not barely more work in Eclair. You guys always have such compact ways to get things done. That's a compliment.

But wouldn't say this is true for all implementations? LND team seems to be focusing on the failure case initially only. The main problem to solve is attribution for failures, and hold times are an extra.

04-onion-routing.md
```diff
@@ -1560,6 +1560,12 @@ The _origin node_:
   - MAY use the data specified in the various failure types for debugging
   purposes.
 
+# Hold times for successful payments
+
+The `update_fulfill_htlc` message contains an optional `attribution_data` field, similar to the one used in the failure
+case described above. The only difference is that there is no failure message that is covered by the HMACs. With
+attribution data, the sender can obtain hold times as reported and committed to by each hop along the payment path.
```
thomash-acinq, line 1567: We should mention blinded routes, nodes inside a blinded must not set `attribution_data`.

joostjager:
I wonder what to call this feature that combines attributable failures, failure hold times and success hold times. `option_htlc_telemetry`?

04-onion-routing.md
joostjager: Added

joostjager:
Merged this PR into https://github.com/lightning/bolts/pull/1044

PR #1263, opened 2025-05-30
Add recommendations for receiver-side random delays

tnull:
After previous discussions on #1044, the last spec meeting, and [Delving](https://delvingbitcoin.org/t/latency-and-privacy-in-lightning/1723), it seems there was agreement from multiple sides that receiver-side delays were preferable to forwarding delays. Here, we propose to add such delays to the spec:

Timing attacks on privacy potentially allow on-path adversaries to determine the destination of an observed payment. To this end, they might merely need to record the time difference between sending `update_add_htlc` and receiving the corresponding `update_{fulfill,fail,fail_malformed}_htlc` response to be able to guess their distance to the destination and potentially deanonymize the latter.

To counteract this, different implementations previously implemented a variety of mitigation, most commonly adding some kind of (at times randomized) delays when *forwarding* HTLCs. However, here we propose to instead (or additionally, if the implementations prefer) add a receiver-side delay before claiming or failing HTLCs, which puts the
receiver's privacy improvements in their own hands. Additionally, adding a randomized delay on the receiver's end makes it easier to argue about how much delay should be added to somewhat reliably increase the anonymity set, as previously multiple intermediate hops (or none) could potentially add forwarding delays, which made picking a reasonable per-hop median delay rather hard.

In the second commit, we add similar recommendations to the BOLT12 flow, which might be a bit more controversial.

(cc @joostjager @t-bast @carlaKC @GeorgeTsagk)

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1446,7 +1446,7 @@ The _origin node_:
       - MUST NOT expose the `channel_update` to third-parties in any other
         context, including applying the `channel_update` to the local network
         graph, send the `channel_update` to peers as gossip, etc.
-    - SHOULD then retry routing and sending the payment.
+    - SHOULD then retry routing and sending the payment over a different path.
   - MAY use the data specified in the various failure types for debugging
   purposes.
 


12-offer-encoding.md
```diff
@@ -714,6 +715,7 @@ A writer of an invoice:
     `invreq_chain`.
   - if the invoice is in response to an `invoice_request`:
     - MUST copy all non-signature fields from the invoice request (including unknown fields).
+    - SHOULD add a random delay before sending the invoice.
```
carlaKC, line 718: This is a little less obvious to me because the `invoice_request` + `invoice` aren't necessarily sent over the same path? I'd lean towards making this `MAY` and adding rationale that explains why this is less straightforward.

07-routing-gossip.md
```diff
@@ -1017,6 +1017,11 @@ This effectively creates a _shadow route extension_ to the actual route and
 provides better protection against this attack vector than simply picking a
 random offset would.
 
+Similarly, senders that consider the historical payment latency over candidate
+hops as input for their routing algorithm should exclude the final hop from
+such scoring, to account for the receiver-induced delay (see [BOLT
+#2](02-peer-protocol.md#removing-an-htlc-update_fulfill_htlc-update_fail_htlc-and-update_fail_malformed_htlc)).
+
```
carlaKC, line 1024: Should we also add a recommendation to introduce a retry delay or make sure to use different paths when retrying a path? Perhaps adding to the [timing analysis](https://github.com/lightning/bolts/blob/master/04-onion-routing.md#rationale-5) note in bolt 04.

(maybe this exists, I just grepped for delay + retry and couldn't find it)

02-peer-protocol.md
```diff
@@ -2417,6 +2417,8 @@ A receiving node:
     - MUST return an error in the `update_fail_htlc` sent to the link which
       originally sent the HTLC, using the `failure_code` given and setting the
       data to `sha256_of_onion`.
+  - SHOULD add a random delay before sending `update_fulfill_htlc`,
+    `update_fulfill_htlc`, `update_fail_malformed_htlc`.
```
t-bast:
```suggestion
    `update_fail_htlc`, `update_fail_malformed_htlc`.
```

carlaKC (commented):
cACK

02-peer-protocol.md
tnull: Ah, heh, whoops.

12-offer-encoding.md
tnull: Well, but for BOLT 12 the adversary model would also include the sender trying to identify the receiver hidden behind a blinded path - and the sender would of course always be able to time the difference of sending the `invoice_request` and receiving the `invoice`.

a-mpch, line 718: I think we should add a rationale on this point as reading it from top (this BOLT) it is not easy to get why should be added.
Maybe pointing it out to BOLT 2 rationale or just a sentence.

a-mpch (commented):
cACK

02-peer-protocol.md
```diff
@@ -2439,6 +2441,14 @@ such detection is left as an option.
 Nodes inside a blinded route must use `invalid_onion_blinding` to avoid
 leaking information to senders trying to probe the blinded route.
 
+Receiving nodes should wait for a random amount of time before responding to
+incoming HTLC with `update_fulfill_htlc` / `update_fail_malformed_htlc` /
+`update_fail_htlc` to impair the capabilities of a potential adversary trying
+to deanonymize them based on message timing analysis. The delay distribution and
+parameters should be be chosen so that the response messages could have
+plausibly originated from a node downstream in 1-2 hops distance from the
+receiving node.
```
nothingmuch: The description of the recommended behavior is possibly ambiguous, and one possible interpretation is to fix a number of hops a priori and then to sample from a distribution of plausible delays for that number of hops, which would not achieve the intended goal.

I'm not sure how to phrase this more clearly but maybe the recommendation should be to sample a simulated number of hops, and then to sample a latency for each simulated hop?

t-bast already alluded to why this might be problematic, but to illustrate more clearly imagine the pathological case which, where all nodes on the network were to simulate just one additional hop. In this scenario latency increases but privacy does not improve.

tnull: > and one possible interpretation is to fix a number of hops a priori and then to sample from a distribution of plausible delays for that number of hops, which would not achieve the intended goal.

True, it would at least be sub-optimal.

> In this scenario latency increases but privacy does not improve.

I tend to partially disagree because it would still increase the anonymity set when compared to the raw network latency which is easier to measure/retrieve out-of-bounds. But yes, an adversary that would create its model based on probes might indeed not classify based on the raw latency but on an intra-network latency model, i.e., might end up just 'pricing in' that +1 latency. It's not as clear cut if we'd actually sample from a decent random distribution instead of uniform or even static distributions, but yes, we should clarify that the receiver should strive for maximizing uncertainty per millisecond artificial delay added.

Thanks, will see if I can tweak the formulation to make it more clear.

tnull: Now added a fixup to make the recommendations a bit more detailed.

nothingmuch: > I tend to partially disagree because it would still increase the anonymity set when compared to the raw network latency which is easier to measure/retrieve out-of-bounds.

To be clear I don't consider the scenario I was describing realistic at all, just used as an edge case for illustration purposes, so I have to agree with you disagreeing =)

> we should clarify that the receiver should strive for maximizing uncertainty per millisecond artificial delay added.

The updated version is much clearer to me and IMO this comment can be marked resolved

07-routing-gossip.md
tnull: I don't think adding sender-side delays is necessary, as generally for most nodes it makes a lot of sense to avoid previously-failed paths on retry. I think LND is the only exception to this as it still employs some carve-outs for nodes that respond with one of a specific set of failure codes. As we showed in the paper this behavior is exploitable and likely should be dropped, but I don't know if we need to add a recommendation/requirement for that?

12-offer-encoding.md
tnull: Now added a brief paragraph to the rationale section.

07-routing-gossip.md
carlaKC: > as generally for most nodes it makes a lot of sense to avoid previously-failed paths on retry

Perhaps we just add a recommendation to not retry the same path [here](https://github.com/lightning/bolts/blame/master/04-onion-routing.md#L1449)?

tnull: Now added a commit making that line a bit more specific. Let me know if you think it's sufficient.

PR #1264, opened 2025-05-30
Update "invalid currency UTF-8" test vector 

erickcestari:
This PR updates the "Malformed: invalid currency UTF-8" test vector to use 3 bytes of invalid UTF-8 instead of 2 bytes, since some applications like rust-lightning only validate encoding/size with proper length encoding (03 length with 3-byte value).

This commit replaces it with a test vector that:
- Uses length encoding (03 length with 3-byte of invalid UTF-8)

### DIFF
--- a/bolt12/offers-test.json
+++ b/bolt12/offers-test.json
@@ -491,7 +491,7 @@
   {
     "description": "Malformed: invalid currency UTF-8",
     "valid": false,
-    "bolt12": "lno1qcpgqsg2q4q5cj2rg5tzzqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqgpqyqszqg"
+    "bolt12": "lno1qcplllhapqpq86q2q4qkc6trv5tzzq6muh550qsfva9fdes0ruph7ctk2s8aqq06r4jxj3msc448wzwy9s"
   },
   {
     "description": "Malformed: truncated description UTF-8",


thomash-acinq:
The current test vector uses the following TLV stream:
```
06 02 8041  // Type: 6 (offer_currency), Length: 2, Value: two bytes that form an invalid utf8 string
0a 05 414c494345  // Type: 10 (offer_description), Length: 5, Value: "ALICE"
16 21 020202020202020202020202020202020202020202020202020202020202020202  // Type: 22 (offer_issuer_id), Length: 33, Value : some public key
```
It does what it claims to do (test that the `offer_currency` is validated as a utf8 string) and I don't see any need to change it.

Your proposed change is
```
06 03 4d2aa4  // Type: 6 (offer_currency), Length: 3, Value: three bytes that form an invalid utf8 string
08 02 2710  // Type: 8 (offer_amount)
0a 0c 5465737420766563746f7273 // Type: 10 (offer_description)
16 21 02eec7245d6b7d2ccb30380bfbe2a3648cd7a942653f5aa340edcea1f283686619  // Type: 22 (offer_issuer_id)
```
which does not seem to be an improvement.

erickcestari:
> The current test vector uses the following TLV stream:
> 
> ```
> 06 02 8041  // Type: 6 (offer_currency), Length: 2, Value: two bytes that form an invalid utf8 string
> 0a 05 414c494345  // Type: 10 (offer_description), Length: 5, Value: "ALICE"
> 16 21 020202020202020202020202020202020202020202020202020202020202020202  // Type: 22 (offer_issuer_id), Length: 33, Value : some public key
> ```
> 
> It does what it claims to do (test that the `offer_currency` is validated as a utf8 string) and I don't see any need to change it.
> 
> Your proposed change is
> 
> ```
> 06 03 4d2aa4  // Type: 6 (offer_currency), Length: 3, Value: three bytes that form an invalid utf8 string
> 08 02 2710  // Type: 8 (offer_amount)
> 0a 0c 5465737420766563746f7273 // Type: 10 (offer_description)
> 16 21 02eec7245d6b7d2ccb30380bfbe2a3648cd7a942653f5aa340edcea1f283686619  // Type: 22 (offer_issuer_id)
> ```
> 
> which does not seem to be an improvement.

Some applications (rust-lightning) validate currency codes by first checking for exactly 3 bytes before parsing the UTF-8 content. Since this test vector isn't 3 bytes long, those applications would reject it based on length rather than detecting the invalid UTF-8 encoding. Using a 3-byte invalid UTF-8 sequence would better test the UTF-8 validation logic specifically. I'll also remove the other fields to keep this test vector focused solely on `offer_currency`.

thomash-acinq:
OK, I understand the problem now. But if we really want to check that this field is a valid ISO 4217 code we should also add test vectors with symbols (US$), lowercase letters (usd) or even codes that look good but are not part of the standard (ABC).
Eclair is not affected as it doesn't support currency conversion so it will reject any offer that sets `offer_currency`.

thomash-acinq:
Eclair now supports the currency field and properly validates it.
The new test case is correct (as was the old one), however I think it should be replaced by something that looks a lot more like a real currency code, like "usd" or "ABC". There is only a small number of acceptable values and checking that it's valid UTF8 feels like checking that a number is nonnegative when the only valid values are 0 and 1.

thomash-acinq:
Yes it's inconvenient to have to keep the list up to date, however it's necessary to be able to understand the offer. The list is important not just to check that the currency is valid (we could just show the currency code to the end user who we can assume is already familiar with such codes) but to display the amount. If I create an offer that costs 50$ and you don't validate against the list, you will display 5000 USD instead of the correct 50.00 USD.

PR #1270, opened 2025-06-17
Increase channel close delay to 72 blocks

t-bast:
In #1044, we introduced a 12-blocks delay before considering a channel closed when we see a spending confirmation on-chain. This ensures that if the channel was spliced instead of closed, the channel participants are able to broadcast a new `channel_announcement` to the rest of the network. If this new `channel_announcement` is received by renote nodes before the 12-blocks delay, the channel can keep its history in path finding scoring/reputation, which is important.

We then realize that 12 blocks probably wasn't enough to allow this to happen: some implementations default to 8 confirmations before sending `splice_locked`, and allow node operators to increase this value. We thus bump this delay to 72 blocks to allow more time before channels are removed from the local graph.

We also add requirements in the 2nd commit to avoid rebroadcasting announcement for spent channels, because our peers may otherwise think we're trying to spam them.

### DIFF
--- a/07-routing-gossip.md
+++ b/07-routing-gossip.md
@@ -228,7 +228,7 @@ The receiving node:
         - SHOULD store this `channel_announcement`.
   - once its funding output has been spent OR reorganized out:
     - SHOULD forget a channel after a 72-block delay.
-    - MUST NOT rebroadcast this `channel_announcement` to its peers.
+    - SHOULD NOT rebroadcast this `channel_announcement` to its peers.
 
 ### Rationale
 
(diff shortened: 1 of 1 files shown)


Roasbeef (approved):
LGTM ⛩️

07-routing-gossip.md
```diff
@@ -228,6 +228,7 @@ The receiving node:
         - SHOULD store this `channel_announcement`.
   - once its funding output has been spent OR reorganized out:
     - SHOULD forget a channel after a 72-block delay.
+    - MUST NOT rebroadcast this `channel_announcement` to its peers.
```
Roasbeef: SHOULD NOT?

t-bast: Done in https://github.com/lightning/bolts/pull/1270/commits/5dc8db03bf61f7832069b4a6b0a84d30808bb9a0

PR #1071, opened 2023-04-28
draft: HTLC Endorsement to Mitigate Channel Jamming

carlaKC:
This PR introduces an `endorsed` TLV to `update_add_htlc` as a way for nodes to indicate whether they expect a HTLC to resolve "honestly". Nodes are advised to allocate a limited portion of their outbound liquidity and slots to HTLCs that are not endorsed by peers that they consider to have high reputation.

Opening early for discussion on structure, not ready for review - discussions around recommendations for local reputation scoring are ongoing.

For additional background see [here](https://gist.github.com/carlaKC/be820bb638624253f3ae7b39dbd0e343).

PR #1280, opened 2025-08-13
Outgoing reputation and HTLC Accountability

carlaKC:
This PR replaces #1071 with our latest proposal for a reputation scheme and resource management to mitigate slow jamming attacks.

Key differences for the new approach are:
* **Reputation is tracked for outgoing channels**, as only downstream peers are capable of slowing down resolution of htlcs.
* **Resources are limited on incoming channels**, as we're defending against a malicious downstream peer slowing resolution down.
* We now pass an `accountable` signal to our outgoing peer to indicate whether we'll **hold their reputation accountable** because scarce resources have been used (rather than an `endorsed` signal that indicates scarce resources _may_ be used).
* This signal **originates at the receiving node**, as they're ultimately in charge of resolution time, and is **propagated in the onion** to prevent tampering.
* General bucket resources have some DOS protections, and we introduce a congestion bucket to help facilitate honest payments during attack.


recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,516 @@
...
+  classifies as "good" behavior, expressed in seconds. The recommended default
+  is 90 seconds (given that the protocol allows for a 60 second MPP timeout).
+- `resolution_time`: the amount of time elapsed in seconds between a HTLC being
+  added to and removed from the local node's commitment transaction.
+- `fees`: the fees that are offered to the local node to forward the HTLC,
+  expressed in milli-satoshis.
+
+We define the `opportunity_cost` of the time a HTLC takes to resolve:
+
+`opportunity_cost = ceil ( (resolution_time - resolution_period) / resolution_period) * fees`
```
thomash-acinq: Why do you use `ceil` here? We want it to be at least 0 but we don't need it to be an integer.

```suggestion
`opportunity_cost = max(0,  (resolution_time - resolution_period) / resolution_period) * fees`
```

11-payment-encoding.md
```diff
@@ -203,6 +204,10 @@ A writer:
   - MUST pad field data to a multiple of 5 bits, using 0s.
   - if a writer offers more than one of any field type, it:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
+  - if the invoice will not be resolved within 90 seconds of HTLC(s) arrival:
+    - MUST NOT include a `a` field
+  - otherwise:
+    - MAY include a single `a` field
```
morehouse: Nit: it's clearer to invert the if-otherwise:

```suggestion
  - if the invoice will be resolved within 90 seconds of HTLC(s) arrival:
    - MAY include a single `a` field
  - otherwise:
    - MUST NOT include a `a` field
```

Also applies to the equivalent requirement for offers.

12-offer-encoding.md
```diff
@@ -745,6 +746,10 @@ A writer of an invoice:
     - MUST include `invoice_blindedpay` with exactly one `blinded_payinfo` for each `blinded_path` in `paths`, in order.
     - MUST set `features` in each `blinded_payinfo` to match `encrypted_data_tlv`.`allowed_features` (or empty, if no `allowed_features`).
     - SHOULD ignore any payment which does not use one of the paths.
+  - if the invoice will not be resolved within 90 seconds of HTLC(s) arrival:
+    - MUST NOT set `accountable`.
+  - otherwise:
+    - SHOULD set `accountable`.
```
morehouse: Why is it a SHOULD for offers, but a MAY for invoices?

recommendations/local-resource-conservation.md
carlaKC: For our purposes an integer was good enough because we just care about which "bucket" a htlc falls into.
It could also be a float, but I don't think it adds that much?

thomash-acinq: Then it means that we get a big jump from `0` to `fees` which can be exploited by an attacker:
- On a payment A -> B -> C, C can hold the HTLC  for `resolution_period - ε` before failing the HTLC
- B computes `opportunity_cost = 0` and C can keep its reputation
- A computes `opportunity_cost = fees` and slashes B's reputation.

carlaKC: Good point! 

I think it practice it this be tricky to pull this off, but no problem to change to allow a gradual grade.

carlaKC: Updated [here](https://github.com/lightning/bolts/compare/4c35777f44689e825bac0e44cae0150897e9abf4..1bab2ce8ce6403ee580b4b1631898e8a17df06fe).

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+  - otherwise: 
+    - `effective_fees` = 0
+
+##### Rationale
+
+Reputation relies on the fees charged by the local node rather than the fee
+offered by the sender to prevent over-payment of advertised fees from
+contributing to reputation. This is unlikely to impact honest senders who will
+abide by advertised fee policies, and complicates (but does not prevent)
+attempts to artificially inflate reputation through excessive fee payments.
```
morehouse, line 207: Why should we care about reputation inflation if it's paid for?  Fees paid are sunk costs to an attacker regardless of whether they happen in one payment or many.

Basing reputation on charged rather than offered fees could be good for a different reason though -- to avoid certain reputation destruction attacks.  Suppose payments like this:  `M1 -> A -> B -> M2`, where the attacker is trying to destroy `B`'s reputation with `A`.  The attacker could offer insanely high fees to `A` and minimal fees to `B`, then hold HTLCs until expiry.  As a result, `B`'s reputation would drop way more than `M2`'s.

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+The resolution of HTLCs with no `accountable` marker set does not have a negative
+impact on the outgoing channel's reputation, because they were not informed that
+they would be held liable. They are allowed to build reputation when they resolve
+successfully within the specified `resolution_period` to allow new nodes in the
+network to bootstrap reputation.
+
+#### Outgoing Channel Reputation
+
+Outgoing channel reputation measures the unforgeable history that the outgoing
+channel has built with the local node via payment of forwarding fees. As this
```
morehouse: Nit: outgoing channels don't pay fees to the local node

```suggestion
Outgoing channel reputation measures the unforgeable history that the outgoing
channel has built with the local node by successfully resolving payments. As this
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+- `general_bucket_slot_count`:
+  - If the channel type allows a maximum of 483 HTLCs: 20
+  - If the channel type allows a maximum of 120 HTLCs: 5
+- `general_bucket_liquidity_allocation` =
+  `general bucket capacity * general_bucket_slot_count / general bucket slot total`
+
+Each `(incoming scid, outgoing scid)` is deterministically assigned slots:
+```
+for i = 0; i < general_bucket_slot_count {
+  slot = sha256(concat(salt, incoming scid, outgoing scid, i) % general bucket slot total
```
morehouse:
```suggestion
  slot = sha256(concat(salt, incoming scid, outgoing scid, i)) % general bucket slot total
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+
+### General Bucket
+
+To protect the general bucket from trivial denial of service, each forwarding
+pair of channels is restricted to a subset of slots and liquidity.
+
+We define the following:
+- `general_bucket_slot_count`:
+  - If the channel type allows a maximum of 483 HTLCs: 20
+  - If the channel type allows a maximum of 120 HTLCs: 5
```
morehouse: Most implementations set `max_accepted_htlcs` to around 40 by default.  Based on the above recommendation, that would be 16 total general slots.

With `general_bucket_slot_count` at 20 or even 5 this simple DoS protection becomes much weaker -- all ~16 general slots can be occupied with relatively few channels.

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+```
+
+Where `salt`:
+- MUST be randomly chosen and unique per channel.
+- SHOULD be persisted across restarts to restore slot allocations.
+
+A HTLC is eligible to use the general bucket if for its
+`(incoming scid, outgoing scid)`'s assigned resources:
+- Currently occupied slots < `general_bucket_slot_count`
+- Currently occupied liquidity + `amt_msat` <= `general_bucket_liquidity_allocation`
```
morehouse, line 366: We should explicitly state that the general bucket can only be used when one of the assigned slots for that channel pair is unoccupied.

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+a cost, while still allowing reasonable usage by honest peers. Salting 
+resource assignment ensures that the attacker cannot detect which resources
+they will be assigned, and thus cannot strategically open new channels to
+manipulate assignment.
+
+The default slot allocations provided are chosen such that it it highly
+improbable that any two channels are granted the same set of resources, making
+it difficult for an attacker to crowd out honest traffic. With these defaults,
+an attacker will need to open approximately 50 channels in expectation to gain
+access to all general resources.
```
morehouse: As noted above, the number of channels required is a lot less with typical values for `max_accepted_htlcs`.  If we have 16 general slots, we'd expect them to be fully occupied after ~11 or ~3 opened channels for `general_bucket_slot_count`s of 5 and 20, respectively (coupon collector expectation).

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+  - Otherwise:
+    - If the outgoing peer has sufficient reputation:
+      - SHOULD forward the HTLC, assigning usage to the protected bucket.
+    - Otherwise:
+      - If the HTLC is eligible to use congestion resources:
+        - SHOULD forward the HTLC, assigning usage to the congestion bucket.
+      - Otherwise:
+        - SHOULD fail the HTLC with `temporary_channel_failure`.
+- MUST set `accountable` in the offered `update_add_htlc` as outlined in
+  [HTLC Accountability](#htlc-accountability).
```
morehouse: Nit:  this requirement seems separate from the preceding resource assignment instructions.  Maybe we should move it to a concluding paragraph to this section.

```suggestion

When forwarding HTLCs, nodes MUST set `accountable` in the offered `update_add_htlc` as described in
  [HTLC Accountability](#htlc-accountability).
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,529 @@
...
+average to minimize the amount of data that needs to be stored per-channel. In
+flight HTLCs can be accounted for separately to this calculation, as the node
+will already have data for these HTLCs available.
+
+Track the following values for each rolling window:
+- `last_update`: stores the timestamp of the last update to the decaying
+    average, expressed in seconds.
+- `decaying_average`: stores the value of the decaying average.
+- `decay_rate`: a constant rate of decay based on the rolling window chosen,
+  calculated as: `((1/2)^(2/window_length_seconds))`.
```
morehouse:
```suggestion
  calculated as: `(1/2)^(2/window_length_seconds)`.
```

morehouse: Should we provide a recommendation for `window_length_seconds`?

It looks like this will decay by 75% every window, so I'm guessing the same size as the fixed window would be reasonable here...

morehouse (commented):
Concept ACK.

This is a big improvement over https://github.com/lightning/bolts/pull/1071:

- The inverted reputation scheme protects against [sink attacks](https://delvingbitcoin.org/t/hybrid-jamming-mitigation-results-and-updates/1147#p-3212-manipulation-sink-attack-9).
- The inverted reputation scheme enables new nodes to immediately send payments to reputable destinations when the network is under attack (better UX).
- The hash-based general slot assignment provides some discouragement for general jamming attacks.
- The tit-for-tat congestion bucket provides a mechanism to build reputation even when being general-jammed.

We probably also want to tweak the general slot assignment recommendations a bit to improve the strength of the hash-based defense:

- Scale `general_bucket_slot_count` based on `max_accepted_htlcs` instead of the commitment type.
- Increase the recommended percentage of total slots allocated to the general bucket.

There's also still an open question about how to address the issues with [high payment fanout](https://delvingbitcoin.org/t/hybrid-jamming-mitigation-results-and-updates/1147/4).  For example, LSPs often have high payment fan-out, which would prevent their clients from *ever* getting a good reputation with the current algorithm.

recommendations/local-resource-conservation.md
carlaKC: > Why should we care about reputation inflation if it's paid for?

We generally don't want an attacker to be able to take unusual actions that give them an advantage over honest behaving nodes. This just forces an attacker to align more with the behavior of honest nodes, even if it's trivial for them to do (multiple payments).

> to avoid certain reputation destruction attacks

Good point! I'll include this in rationale.

recommendations/local-resource-conservation.md
carlaKC: This is covered by the `general_bucket_slot_count` check above?
It's badly named, I"ll change it to `general_bucket_slot_allocation` so that it's clearer.

recommendations/local-resource-conservation.md
carlaKC: `window_length_seconds` is a placeholder for the period that you want a rolling average over - for revenue you'd use `revenue_window` and for reputation you'd use `revenue_window * reputation_multiplier`, for example.

I'll update the wording - let me know if it clarifies it!

recommendations/local-resource-conservation.md
carlaKC: [LND](https://github.com/lightningnetwork/lnd/blob/f293566849f2fe940aeb03f3b05514590bd54eba/config.go#L184) and [LDK](https://github.com/lightningdevkit/rust-lightning/blob/9514637f8732c66450ec3a5cc2c64c8708d36d67/lightning/src/ln/chan_utils.rs#L69) both have defaults on 483, so most of the network is _probably_ running with this default.

Assuming that these values are set to protect nodes from the on-chain cost of resolution, I'm hopeful that other impls will be able to increase the number of slots they allow now that they can't be so trivially filled up.

11-payment-encoding.md
```diff
@@ -203,6 +204,10 @@ A writer:
   - MUST pad field data to a multiple of 5 bits, using 0s.
   - if a writer offers more than one of any field type, it:
     - MUST specify the most-preferred field first, followed by less-preferred fields, in order.
+  - if the invoice will be resolved within 90 seconds of HTLC(s) arrival:
+    - SHOULD include a `a` field
+  - otherwise:
+    - MUST NOT include a single `a` field
```
morehouse: Nit
```suggestion
    - SHOULD include a single `a` field
  - otherwise:
    - MUST NOT include an `a` field
```

recommendations/local-resource-conservation.md
morehouse: > [LND](https://github.com/lightningnetwork/lnd/blob/f293566849f2fe940aeb03f3b05514590bd54eba/config.go#L184) and [LDK](https://github.com/lightningdevkit/rust-lightning/blob/9514637f8732c66450ec3a5cc2c64c8708d36d67/lightning/src/ln/chan_utils.rs#L69) both have defaults on 483, so most of the network is probably running with this default.

That's true for LND, though it looks like LDK [defaults](https://github.com/lightningdevkit/rust-lightning/blob/81495c6040647573a750eefcafb686edb770ab0d/lightning/src/util/config.rs#L254) to 50.

> Assuming that these values are set to protect nodes from the on-chain cost of resolution, I'm hopeful that other impls will be able to increase the number of slots they allow now that they can't be so trivially filled up.

Good point.  Using the suggested defaults with zero-fee commits would put the general bucket size at `120 * .4 = 48`, which is  essentially the same as current `max_accepted_htlcs` defaults.  Usage above that amount would be possible but would require reputation.

02-peer-protocol.md
```diff
@@ -2233,6 +2233,7 @@ is destined, is described in [BOLT #4](04-onion-routing.md).
     1. type: 0 (`blinded_path`)
     2. data:
         * [`point`:`path_key`]
+    1. type: 1 (`accountable`)
```
thomash-acinq, line 2236: Why not allow multiple accountability level instead of it being just a boolean flag? Just as with the endorsement before.

carlaKC: The `accountable` signal being set should be interpreted as: "I will hold your reputation responsible for the timely resolution of this HTLC", which is a boolean statement.

By contrast, endorsement signals meant "I think that this HTLC will resolve in a timely manner", which makes more sense to interpret as a range.

11-payment-encoding.md
carlaKC: [fixed!](https://github.com/lightning/bolts/compare/c25f7b11406c6485c9980ec9177ec62c2c35c19c..84ab329edbfbe687f557eb8569f11bad1bf075e7)

12-offer-encoding.md
```diff
@@ -644,6 +644,7 @@ the `onion_message` `invoice` field.
         * [`name_len*byte`:`name`]
         * [`u8`:`domain_len`]
         * [`domain_len*byte`:`domain`]
+    1. type: 93 (`accountable`)
```
t-bast: The range 80-159 is for `invoice_request`, you cannot use `93` for an invoice-only field (see https://github.com/lightning/bolts/blob/master/12-offer-encoding.md#invoice-requests). This must be in the range `160-240` (unless you want this field to be set in the `invoice_request`, in which case its name should probably be `invreq_accountable`?).

carlaKC: > This must be in the range 160-240

Right - only needs to be in the invoice! Addressed in https://github.com/lightning/bolts/pull/1280/commits/dd99be33aaaeb91c343321e45ddf1d62a1656b76

04-onion-routing.md
```diff
@@ -339,6 +347,10 @@ The reader:
       - incoming `amount_msat` < `amt_to_forward`.
```
thomash-acinq: A few lines above (old line 322: "MUST return an error if the payload contains other tlv fields than"): we need to add `upgrade_accountability` to the list of TLVs allowed.

thomash-acinq (commented):
Passing `upgrade_accountability` in the payload is not compatible with blinded routes, I've fixed it in https://github.com/carlaKC/lightning-rfc/pull/6

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,533 @@
...
+
+We define the `outgoing_channel_reputation`:
+- for each `update_add_htlc` offered on the outgoing channel over
+  the rolling window [`now` - `revenue_window` *  `reputation_multiplier`; `now`]:
+  - if the HTLC has been resolved:
+    - `outgoing_channel_reputation` += `effective_fee(HTLC)`
+
+##### Rationale
+
+Fees are used to accumulate reputation because they are an unforegable,
```
elnosh: *unforegable/unforgeable

carlaKC:
> Letting the recipient add upgrade_accountability in the blinded route has another advantage: it allows the HTLC to become accountable inside the blinded route even if the sender does not support accountability.

This seems like the right approach to me, since it's a recipient chosen value. Thanks for the PR, I'll upstream it!

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+  are quickly failed, sent at a rate that continuously consumes all resources
+  along the targeted route.
+- Slow Jamming: sending a slow stream of payments through a route and only
+  failing them at the latest possible timeout, blocking all of the liquidity
+  or HTLC slots along the route for the duration.
+
+This document outlines recommendations for implementing local resource
+conservation to mitigate slow jamming attacks. It is part of a hybrid solution
+to mitigating channel jamming, intended to be followed by the introduction of
+unconditional fees to mitigate fast jamming.
```
erickcestari: nit++: Is quick or fast jamming?

```suggestion
unconditional fees to mitigate quick jamming.
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+on-chain and the opportunity cost of capital committed to the attack.
+
+A channel can be fully jammed by depleting its available HTLC slots by holding
+`max_accepted_htlcs` or its available liquidity by holding the
+`max_htlc_value_in_flight`. An economically rational attacker will pursue the
+attack method with the lowest cost and highest utility. An attack can exhaust
+the resources along the target route in two ways, though the difference
+between the two behaviors is not rigorously defined:
+- Quick Jamming: sending a continuous stream of payments through a route that
+  are quickly failed, sent at a rate that continuously consumes all resources
```
erickcestari: What exactly are resources in this case?

If the HTLC is being held for less than 90 seconds, would this be considered a quick jamming attack?

11-payment-encoding.md
```diff
@@ -161,6 +158,7 @@ Currently defined tagged fields are:
 * `9` (5): `data_length` variable. One or more 5-bit values containing features
   supported or required for receiving this payment.
   See [Feature Bits](#feature-bits).
+* `a` (31): `data_length` 0. A marker field that indicates whether the recipient is willing to be held accountable for the timely resolution of the invoice upon HTLC(s) receipt.
```
erickcestari: According to BIP-0173 (bech32), the number `31` is the letter `l`. To have the letter `a`, we should sign number `29`.

https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32

```suggestion
* `a` (29): `data_length` 0. A marker field that indicates whether the recipient is willing to be held accountable for the timely resolution of the invoice upon HTLC(s) receipt. 
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+
+The `accountable` signal indicates that the receiving node's reputation will
+be held accountable for the fast resolution of the HTLC. This may only be set
+if the final recipient indicates that they are `accountable` in their invoice,
+as resolution is ultimately dependent on their hold time. 
+
+The original sender:
+- SHOULD NOT set `accountable` in their offered `update_add_htlc`.
+- MAY set `accountable` to mimic patterns observed in their forwarded traffic
+  if `upgrade_accountability` is set in the recipient's invoice.
```
erickcestari: nit: The invoice has the `accountable` field.

```suggestion
  if the recipient's invoice set `accountable`.
```

02-peer-protocol.md
```diff
@@ -2275,6 +2272,18 @@ A sending node:
   - MUST increase the value of `id` by 1 for each successive offer.
   - if it is relaying a payment inside a blinded route:
     - MUST set `path_key` (see [Route Blinding](04-onion-routing.md#route-blinding))
+  - if it is the original source of the HTLC:
+    - SHOULD omit `accountable`.
+    - MAY set `accountable` to mimic patterns of HTLCs it has forwarded.
```
erickcestari: A payer cannot set `accountable` to the HTLC if the invoice does not have the `accountable` field set because the receiver would reject the payment.

>  - If it is the final node:
    - MUST return an error if:
      ...
      - `accountable` is set in the incoming `update_add_htlc` but `accountable` was not set in the invoice.

```suggestion
    - if the recipient's invoice set `accountable`:
      - MAY set `accountable` to mimic patterns of HTLCs it has forwarded.
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
+# Local Resource Conservation
+
+## Table of Contents
+
+- [Introduction](#introduction)
+- [Overview](#overview)
```
erickcestari, line 6: nit:

```suggestion
- [Overview](#overview)
- [HTLC Accountability](#htlc-accountability)
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+- MUST NOT allocate more liquidity than their `max_htlc_value_in_flight_msat`
+  across buckets.
+
+Recommended Defaults:
+- General: 40% of `max_accepted_htlcs` and `max_htlc_value_in_flight_msat`.
+- Congestion: 20% of `max_accepted_htlcs` and `max_htlc_value_in_flight_msat`.
+- Protected: 40% of `max_accepted_htlcs` and `max_htlc_value_in_flight_msat`.
+
+Note that if a channel has a sufficiently small `max_accepted_htlcs` or
+`max_htlc_value_in_flight_msat`, these buckets cannot provided meaningful
```
erickcestari: nit:
```suggestion
`max_htlc_value_in_flight_msat`, these buckets cannot provide meaningful
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+The protected bucket is reserved for peers with sufficient outgoing reputation.
+There are no restrictions on usage so that channels that have built sufficient
+reputation can operate as usual during periods of attack. If this bucket is
+full, HTLCs that were eligible to use it may utilize any other available 
+resources in general or congestion buckets (using them partially if need be).
+
+#### Rationale
+
+The protected bucket aims to project unrestricted resource access to peers that
+have build sufficient reputation. In the unlikely event that the protected
```
erickcestari: nit:
```suggestion
have built sufficient reputation. In the unlikely event that the protected
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+        - SHOULD fail the HTLC with `temporary_channel_failure`.
+
+When forwarding HTLCs, nodes MUST set `accountable` in the offered
+`update_add_htlc` as outlined in [HTLC Accountability](#htlc-accountability).
+
+### Rationale
+
+In times of peace when general resources are not saturated, there is no need
+for nodes to hold any HTLCs `accountable`. When resources become scarce, it is
+only safe to allow nodes that have build reputation to access protected
```
erickcestari: nit:
```suggestion
only safe to allow nodes that have built reputation to access protected
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+
+Local resource conservation combines the following elements to provide nodes
+with a mechanism to protect themselves against slow-jamming attacks:
+
+- [Local Reputation](#local-reputation): local tracking of the historical
+  forwarding behavior of downstream peers, used to make forwarding decisions.
+  Behavior of downstream peers is observed because a malicious downstream peer
+  is necessary to perform a slow-jamming attack (the upstream may or may not be
+  malicious).
+- [HTLC accountability](#../02-peer-protocol.md#adding-an-htlc-update_add_htlc):
```
erickcestari, line 62: nit:
```suggestion
- [HTLC accountability](../02-peer-protocol.md#adding-an-htlc-update_add_htlc):
```

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,551 @@
...
+as resolution is ultimately dependent on their hold time. 
+
+The original sender:
+- SHOULD NOT set `accountable` in their offered `update_add_htlc`.
+- MAY set `accountable` to mimic patterns observed in their forwarded traffic
+  if `upgrade_accountability` is set in the recipient's invoice.
+
+The forwarding node:
+- If `accountable` is set in the received `update_add_htlc`:
+  - MUST set `accountable` in the offered `update_add_htlc` (see [BOLT-02](#../02-peer-protocol.md#adding-an-htlc-update_add_htlc)).
```
erickcestari, line 92: nit:
```suggestion
  - MUST set `accountable` in the offered `update_add_htlc` (see [BOLT-02](../02-peer-protocol.md#adding-an-htlc-update_add_htlc)).
```

erickcestari (commented):
Did a pass over the recommendation doc and the BOLT 02/04/11/12 changes, design reads well and the permission (BOLT-02) vs. policy (recommendation) split is coherent.

One real issue: the BOLT-11 `a` field is numbered `31`, but `31` is `l` in `bech32`; `a` is `29`. Worth fixing before merge.

Everything else is nits: typos, broken cross-file links (leading #), missing TOC entry, and tightening the sender-mimicry wording.

recommendations/local-resource-conservation.md
carlaKC: > What exactly are resources in this case?

The resources are slots + liquidity. Can add clarification to the first sentence of this paragraph if you think it needs it?

> If the HTLC is being held for less than 90 seconds, would this be considered a quick jamming attack?

Yes, but as handwaved in "not rigorously defined" we don't have a strict classification line between the two.

carlaKC:
Addressed @erickcestari's comments - [diff here](https://github.com/lightning/bolts/compare/4acb09ccfb2eb0967fe6b89defba8645a732cf9d..c2771fb7cfb9a3dcc155e55f8062f85b17c627c7) + [here](https://github.com/lightning/bolts/compare/2e09bcdfdf0a1670c9df041bd703fd6f9a6e4ecd..b881c0b520ddd89eed08102dd0b0faecb3b52dc0) Squashed fixups because they were getting a bit unwieldy.

11-payment-encoding.md
carlaKC: Nice catch!

recommendations/local-resource-conservation.md
erickcestari: > The resources are slots + liquidity. Can add clarification to the first sentence of this paragraph if you think it needs it?

I think it would be clearer, since the Slow Jamming definition explicitly names the consumed resources.

carlaKC: How's [this](https://github.com/lightning/bolts/compare/2e09bcdfdf0a1670c9df041bd703fd6f9a6e4ecd..b881c0b520ddd89eed08102dd0b0faecb3b52dc0)?

erickcestari: It looks great! Thanks!

GeorgeTsagk:
Woops apparently a commit that mentions this PR notifies this page, sorry for the spam, removing the URL from the above commit.

recommendations/local-resource-conservation.md
```diff
@@ -0,0 +1,552 @@
...
+flight HTLCs can be accounted for separately to this calculation, as the node
+will already have data for these HTLCs available.
+
+To track a rolling window over the period `window_length_seconds` as a decaying
+average, track the following values:
+- `last_update`: stores the timestamp of the last update to the decaying
+    average, expressed in seconds.
+- `decaying_average`: stores the value of the decaying average.
+- `decay_rate`: a constant rate of decay based on the rolling window chosen,
+  calculated as: `(1/2)^(1/ln(2)*window_length_seconds)`
```
erickcestari, line 502: I think the `decay_rate` formula in the spec is missing a parenthesis: `(1/2)^(1/ln(2)*window_length_seconds)` as written parses as `0.5^(W/ln 2) = e^(-W)`, which is essentially zero for any realistic window. I guess the intended formula was `(1/2)^(1/(ln(2)*window_length_seconds))`, which simplifies to `e^(-1/W)`.

One thing I noticed is that LND and LDK are currently using `0.5^(2/W)` (half-life = W/2), which decays a bit faster than `e^(-1/W)`. With the 24-week window, `0.5^(2/W)` leaves 25% weight at the end vs 37% for `e^(-1/W)`.

So which one should be the recommended `decay_rate` here? Should the spec just get the parenthesis fix and keep `e^(-1/W)`, or would it make sense to align it with what the implementations are already doing, i.e. `(1/2)^(2/window_length_seconds)`?

```suggestion
- `decay_rate`: a constant rate of decay based on the rolling window chosen,
  calculated as: `(1/2)^(1/(ln(2)*window_length_seconds))`
```

PR #1284, opened 2025-08-21
bolt11: clarify signature normalization requirements

erickcestari:
Explicitly document that signature validation requirements differ based on the presence of the `n` field:

- With `n` field: signature MUST be normalized lower-S form
- Without `n` field: signature recovery accepts both high-S and low-S

This clarifies existing implementation behavior where `secp256k1_ecdsa_verify` (used with `n` field) requires normalized signatures, while `secp256k1_ecdsa_recover` (used without `n` field) accepts both normalized
and non-normalized signatures.

cc @real-or-random

### DIFF
--- a/11-payment-encoding.md
+++ b/11-payment-encoding.md
@@ -221,6 +221,10 @@ A reader:
   description.
   - if a valid `n` field is provided:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
+    - the signature MUST be normalized lower-S form.
+  - otherwise:
+    - MUST perform signature recovery.
+    - signature recovery accepts both high-S and low-S signatures.
   - if a valid `s` field is not provided:
     - MUST fail the payment.
   - otherwise:


11-payment-encoding.md
```diff
@@ -221,6 +221,10 @@ A reader:
   description.
   - if a valid `n` field is provided:
     - MUST use the `n` field to validate the signature instead of performing signature recovery.
+    - the signature MUST be normalized lower-S form.
```
Roasbeef: Makes sense. 

Before merge, we should also go through the test vectors to see if any of them are high-s, regenerating if so.

11-payment-encoding.md
```diff
@@ -220,7 +220,10 @@ A reader:
   - MUST check that the SHA2 256-bit hash in the `h` field exactly matches the hashed
   description.
   - if a valid `n` field is provided:
-    - MUST use the `n` field to validate the signature instead of performing signature recovery.
+    - MUST use the `n` field to validate the signature instead of performing public-key recovery.
+    - the signature MUST be normalized low-S.
```
t-bast: nit: it's slightly better to explicitly say what to *do* rather than how things should *be*:

```suggestion
    - If the signature is not compliant with the low-S standard rule<sup>[low-S](https://github.com/bitcoin/bitcoin/pull/6769)</sup>:
      - MUST fail the payment.
```

Not sure if "low-S-normalized" is a standard term?

erickcestari: Updated! It's mentioned in the secp256k1 as normalized lower-S form.

https://github.com/bitcoin-core/secp256k1/blob/ce7923874f3f51dd87f396e0be655e2b25ad4629/include/secp256k1.h#L585-L630

11-payment-encoding.md
erickcestari: There are no BOLT-11 test vectors that cover signature validation using the `n` field. We should probably add some.

PR #1160, opened 2024-05-02
Channel Splicing (feature 62/63)

t-bast:
Splicing allows spending the current funding transaction to replace it with a new one that changes the capacity of the channel, allowing both peers to add or remove funds to/from their channel balance.

Splicing takes place while a channel is quiescent, to ensure that both peers have the same view of the current commitments.

We don't want channels to be unusable while waiting for transactions to confirm, so channel operation returns to normal once the splice transaction has been signed and we're waiting for it to confirm. The channel can then be used for payments, as long as those payments are valid for every pending splice transactions. Splice transactions can be RBF-ed to speed up confirmation.

Once one of the pending splice transactions confirms and reaches acceptable depth, peers exchange `splice_locked` to discard the other pending splice transactions and the previous funding transaction. The confirmed splice transaction becomes the channel funding transaction.

Nodes then advertise this spliced channel to the network, so that nodes keep routing payments through it without any downtime.

This PR replaces #863 which contains a lot of legacy mechanisms for early versions of splicing, which didn't work in some edge cases (detailed in the test vectors provided in this PR). It can be very helpful to read the protocol flows described in the test vector: they give a better intuition of how splicing works, and how it deals with message concurrency and disconnections.

This PR requires the quiescence feature (#869) to start negotiating a splice.

Credits to @rustyrussell and @ddustin will be added in the commit messages once we're ready to merge this PR.

PR #1289, opened 2025-09-09
Explicit `commit_sig` retransmission for `interactive-tx`

t-bast:
When disconnecting in the middle of the signing steps of an `interactive-tx` transaction, we must retransmit signatures on reconnection to complete the `interactive-tx` protocol.

Nodes first exchange `commitment_signed`, followed by `tx_signatures` once they have both sent and received `commitment_signed`.

We previously always retransmitted `commitment_signed`, even when our peer had already received it. We now include an explicit bitfield that lets nodes request `commitment_signed` if they haven't received it.

Note that this is a breaking change, and we are thus changing the TLV type to make it easier to detect. In practice it should be fine, since only `eclair` and `cln` have shipped support for dual funding, and those two implementations are already incompatible on reconnection because `eclair` implements #1214 but `cln` doesn't. This edge case only creates an issue when nodes disconnect after exchanging `tx_complete` but before receiving signatures, which should happen very infrequently.

Replaces #1214.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -2742,7 +2742,7 @@ A node:
     - MUST reuse the same commitment number for its next `commitment_signed`.
   - otherwise:
     - if `next_commitment_number` is not equal to the commitment number of the
-      next `commitment_signed` the receiving node would send:
+      next `commitment_signed` that the receiving node would send:
       - SHOULD send an `error` and fail the channel.
   - if `next_revocation_number` is equal to the commitment number of
   the last `revoke_and_ack` the receiving node sent, AND the receiving node


02-peer-protocol.md
```diff
@@ -2653,16 +2653,25 @@ messages are), they are independent of requirements here.
 
 1. `tlv_stream`: `channel_reestablish_tlvs`
 2. types:
-    1. type: 0 (`next_funding`)
+    1. type: 1 (`next_funding`)
     2. data:
         * [`sha256`:`next_funding_txid`]
+        * [`byte`:`retransmit_flags`]
```
ddustin, line 2659: In #1160 the diff currently shows:
```
    1. type: 0 (`next_funding`)
    2. data:
        * [`sha256`:`next_funding_txid`]
    1. type: 5 (`my_current_funding_locked`)
    2. data:
        * [`sha256`:`my_current_funding_locked_txid`]
        * [`byte`:`retransmit_flags`]
```

While this PR is showing
```
    1. type: 1 (`next_funding`)
    2. data:
        * [`sha256`:`next_funding_txid`]
        * [`byte`:`retransmit_flags`]
```

So the final goal is to have:
```
    1. type: 1 (`next_funding`)
    2. data:
        * [`sha256`:`next_funding_txid`]
        * [`byte`:`retransmit_flags`]
    1. type: 5 (`my_current_funding_locked`)
    2. data:
        * [`sha256`:`my_current_funding_locked_txid`]
        * [`byte`:`retransmit_flags`]
```

t-bast:
> Can you remind me again why we can't just retransmit the commitment_signed message? It seems much simpler to implement and requires less logic in general, but maybe I'm missing an edge case here where having a duplicate message is problematic.

Because it's hacky, why retransmit a message that doesn't need retransmission? Also, it's trivial without taproot, because the sender could simply store their `commit_sig` and re-send it on reconnection, but with taproot it's much more annoying:

- nonces need to be shared before-hand (your peer must send you nonces for this specific commitment)
- you cannot simply retransmit, you have to re-sign because nonces have changed

If we can avoid an unnecessary musig2 round, I think it's worth it, especially since it makes the protocol conceptually cleaner.

02-peer-protocol.md
t-bast: Yes that's exactly it! I'll rebase #1160 once this PR and #1236 have been merged.

ddustin: Are we still decrementing the sent `next_commitment_number` when splice needs commitments sent?

ddustin: For context, this is what we currently do when sending channel_reestablish:
```
			/* Eclair wants us to decrement commitment number to
			 * indicate that we would like them to re-send
			 * commitment signatures */
			if (!inflight->last_tx)
				send_next_commitment_number--;
```

When recieving channel_reestablish, we use this as the condition for when we resent the `commitment_siged`:
```
next_commitment_number == peer->next_index[REMOTE] - 1
```

When we send (and need more info to build splice tx), do we:
1) Continue decrementing commitment number AND set retransmit_flags bit 0.
2) Stop decrementing commitment number and ONLY set retransmit_flags bit 0.

When receiving, do we:
1) Resend commitment_signed when noticing a decremented commitment number OR retransmit_flags bit 0 being set
2) Resend commitment_signed ONLY when retransmit_flags bit 0 is set

ddustin: Also it appears we're dropping `your_last_funding_locked_txid` entirely -- correct?

ddustin: We were using `your_last_funding_locked_txid` to determine if we needed to resend `splice_locked`. It looks like this new spec does not allow re-sending of `splice_locked` at all. Is that correct?

t-bast: > Are we still decrementing the sent next_commitment_number when splice needs commitments sent?

No we're not, that has been deprecated in favor of this PR! So we're back to the simpler case where `next_commitment_number` really is the next commitment number :)

> Also it appears we're dropping your_last_funding_locked_txid entirely -- correct?

Yes, this has been removed entirely, as it's now unnecessary. You don't need to re-send `splice_locked` at all, it is "contained" in `my_current_funding_locked`, which is simpler: you always send in `my_current_funding_locked` the latest thing you've locked.

I think your comment [here](https://github.com/lightning/bolts/pull/1289#discussion_r2473766377) is thus obsolete, the logic becomes much simpler now that we don't need to decrement the commitment number? You simply re-send `commitment_signed` if your peer requests it in the `retransmit_flags` you receive. And you simply ask for retransmission in the `retransmit_flags` you send if you haven't received the remote `commit_sig` yet for this interactive-tx.

02-peer-protocol.md
```diff
@@ -2726,12 +2741,8 @@ A node:
...
     - MUST reuse the same commitment number for its next `commitment_signed`.
   - otherwise:
-    - if `next_commitment_number` is not 1 greater than the
-  commitment number of the last `commitment_signed` message the receiving
-  node has sent:
-      - SHOULD send an `error` and fail the channel.
-    - if it has not sent `commitment_signed`, AND `next_commitment_number`
-    is not equal to 1:
+    - if `next_commitment_number` is not equal to the commitment number of the
+      next `commitment_signed` the receiving node will send:
```
jkczyz: The "receiving node will send" wording here doesn't read well with "SHOULD send" in the line that follows.

t-bast: I'm not sure what you'd like to say then, those are two distinct messages that the receiving node would send: a `commitment_signed` in the future after the next commitment update, and an `error` that should be sent right now if we end up in that situation.

I changed that to `would` in https://github.com/lightning/bolts/pull/1289/commits/314e79e3c8f97822985739525c94dcba738b504e, but otherwise please let me know how you'd like to rephrase that.

t-bast:
@niftynei @ddustin can you please take a look and ACK this PR? We'd really like to move forward and get it merged since we have cross-compat with LDK and it works with `cln` for splicing (but I haven't tested dual-funding).

02-peer-protocol.md
jkczyz: Oh, lol I mentally inserted a comma after "the commitment number of the next `commitment_signed`". Yeah, "would" instead of "will" helps. Could also insert "that" before "the receiving node".

t-bast: Oh ok, I understand better then! Agreed that adding "that" improves readability, done in https://github.com/lightning/bolts/pull/1289/commits/7646fc2ed2ada63cfc24eb1c772eccf2b9ada159

jkczyz (commented):
ACK 7646fc2ed2ada63cfc24eb1c772eccf2b9ada159

PR #1292, opened 2025-09-22
BOLT 4: add minitramp support.

rustyrussell:
We use the same "encrypted blob inside onion" trick as blinded paths, but using TLV 1, so it's ignored for backward compatibility.  In this case, we need to tell the trampoline the `payment_secret` (and optional `payment_metadata`) to put in the onion, which gives it the possibility of probing, or even underpaying amountless invoices.  However, if the final hop *does* support TLV 1, it can detect interference and refuse attempts to pay that same invoice (since there's a dishonest trampoline), providing *some* honesty incentive.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -189,6 +189,10 @@ This is formatted according to the Type-Length-Value format defined in [BOLT #1]
 
 1. `tlv_stream`: `payload`
 2. types:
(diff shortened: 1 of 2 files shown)


rustyrussell:
I think this is much simpler, though?  But I'll know more once I've implemented it...

PR #1299, opened 2025-11-17
Remove ambiguous note about use of `localpubkey` in `to_remote`

carlaKC:
* Remove ambiguous note about `localpubkey` being used in `to_remote`
* While we're here, use proper htlc pubkey names in expected weights section (rather than overloading `{local/remote}pubkey`


t-bast:
> Remove ambiguous note about localpubkey being used in to_remote

Do you really find it ambiguous? I find it nice to highlight this symmetry between the local and remote commitment transactions, as this is one of the things that newcomers have a hard time wrapping their head around. But I may be biased since I've known this stuff for a while, if people find it confusing, happy to remove it!

PR #1301, opened 2025-11-19
Add details for post-anchor `dust_limit_satoshis`

t-bast:
Since the introduction of `option_anchors`, pre-signed HTLC transactions don't include any fees, and thus don't contribute to trimming. This results in HTLC outputs in the commitment transactions that may be forever uneconomical to spend, and will pollute the utxo set.

We should use a `dust_limit_satoshis` that is higher than Bitcoin Core's standard values to take into account the cost of these second-stage txs at a reasonable (TM) feerate. If we want for example to be able to claim HTLC outputs at `20 sat/byte`, we need a `dust_limit_satoshis` of around `4 000 sats`.

### DIFF
--- a/03-transactions.md
+++ b/03-transactions.md
@@ -573,6 +573,11 @@ second-stage HTLC transactions, to ensure that outputs added to the commitment
 transaction can actually be claimed on-chain, otherwise they may pollute the
 utxo set indefinitely. At a minimum, nodes should allow their peer to use a
 `dust_limit_satoshis` that is higher than the values defined by Bitcoin Core.
+We cannot predict future feerates, so this will not always work and can still
+result in HTLC outputs that are unspendable if the on-chain fees are too high.
+We cannot use very large `dust_limit_satoshis` values either since it would
+create too much dust exposure in the commitment transaction (more details
+[here](/02-peer-protocol.md#bounding-exposure-to-trimmed-in-flight-htlcs-max_dust_htlc_exposure_msat)).
 
 ### Pay to pubkey hash (p2pkh)
 


03-transactions.md
```diff
@@ -551,6 +555,7 @@ Bitcoin Core defines the following dust thresholds:
 - pay to script hash (p2sh): 540 satoshis
 - pay to witness pubkey hash (p2wpkh): 294 satoshis
 - pay to witness script hash (p2wsh): 330 satoshis
+- pay to anchor (p2a): 240 satoshis
```
ziggie1984, line 558: Nit: should probably be part of: https://github.com/lightning/bolts/pull/1228

03-transactions.md
```diff
@@ -560,6 +565,15 @@ is explained in the following sections.
...
 as per Bitcoin Core's implementation.
 
+Note that since the introduction of `option_anchors`, the second-stage HTLC
+transaction's weight is not taken into account when deciding whether outputs
+should be included in the commitment transaction or not. It thus makes sense
+to use a `dust_limit_satoshis` that takes into account the cost of those
+second-stage HTLC transactions, to ensure that outputs added to the commitment
+transaction can actually be claimed on-chain, otherwise they may pollute the
+utxo set indefinitely. At a minimum, nodes should allow their peer to use a
+`dust_limit_satoshis` that is higher than the values defined by Bitcoin Core.
```
ziggie1984, line 575: could you maybe also add a note describing that it can basically not be perfect, if the transaction fee of the base-layer stays for example very high for a long period of time HTLCs will be unspendable and on the other hand the dust limit cannot be too high because of the griefing attacks.

ziggie1984 (approved):
ACK, had some minor comments.

03-transactions.md
t-bast: I'd rather have it here since we "mirror" the dust limits set by Bitcoin Core, even for output types we don't use. For example, we don't use `p2sh` in lightning, but we do list it here for reference.

03-transactions.md
t-bast: Done in https://github.com/lightning/bolts/pull/1301/commits/68c6544f38bd32d7ea56c4b3634294232b7a72d9

PR #1316, opened 2026-01-26
BOLT 12: clarify that offer_amount must be greater than zero

vincenzopalazzo:
Add explicit requirements that:
- Writers MUST set offer_amount greater than zero when present
- Readers MUST NOT respond to offers where offer_amount is zero

This addresses ambiguity about whether offer_amount=0 is valid. Since omitting offer_amount indicates no minimum is required, a zero value when present would be semantically incorrect.

Fixes #1314

### DIFF
--- a/12-offer-encoding.md
+++ b/12-offer-encoding.md
@@ -240,6 +240,7 @@ A writer of an offer:
     - SHOULD omit `offer_chains`, implying that bitcoin is only chain.
   - if a specific minimum `offer_amount` is required for successful payment:
     - MUST set `offer_amount` to the amount expected (per item).
+    - MUST set `offer_amount` greater than zero.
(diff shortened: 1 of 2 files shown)


t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1316/changes/860577233ba3624f6f992b6fc5292b2aa941e5d8, this is obvious and I think it's a bit spammy to explicitly spell it out, but why not.

@rustyrussell WDYT?

rustyrussell (approved):
Ack

Test vector please!

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1316/changes/c0925547a87755db56484e477c6de1cbace462bb, added to eclair in https://github.com/ACINQ/eclair/pull/3265.

PR #1318, opened 2026-02-06
BOLT-4 clarifications

w3irdrobot:
Lots of small clarifications in BOLT-4 found while attempting to implement onion routing:

- The mix header is called a few different things throughout the document. A note was added to explain that
- When discussing right-shifting the header during creation, `hop_payload` was used instead of `hop_payloads`
- Clarified the description of the location of copying data after right-shifting
- Lightning uses a mix-header size of 1300. The sample packet creation code used some variables that weren't actually in the source, making it a bit confusing when trying to read it to understand the process. I hardcoded the values with constants as well as fixed some formatting
- The header creation description makes it clear the header should be generated inside-out, or the last hop first. However, the filler is supposed to be created in regular order but is not mentioned. So I added a small note just to make the clear
- The last commit might be off, but there's nothing in the BOLTs detailing how the associated data is sent to each hop to ensure they are able to calculate the HMACs correctly. This leads me to believe there is a standard piece of data that is assumed to be used by each implementation. However, I didn't see anything in the BOLTs detailing this. Looking at [the LDK source](https://github.com/lightningdevkit/rust-lightning/blob/9df02807c3a921657ca32029ae2054b9def86283/lightning/src/ln/onion_utils.rs#L707), it seems the payment hash might be that standard. So I added that note in here for clarification. However, if this is wrong, I can remove that from the PR

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -819,6 +819,8 @@ the obfuscated `hop_payloads`.
 The following Go code is an example implementation of the packet construction:
 
 ```Go
+const RoutingInfoSize int32 = 1300
+
 func NewOnionPacket(paymentPath []*btcec.PublicKey, sessionKey *btcec.PrivateKey,
 	hopsData []HopData, assocData []byte) (*OnionPacket, error) {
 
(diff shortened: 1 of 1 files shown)


rustyrussell (approved):
Ack 27952baf99bee96092cd83d2ec834631f7e95541

t-bast (approved):
LGTM, thanks!

PR #1324, opened 2026-03-05
Add TLVs required for splicing taproot channels

t-bast:
We add support for splicing and dual-funding taproot channels. Each commit adds a specific feature that leverages taproot and splicing:

- the first commit allows omitting the `prevtx` field when using taproot: there are caveats because of malleability issues (see https://delvingbitcoin.org/t/malleability-issues-when-creating-shared-transactions-with-segwit-v0/497 for more details)
- the second commit adds nonces in various messages to allow creating musig2 partial signatures during dual-funding, splicing and on reconnection

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -213,6 +213,11 @@ This message contains a transaction input.
    1. type: 0 (`shared_input_txid`)
    2. data:
      * [`sha256`:`funding_txid`]
+   1. type: 2 (`prevtx_details`)
+   2. data:
+     * [`sha256`:`prevtx_txid`]
+     * [`u64`:`amount_satoshis`]
+     * [`...*byte`:`scriptpubkey`]
 
 #### Requirements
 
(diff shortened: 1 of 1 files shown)


02-peer-protocol.md
```diff
@@ -221,6 +226,9 @@ The sending node:
   - MUST use a unique `serial_id` for each input currently added to the
     transaction
   - MUST set `sequence` to be less than or equal to 4294967293 (`0xFFFFFFFD`)
+  - If the transaction splices an existing taproot channel or only uses taproot inputs:
+    - MAY omit `prevtx` (and set `prevtx_len` to `0`) and provide the amount and script
```
Roasbeef, line 230: Prob too late to also make `prevtx` a TLV given already deployed?

Roasbeef (commented):
Nice to have this follow up!

02-peer-protocol.md
t-bast: Yeah, unfortunately this shipped has sailed, but it's only two bytes wasted so it's probably fine even though it isn't very elegant...

PR #1325, opened 2026-03-09
bolt07: fix indentation in spec of final reply_channel_range msg

ekzyis:
I believe this is the intended indentation.


t-bast (approved):
Good catch, thanks!

PR #1327, opened 2026-03-19
Change RBF feerate bump rule to ensure BIP125 relay conformance

jkczyz:
The 25/24 multiplicative feerate bump for `tx_init_rbf` can produce increments too small for Bitcoin Core to relay the replacement transaction under BIP125's minimum relay fee policy (default `incrementalRelayFee` of 0.1 sat/vB since Bitcoin Core v30.0). Add a minimum additive increment of 25 sat/kw alongside the existing multiplicative rule, using whichever produces a higher feerate.

### DIFF
--- a/02-peer-protocol.md
+++ b/02-peer-protocol.md
@@ -458,8 +458,11 @@ completed.
 #### Requirements
 
 The sender:
-  - MUST set `feerate` greater than or equal to 25/24 times the `feerate`
-    of the previously constructed transaction, rounded down.
+  - MUST set `feerate` greater than or equal to the maximum of:
+    - 25/24 times the `feerate` of the previously constructed transaction,
+      rounded down.
+    - 25 sat per kw greater than the `feerate` of the previously
+      constructed transaction.
   - If it contributes to the transaction's funding output:
     - MUST set `funding_output_contribution`
   - If it requires the receiving node to only use confirmed inputs:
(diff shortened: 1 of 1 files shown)


jkczyz:
Alternatively, we could do the max of 25/24 and +25 sat/kwu?

t-bast:
> Alternatively, we could do the max of 25/24 and +25 sat/kwu?

I like that, it's too cheap to only bump by `0.1 sat/vB` if the previous feerate wasn't negligible (e.g. `10 sat/vB`). We want those transactions to confirm quickly, otherwise we're not benefiting from the channel capacity changes.

rustyrussell (commented):
I think we should do max of the two.  I missed that 25 perkw is "0.1 sat per vbyte" on my original read.

jkczyz:
Great, I'll update the PR and clarify. For the math, if you start with a fee rate of 25 sat / kwu, using the minimum RBF needed would at first double (100%) to 50 sat / kwu. Then, for each additional RBF round, the percentage increase would fall until reaching 25/24 (4.167%) at the 24th round, which would be going from 600 to 625 sat / kwu. Then going forward the 25/24 multiplicative dominates leading to a fixed 4.167% increase for each successive round instead of the fixed 25 sat / kwu increase.

This assumes using the minimum needed increase each round. If a larger amount is used, the switchover would happen sooner.

jkczyz:
@t-bast @rustyrussell I've updated the PR to use the maximum of the previous requirement and the new requirement. PTAL.

t-bast (approved):

PR #995, opened 2022-05-30
extension-bolt: simple taproot channels (feature 80/81)

Roasbeef:
This PR puts forth two concepts: 
  1. The concept of an "extension BOLT", which is a single documentation extension to the _base_ BOLT spec that references existing "base" BOLTs with new slightly modified functionality. This presents an alternative to littering the main set of BOLTs with a series of "if" statements, which can be somewhat unwieldy for larger changes, and also harder to review/parse. 
  2. A new set set of feature bit, channel type, funding output changes, commitment changes, and HTLC script changes under the umbrella of "simple taproot channels", a.k.a the minimal amount of changes needed to get us to "taprooty level 1". 

The extensions described in this document have purposefully excluded any gossip related changes, as the there doesn't yet appear to be a predominant direction we'd all like to head in (nu nu gossip vs kick the can and add schnorr). 

Most of the changes here described are pretty routine: use musig2 when relevant, and create simple tapscript trees to fold in areas where the script has multiple conditional paths. The main consideration with `musig2` is ofc: how to handle nonces. This document takes a very conservative stance, and simply proposes that all nonces be 100% ephemeral, and forgotten, even after a connection has been dropped. This has some non-obvious implications w.r.t the retransmission flow. Beyond that, it's mostly: piggy back the nonce set of nonces (4 public nonces total, since there're "two" messages) on a message to avoid having to add additional round trips. 

The other "new" thing this adds is the generation/existence of a NUMs point, which is used to ensure that certain paths can _only_ be spent via the script spend path (like the to remote output for the remote party, as this inherits anchor outputs semantics). 

This is still marked as draft, as it's just barely to the point of being readable, and still has a lot of clean ups to be done w.r.t notation, clarify, wording, and full specification.

PR #1330, opened 2026-04-02
Zero-fee taproot channels using v3 transactions (feature 82/83)

t-bast:
We introduce a new `channel_type` that leverages v3 (TRUC) transactions, taproot, pay-to-anchor outputs and ephemeral dust. This `channel_type` unifies `zero_fee_commitments` (which are defined for segwit v0) and `option_simple_taproot` (which is defined for `nVersion = 2`), by building on top of #995 and #1228.

Combining those two channel types results in a cleaner transaction architecture:

- by using a single pay-to-anchor instead of keyed segwit v0 anchors, we don't need to reveal funding keys when spending output to allow late spending of the anchor outputs: this lets us use the revocation key as internal key *everywhere*, which gets rid of the NUMS point introduced by `option_simple_taproot`
- by using TRUC replacement rules, we don't need the `CSV 1` on outputs and the main remote output can directly be a key-path spend, which can be used to CPFP the commitment transaction
- similarly, HTLC outputs can be used to CPFP the commitment transaction
- it also allows an extension that provides better mobile wallet safety against malicious LSP, described here: https://delvingbitcoin.org/t/zero-fee-commitments-for-mobile-wallets/1453

It should be preferred to all existing channel types, once enough bitcoin nodes correctly relay v3 transactions.

A prototype implementation can be found in https://github.com/ACINQ/eclair/pull/3285


rustyrussell:
Please include https://github.com/lightning/bolts/issues/873.  So either include that change in dust htlc handling, or not.  I don't really want YA variant of all channel types, so if it's a good idea, let's do this.

t-bast:
Good point, I'll look into it, it could be a good opportunity to bundle those two.

bolt-simple-taproot.md
```diff
@@ -0,0 +1,2025 @@
...
+          ```
+          OP_SIZE 32 OP_EQUALVERIFY OP_HASH160 <RIPEMD160(payment_hash)> OP_EQUALVERIFY
+          <remote_htlcpubkey> OP_CHECKSIGVERIFY
+          1 OP_CHECKSEQUENCEVERIFY
+          ```
+      * if `nVersion = 3`:
+          ```
+          <remote_htlcpubkey> OP_CHECKSIGVERIFY
+          OP_SIZE 32 OP_EQUALVERIFY OP_HASH160 <RIPEMD160(payment_hash)> OP_EQUAL
+          ```
```
darosior, line 1446: What's the reason for swapping the signature and preimage checks between v2 and v3?

t-bast: It's simply what miniscript yields, the idea was to be miniscript-compatible unless I got it wrong!

darosior: Ok. Doesn't really matter besides readability, but for what it's worth either ordering is Miniscript compatible.

`nVersion` == 2 Miniscript is `and_v(v:hash256(H),and_v(v:pk(K),older(1)))`.

`nVersion` == 3 Miniscript preserving the same ordering is `and_v(v:hash256(H),pk(K))`.

`nVersion` == 3 Miniscript with ordering swapped (above) is `and_v(v:pk(K),hash256(H))`.

t-bast: Oh interesting, I hadn't tried reversing the miniscript policy. Thanks for pointing it out, I'll change that in the next batch of updates to that PR then.

PR #1331, opened 2026-04-10
BOLT 7: port zero always bad.

rustyrussell:
This language dates back to when there were only IPv4 and v6 descriptors. DNS got added with that change, but Tor was never updated.

Zero ports never make sense, always disallow.  I assume everyone does this.


vincenzopalazzo (approved):

t-bast (approved):

vincenzopalazzo (approved):

Roasbeef (approved):
LGTM 🐸

PR #1333, opened 2026-04-24
BOLT 7: add reader requirement on canonical node id ordering.

rustyrussell:
We require this for writers, but omitted it for readers:

```
  - MUST set `node_id_1` and `node_id_2` to the public keys of the two nodes
  operating the channel, such that `node_id_1` is the lexicographically-lesser of the
  two compressed keys sorted in ascending lexicographic order.
```


t-bast (approved):

vincenzopalazzo (approved):

Roasbeef (approved):
LGTM 🍇

PR #1336, opened 2026-05-26
BOLT 12: feature to allow requests of a bolt11 invoice.

rustyrussell:
This would allow a bridge to create a LNURL adapter for a BIP353. for simple offers.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1658,6 +1658,9 @@ even, of course!).
     1. type: 68 (`invoice_error`)
     2. data:
         * [`tlv_invoice_error`:`inverr`]
+    1. type: 70 (`bolt11_invoice`)
+    2. data:
+        * [`...*utf8`:`bolt11`]
 
 #### Requirements
 
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1658,7 +1658,7 @@ even, of course!).
     1. type: 68 (`invoice_error`)
     2. data:
         * [`tlv_invoice_error`:`inverr`]
-    1. type: 70 (`bolt11_invoice`)
+    1. type: 72 (`bolt11_invoice`)
     2. data:
         * [`...*utf8`:`bolt11`]
 
--- a/09-features.md
+++ b/09-features.md
@@ -28,12 +28,14 @@ The Context column decodes as follows:
 * `C-`: presented in the `channel_announcement` message, but always odd (optional).
 * `C+`: presented in the `channel_announcement` message, but always even (required).
 * `9`: presented in [BOLT 11](11-payment-encoding.md) invoices.
+* `2`: presented in [BOLT 12](12-offer-encoding.md) offers.
 * `B`: presented in the `allowed_features` field of a blinded path.
 * `T`: used in the `channel_type` field [when opening channels](02-peer-protocol.md#the-open_channel-message).
 
 | Bits  | Name                              | Description                                               | Context  | Dependencies                | Link                                                                  |
 |-------|-----------------------------------|-----------------------------------------------------------|----------|-----------------------------|-----------------------------------------------------------------------|
 | 0/1   | `option_data_loss_protect`        | ASSUMED                                                   |          |                             |                                                                       |
+| 2/3   | `option_bolt11_request`           | Invoice request for a bolt11 invoice                      | 2        |                             | [BOLT #12][bolt12-bolt11-req]                                         |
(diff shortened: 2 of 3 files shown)


04-onion-routing.md
```diff
@@ -1658,6 +1658,9 @@ even, of course!).
     1. type: 68 (`invoice_error`)
     2. data:
         * [`tlv_invoice_error`:`inverr`]
+    1. type: 70 (`bolt11_invoice`)
```
tnull: nit: This conflicts with `static_invoice` in https://github.com/lightning/bolts/pull/1149

rustyrussell: Fixed!  Good spotting!

09-features.md
```diff
@@ -113,4 +115,5 @@ This work is licensed under a [Creative Commons Attribution 4.0 International Li
 [bolt04-mpp]: 04-onion-routing.md#basic-multi-part-payments
 [bolt04-route-blinding]: 04-onion-routing.md#route-blinding
 [bolt04-attributable-errors]: 04-onion-routing.md
+[bolt12-bolt11-req]: 12-offer-encoding.md#requrements-for-invoice-requests
```
t-bast, line 118: nit:


```suggestion
[bolt12-bolt11-req]: 12-offer-encoding.md#requirements-for-invoice-requests
```

12-offer-encoding.md
```diff
@@ -580,6 +590,8 @@ informative for the payer to know how the sender claims
 
 The requirement to use `offer_paths` if present, ensures a node does not reveal it is the source of an offer if it is asked directly.  Similarly, the requirement that the correct path is used for the offer ensures that cannot be made to reveal that it is the same node that created some other offer.
 
+The `option_bolt11_request` field allows a backwards-compatible mechanism for simple offers, to simplify transition.
```
yyforyongyu, line 593: Looks like we missed the part to bind the bolt11 invoice. For a bolt12 invoice there's a cryptographic binding, but for bolt11 invoice there is no proof that it is coming from the offer? Like the payer cannot prove it corresponds to the exact request or offer version.

12-offer-encoding.md
```diff
@@ -480,6 +481,8 @@ The writer:
         - MUST set `invreq_quantity` less than or equal to `offer_quantity_max`.
     - otherwise:
       - MUST NOT set `invreq_quantity`
+    - if `offer_features` contains `option_bolt11_request`:
```
yyforyongyu, line 484: Guess these if-else's are a bit difficult to follow😂

Given that, when the offer has `offer_quantity_max`, the writer (payer) must set `invreq_quantity`, and may set `invreq_bolt11` when `option_bolt11_request` is found in the `offer_features`, the payer is allowed to include these in a single request,
- `invreq_bolt11`
- `invreq_quantity`

which will be rejected by the reader. So we need to add a writer rule: if `offer_quantity_max` is present, the writer MUST NOT set `invreq_bolt11`.

vincenzopalazzo:
Mh, I do not recall the main reason why this is considered sanity? Could you please help me to understand what the workflow of LNURL is that will require using bolt12 to fetch a bolt11 invoice?

12-offer-encoding.md
rustyrussell: Indeed.  That's bolt11 for you.

12-offer-encoding.md
rustyrussell: Good catch.  We should restrict that at offer creation time.  You can't set offer_quantity_max and also offer the bolt11 feature.

rustyrussell:
> Mh, I do not recall the main reason why this is considered sanity? Could you please help me to understand what the workflow of LNURL is that will require using bolt12 to fetch a bolt11 invoice?

With this, you can create a generic bip353 server (which has to return a bolt12) in front of a LNURL service (which returns a bolt11).  It's not particularly hard to add, or support.  It might be considered a bad idea, but it is a bridge which may help ease transition.

vincenzopalazzo:
>With this, you can create a generic bip353 server (which has to return a bolt12) in front of a LNURL service (which returns a bolt11). It's not particularly hard to add, or support. It might be considered a bad idea, but it is a bridge which may help ease transition.

I see, I am wondering if the transition would be much smoother now that payer proof is getting inside Nostr, but I see now! However, do you/we already have someone who will try to use this?

I am worried that this will be the bridge to slow down the transition

PR #1337, opened 2026-05-27
BOLT 4: further restrictions on overpayment

rustyrussell:
Reports from the field (thanks @michael1011) of gross overpayment in reasonable user scenarios: a payment is taking too long, so they try with another wallet.

Two changes to avoid this.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -395,7 +395,8 @@ The final node:
       all HTLCs in the set.
     - if the total `amt_to_forward` of this HTLC set is equal to or greater
       than `total_msat`:
-      - SHOULD fulfill all HTLCs in the HTLC set
+      - SHOULD fulfill the minimal subset of those HTLCs in the HTLC set which are required for payment
+      - SHOULD reject other "unnecessary" HTLCs in the HTLC set.
     - otherwise, if the total `amt_to_forward` of this HTLC set is less than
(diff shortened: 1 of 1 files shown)


ziggie1984:
I think this leaves the overpayment rules inconsistent with the existing 2x language elsewhere in the spec.

Before this PR, the 2x rule appears to line up across the relevant sections:

- BOLT 11 says BOLT 4 specifies that the payee SHOULD accept up to twice the expected `amount`, so the payer can add small variations.
- BOLT 4's basic MPP writer rules say that if the invoice specifies an `amount`, the sender MUST set `total_msat` to at least that amount and less than or equal to twice that amount.
- BOLT 4's final-node failure handling currently only says the final node SHOULD fail if the amount paid is more than twice the amount expected.

This PR changes only the last rule to a much tighter threshold: more than `max(5000msat, 1%)` above the expected amount. That means a payer could still be compliant with the unchanged BOLT 4 MPP sender rule by setting `total_msat` anywhere up to 2x the invoice amount, while the final node would now also be compliant by failing that payment once it exceeds the new threshold.

So I think this needs matching updates to the BOLT 11 text and the BOLT 4 MPP sender rule, or the new failure threshold needs to be scoped more narrowly so it does not contradict the existing 2x allowance.

PR #1343, opened 2026-06-02
bolt04/09: add option_onion_messages_only_channels (feature 66/67)

erickcestari:
Some senders look up nodes advertising `option_onion_messages` and assume they can relay through them without holding a channel.  The spec was silent on accepting onion messages from channelless peers, so a node that restricts acceptance to channel peers was still compliant yet silently dropped those messages, with no feedback to the sender and no way for the sender to know in advance.

Make accepting onion messages from any connected peer the baseline: a node advertising `option_onion_messages` `SHOULD` accept them from channelless peers. This is a `SHOULD` rather than a `MUST` because acceptance is always best effort: a node may still drop messages due to rate-limiting.  Add `option_onion_messages_only_channels` (66/67), depending on `option_onion_messages`, as the explicit opt-out for a node that wants to restrict acceptance to channel peers (e.g. to limit the DoS surface of free bandwidth/CPU for any node that can connect). A restrictive node should not silently drop these messages while advertising only the baseline: it has to advertise the bit, which makes the pair a reliable signal.  Senders can detect it and simply not attempt a relay they know will fail, rather than hoping a compliant node happens to be lenient.

### DIFF
--- a/04-onion-routing.md
+++ b/04-onion-routing.md
@@ -1785,7 +1785,10 @@ The writer:
 
 The reader:
 
-- SHOULD accept onion messages from peers without an established channel.
+- if it advertises `option_onion_messages_only_channels`:
+  - MUST NOT accept onion messages from peers without an established channel.
+- otherwise:
+  - SHOULD accept onion messages from peers without an established channel.
 - MAY rate-limit messages by dropping them.
 - MUST decrypt `onion_message_packet` using an empty `associated_data`, and `path_key`, as described in [Onion Decryption](04-onion-routing.md#onion-decryption) to extract an `onionmsg_tlv`.
 - If decryption fails, the result is not a valid `onionmsg_tlv`, or it contains unknown even types:
@@ -1826,6 +1829,19 @@ The reader:
 
 
 #### Rationale
+
+Accepting onion messages from channelless peers is the baseline: a node that
+advertises `option_onion_messages` is expected to accept onion messages from
+any connected peer.  This is a SHOULD rather than a MUST because acceptance
+is always best effort: a node may still drop messages due to rate-limiting.
+
+A node that wants to restrict acceptance to channel peers (e.g. to limit the
+DoS surface of free bandwidth/CPU for any node that can connect) advertises
+`option_onion_messages_only_channels` as an explicit opt-out.  A restrictive
+node should not silently drop these messages while advertising only the
+baseline: it has to advertise the bit, which makes the bit a reliable signal.
+A sender can therefore detect it and simply not attempt a relay it knows will
+fail, rather than sending a message that is silently dropped.
 
 Care must be taken that replies are only accepted using the exact
 reply_path given, otherwise probing is possible.  That means checking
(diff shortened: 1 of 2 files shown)


TheBlueMatt:
Given this is really intended as a short-term thing until onion messages are broadly supported, it seems really weird to make this a separate feature bit.

t-bast:
I'm not a fan of introducing a new feature bit either, since I don't think it really improves reliability in practice. Senders should preferably go through a route that has channels anyway: if they can't find one, it will be a best effort and rate-limits may be strict, so even if nodes advertise `option_onion_messages_anywhere` it doesn't necessarily improve reliability.

The best improvement in practice for onion messages reliability will be when most nodes on the network do support it, at which point there will be a lot of routes with channels between nodes that can relay onion messages.

erickcestari:
> Given this is really intended as a short-term thing until onion messages are broadly supported, it seems really weird to make this a separate feature bit.

I don't think it's actually short-term. Even once onion messages are universal, accepting from channelless peers stays useful: a sender can inject from diverse entrypoints to try k different paths to the receiver instead of funneling everything through its own channeled peer, which matters when each hop adds some chance the message is rate-limited and dropped.

We also could avoid the new feature bit if we changed to a MUST route onion messages from channelless peers. But it would require changes in both LND and Eclair.

Roasbeef:
See https://delvingbitcoin.org/t/reimagining-onion-messages-as-an-overlay-layer/1799

TheBlueMatt (commented):
Meeting discussion: if we flip the feature bit to `onion_messages_only_channels` to invert the meaning it would better match reality. Then at least we'd be able to detect the feature bit and just not bother trying cause its clear that it won't work.

erickcestari:
> Meeting discussion: if we flip the feature bit to `onion_messages_only_channels` to invert the meaning it would better match reality. Then at least we'd be able to detect the feature bit and just not bother trying cause its clear that it won't work.

I've flipped the feature bit to `onion_messages_only_channels`.

04-onion-routing.md
```diff
@@ -1703,7 +1703,10 @@ The writer:
 
 The reader:
 
-- SHOULD accept onion messages from peers without an established channel.
+- if it advertises `option_onion_messages_only_channels`:
+  - MUST NOT accept onion messages from peers without an established channel.
+- otherwise:
+  - MUST accept onion messages from peers without an established channel.
```
TheBlueMatt: This kinda reads as if it needs to accept all OMs, but of course its expected it can ratelimit as well. Might be worth a SHOULD but its not all that critical honestly.

TheBlueMatt (approved):

04-onion-routing.md
cdecker: The way that I read it, `to accept` in this case is being used as `read message, then act`, which has to be best effort always. Either works I think, but `SHOULD` is likely closer to the intent of this change.

cdecker (approved):

04-onion-routing.md
erickcestari: My initial rationale for `MUST` was to prevent implementations from rejecting OMs from channel-less peers while staying compliant without signaling the bit. But since drops are unobservable, a `MUST` here isn't enforceable in practice. It's a network-health policy rather than an invariant. I'll update it with the `SHOULD`.

t-bast (approved):

PR #1344, opened 2026-06-15
Add fulfillment_payload to attribution data

joostjager:
This adds an optional `fulfillment_payload` to the `attribution_data` field in `update_fulfill_htlc`, letting the recipient return end-to-end data to the sender on a successful payment. It is the success-case counterpart of the failure message the receiver can already return via the `reason` field of `update_fail_htlc`.

Attribution data (`option_attribution_data`) is not yet widely deployed, so this is a good moment to make a deliberately breaking change in how `attribution_data` is computed.

One motivating use case is proof of payment for spontaneous (keysend) payments, where the sender picks the preimage so it cannot serve as proof, and routing nodes learn it anyway. The payload gives the recipient a channel to return a signed receipt instead, the proof such payments otherwise lack.


t-bast:
Concept ACK, this seems clear enough. At first glance, this should work correctly with trampoline, similarly to how we handle failures (with a double wrapping step at each trampoline node), but I'll need to write some code and tests to be 100% sure.

joostjager:
Hmm I indeed wonder whether the chacha-poly that the final node uses is going to be a problem. But it should be okay, because the double wrapping happens with the normal chacha streams.

04-onion-routing.md
```diff
@@ -1597,11 +1598,81 @@ The _origin node_:
...
+each hop along the payment path.
+
+In addition, the final node can return data to the origin node in an optional `fulfillment_payload` field. It is the
+success counterpart of the failure message that an erring node returns in the `reason` field of `update_fail_htlc`. The
+final node builds the following plaintext:
+
+1. data:
+   * [`u16`:`tlvs_len`]
+   * [`tlvs_len*byte`:`fulfillment_payload_tlvs`]
+   * [`...*byte`:`pad`]
```
TheBlueMatt: Is it simpler to just define TLV N as the padding data rather than doing it with a separate len?

04-onion-routing.md
```diff
@@ -1597,11 +1598,81 @@ The _origin node_:
...
+
+`fulfillment_payload_tlvs` is a [TLV stream](01-messaging.md#type-length-value-format), so that record types
+can be defined without changing a container format; none are defined yet.
+`pad` conceals the length of the conveyed data; it runs to the end of the plaintext, with `tlvs_len` marking the
+boundary.
+
+The final node generates a key using key type `fulfillment` and its shared secret with the origin node, then encrypts
+this plaintext with ChaCha20-Poly1305 using that key and an all-zero nonce, as for `encrypted_recipient_data` in route
+blinding. The ciphertext and its 16-byte tag are the `fulfillment_payload`. Each intermediate node then obfuscates the
+`fulfillment_payload` with its `ammag` key, exactly as it obfuscates a failure return packet; the final node does not,
```
TheBlueMatt, line 1623: Doing this separately with the same key reuses the same XOR stream which seems bad. We should either derive a new key or use a different nonce.

04-onion-routing.md
```diff
@@ -1597,11 +1598,81 @@ The _origin node_:
...
+## Requirements
+
+`attribution_data` is initialized, transformed, updated and verified exactly as in the failure case.
+
+The _final node_:
+  - if `option_attribution_data` is advertised:
+    - MAY include a `fulfillment_payload`, even when reached through a blinded path
+  - if it includes a `fulfillment_payload`:
+    - MUST set `fulfillment_payload_tlvs` to a serialized TLV stream
+    - MUST pad such that the `fulfillment_payload` is at least 256 bytes
```
TheBlueMatt: Should we require that its a multiple of NN bytes?

TheBlueMatt (commented):
Thanks!

04-onion-routing.md
joostjager: I think the problem with that is that you can't pad to any length because of the var length length encoding?

04-onion-routing.md
joostjager: Attribution data uses `ammagext`, so I don't think it is reused in the fulfill message?

04-onion-routing.md
joostjager: Good point. Updated to multiple of 256

04-onion-routing.md
t-bast: If I understand Matt's point correctly, he means that you would do:

```
1. data:
    * [`...*byte`:`fulfillment_payload_tlvs`]
```

Then each TLV inside `fulfillment_payload_tlvs` has the standard TLV format: (`tag`, `length`, `value`).
What Matt proposes is that you simply define a padding TLV field *inside* `fulfillment_payload_tlvs` (instead of outside, which is what you did), for example:

```
1. type: 1 (`fulfillment_payload_padding`)
2. data:
    * [`...*byte`:`padding`]
```

This saves one `u16` and better encapsulates the padding. Depending on your implementation, it makes it slightly more painful to pad the whole `fulfillment_payload` to a constant size (because the padding is inside this TLV stream) but it isn't too hard to do either.

I think this would be cleaner than an outside padding.

joostjager: Yes I got that, but I tried to point out that it might be that you cannot pad to all lengths. When the length field of the padding tlv record jumps to a larger size (BigSize), you skip one possible padding length?

t-bast: Right, I didn't understand that this was what you were pointing out. In that case you "simply" need to pad to the next multiple of 256? But it can feel somewhat hacky, so not sure which option is best.

joostjager: Maybe that can work. It probably ~never happens that you'd actually run into this limitation.

joostjager: I think I prefer the in-tlv padding. Updated PR and test vectors.

t-bast:
I have verified that this works correctly with trampoline :+1: 

Nothing weird came up while implementing this in eclair: I need to verify the test vectors, but otherwise it was quite straightforward.

joostjager:
That's good to hear. And you are FAST! Awesome

t-bast:
I'm able to get the test vectors to pass. I have one comment though: in your current implementation, the final node uses the *encrypted* fulfillment payload in its attribution's HMACs, which is the same thing that the next-to-last node also uses in its attribution HMACs. I would find it more consistent if the final node used the *non-encrypted* fulfillment payload in its attribution's HMACs, then encrypts and passes it to the next-to-last node. This also makes the eclair implementation slightly more elegant. What do you think?

joostjager:
I think one downside is that it would no longer be possible to detect the case where the final node committed, in its attribution data, to an invalid encrypted `fulfillment_payload`. But maybe that is only useful for diagnostics: the origin already knows who the final node is, and the Poly1305 tag is what determines whether the payload is usable.

Thinking about it more, perhaps it is cleaner to not cover the payload at all in the final node attribution HMAC? Intermediate nodes would still cover the encrypted `fulfillment_payload` they receive and relay, which is what matters for attributing return-path tampering.

An even more minimal option would be to omit attribution data for the final node entirely. That would lose the final node hold time, but I am not sure that value is very useful: final nodes are incentivized to report zero anyway.

Also wondering whether that last simplification could work for the failure case too.

Let me know what you think!

t-bast:
> Thinking about it more, perhaps it is cleaner to not cover the payload at all in the final node attribution HMAC? Intermediate nodes would still cover the encrypted fulfillment_payload they receive and relay, which is what matters for attributing return-path tampering.

Yes, I think that makes sense. For the final node, we shouldn't include the `fulfillment_payload` in the attribution HMACs: we already have the Poly1305 MAC on the encrypted `fulfillment_payload` anyway.

> An even more minimal option would be to omit attribution data for the final node entirely. That would lose the final node hold time, but I am not sure that value is very useful: final nodes are incentivized to report zero anyway.

I think it's worth keeping the attribution data for the final node: eclair honestly reports its hold time, and it may be useful in the future. Also, if the final node doesn't include a hold time, the next-to-final node immediately knows that the next node is the final node, which is bad!

joostjager:
Good point on keeping the final node hold time.

I don't think the next-to-final node learns anything from the final node not contributing a hold time though. The final node could still return an `attribution_data` blob with random/masked contents of the expected size. Since the next-to-final node only sees opaque bytes before applying its own transformation, it cannot tell whether those bytes contain a valid final-node hold time/HMAC contribution or just filler. The origin would simply know not to verify/use the final-node attribution entry under that design.

joostjager:
Updated PR to exclude payload from final hop attribution. Updated test vectors.

t-bast (approved):
ACK https://github.com/lightning/bolts/pull/1344/changes/8ed85ba483b65951c47d1c3728f0ce39e3f2b6d9, the intermediate relaying part is implemented in https://github.com/ACINQ/eclair/pull/3321 and matches the test vectors.

joostjager:
I vibe-coded a passing ldk-server interop repro here: https://github.com/lightningdevkit/ldk-server/compare/main...joostjager:ldk-server:fulfillment-payload-interop

It spins up regtest bitcoind, Eclair, and ldk-server nodes and checks that the payment sender can recover the fulfillment payload and validate the attribution data set by the receiver. The current matrix is

`eclair -> ldk -> ldk`
`ldk -> eclair -> ldk`
`eclair -> eclair -> ldk`

joostjager:
I do not have permission to request a review through GitHub, but just noting that @carlaKC  is planning to take a review pass on this.

joostjager:
No news on this spec change. Waiting for more review.

joostjager:
Great to see this merged! I could have squashed the commits, but too late for that now.

PR #1352, opened 2026-08-14
bolt-simple-taproot: fix the output key derivations, signature rules and test vectors

erickcestari:
Corrections to the Simple Taproot Channels extension, found by re-deriving every
published value from the seed and checking the document against the BIP 340,
341, 342 and 327 reference code and against lnd and eclair.

No protocol change is intended: every fix moves the document toward what the two
implementations already do.

### Output keys and control blocks

- Five output key formulas (anchors, offered and accepted HTLCs, both second
  level outputs) added the tweak hash straight to the internal key. BIP 341
  computes `Q = P + int(hashTapTweak(...))G`.
- BIP 341 consumes an internal key in its 32-byte x-only form, but only the NUMS
  point said so. `x_only()` is now defined once and used for every internal key,
  including in the control block serialization.
- `taproot_nums_point` was used throughout the commitment sections but never
  defined, and the `to_remote` section defined its own value for it, which
  matched neither the definition earlier in the document nor the test vectors.
  Its control block also named `combined_funding_key` as the internal key.
- The `to_local` control blocks carried the wrong data and the delay path claimed
  to be 33 bytes when it needs the sibling leaf hash.
- The anchor outputs' key path spend was undocumented.
- Both second level outputs pay to the very same script as `to_local`, but it was
  called `htlc_success` in one place and `htlc_timeout` in the other, reading as
  three distinct scripts. An unclosed code fence in the `to_local` section had
  also swallowed the revoke script into the delay script's block.

### Signatures and nonces

- `commitment_signed` did not say which sighash type the HTLC signatures use.
  They are `SIGHASH_SINGLE|SIGHASH_ANYONECANPAY`, so the receiver has to append
  the `0x83` byte to the 64-byte field when building the witness.
- `SIGHASH_SINGLE` commits to the output at the same index as the signed input,
  which constrains how second level transactions may be aggregated.
- A BIP 327 secret nonce is 97 bytes, not two scalars.
- Names the HMAC arguments in the counter based nonce derivation, documents
  `next_closee_nonce`, and fixes the BIP 327 link.

### Test vectors

- The `accepted_htlc_*` entries carried the local and remote HTLC keys in the
  wrong positions, disagreeing with the script templates here, with the
  transaction vectors in this same document, and with both implementations.
- The `*_remote_commit` entries were byte-identical copies of the local ones, so
  they exercised nothing. They are now derived from `remote_per_commit_point`
  and `local_revocation_basepoint`.
- `combined_key` actually held the BIP 86 tweaked output key, so it is split into
  `combined_funding_key` and `funding_key`.

### Verification

Every key, script, leaf hash, tapscript root and output key re-derived from the
seed; every signature checked against the BIP 341 sighash using the BIP 340 and
BIP 327 reference implementations, including a full MuSig2 replay of each
commitment signature; every witness executed under BIP 342 rules.

Both implementations ship a copy of these vectors, byte-identical to each other
and, in their `transactions` section, to this document. Running each project's
own suite against the JSON from this branch:

- eclair's `SimpleTaprootCommitmentsTestVectorSpec` passes with no source
  changes. It re-derives the keys, rebuilds all three commitment transactions
  and every HTLC transaction, reproduces `remote_partial_sig`, and runs
  `correctlySpends` on each. It reads only `params`, `scripts.funding` and
  `transactions`, so the script entries are covered indirectly, through the leaf
  scripts embedded in the resolution witnesses.
- lnd's `TestTaprootVectors` passes on `params`, the three transactions, the
  signature verification, the MuSig2 replay and 5 of the 11 script entries. The
  6 failures are the changes above: three are lnd's vector generator, and three 
  are the `combined_key` and second level leaf renames its test structs would 
  follow.
  
The 15 commits are split for review only, one defect each, so they can be read
against the BIP text one at a time and any single one can be dropped without
disturbing the rest. They might be squashed into fewer commits before landing.

I've also asked claude to write a script to verify if the new test vectors are correct.

https://gist.github.com/erickcestari/8cd853821529c57aaf6109ee77409dd8

### DIFF
--- a/bolt-simple-taproot.md
+++ b/bolt-simple-taproot.md
@@ -1477,10 +1477,12 @@ A HTLC-Timeout transaction has the following structure:
 All test vectors are derived deterministically from a single 32-byte seed:
 `000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f`. Private
 keys are produced via `SHA256(seed || label)` where `label` is a fixed ASCII
-string for each key (e.g. `"local-funding"`, `"remote-funding"`, etc.). The
-per-commitment point is derived by computing `SHA256(seed ||
-"local-per-commit-secret")` and using the result as the per-commitment secret
-scalar, then multiplying by the generator to obtain the per-commitment point.
+string for each key (e.g. `"local-funding"`, `"remote-funding"`, etc.). Each
+per-commitment point is derived the same way: `SHA256(seed ||
+"local-per-commit-secret")` for the local commitment and `SHA256(seed ||
+"remote-per-commit-secret")` for the remote one give the per-commitment secret
+scalar, which is multiplied by the generator to obtain the per-commitment
+point.
 
 The vectors cover two areas:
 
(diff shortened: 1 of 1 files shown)


t-bast (commented):
Thanks, that looks good to me! I never bothered checking the low-level details of the test vectors in eclair, because checking the entire transactions (with witnesses) was enough proof that the implementation was correct. That's probably why those were merged without being verified.

I'll let @Roasbeef also review this to make sure I didn't miss anything before we integrate it.

PR #1355, opened 2026-08-27
BOLT 4: clarify final onion payload must always contain total_msat

SomberNight:
Now that [`option_payment_secret` is assumed](https://github.com/lightning/bolts/commit/872c6eca9c813a543ec8f0f79dcffc5df03f4bc3), the `total_msat` field must always be present for final onions.

Either the `payment_data` TLV or the `total_amount_msat` TLV must be present, which implies `total_msat` being present.


t-bast (approved):
ACK, agreed that we can get rid of this legacy case now.

Roasbeef (approved):
LGTM

TheBlueMatt (approved):

PR #1357, opened 2026-09-10
bolt11: reject duplicate payment hashes in invoices

Roasbeef:
In this PR, we modify the spec to _reject_ an invoice with duplicate payment hashes. 

Unfortunately there was a recent hack o fa popular telegram p2p bot that exploited this ambiguity. Some libraries took the first payment hash, while some took the last. If you were running a system that exhibited both variants of the behavior (diff between libraries, or nodes, etc), then your system could be tricked into thinking it never paid out a withdrawl, thereby leading to a vuln that could drain the system. 

IMO there's no reason an invoice should have > 1 payment hash, so we should just reject it. 

Test vectors have been updated accordingly.


11-payment-encoding.md
```diff
@@ -364,12 +368,14 @@ https://github.com/rustyrussell/lightning-payencode
 
 # Examples
 
+Machine-readable invoice test vectors for all examples below are provided in
```
TheBlueMatt, line 371: Seems weird to move the invoice data to a json file but leave the descriptions here, makes it much harder to parse.

TheBlueMatt (commented):
Ugh. Yea, should do, obviously.

MPins:
Concept ACK on the duplicate rule.

One suggestion on the test vectors. Marking the "fields which must be ignored" invoice as invalid is correct, but it was also the only example exercising two forward-compatibility paths. I walked the tagged fields of all 28 vectors in the proposed `bolt11/invoice-test.json`: the unknown field type `2` and the `f` field with unknown `version` 19 appear in exactly one invoice — the one this PR marks `"valid": false` — and none of the 15 valid vectors carries an unknown field type or an unknown fallback version. So after this PR nothing positive covers either path.

11-payment-encoding.md
```diff
@@ -691,11 +685,10 @@ Breakdown:
...
-> LNBC25M1PVJLUEZPP5QQQSYQCYQ5RQWZQFQQQSYQCYQ5RQWZQFQQQSYQCYQ5RQWZQFQYPQDQ5VDHKVEN9V5SXYETPDEESSP5ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYG3ZYGS9Q5SQQQQQQQQQQQQQQQQSGQ2A25DXL5HRNTDTN6ZVYDT7D66HYZSYHQS4WDYNAVYS42XGL6SGX9C4G7ME86A27T07MDTFRY458RTJR0V92CNMSWPSJSCGT2VCSE3SGPZ3UAPA
+> ### Same, but with unknown fields and invalid fixed-length fields.
 
-> ### Same, but including fields which must be ignored.
-> lnbc25m1pvjluezpp5qqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqqqsyqcyq5rqwzqfqypqdq5vdhkven9v5sxyetpdeessp5zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zyg3zygs9q5sqqqqqqqqqqqqqqqqsgq2qrqqqfppnqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqppnqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqpp4qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqhpnqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqhp4qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqspnqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqsp4qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqnp5qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqnpkqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqz599y53s3ujmcfjp5xrdap68qxymkqphwsexhmhr8wdz5usdzkzrse33chw6dlp3jhuhge9ley7j2ayx36kawe7kmgg8sv5ugdyusdcqzn8z9x
```
MPins, line 698: Marking this invoice as invalid is correct, but it was also the only example exercising two forward-compatibility paths:
- the unknown field type 2
- the f field with unknown version 19

saubyk (approved):
Concept Ack

ekzyis (commented):
Concept ACK

Fwiw, I was confused about the sender vs. receiver requirements for parsing fields with a fixed length (https://github.com/lightning/bolts/pull/1305). I mentioned in the LDK Discord that these two statements seemed to be in conflict if the invoice contained both a valid and an invalid `p` field:

> [a payer] SHOULD use the first `p` field as the payment hash.

> [a reader] MUST fail the payment if any field with fixed data_length (p, h, s, n) does not have the correct length (52, 52, 52, 53).

It was confusing because it wasn't clear whether the fixed-length check applies _after_ we pick the first `p`, or whether it applies to _any_ `p`, which implies there might be multiple.

I mentioned that I would try to make this less confusing. Maybe I would have found this vulnerability if I had done that 😅

I think that with "MUST use the first `p` field as the payment hash" and "MUST fail the payment if more than one `p` field is present" from this PR, it is now clearer how to handle `p`.

ziggie1984:
Concept ACK. Two things, one substantive and one about the vector format.

### The same rule arguably belongs on `s`

The writer requirements already mandate both:

> - MUST include exactly one `p` field.
> - MUST include exactly one `s` field.

but this PR only adds the matching reader rule for `p`. The reader side for `s` is still phrased in the singular with no rule for what happens when several are present:

> - if a valid `s` field is not provided:
>   - MUST fail the payment.
> - otherwise:
>   - MUST use the `s` field as `payment_secret`

That is the same ambiguity being closed for `p`: "the `s` field" does not say which one, so one implementation takes the first and another takes the last, and two components fed the same signed invoice disagree on the `payment_secret`. The consequences differ from the payment-hash case — a mismatched secret mostly means the payment is rejected at the recipient rather than a bookkeeping split — but a system that decides "paid" or "not paid" by matching secrets has the same shape of bug available to it, and @ekzyis's point above is that the `p` wording was already confusing enough to hide one.

Since the writer rule exists for both fields, the cheapest fix is to extend the new line:

> - MUST fail the payment if more than one `p` or `s` field is present.

Or, more generally, fail on any repeat of a field the writer section specifies as "exactly one" or "one" (`p`, `s`, `d`/`h`, `x`, `c`, `n`, `m`), leaving `f`, `r` and `b` repeatable. I'd understand keeping this PR to `p` if the broader rule needs its own discussion, but `s` seems in scope given the writer rule is already there.

### The vectors record the verdict but not the reason

Building on @TheBlueMatt's point about the split between the JSON and the descriptions: as it stands each entry carries only `valid: true|false`, which lets an implementation pass a vector by rejecting the invoice for an unrelated reason.

Concretely, I ran the 28 proposed vectors through lnd's BOLT 11 decoder, before and after lightningnetwork/lnd#11190 (which implements this PR's `p` rule):

- base: 22/28 agree, after: 25/28. The three that flip are exactly the duplicate-`p` cases, which is the intended effect.
- `Unknown fields and invalid fixed-length fields` is marked invalid here because several fixed-length fields have the wrong `data_length`. lnd rejects it with `invoice contains multiple payment hashes` — the right verdict for the wrong reason, and it would keep passing even if lnd never implemented the length rule. (That invoice now violates two separate MUSTs, so it can't distinguish them.)
- Three `valid: false` vectors are accepted outright by lnd today: `Unknown required feature 100`, `Missing required s field`, and `Non canonical signature (high-S) with 'n' field defined`. Those are pre-existing divergences, not something this PR causes — but they're invisible in a pass/fail-only format until someone reads the output by hand.

Suggestion: give each invalid vector an expected-failure tag (e.g. `"failure": "duplicate_payment_hash"` / `"invalid_field_length"` / `"unknown_required_feature"`), and give the valid ones the expected decoded values (`payment_hash`, `amount_msat`, `payment_secret`, …). If the JSON is going to be the machine-readable source of truth, it may as well carry the assertions rather than just the verdict — that also gives the duplicate-`p` rule a vector that can only be passed by implementing it.

